PR #23, opened 2023-04-14
bugfix for LanceTable.add to convert python lists

gsilvestrin:
- Fixed `add` unit test to create the correct expected result
- Added a unit test for LanceTable.add
- Need to discuss if len(LanceTable) is handled correctly

### DIFF
--- a/python/tests/test_table.py
+++ b/python/tests/test_table.py
@@ -46,7 +46,7 @@ def test_basic(db):
     assert table.to_lance().to_table() == ds.to_table()
 
 
-def test_add(db):
+def test_create_table(db):
     schema = pa.schema(
         [
             pa.field("vector", pa.list_(pa.float32(), 2)),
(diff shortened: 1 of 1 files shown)


python/tests/test_table.py
```diff
@@ -49,14 +49,14 @@ def test_basic(db):
 def test_add(db):
```
changhiskhan, line 49: should this be renamed `test_create_table` or smth?

python/tests/test_table.py
```diff
@@ -79,3 +79,35 @@ def test_add(db):
...
+            {"vector": [5.9, 26.5], "item": "bar", "price": 20.0},
+        ],
+    )
+
+    # table = LanceTable(db, "test")
+    assert len(table) == 2
+
+    count = table.add([{"vector": [6.3, 100.5], "item": "new", "price": 30.0}])
+    assert count == 3
+    #assert len(table) == 3 #FAILS! len(table) == 2, since add creates a new ds
```
changhiskhan, line 99: the table instance holds a reference to a lance dataset right? 
the call to `lance.write_dataset(mode="append")` gives a LanceDataset as output. the table should update it's reference to use that new instance

changhiskhan (approved):

PR #25, opened 2023-04-14
Update README.md

gsilvestrin:


changhiskhan (commented):
You're not connecting to the lance directory. The local connection is to the directory above the *.lance datasets

changhiskhan (approved):

PR #26, opened 2023-04-18
invalidate cached dataset after create_index and add

changhiskhan:


python/lancedb/table.py
```diff
@@ -92,12 +98,13 @@ def create_index(self, num_partitions=256, num_sub_vectors=96):
             The number of PQ sub-vectors to use when creating the index.
             Default is 96.
         """
-        return self._dataset.create_index(
```
eddyxu, line 95: before the PR, this `create_index` does not return new dataset? if so, is it a bug from `lance` side?

python/lancedb/table.py
```diff
@@ -124,7 +131,8 @@ def add(self, data: DATA, mode: str = "append") -> int:
         """
         data = _sanitize_data(data, self.schema)
         ds = lance.write_dataset(data, self._dataset_uri, mode=mode)
-        return ds.count_rows()
+        self._reset_dataset()
+        len(self)
```
eddyxu, line 135: this does not return anything, right? (this is python)

python/lancedb/table.py
changhiskhan: it does, but we don't necessarily want it to return a lance dataset because that's implementation detail from the db's perspective

python/lancedb/table.py
changhiskhan: yeah my mistake

PR #37, opened 2023-04-20
skipping embeddings rate limit when python version > 3.10

gsilvestrin:
prints a warning if the python version > 3.10

```
>>> data = with_embeddings(embed_func, df, show_progress=True)
WARNING: rate limit only support up to 3.10, proceeding without rate limiter
```

Tested with Python 3.11.2 (macos) and  Python 3.10.6 (linux)

Long term we should look into another rate limiter library, possibly https://pypi.org/project/throttler/

### DIFF
--- a/python/lancedb/embeddings.py
+++ b/python/lancedb/embeddings.py
@@ -12,6 +12,8 @@
 #  limitations under the License.
 
 import math
+import sys
+
 from retry import retry
 from typing import Callable, Union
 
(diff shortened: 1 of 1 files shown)


python/lancedb/embeddings.py
```diff
@@ -64,13 +66,17 @@ def embed_func(c):
...
         if len(self.rate_limiter_kwargs) > 0:
-            import ratelimiter
-
-            max_calls = self.rate_limiter_kwargs["max_calls"]
-            limiter = ratelimiter.RateLimiter(
-                max_calls, period=self.rate_limiter_kwargs["period"]
-            )
-            embed_func = limiter(embed_func)
+            v = int(sys.version_info.minor)
+            if v >= 11:
```
eddyxu, line 70: will `ratelimiter` throw an execption or something? if so, just catch the exception?

changhiskhan (approved):

PR #47, opened 2023-04-25
Make distance metric configurable in LanceDB

### LINKED ISSUE #41: Expose MetricType in LanceDB
Allow users to customize the metric / distance type when building index and perform queries.

changhiskhan:
Closes #41

### DIFF
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -24,7 +24,7 @@ class LanceQueryBuilder:
     """
 
     def __init__(self, table: "lancedb.table.LanceTable", query: np.ndarray):
-        self._metric = "l2"
+        self._metric = "L2"
         self._nprobes = 20
         self._refine_factor = None
         self._table = table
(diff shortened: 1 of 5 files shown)


changhiskhan:
i think there's a bug in the cosine distance function in lance

python/lancedb/query.py
```diff
@@ -118,6 +134,7 @@ def to_df(self) -> pd.DataFrame:
                 "column": VECTOR_COLUMN_NAME,
                 "q": self._query,
                 "k": self._limit,
+                "metric": self._metric,
```
eddyxu, line 137: does it need to normalize to lower / upper case ?

python/lancedb/table.py
```diff
@@ -121,6 +124,7 @@ def create_index(self, num_partitions=256, num_sub_vectors=96):
         self._dataset.create_index(
             column=VECTOR_COLUMN_NAME,
             index_type="IVF_PQ",
+            metric=metric,
```
eddyxu, line 127: do we need to do checks on metric string? or normalize it to upper / lower case?

eddyxu (approved):
Minor questions.

python/lancedb/table.py
changhiskhan: we do this in Lance already. And the Rust code converts it to lowercase

PR #57, opened 2023-05-04
S3 Lambda example

jaichopra:


rust/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub fn add(left: usize, right: usize) -> usize {
```
changhiskhan, line 1: ?

docs/src/examples/langchain.md
```diff
@@ -0,0 +1,138 @@
+# Code Documentation Q&A Bot 2.0
```
changhiskhan, line 1: this all references lance format

rust/src/lib.rs
jaichopra: Hmm residual code from a rebase, I'll clean up

docs/src/examples/langchain.md
jaichopra: I'll remove this example and fix in another PR

changhiskhan (approved):

PR #59, opened 2023-05-04
[python] Fix adding data to empty table

### LINKED ISSUE #48: Integrate black for code style
Implement [black](https://pypi.org/project/black/) during CI to enforce a consistent code style

### LINKED ISSUE #54: BUG: adding data to empty table failing
```
In [1]: import lancedb

In [2]: db = lancedb.connect("/tmp/testdb")

In [3]: tbl = db.create_table("test")

In [4]: tbl.add(None)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 tbl.add(None)

File ~/.venv/lance/lib/python3.10/site-packages/lancedb/table.py:156, in LanceTable.add(self, data, mode)
    141 def add(self, data: DATA, mode: str = "append") -> int:
    142     """Add data to the table.
    143
    144     Parameters
   (...)
    154     The number of vectors added to the table.
    155     """
--> 156     data = _sanitize_data(data, self.schema)
    157     lance.write_dataset(data, self._dataset_uri, mode=mode)
    158     self._reset_dataset()

File ~/.venv/lance/lib/python3.10/site-packages/lancedb/table.py:65, in LanceTable.schema(self)
     62 @property
     63 def schema(self) -> pa.Schema:
     64     """Return the schema of the table."""
---> 65     return self._dataset.schema

File /opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/functools.py:981, in cached_property.__get__(self, instance, owner)
    979 val = cache.get(self.attrname, _NOT_FOUND)
    980 if val is _NOT_FOUND:
--> 981     val = self.func(instance)
    982     try:
    983         cache[self.attrname] = val

File ~/.venv/lance/lib/python3.10/site-packages/lancedb/table.py:135, in LanceTable._dataset(self)
    133 @cached_property
    134 def _dataset(self) -> LanceDataset:
--> 135     return lance.dataset(self._dataset_uri, version=self._version)

File ~/code/eto/lance/python/python/lance/__init__.py:50, in dataset(uri, version, asof)
     32 def dataset(
     33     uri: Union[str, Path],
     34     version: Optional[int] = None,
     35     asof: Optional[Union[datetime, pd.Timestamp, str]] = None,
     36 ) -> LanceDataset:
     37     """
     38     Opens the Lance dataset from the address specified.
     39
   (...)
     48         If a version is already specified, this arg is ignored.
     49     """
---> 50     ds = LanceDataset(uri, version)
     51     if version is None and asof is not None:
     52         ts_cutoff = sanitize_ts(asof)

File ~/code/eto/lance/python/python/lance/dataset.py:41, in LanceDataset.__init__(self, uri, version)
     39 uri = os.fspath(uri) if isinstance(uri, Path) else uri
     40 self._uri = uri
---> 41 self._ds = _Dataset(uri, version)

ValueError: LanceError(I/O): Object at location /private/tmp/testdb/test.lance/_latest.manifest not found: No such file or directory (os error 2)
```

There's no schema on an empty table

changhiskhan:
Previously adding data used the table schema to sanitize incoming data. But if the Table was empty to start with then there's no existing schema.

Closes #54 
Closes #48


python/lancedb/table.py
```diff
@@ -183,13 +194,24 @@ def search(self, query: VEC) -> LanceQueryBuilder:
...
-    def create(cls, db, name, data, schema=None, mode="create"):
+    def create(cls, db, name, data=None, schema=None, mode="create"):
         tbl = LanceTable(db, name)
-        data = _sanitize_data(data, schema)
-        lance.write_dataset(data, tbl._dataset_uri, mode=mode)
+        if data is not None:
+            data = _sanitize_data(data, schema)
+            lance.write_dataset(data, tbl._dataset_uri, mode=mode)
+        else:
+            # If we're not writing data, we have to manually check that the table doesn't exist
```
eddyxu, line 203: not sure what logic is here. So it allows user to create table / database without data?
What if the mode == `append` or `overwrite`?

python/tests/test_table.py
```diff
@@ -90,7 +90,20 @@ def test_add(db):
             {"vector": [5.9, 26.5], "item": "bar", "price": 20.0},
         ],
     )
+    _add(table)
 
+    # from empty table GH#54
+    table = LanceTable.create(db, "test2")
```
eddyxu, line 96: in this moment, what would be the behavior to call `table.schema()` `table.count_rows()` and etc?

s7726:
Agree this pr could be better with more use cases accounted for through either direct handling, errors, warnings, etc.

But it does solve the stated issue of not being able to add content to an empty database. Since the existing implementation allows the creation of an empty database (without schema), I think solving just that use case in the near term is worthwhile.

python/tests/test_table.py
s7726: That behavior is already allowed by having the data and schema be optional to the creation of the table.

python/lancedb/table.py
s7726: Overwrite is only for the case that days exists, so not having data isn't an issue. Append just adds data after any potentially existing data. There's no conflict here.

s7726:
Is there something else that needs to be done to get this merged?

skyfallsin:
@eddyxu this PR should fix one of the onboarding issues we discussed

changhiskhan:
Github PR is borked. I'm going to reopen a PR for this branch so the diff is not totally crazy

PR #62, opened 2023-05-07
Basic full text search capabilities

changhiskhan:
This is v1 of integrating full text search index into LanceDB.

# API
The query API is roughly the same as before, except if the input is text instead of a vector we assume that its fts search.

## Example
If `table` is a LanceDB LanceTable, then:

Build index: `table.create_fts_index("text")`

Query: `df = table.search("puppy").limit(10).select(["text"]).to_df()`

# Implementation
Here we use the tantivy-py package to build the index. We then use the row id's as the full-text-search index's doc id then we just do a Take operation to fetch the rows.

# Limitations

1. don't support incremental row appends yet. New data won't show up in search
2. local filesystem only 
3. requires building tantivy explicitly

### DIFF
--- a/docs/src/fts.md
+++ b/docs/src/fts.md
@@ -1,6 +1,8 @@
 # [EXPERIMENTAL] Full text search
 
 LanceDB now provides experimental support for full text search.
+This is currently Python only. We plan to push the integration down to Rust in the future
+to make this available for JS as well.
 
 ## Installation
 
--- a/docs/src/fts.md
+++ b/docs/src/fts.md
@@ -1,0 +1,48 @@
+# [EXPERIMENTAL] Full text search
+
+LanceDB now provides experimental support for full text search.
+
+## Installation
+
+To use full text search, you must install the fts optional dependencies:
+
+`pip install lancedb[fts]`
+
+
+## Quickstart
+
+Assume:
+1. `table` is a LanceDB Table
+2. `text` is the name of the Table column that we want to index
+
+To create the index:
+
+```python
+table.create_fts_index("text")
+```
+
+To search:
+
+```python
+df = table.search("puppy").limit(10).select(["text"]).to_df()
+```
+
+LanceDB automatically looks for an FTS index if the input is str.
+
+## Multiple text columns
+
+If you have multiple columns to index, pass them all as a list to `create_fts_index`:
+
+```python
+table.create_fts_index(["text1", "text2"])
+```
+
+Note that the search API call does not change - you can search over all indexed columns at once.
+
+## Current limitations
+
+1. Currently we do not yet support incremental writes.
+If you add data after fts index creation, it won't be reflected
+in search results until you do a full reindex.
+
+2. We currently only support local filesystem paths for the fts index.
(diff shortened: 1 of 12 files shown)


python/lancedb/fts.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # create a tantivy writer
+    writer = index.writer()
+    # write data into index
+    dataset = table.to_lance()
+    count = 0
+    for b in dataset.to_batches(columns=fields):
+        for i in range(b.num_rows):
+            doc = tantivy.Document()
+            doc.add_integer("doc_id", i)
```
eddyxu, line 81: This id is the id in a RecordBatch, not in the dataset.

changhiskhan: fixed

docs/src/fts.md
```diff
@@ -0,0 +1,48 @@
...
+table.create_fts_index("text")
+```
+
+To search:
+
+```python
+df = table.search("puppy").limit(10).select(["text"]).to_df()
+```
+
+LanceDB automatically looks for an FTS index if the input is str.
```
eddyxu, line 30: This is a python only feature, right?

python/lancedb/query.py
```diff
@@ -145,3 +145,26 @@ def to_df(self) -> pd.DataFrame:
...
+            raise ImportError(
+                "You need to install the `lancedb[fts]` extra to use this method."
+            )
+
+        from .fts import search_index
+
+        # get the index path
+        index_path = self._table._get_fts_index_path()
+        # open the index
+        index = tantivy.Index.open(index_path)
```
eddyxu, line 164: this does not support multi-column?

python/lancedb/table.py
```diff
@@ -130,6 +133,24 @@ def create_index(self, metric="L2", num_partitions=256, num_sub_vectors=96):
         )
         self._reset_dataset()
 
+    def create_fts_index(self, field_names: Union[str, List[str]]):
```
eddyxu, line 136: can we share the `create_index()` API ?

docs/src/fts.md
changhiskhan: For now. Ideally we can integrate Tantivy from Rust but I think that takes more architectural clarity than we currently have (since we don't really know how the usage will look)

python/lancedb/query.py
changhiskhan: So with fts indices you can have a single index that covers multiple columns.

docs/src/fts.md
eddyxu: Add some doc w.r.t. of this python only API?

python/lancedb/table.py
changhiskhan: We can, it just makes that api harder to use because now you have to figure out the index type name and also which kwargs go with which index type etc.

eddyxu: would that make a bunch of `create_foo_index()` ? i.e., a `create_btree_index()` or `create_vector_index()`

changhiskhan: I think we can refactor this into an Index/IndexBuilder abstraction. I just didn't want to commit to it now because this is an experimental feature.

docs/src/fts.md
changhiskhan: Yup will do

eddyxu (approved):

PR #72, opened 2023-05-11
javascript example improvements

gsilvestrin:


changhiskhan (approved):
Please add a `limit` clause to the example

PR #73, opened 2023-05-11
Exposing limit parameter

gsilvestrin:


node/package-lock.json
```diff
@@ -1,13 +1,12 @@
 {
   "name": "vectordb",
-  "version": "0.0.1",
+  "version": "0.0.5",
```
changhiskhan, line 3: Aren't we on 0.0.6 now?

changhiskhan (approved):

node/src/index.ts
```diff
@@ -89,13 +89,17 @@ export class Table {
...
   private readonly _query_vector: number[]
-  private readonly _limit: number
+  private _limit: number
   private readonly _refine_factor?: number
   private readonly _nprobes: number
   private readonly _columns?: string[]
   private readonly _where?: string
   private readonly _metric = 'L2'
 
+  set limit(value: number) {
```
eddyxu, line 99: should we keep the constructor to be the first method in the class?

And why only limit is not readonly? 

Also what's the reason that use `_` as prefix for parameters here?

PR #74, opened 2023-05-11
updated node example

gsilvestrin:


node/examples/simple/index.js
```diff
@@ -22,6 +22,7 @@ async function example() {
 
     const tbl = await db.openTable('my_table');
     const query = tbl.search([0.1, 0.3]);
+    query.limit = 20;
```
changhiskhan, line 25: Can we do this as a chained call so search/limit/execute can a single statement?

jaichopra: I think there are two usability areas to discuss:
1. should `.search` also execute by default, and return results?
2. are we able to chain calls via way of the promises API (in addition to using async/await), i.e:

```
 const results = await db.openTable('my_table')
        .then(tbl => tbl.search([0.1, 0.3]))
        .then(query => {     
            query.limit = 20;
            return query.execute();
        });
```

@changhiskhan are you asking about (2)? 

I was curious to get folks thoughts on (1) -- should we have search execute the results, and then its one less call to make?

changhiskhan (approved):

PR #75, opened 2023-05-11
nodejs create_table

gsilvestrin:
Exposes a `create_table` method in the js in JS interface. Example usage

```javascript
const data = [
     { id: 1, vector: [0.1, 0.2], price: 10 },
     { id: 2, vector: [1.1, 1.2], price: 50 }
]

const table = await con.createTable(tableName, data)
```

### DIFF
--- a/node/examples/simple/index.js
+++ b/node/examples/simple/index.js
@@ -16,12 +16,17 @@
 
 async function example() {
     const lancedb = require('vectordb');
-    const db = lancedb.connect('../../sample-lancedb');
+    const db = await lancedb.connect('data/sample-lancedb');
 
-    console.log(db.tableNames());
+    const data = [
+      { id: 1, vector: [0.1, 0.2], price: 10 },
+      { id: 2, vector: [1.1, 1.2], price: 50 }
+    ]
 
-    const tbl = await db.openTable('my_table');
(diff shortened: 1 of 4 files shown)


rust/ffi/node/src/arrow.rs
```diff
@@ -0,0 +1,33 @@
+use std::ops::Deref;
```
eddyxu, line 1: license

node/src/test/test.ts
```diff
@@ -14,67 +14,84 @@
...
-        const builder = table.search([0.1, 0.3])
-        const results = await builder.execute_cast<MyResult>()
-        assert.equal(results.length, 2)
-        assert.equal(results[0].item, 'foo')
-        assert.equal(results[0].price, 10)
-        assert.approximately(results[0].vector[0], 3.1, 0.1)
-        assert.approximately(results[0].vector[1], 4.1, 0.1)
-      })
+      const data = [
+        { id: 1, vector: [0.1, 0.2], price: 10 },
```
eddyxu, line 72: Are there limitation of what other field types can be? 

Can we test string, float , int as well. 

Might be good to add document later for the rules of schema inference?

eddyxu (commented):
Good for a public announcement.

eddyxu (approved):

PR #84, opened 2023-05-16
JavaScript client initial linux support

gsilvestrin:
The native modules are built manually in linux / macos boxes.

This branch was published to npm as a beta version (won't upgrade existing users). Can be tested with

```
npm install vectordb@0.1.0-linux
```

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vectordb",
-  "version": "0.1.0-linux",
+  "version": "0.1.1",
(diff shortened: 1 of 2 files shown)


node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vectordb",
-  "version": "0.1.0",
+  "version": "0.1.0-linux",
```
eddyxu, line 3: you are checking the system in `node/native.js`, do you still need a system name in version tag

eddyxu (approved):
Yolo

PR #89, opened 2023-05-18
add create index to nodejs client

gsilvestrin:
The api looks like this

```javascript
      const table = await con.openTable('vectors')
      await table.create_index({ type: 'ivf_pq', column: 'vector', num_partitions: 10, max_iters: 2 })
```

In the future we will have other types such as diskann.

### DIFF
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -138,7 +138,7 @@ describe('LanceDB client', function () {
       const con = await lancedb.connect(uri)
       const table = await con.openTable('vectors')
       await table.create_index({ type: 'ivf_pq', column: 'vector', num_partitions: 2, max_iters: 2 })
-    }).timeout(5_000)
+    }).timeout(10_000) // Timeout is high partially because GH macos runner is pretty slow
   })
 })
 


rust/vectordb/src/table.rs
```diff
@@ -83,6 +85,23 @@ impl Table {
...
+    pub async fn create_index_ivf(&self, metric_type: MetricType, num_bits: u8, num_partitions: usize, num_sub_vectors: usize, use_opq: bool, metric_type, max_iterations: usize) {
+        let params: &dyn IndexParams = &VectorIndexParams::ivf_pq(num_partitions, num_bits, num_sub_vectors, use_opq, metric_type, max_iterations);
+        self.dataset.create_index(&*[VECTOR_COLUMN_NAME],IndexType::Vector, None, params)
+    }
+    pub async fn create_index_diskann(&self, /* params here */) {}
+
+    // Option 2 - Single API, accepts IndexParams defined in the lance crate
+    pub async fn create_index(&self, params: &dyn IndexParams) {
+        self.dataset.create_index(&*[VECTOR_COLUMN_NAME],IndexType::Vector, None, params)
+    }
```
wjones127, line 102: @gsilvestrin  What if the API were like:

```js
await table.create_index() // Returns "IndexBuilderSelector" or something
  .ivf({metric_type: my_metric_type, ...})
  .execute();

# or maybe just 
await table.create_index()
  .diskann({...});
```

node/src/index.ts
```diff
@@ -113,8 +113,67 @@ export class Table {
...
+
+  type: 'ivf'
+}
+
+export enum MetricType {
+  L2 = 'l2',
+  Cosine = 'cosine'
+}
+
+export type VectorIndexParams = IvfIndexParamsConfig
```
eddyxu, line 175: Does this mean that it only supports `IvfPQ`? How should it take diskann parameters?

rust/vectordb/src/index/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod vector;
```
eddyxu, line 1: New rust does not need a mod file,

Just put this line into `src/index.rs`

eddyxu (approved):
Pending CI

PR #92, opened 2023-05-22
Lint and test vectordb node in CI

### LINKED ISSUE #90: Test Node module in CI

wjones127:
Closes #90.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -1,0 +1,101 @@
+name: Node
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - node/**
+      - rust/ffi/node/**
+      - .github/workflows/node.yml
+
+env:
+  # Disable full debug symbol generation to speed up CI build and keep memory down
+  # "1" means line tables only, which is useful for panic tracebacks.
+  RUSTFLAGS: "-C debuginfo=1"
+  RUST_BACKTRACE: "1"
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: node
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        lfs: true
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+        cache: 'npm'
+        cache-dependency-path: node/package-lock.json
+    - name: Lint
+      run: |
+        npm ci
+        npm run lint
+  linux:
+    name: Linux (Node ${{ matrix.node-version }})
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        node-version: [ "14", "16", "18", "19" ]
+    runs-on: "ubuntu-22.04"
+    defaults:
+      run:
+        shell: bash
+        working-directory: node
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        lfs: true
+    - uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+        cache-dependency-path: node/package-lock.json
+    - uses: Swatinem/rust-cache@v2
+    - name: Install dependencies
+      run: |
+        sudo apt update
+        sudo apt install -y protobuf-compiler libssl-dev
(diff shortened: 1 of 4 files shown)


node/native.js
```diff
@@ -14,12 +14,26 @@
 
 let nativeLib;
 
-if (process.platform === "darwin" && process.arch === "arm64") {
-    nativeLib = require('./darwin_arm64.node')
-} else if (process.platform === "linux" && process.arch === "x64") {
-    nativeLib = require('./linux-x64.node')
-} else {
-    throw new Error(`vectordb: unsupported platform ${process.platform}_${process.arch}. Please file a bug report at https://github.com/lancedb/lancedb/issues`)
+function getPlatformLibrary() {
```
wjones127, line 17: I'll fill this out in the next PR, this is mostly a placeholder.

Cargo.toml
```diff
@@ -3,3 +3,4 @@ members = [
     "rust/vectordb",
     "rust/ffi/node"
 ]
+resolver = "2"
```
wjones127, line 6: This is the fix for the Linux build.

wjones127:
All CI jobs pass in 2 minutes or less! 🚀

.github/workflows/node.yml
```diff
@@ -0,0 +1,101 @@
...
+    - name: Lint
+      run: |
+        npm ci
+        npm run lint
+  linux:
+    name: Linux (Node ${{ matrix.node-version }})
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        node-version: [ "14", "16", "18", "19" ]
```
eddyxu, line 46: What are the typical number of versions to support in nodejs ecosystem? @gsilvestrin

.github/workflows/node.yml
```diff
@@ -0,0 +1,101 @@
...
+        fetch-depth: 0
+        lfs: true
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+        cache: 'npm'
+        cache-dependency-path: node/package-lock.json
+    - name: Lint
+      run: |
+        npm ci
```
eddyxu, line 39: should we split ci and lint as two steps, let one to fail fast?

.github/workflows/node.yml
```diff
@@ -0,0 +1,101 @@
...
+    - name: Build
+      run: |
+        npm ci
+        npm run build
+        npm run tsc
+    - name: Test
+      run: npm run test
+  macos:
+    timeout-minutes: 30
+    runs-on: "macos-11"
```
eddyxu, line 76: is there a reason to use `macos-11` instead of 12 or even 13?

.github/workflows/node.yml
```diff
@@ -0,0 +1,101 @@
...
+      with:
+        node-version: 18
+        cache: 'npm'
+        cache-dependency-path: node/package-lock.json
+    - uses: Swatinem/rust-cache@v2
+    - name: Install dependencies
+      run: brew install protobuf
+    - name: Build
+      run: |
+        npm ci
```
eddyxu, line 96: Not clear to me what `npm ci` does.

.github/workflows/node.yml
wjones127: I assumed `npm ci` is required for lint. It installs the dev dependencies needed for `npm lint`, right?

.github/workflows/node.yml
wjones127: `npm ci` installs all the dependencies, but is optimized for CI use cases: https://docs.npmjs.com/cli/v9/commands/npm-ci

.github/workflows/node.yml
wjones127: TBH that's just what Co-pilot suggested. Happy to change to 13.

node/native.js
wjones127: For releases, that's still to be decided. But for local development, I think you'll just build index.node for your local platform and it will pick that up. In release it will instead download the the right platform build named `<platform>.node`, and that's what this function is starting to handle. But that will be filled in the next PR.

.github/workflows/node.yml
wjones127: I think I'll cut this down to 16 and 18. 14 is out of support, and I don't think we yet have much reason to test 19. We can always add it back if we discover it's important.

PR #95, opened 2023-05-24
add embedding functions to the nodejs client

gsilvestrin:
Creates an Embeddings trait that can be used to seamless vectorize one of the columns of a Table

The base Embeddings trait is very simple - just the column that should be vectorized and and `embed`function

```javascript
export interface Embeddings<T> {
  sourceColumn: string
  embed: (data: T[]) => number[][]
}
```

It can be used when creating a table as well as when executing queries

```javascript
const embeddings = new MyEmbedding('name')

// No vectors here, they will be created on the fly
const data = [
{ price: 10, name: 'foo' },
{ price: 50, name: 'bar' }
]
// Embedding function is passed as the last element
const table = await con.createTable('vectors', data, embeddings)

// Use 
const results = await table.search('foo').execute()
```

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -134,7 +134,6 @@ export class Table<T = number[]> {
    * Creates a search query to find the nearest neighbors of the given search term
    * @param query The query search term
    */
-  // search (query: T[]): Query
   search (query: T): Query {
     let queryVector: number[]
     if (this._embeddings !== undefined) {


node/src/index.ts
```diff
@@ -92,33 +92,53 @@ export class Table {
...
+    * @param queryVector The query vector.
+    */
+  search (queryVector: number[]): Query
+  /**
+   * Creates a search query to find the nearest neighbors of the given search term
+   * @param query The query search term
+   * @param embeddings An embedding function used to vectorize the query vector
+   */
+  search<T> (query: T, embeddings: EmbeddingFunction<T>): Query
+  search<T> (query: T | number[], embeddings?: EmbeddingFunction<T>): Query {
```
changhiskhan, line 105: Can the Table just hold a reference to the embedding function? I'm not sure what the advantage here just saving one function call.

node/src/index.ts
```diff
@@ -92,33 +92,53 @@ export class Table {
...
   /**
    * Insert records into this Table.
    *
    * @param data Records to be inserted into the Table
+   * @param embeddings An optional embedding function to use
    * @return The number of rows added to the table
    */
-  async add (data: Array<Record<string, unknown>>): Promise<number> {
-    return tableAdd.call(this._tbl, await fromRecordsToBuffer(data), WriteMode.Append.toString())
+  async add<T> (data: Array<Record<string, unknown>>, embeddings?: EmbeddingFunction<T>): Promise<number> {
```
changhiskhan, line 122: same here. can the Table hold a reference?

node/src/index.ts
```diff
@@ -55,17 +55,50 @@ export class Connection {
...
+   * @param name The name of the table.
+   */
+  async openTable (name: string): Promise<Table>
+  /**
+   * Open a table in the database.
+   *
+   * @param name The name of the table.
+   * @param embeddings An embedding function to use on this Table
+   */
+  async openTable<T> (name: string, embeddings: EmbeddingFunction<T>): Promise<Table<T>>
```
eddyxu, line 69: is this a merge leftover? why there are two openTable with embeddings, one optional the other is not.

Also `openTable(name:string)` can be covered by `opentable(name, embeddings?)`?

node/src/index.ts
```diff
@@ -75,27 +108,40 @@ export class Connection {
...
+  private readonly _embeddings?: EmbeddingFunction<T>
 
-  constructor (tbl: any, name: string) {
+  constructor (tbl: any, name: string)
+  /**
+   * @param tbl
+   * @param name
+   * @param embeddings An embedding function to use when interacting with this table
+   */
+  constructor (tbl: any, name: string, embeddings: EmbeddingFunction<T>)
```
eddyxu, line 122: why you need two signatures for `embeddings` and `embeddings?`?

node/src/index.ts
```diff
@@ -75,27 +108,40 @@ export class Connection {
...
 
   /**
-     * Create a search query to find the nearest neighbors of the given query vector.
-     * @param queryVector The query vector.
-     */
-  search (queryVector: number[]): Query {
+   * Creates a search query to find the nearest neighbors of the given search term
+   * @param query The query search term
+   */
+  // search (query: T[]): Query
```
eddyxu, line 137: does this comment serve anything?

changhiskhan (approved):
letsgoooooo

PR #107, opened 2023-05-31
add openai embedding function to nodejs client

gsilvestrin:
- openai is an optional dependency for lancedb
- added an example to show how to use it

### DIFF
--- a/.github/workflows/make_release_commit.yml
+++ b/.github/workflows/make_release_commit.yml
@@ -1,0 +1,48 @@
+name: Create release commit
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Just create the local commit/tags but do not push it'
+        required: true
+        default: "false"
+        type: choice
+        options:
+          - "true"
+          - "false"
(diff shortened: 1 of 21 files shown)


node/src/embedding/embedding_function.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface EmbeddingFunction<T> {
+  /**
+     * The name of the column that will be used as input for the Embedding Function.
+     */
+  sourceColumn: string
+
+  /**
+     * Creates a vector representation for the given values.
+     */
+  embed: (data: T[]) => Promise<number[][]>
```
eddyxu, line 27: do you need a `async` keyword here?

node/src/embedding/openai.ts
```diff
@@ -0,0 +1,38 @@
+
```
eddyxu, line 1: license

eddyxu (commented):
Minor issue

eddyxu (approved):
LGTM

PR #111, opened 2023-05-31
Generate image embeddings 

eddyxu:


notebooks/diffusiondb/datagen.py
```diff
@@ -0,0 +1,110 @@
...
+
+device = "cuda"
+
+tokenizer = CLIPTokenizerFast.from_pretrained(MODEL_ID)
+model = CLIPModel.from_pretrained("openai/clip-vit-base-patch32").to(device)
+processor = CLIPProcessor.from_pretrained("openai/clip-vit-base-patch32")
+
+schema = pa.schema(
+    [
+        pa.field("prompt", pa.string()),
```
changhiskhan, line 40: there are some rows where the prompt is null, which causes problems during index creation. I dropped the NA's manually.

eddyxu: use nullable here? `nullable=True` is the default value.

PR #112, opened 2023-05-31
add existence check

jaichopra:


python/lancedb/table.py
```diff
@@ -225,7 +225,7 @@ def _sanitize_vector_column(data: pa.Table, vector_column_name: str) -> pa.Table
     vector_column_name: str
         The name of the vector column.
     """
-    i = data.column_names.index(vector_column_name)
+    i = data.column_names.index(vector_column_name) if vector_column_name in data.column_names else -1
```
eddyxu, line 228: do we use `i` later?

if just for raise an error, can just do

```python
if vector_column_name not in data.column_names:
     raise ValueError(...)
```?

jaichopra: Agree, this is a cleaner way -- will fix.

changhiskhan (approved):

PR #116, opened 2023-06-01
Check if table exists before opening a table

xudong963:
The `open` api will open an existing table
```rust
impl Table {
    /// Opens an existing Table
    ///
    /// # Arguments
    ///
    /// * `base_path` - The base path where the table is located
    /// * `name` The Table name
    ///
    /// # Returns
    ///
    /// * A [Table] object.
    pub async fn open(base_uri: &str, name: &str) -> Result<Self> {
```

So it's better to check if a table exists before opening it.

Btw, the example in https://lancedb.com/ will be better to create table before opening table
image


changhiskhan (approved):
Thanks!

PR #117, opened 2023-06-01
Use fsspec to implement table_names with cloud storage support

gsilvestrin:
See `python/tests/test_io.py` for an example of usage with s3

### DIFF
--- a/python/lancedb/db.py
+++ b/python/lancedb/db.py
@@ -49,17 +49,11 @@ class LanceDBConnection:
         -------
         A list of table names.
         """
-        filesystem = None
-        scheme = get_uri_scheme(self.uri)
-        if scheme == "file":
-            filesystem = fs.LocalFileSystem()
-        elif scheme == "s3":
-            filesystem = fs.S3FileSystem()
-        elif scheme == "gs":
-            filesystem = fs.GcsFileSystem()
-        else:
+        try:
+            filesystem, path = fs.FileSystem.from_uri(self.uri)
+        except pa.ArrowInvalid:
             raise NotImplementedError(
-                "Unsupported scheme: " + scheme
+                "Unsupported scheme: " + self.uri
             )
 
         try:


python/lancedb/util.py
```diff
@@ -41,3 +41,11 @@ def get_uri_scheme(uri: str) -> str:
         # So we add special handling here for schemes that are a single character
         scheme = "file"
     return scheme
+
+
+# A function that removes the suffix of a string
+# This helper exist because String.removesuffix is only available for Python > 3.9
+def removesuffix(string, suffix):
```
changhiskhan, line 48: does `os.path.splitext` work?

python/lancedb/db.py
```diff
@@ -47,11 +49,26 @@ def table_names(self) -> list[str]:
...
+        if scheme == "file":
+            filesystem = fs.LocalFileSystem()
+        elif scheme == "s3":
+            filesystem = fs.S3FileSystem()
+        elif scheme == "gs":
+            filesystem = fs.GcsFileSystem()
+        else:
+            raise NotImplementedError(
+                "Unsupported scheme: " + scheme
+            )
```
wjones127, line 63: We can let PyArrow get the path for us if we want:

```suggestion
        try:
            filesystem, path = fs.FileSystem.from_uri(self.uri)
        except pa.ArrowInvalid:
            raise NotImplementedError(
                "Unsupported scheme: " + self.uri
            )
```

wjones127 (approved):
Seems good. Had a comment where we might simplify, but given this is temporary I think it's fine to merge as is.

changhiskhan (approved):

python/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "lancedb"
 version = "0.1.4"
-dependencies = ["pylance>=0.4.17", "ratelimiter", "retry", "tqdm"]
+dependencies = ["pylance>=0.4.17", "ratelimiter", "retry", "tqdm", "fsspec", "s3fs", "gcsfs"]
```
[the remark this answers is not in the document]
eddyxu: Isn't pyarrow already support s3 and gcs  natively?

PR #120, opened 2023-06-01
Bumping vectordb to v0.1.3

gsilvestrin:


wjones127:
Did you run the script at ci/bump_versions.sh? This PR seems to be missing the native dependency bumps, which that script should handle.

https://github.com/lancedb/lancedb/blob/391de1644b6caf306d1aea8f1e97ca1eeac8bd47/node/package.json#L74-L77

PR #159, opened 2023-06-07
rename examples for slugs

jaichopra:


chebbyChefNEQ (approved):
quick nit: may double check if there are references to these renamed docs and fix the links.

jaichopra:
> quick nit: maybe double check if there are references to these renamed docs and fix the links.

will do

PR #172, opened 2023-06-10
feat(python): add action to tag python releases

gsilvestrin:
- Bumps Python package version
- Creates a tag in the format python-vX.Y.Z
- Commits the changes to the repo

Uses [bump2version](https://pypi.org/project/bump2version/) to simplify the process.


wjones127 (approved):
This all seems reasonable.

PR #174, opened 2023-06-11
[Python] Updated to_df implementation in Contextualizer class

### LINKED ISSUE #165: Contextualizer drops windows that don't have enough rows
It seems like contextualizer will only return rows with the exactly `window` source rows, which means the last window is always chopped off. In some cases, like if there's just one source row left, we might want to leave it out. But others, like we have just one less than a full window, it would be nice to keep in. Perhaps we should have an optional threshold for this?

```python
from lancedb.context import contextualize
import pandas as pd
data = pd.DataFrame({
   'token': ['The', 'quick', 'brown', 'fox', 'jumped', 'over', 'the', 'lazy', 'dog', 'I', 'love', 'sandwiches'],
   'document_id': [1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2]
})

(
    contextualize(data)
    .window(6)
    .stride(3)
    .text_col('token')
    .groupby('document_id')
    .to_df()['token']
    .to_list()
)
```

Got:
```
['The quick brown fox jumped over']
```

But expected:

```
['The quick brown fox jumped over', 'fox jumped over the lazy dog', 'I love sandwiches']
```

trueutkarsh:
Changes include:
- Contexts of sizes less than window param to be included as well
- Added optional threshold parameter to to_df in Contextualizer 
This should close #165 
- If maintainers satisfied with the implementation will add more examples and test cases and update the documentations as well.

### DIFF
--- a/python/lancedb/context.py
+++ b/python/lancedb/context.py
@@ -137,17 +137,23 @@ class Contextualizer:
         self._text_col = text_col
         return self
 
-    def to_df(self) -> pd.DataFrame:
+    def to_df(self, threshold: int = 1) -> pd.DataFrame:
         """Create the context windows and return a DataFrame."""
 
         def process_group(grp):
             # For each group, create the text rolling window
+            # with values greater than threshold
             text = grp[self._text_col].values
-            contexts = grp.iloc[: -self._window : self._stride, :].copy()
-            contexts[self._text_col] = [
-                " ".join(text[start_i : start_i + self._window])
-                for start_i in range(0, len(grp) - self._window, self._stride)
+            contexts = grp.iloc[: : self._stride, :].copy()
+            windows = [
+                " ".join(text[start_i : min(start_i + self._window, len(grp))])
+                for start_i in range(0, len(grp), self._stride)
+                if start_i + self._window <= len(grp) or len(grp) - start_i > threshold
             ]
+            # if last few rows dropped
+            if len(windows) < len(contexts):
+                contexts = contexts.iloc[:len(windows)]
+            contexts[self._text_col] = windows
             return contexts
 
         if self._groupby is None:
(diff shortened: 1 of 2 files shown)


python/lancedb/context.py
```diff
@@ -137,17 +137,23 @@ def text_col(self, text_col: str) -> Contextualizer:
         self._text_col = text_col
         return self
 
-    def to_df(self) -> pd.DataFrame:
+    def to_df(self, threshold: int = 1) -> pd.DataFrame:
```
wjones127, line 140: I think this should be a parameter on the builder and not in the `to_df` method. This is mostly because we might offer a set of other `to_<something>` methods and we don't want to have to replicate those arguments over all of them.

wjones127 (commented):
This looks good. I appreciate the tests so far. Just had one piece of feedback on the placement of the parameter.

python/lancedb/context.py
trueutkarsh: Hi Will, 
Thanks for pointing this out. I have updated the implementation accordingly.  Please let me know your thoughts over it.

python/lancedb/context.py
```diff
@@ -139,6 +164,18 @@ def text_col(self, text_col: str) -> Contextualizer:
         self._text_col = text_col
         return self
 
+    def threshold(self, threshold: int) -> Contextualizer:
```
wjones127, line 167: For clarity, could we rename `threshold` to `min_window_size`? And also change the inequality to reflect that?

trueutkarsh: Sure, definitely. That definitely makes it more clear.

wjones127 (approved):
Thanks for working on this!

PR #179, opened 2023-06-12
feat(python): add action to build and publish wheel

gsilvestrin:
Tested by building and publishing to test.pypi. The version over there doesn't match our own since I had to repeat my tests a couple of times: https://test.pypi.org/project/lancedb/0.1.20/

It can be installed locally by 

```
pip install --index-url https://test.pypi.org/simple/ --no-deps lancedb
```

Since we don't push pylance to the test repo often, you need to install the dependencies manually.

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -1,0 +1,31 @@
+name: PyPI Publish
+
+on:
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Looks good!

PR #183, opened 2023-06-13
feat(node): add where method to query builder

### LINKED ISSUE #181: [nodejs] Expose where method
Currently the JS library exposes a `filter` method, but the Python API calls it `where` instead. We should create an alias that exports `filter` as `where` to make things consistent.

gsilvestrin:
Closes #181


wjones127 (approved):
Looks good!

PR #194, opened 2023-06-14
port remote connection client into lancedb

chebbyChefNEQ:
* to_df() is now async, added `to_df_blocking` to convenience
* add remote lancedb client to public lancedb
* make lancedb connection class understand url scheme `lancedb+<connection_type>://<host>:<port>`.

### DIFF
--- a/python/lancedb/db.py
+++ b/python/lancedb/db.py
@@ -12,6 +12,7 @@
 #  limitations under the License.
 
 from __future__ import annotations
+import functools
 
 import os
 from pathlib import Path
@@ -56,7 +57,16 @@ class LanceDBConnection:
     """
 
     def __init__(self, uri: URI):
-        is_local = isinstance(uri, Path) or get_uri_scheme(uri) == "file"
+        if not isinstance(uri, Path):
+            scheme = get_uri_scheme(uri)
+        is_local = isinstance(uri, Path) or scheme == "file"
+        # managed lancedb remote uses schema like lancedb+[http|grpc|...]://
+        self._is_managed_remote = not is_local and scheme.startswith("lancedb")
+        if self._is_managed_remote:
+            if len(scheme.split("+")) != 2:
+                raise ValueError(
+                    f"Invalid LanceDB URI: {uri}, expected uri to have scheme like lancedb+<flavor>://..."
+                )
         if is_local:
             if isinstance(uri, str):
                 uri = Path(uri)
@@ -64,9 +74,48 @@ class LanceDBConnection:
             Path(uri).mkdir(parents=True, exist_ok=True)
         self._uri = str(uri)
 
+        self._entered = False
+
     @property
     def uri(self) -> str:
         return self._uri
+
+    @functools.cached_property
+    def is_managed_remote(self) -> bool:
+        return self._is_managed_remote
+
+    @functools.cached_property
+    def remote_flavor(self) -> str:
+        if not self.is_managed_remote:
+            raise ValueError(
+                "Not a managed remote LanceDB, there should be no server flavor"
+            )
+        return get_uri_scheme(self.uri).split("+")[1]
+
+    @functools.cached_property
+    def _client(self) -> "lancedb.remote.LanceDBClient":
+        if not self.is_managed_remote:
+            raise ValueError("Not a managed remote LanceDB, there should be no client")
+
+        # don't import unless we are really using remote
+        from lancedb.remote.client import RestfulLanceDBClient
+
+        if self.remote_flavor == "http":
+            return RestfulLanceDBClient(self._uri)
+
+        raise ValueError("Unsupported remote flavor: " + self.remote_flavor)
+
+    async def close(self):
+        if self._entered:
+            raise ValueError("Cannot re-enter the same LanceDBConnection twice")
+        self._entered = True
+        await self._client.close()
+
+    async def __aenter__(self) -> LanceDBConnection:
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        await self.close()
 
     def table_names(self) -> list[str]:
         """Get the names of all tables in the database.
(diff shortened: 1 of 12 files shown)


python/tests/test_e2e_remote_db.py
```diff
@@ -0,0 +1,15 @@
+from lancedb import LanceDBConnection
```
eddyxu, line 1: add license to each file.

python/lancedb/table.py
```diff
@@ -159,10 +174,12 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return self.__repr__()
 
+    @_not_supported_on_remote
     def head(self, n=5) -> pa.Table:
         """Return the first n rows of the table."""
         return self._dataset.head(n)
 
+    @_not_supported_on_remote
```
eddyxu, line 182: why is `to_pandas()/to_arrow()` not supported?

python/lancedb/db.py
```diff
@@ -68,6 +92,41 @@ def __init__(self, uri: URI):
...
+    @functools.cached_property
+    def remote_flavor(self) -> str:
+        if not self.is_managed_remote:
+            raise ValueError(
+                "Not a managed remote LanceDB, there should be no server flavor"
+            )
+        return get_uri_scheme(self.uri).split("+")[1]
+
+    @functools.cached_property
+    def client(self) -> "RestfulLanceDBClient":
```
eddyxu, line 108: This is a private property, i.e., users dont usually access `db.client` by themselves?

we can probally use `from futures import __annotations__` to make the type declaration better. Also should it return a `LanceDBClient` base class?

python/lancedb/db.py
```diff
@@ -68,6 +92,41 @@ def __init__(self, uri: URI):
...
+
+        if self.remote_flavor == "http":
+            return RestfulLanceDBClient(self._uri)
+
+        raise ValueError("Unsupported remote flavor: " + self.remote_flavor)
+
+    async def close(self):
+        await self.client.close()
+
+    async def __aenter__(self) -> LanceDBConnection:
```
eddyxu, line 123: is there a flag to check this connection is "closed".  It can not enter the `__aenter__` twice

python/lancedb/db.py
```diff
@@ -104,6 +163,7 @@ def __contains__(self, name: str) -> bool:
     def __getitem__(self, name: str) -> LanceTable:
         return self.open_table(name)
 
+    @_not_supported_on_remote
```
eddyxu, line 166: How can user to create table when they are using remote server ? :)

python/lancedb/query.py
```diff
@@ -183,6 +193,33 @@ def to_df(self) -> pd.DataFrame:
...
+        and also the "score" column which is the distance between the query
+        vector and the returned vector.
+        """
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = asyncio.get_event_loop()
+        return loop.run_until_complete(self.to_df_async())
+
+    def to_remote_query(self) -> "VectorQuery":
```
eddyxu, line 209: is there a difference between remove vs local query?

python/lancedb/remote/client.py
```diff
@@ -0,0 +1,107 @@
+import functools
```
eddyxu, line 1: add license.

python/lancedb/table.py
chebbyChefNEQ: these two are fetching the entire table. 

we don't have an API for this yet.

python/lancedb/db.py
chebbyChefNEQ: this one need to stat quoted because we don't want to import any thing under `lancedb.remote` unless we are using remote. Refactored to return an interface

python/lancedb/db.py
chebbyChefNEQ: right now create table is not supported over the remote. we can add that api.

python/pyproject.toml
```diff
@@ -37,14 +37,17 @@ repository = "https://github.com/lancedb/lancedb"
...
-    "pytest", "pytest-mock", "doctest"
+    "pytest", "pytest-mock", "doctest", "pytest-asyncio", "aiohttp", "pydantic", "attr"
 ]
 dev = [
     "ruff", "pre-commit", "black"
 ]
 docs = [
     "mkdocs", "mkdocs-jupyter", "mkdocs-material", "mkdocstrings[python]"
 ]
+managed = [
```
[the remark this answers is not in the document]
chebbyChefNEQ: not too big, these three are all very light weight

python/lancedb/remote/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+try:
+    import aiohttp
+    import pydantic
+except ImportError:
+    raise ImportError(
+        "LanceDB remote client requires aiohttp and pydantic, please install lancedb[managed]"
```
eddyxu, line 19: This `ImportError` is not valid anymore, right?

python/lancedb/remote/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+import abc
+from typing import List, Optional
```
eddyxu, line 23: shall we run `isort` over this file?

python/lancedb/remote/client.py
```diff
@@ -0,0 +1,80 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+import functools
+
+import aiohttp
+import attr
+import pyarrow as pa
+import urllib.parse
```
eddyxu, line 20: lets run a isort here?

eddyxu (approved):
LGTM overall, pending a few small nits.

python/lancedb/remote/__init__.py
chebbyChefNEQ: yes, after this PR. will submit an isort PR. there are some left over stuff from previous PR.

PR #196, opened 2023-06-15
Add docs code testing & documentation syntax changes

### LINKED ISSUE #170: Run user guide examples in CI
Extract all the Python and Node codeblocks and run them through in a test environment, like the Python doctests. This ensures they are always up-to-date.

TevinWang:
- Creates testing files `md_testing.py` and `md_testing.js` for testing python and nodejs code in markdown files in the documentation
This listens for HTML tags as well: `<!--[language] code code code...-->` will create a set-up file to create some mock tables or to fulfill some assumptions in the documentation.
- Creates a github action workflow that triggers every push/pr to `docs/**`
- Modifies documentation so tests run (mostly indentation, some small syntax errors and some missing imports)

A list of excluded files that we need to take a closer look at later on: 
```javascript
const excludedFiles = [
  "../src/fts.md",
  "../src/embedding.md",
  "../src/examples/serverless_lancedb_with_s3_and_lambda.md",
  "../src/examples/serverless_qa_bot_with_modal_and_langchain.md",
  "../src/examples/youtube_transcript_bot_with_nodejs.md",
];
```
Many of them can't be done because we need the OpenAI API key :(. `fts.md` has some issues with the library, I believe this is still experimental? 

Closes #170


wjones127 (changes requested):
I think the two top priorities are:

1. Restore the docs so the tabs aren't broken
2. Get this working in CI with the current dev version rather than the latest released version.

docs/src/search.md
```diff
@@ -30,55 +30,81 @@ the vector column and compute the distance.
...
+    .limit(10) \
+    .to_df()
+```
 
 === "JavaScript"
 
-    ```javascript
-    const vectordb = require('vectordb')
-    const db = await vectordb.connect('data/sample-lancedb')
+<!--javascript
```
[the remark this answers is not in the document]
TevinWang: I was thinking that we should exclude just the metricType section for now instead of the entire file. Do you think its better to just exclude the entire file?

docs/test/md_testing.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if len(lines) > 0:
+            out_path = Path(python_folder) / Path(file).name.strip(".md") / (Path(file).name.strip(".md") + file_ending + python_file)
+            print(out_path)
+            out_path.parent.mkdir(exist_ok=True, parents=True)
+            with open(out_path, "w") as out:
+                out.writelines(lines)
+
+# Setup doc code
+create_code_files("<!--", "-->", "-setup")
```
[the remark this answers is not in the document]
TevinWang: Some documentation assumes that we have a database and table already or data already in the table, etc. I believe we should set that up using some setup code file!

docs/src/ann_indexes.md
```diff
@@ -13,37 +13,37 @@ In the future we will look to automatically create and configure the ANN index.
...
 
 === "Python"
-     Creating indexes is done via the [create_index](https://lancedb.github.io/lancedb/python/#lancedb.table.LanceTable.create_index) method.
+Creating indexes is done via the [create_index](https://lancedb.github.io/lancedb/python/#lancedb.table.LanceTable.create_index) method.
 
-     ```python
-     import lancedb
-     import numpy as np
-     uri = "data/sample-lancedb"
-     db = lancedb.connect(uri)
```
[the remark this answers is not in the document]
TevinWang: I wanted to keep the script getting the test files simpler, but you are totally right I completely overlooked the tab structure. Will fix!

docs/test/md_testing.py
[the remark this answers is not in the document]
wjones127: Got it. Could we add a little more description of that step in a comment here?

docs/src/ann_indexes.md
```diff
@@ -121,7 +122,9 @@ You can select the columns returned by the query using a select clause.
 
 === "Python"
      ```python
-     tbl.search(np.random.random((768))).select(["vector"]).to_df()
+     tbl.search(np.random.random((1536))).select(["vector"]).to_df()
+     ```
+     ```
```
wjones127, line 127:
```suggestion
```

docs/src/fts.md
```diff
@@ -18,6 +18,20 @@ Assume:
...
+
+```python
+import lancedb
+
+uri = "data/sample-lancedb"
+db = lancedb.connect(uri)
+
+table = db.create_table("my_table",
+            data=[{"vector": [3.1, 4.1], "text": "foo"},
+                  {"vector": [5.9, 26.5], "text": "bar"}])
```
wjones127, line 31: Since the example searches for "puppy" later, why don't we have some sentences that include puppy?
```suggestion
            data=[{"vector": [3.1, 4.1], "text": "Frodo was a happy puppy"},
                  {"vector": [5.9, 26.5], "text": "There are several kittens playing"}])
```

wjones127 (commented):
I'll take a closer look tomorrow morning, but two initial comments.

docs/src/search.md
```diff
@@ -30,55 +30,82 @@ the vector column and compute the distance.
...
+import numpy as np
+uri = "data/sample-lancedb"
+db = lancedb.connect(uri)
+
+data = [{"vector": row, "item": f"item {i}"}
+     for i, row in enumerate(np.random.random((10_000, 1536)).astype('float32'))]
+
+db.create_table("my_vectors", data=data)
+-->
+
```
wjones127, line 44: These comments seem to break the tabs. What if we instead brought the setup code up above the tabs?

Also, could we just make the setup blocks normal code blocks that are inside a comment? I think that will make it much clearer to future developers what is going on. Plus IIUC it should reduce the number of steps in the code process. What do you think?

```python

<!-- Setup Code

```python
import lancedb
import numpy as np
uri = "data/sample-lancedb"
db = lancedb.connect(uri)

data = [{"vector": row, "item": f"item {i}"}
     for i, row in enumerate(np.random.random((10_000, 1536)).astype('float32'))]

db.create_table("my_vectors", data=data)
`` ` (<- Hard to escape this in GitHub)
-->

=== "Python"
```

docs/src/ann_indexes.md
TevinWang: I think we might want to keep this to maintain the next code block: 
Screen Shot 2023-06-22 at 10 15 12 AM

docs/src/search.md
TevinWang: This would definitely make it much better and clenaer! I guess the one issue is that we would need to remove the duplicate imports in the actual code blocks, since then all of the code would go into the same file.

TevinWang: Never mind, I'm completely wrong. It works with duplicate imports, I pushed the changes.

wjones127 (commented):
cc @AyushExel does this overlap with what you were talking about with having notebook-based examples?

wjones127 (approved):
This LGTM now. Will merge later today.

AyushExel:
> cc @AyushExel does this overlap with what you were talking about with having notebook-based examples?

Sorry didn't see this. Yeah something similar. I wanted to have all the examples currently in the docs to be available as both notebooks and scripts that users can simply run. And future, also include code from blogs, plugins, etc. that don't necessarily belong in the main repo. Started some work here https://github.com/lancedb/vectordb-recipes

wjones127:
@TevinWang Sorry for the delay. Could you resolve the conflict? Then I can merge.

TevinWang:
> @TevinWang Sorry for the delay. Could you resolve the conflict? Then I can merge.

No worries, resolved!

PR #210, opened 2023-06-20
expose caching in the DB as well

chebbyChefNEQ:


rust/vectordb/src/query.rs
```diff
@@ -74,9 +74,7 @@ impl Query {
         )?;
         scanner.nprobs(self.nprobes);
         scanner.use_index(self.use_index);
-        self.select
-            .as_ref()
-            .map(|p| scanner.project(p.as_slice()));
+        self.select.as_ref().map(|p| scanner.project(p.as_slice()));
```
chebbyChefNEQ, line 77: I think this came from cargo fmt

changhiskhan (commented):
don't check in Cargo.lock since this is a library?

chebbyChefNEQ:
old pr.

Issue #130, opened 2023-06-02
[Doc] Describe a life cycle of a Table


eddyxu:
# Problem Statement

To describe a user journay:

* How to open a database
* How to find table
* How to open a table
* How to create an empty table
* How to append / add / delete data
* How to create indexa and re-index
* How to drop a table

in both `python` and `typescript`.

AyushExel:
Use an external table 
* Create a table from a table
   **Ques** - Table file path or Table object? Or both?
  **Use case** - lanceDB tables could be used as experiment artifacts/assets, so it would need a clean interface to create a working table from an original table without altering it.
  ```
  table = db.import_table(str or LanceTable) # To copy the given table to current connection
  ```
`import_table` could be confusing, maybe call it `copy_table` for explicit implication that the copied table will not be altered

PR #215, opened 2023-06-21
Allow copying table 

AyushExel:
**Story - https://github.com/lancedb/lancedb/issues/130**

**Motivation - Allow simple workflow to copy/import tables**
Usage 1 - Using lanceDB tables as assets, like model weights, to be loaded and shared via path
```
lancedb_art = "path/to/my_table.lance" # lanceDB artifact
db = lancedb.connect(".db/")
db.copy_table(lancedb_art)  
```
Usage 2 - Creating table from a table
```
table_dest = db_dest.copy_table(table_source)
```

Can be used to implement `LanceTable.copy(name)`
A bit conficted about calling this `import_table()` tbh. `copy_table()` makes it explicit that the table won't be altered.

TODO:
- [x] check tests
- [ ] Implement `table.copy()`

### DIFF
--- a/python/lancedb/db.py
+++ b/python/lancedb/db.py
@@ -256,7 +256,9 @@ class LanceDBConnection:
             tbl = LanceTable(self, name)
         return tbl
 
-    def copy_table(self, source: Union[str, Path, LanceTable], name: str = None) -> LanceTable:
+    def copy_table(
+        self, source: Union[str, Path, LanceTable], name: str = None
+    ) -> LanceTable:
         """Copy a table in the database.
 
         Parameters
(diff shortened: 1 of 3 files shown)


python/lancedb/db.py
```diff
@@ -255,6 +256,48 @@ def create_table(
             tbl = LanceTable(self, name)
         return tbl
 
+    def copy_table(
```
eddyxu, line 259: Should we implement this as `Database.create_table(existed_table | dataframe | arrow_table)`? 

Feel it can reduce the surface area of the one-off request.

AyushExel: actually based on the standup meeting, once the table versioning can be based on a namespace so copying will no longer be needed. Closing this

PR #217, opened 2023-06-22
[Vector Column] Specify and Index Column for Vector Search

philipk19238:


changhiskhan (commented):
This looks great thanks!

Would you mind:
1. Adding a test case where the vector column name is different from the default? (easiest probably to just copy one of the existing unit tests and adding a different column name?).
2. Run `black` on the code to pass linting

philipk19238:
Added - let me know if the test case is sufficient.

changhiskhan:
> Added - let me know if the test case is sufficient.

Thanks! Seems like the test is failing. Would mind taking a look?

philipk19238:
Needed to add vector_column_name to ```LanceFtsQueryBuilder``` and integrate it with all the test cases. Ran ```python3 -m pytest``` and passed all the test cases.

python/tests/test_query.py
```diff
@@ -48,30 +48,43 @@ def table(tmp_path) -> MockTable:
...
         abs=1e-6,
     )
     assert 0 <= df_cosine.score[0] <= 1
 
 
+def test_query_builder_with_different_vector_column(table):
+    query = [4, 8]
+    vector_column_name = "foo_vector"
+    builder = LanceQueryBuilder(table, query, vector_column_name)
+    assert builder._vector_column_name == vector_column_name
```
changhiskhan, line 86: my only nitpick here is that this tests the internal details rather than the API.
If the test could cover `create_index` and `search` that would be ideal

changhiskhan (commented):
Thanks! one very last comment and it should be good to go

changhiskhan:
oh also the CI failure is the doctests

philipk19238:
Should be ok now. I added updated test cases in ```test_query``` and ```test_table```. I also ran ```pytest --doctest-modules lancedb``` on my dev env and ensured it passed. Unsure if it will pass linux (I'm running M2 chip).

python/tests/test_query.py
```diff
@@ -48,30 +51,66 @@ def table(tmp_path) -> MockTable:
 
 
 def test_query_builder(table):
-    df = LanceQueryBuilder(table, [0, 0]).limit(1).select(["id"]).to_df()
+    df = LanceQueryBuilder(table, [0, 0], "vector").limit(1).select(["id"]).to_df()
```
eddyxu, line 54: shall we make `vector_column_name` to be a kwargs with default value, so that we don't break API like this.

philipk19238: agreed - added default

eddyxu (approved):
LGTM

Issue #216, opened 2023-06-21
open_table + create_table error implementation


unkn-wn:
Both `open_table()` and `create_table()` don't provide any errors when opening and creating empty/non-existent databases. The errors should show on function call instead of on the `add()` function. Windows, Python 3.8.10

> create_table()
```py
>>> tbl = db.create_table("hi")

>>> test
[{'id': 100, 'vector': [10, 10], 'meta': 'data', 'meta2': 'data2'}, {'id': 101, 'vector': [11, 11], 'meta': 'ATAD', 'meta2': '2ATAD'}]

>>> tbl.add(test)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/leony/.local/lib/python3.8/site-packages/lancedb/table.py", line 256, in add
    data = _sanitize_data(data, self.schema)
  File "/home/leony/.local/lib/python3.8/site-packages/lancedb/table.py", line 107, in schema
    return self._dataset.schema
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/home/leony/.local/lib/python3.8/site-packages/lancedb/table.py", line 234, in _dataset
    return lance.dataset(self._dataset_uri, version=self._version)
  File "/home/leony/.local/lib/python3.8/site-packages/lance/__init__.py", line 56, in dataset
    ds = LanceDataset(uri, version, block_size)
  File "/home/leony/.local/lib/python3.8/site-packages/lance/dataset.py", line 48, in __init__
    self._ds = _Dataset(uri, version, block_size, index_cache_size)
ValueError: Dataset at path home/leony/.lancedb/hi.lance/_latest.manifest was not found: Not found: home/leony/.lancedb/hi.lance/_latest.manifest
```

> open_table()
```py
>>> tbl = db.open_table("nonexistent")
>>> tbl.add(test)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/leony/.local/lib/python3.8/site-packages/lancedb/table.py", line 256, in add
    data = _sanitize_data(data, self.schema)
  File "/home/leony/.local/lib/python3.8/site-packages/lancedb/table.py", line 107, in schema
    return self._dataset.schema
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/home/leony/.local/lib/python3.8/site-packages/lancedb/table.py", line 234, in _dataset
[...]

PR #232, opened 2023-06-27
Error implementations

unkn-wn:
Solves #216 by adding a check on table open for existence of the `.lance` file. Does not check for it for remote connections.

For the readmes, @TevinWang and I were trying to test the UNIX setup, but it seems to only work sometimes? If someone else could also check the python development setup for the readme, that would be great.


python/README.md
```diff
@@ -0,0 +1,84 @@
...
+To run the unit tests:
+
+```bash
+pytest
+```
+
+To run linter and automatically fix all errors:
+
+```bash
+black .
```
chebbyChefNEQ, line 48: nit: also isore

python/lancedb/table.py
```diff
@@ -300,6 +300,19 @@ def create(cls, db, name, data, schema=None, mode="create"):
         lance.write_dataset(data, tbl._dataset_uri, mode=mode)
         return tbl
 
+    @classmethod
+    def open(cls, db, name):
+        tbl = LanceTable(db, name)
```
chebbyChefNEQ, line 305: instead we could do 
```python
cls(db, name)
```

This way child classes can use this method too without having to reimplement

chebbyChefNEQ (approved):

PR #237, opened 2023-06-29
bump lance to 0.5.2, make object store construction hook public

chebbyChefNEQ:
* bump to 0.5.2 to pick up S3 auth fixes
* make `dataset_read_params` a public attribute
* add `open_table_with_params` on `Database`

### DIFF
--- a/rust/vectordb/src/table.rs
+++ b/rust/vectordb/src/table.rs
@@ -43,7 +43,7 @@ impl std::fmt::Display for Table {
 
 #[derive(Default)]
 pub struct OpenTableParams {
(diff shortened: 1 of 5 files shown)


rust/vectordb/src/table.rs
```diff
@@ -43,7 +43,7 @@ impl std::fmt::Display for Table {
 
 #[derive(Default)]
 pub struct OpenTableParams {
-    pub(crate) dataset_read_params: ReadParams,
+    pub dataset_read_params: ReadParams,
```
eddyxu, line 46: do we want to make name consistent between "dataset_read_params" or `open_table_params`?

PR #239, opened 2023-06-29
Add dot product support

### LINKED ISSUE #207: Can LanceDB be used with dot product for a search metric?
From the docs, it looks like you only supports L2 or cosine metrics. Per the [sentence transformers documentation](https://www.sbert.net/docs/pretrained-models/msmarco-v3.html):

```Models tuned for cosine-similarity will prefer the retrieval of shorter passages, while models for dot-product will prefer the retrieval of longer passages.```

Also from that same page it looks like L2 & cosine are grouped together, separately from the dot product models, so I'm guessing L2 is also better for short retrieval than long retrieval.

Do you have any recommendations for using dot product as a metric in LanceDB?

### PR DESCRIPTION
**Pull request by eddyxu**
Closes #207

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -462,5 +462,10 @@ export enum MetricType {
   /**
    * Cosine distance
    */
-  Cosine = 'cosine'
+  Cosine = 'cosine',
+
+  /**
+   * Dot product
+   */
+  Dot = 'dot'
 }
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -17,7 +17,8 @@ Currently, we support the following metrics:
 | Metric      | Description                          |
 | ----------- | ------------------------------------ |
 | `L2`        | [Euclidean / L2 distance](https://en.wikipedia.org/wiki/Euclidean_distance) |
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: node/src/index.ts [hunk 1]
```diff
@@ -462,5 +462,10 @@ export enum MetricType {
   /**
    * Cosine distance
    */
-  Cosine = 'cosine'
+  Cosine = 'cosine',
+
+  /**
+   * Dot product
+   */
+  Dot = 'dot'
```
chebbyChefNEQ, line 470: remember to regen js docs

chebbyChefNEQ (approved):

albertovilla:
I think the file `lancecb/python/lancedb/query.py` needs to be updated to reflect that dot product is supported. Both the documentation of the method and its definition need to include "dot" as an option, otherwise you get errors like the following:

image

```python
def metric(self, metric: Literal["L2", "cosine"]) -> LanceVectorQueryBuilder:
        """Set the distance metric to use.

        Parameters
        ----------
        metric: "L2" or "cosine"
            The distance metric to use. By default "L2" is used.

        Returns
        -------
        LanceVectorQueryBuilder
            The LanceQueryBuilder object.
        """
        self._metric = metric
        return self
```

PR #241, opened 2023-07-02
Handle NaN input data

changhiskhan:
Sometimes LangChain would insert a single `[np.nan]` as a placeholder if the embedding function failed. This causes a problem for Lance format because then the array can't be stored as a FixedSizedListArray.

Instead:
1. By default we remove rows with embedding lengths less than the maximum length in the batch
2. If `strict=True` kwargs is set to True, then a `ValueError` is raised if the embeddings aren't all the same length

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 import os
 from functools import cached_property
-from typing import List, Union, Any
+from typing import Any, List, Union
 
 import lance
 import numpy as np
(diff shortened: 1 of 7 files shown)


python/lancedb/table.py
```diff
@@ -392,6 +408,18 @@ def _sanitize_vector_column(data: pa.Table, vector_column_name: str) -> pa.Table
         return data
     if not pa.types.is_list(vec_arr.type):
         raise TypeError(f"Unsupported vector column type: {vec_arr.type}")
+    if len(vec_arr.values) % len(data) != 0:
+        if strict_mode:
```
eddyxu, line 412: Is this right? Say if the input array has all `[NaN]`  here?

python/lancedb/table.py
```diff
@@ -392,6 +408,18 @@ def _sanitize_vector_column(data: pa.Table, vector_column_name: str) -> pa.Table
...
+            raise ValueError(
+                f"Vector column {vector_column_name} has variable length vectors"
+            )
+        # not a fixed size list, maybe some NaNs
+        lst_lengths = np.unique([len(lst) for lst in vec_arr])
+        ndims = np.max(lst_lengths)
+        mask = pa.array([len(lst) == ndims for lst in vec_arr])
```
AyushExel, line 419: @changhiskhan I'll just echo what said about also handling [nan]*len(ndims). Maybe also include a check for any nan.
```
ask = pa.array([len(lst) == ndims and not numpy.isnan(lst).any()for lst in vec_arr])
```

changhiskhan: I'd rather not introduce numba as dependency. benchmarking this locally, i get ~0.1ms for `np.isnan` on 100,000 element array. So this is unlikely to be a bottleneck

changhiskhan:
Note that the new solution covers way more than just the blocker raised by @AyushExel from LangChain

eddyxu (approved):

PR #251, opened 2023-07-05
NaN handling in TypeScript

changhiskhan:
I haven't changed the interface so currently it defaults to dropping rows that don't have the right number of dimensions or with NaNs.

i think this at least will preven the langchain integration from failing when openai errors out.

oh also need unit tests
i don't know what's the right way to expose these options without declaring a bazillion new interface variants tho

### DIFF
--- a/node/src/arrow.ts
+++ b/node/src/arrow.ts
@@ -15,48 +15,48 @@
 import {
   Field,
   Float32,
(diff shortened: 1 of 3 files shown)


changhiskhan:
not sure why this PR is not updating

Issue #245, opened 2023-07-03
[Doc] Dynamically create tsdoc markdown in CI


eddyxu:
So we dont need (or forget to) to check-in these generated markdowns from time to time.

PR #253, opened 2023-07-05
[docs] dynamic docs generation and deployment

unkn-wn:
Solves #245 , edited docs.yml to run the generation of docs before deployment. Tested on a test repository and it seems to work, but it takes a bit to build node.


changhiskhan (approved):
seems fine?

unkn-wn:
we could always revert if needed, so I’ll just merge

PR #254, opened 2023-07-05
Allow creation of an empty table

changhiskhan:
It's inconvenient to always require data at table creation time.
Here we enable you to create an empty table and add data and set schema later.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -284,20 +284,12 @@ class LanceTable:
             The number of vectors in the table.
         """
         # TODO: manage table listing and metadata separately
-        schema = None if not self._has_data() else self.schema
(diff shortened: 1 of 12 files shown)


python/lancedb/table.py
```diff
@@ -354,7 +357,7 @@ def create(
             The LanceDB instance to create the table in.
         name: str
             The name of the table to create.
-        data: list-of-dict, dict, pd.DataFrame
+        data: list-of-dict, dict, pd.DataFrame, default None
```
eddyxu, line 360: Mention here that at least one of data or schema need to be provided?

eddyxu (commented):
Nit w.r.t of comments

eddyxu (approved):

PR #260, opened 2023-07-07
[Python] Initial support of cloud API

eddyxu:
Support connect with remote database, and implement Search API


python/lancedb/remote/client.py
```diff
@@ -35,29 +35,32 @@ def wrapped(self, *args, **kwargs):
 
 @attr.define(slots=False)
 class RestfulLanceDBClient:
-    url: str
+    db_name: str
+    region: str
+    api_key: str
```
chebbyChefNEQ, line 40: we should make sure this field doesn't get accidentally printed somewhere.

something like 
```
class Password(str):
    def __repr__(self):
        return "****"

   def __str__(self):
        return "****"
```

eddyxu: can i just do __repr__ and __str__ on this Query object?

python/lancedb/remote/db.py
```diff
@@ -0,0 +1,68 @@
...
+        name: str
+            The name of the table.
+
+        Returns
+        -------
+        A LanceTable object representing the table.
+        """
+        from .table import RemoteTable
+
+        return RemoteTable(self, name)
```
chebbyChefNEQ, line 57: let's add a todo to check for existence?

eddyxu: Done.

python/lancedb/__init__.py
```diff
@@ -11,16 +11,24 @@
...
+from typing import Optional
 
+from .db import URI, DBConnection, LanceDBConnection
+from .remote.db import RemoteDBConnection
 
-def connect(uri: URI) -> LanceDBConnection:
-    """Connect to a LanceDB instance at the given URI
+
+def connect(
+    uri: URI, api_key: Optional[str] = None, region: str = "us-west-2"
```
chebbyChefNEQ, line 21: Nit: let's add a * after uri requiring args to be named

python/lancedb/remote/client.py
```diff
@@ -35,29 +36,32 @@ def wrapped(self, *args, **kwargs):
...
     def session(self) -> aiohttp.ClientSession:
-        parsed = urllib.parse.urlparse(self.url)
-        scheme = parsed.scheme
-        if not scheme.startswith("lancedb"):
-            raise ValueError(
-                f"Invalid scheme: {scheme}, must be like lancedb+<flavor>://"
-            )
-        flavor = scheme.split("+")[1]
-        url = f"{flavor}://{parsed.hostname}:{parsed.port}"
+        url = f"https://{self.db_name}.{self.region}.api.lancedb.com"
```
chebbyChefNEQ, line 46: Nit: add a plaintext argument where we use http over https

chebbyChefNEQ (approved):
Just two nits

python/lancedb/__init__.py
eddyxu: Fixed.

PR #270, opened 2023-07-10
[Docs] Add vectordb recipes as submodule

TevinWang:
- Building our website from examples from the vectordb-recipes repo in order to remove duplicate code. Note that this will change the url of our examples in the examples tab of the navigation to `lancedb.github.io/lancedb/examples/[example]/main`. Let me know if this causes any concerns.
- Changed our docs testing to exclude from globs so I can exclude examples and vectordb-recipes folder
- Add `mkdocs-exclude` package in order to exclude vectordb-recipes + copied over files that are not notebooks

Only removed python notebooks so far. Will add new examples as we port more + may add Javascript notebooks later so we can unduplicate javascript examples.


AyushExel:
Okay looks good. I think the node/examples/youtube_bot can also be removed as it is present in the recipes repo?
And have you verified that mkdocs.yaml doesn't need to be moved to top level, i.e, it can access & render the notebooks from that folder? 
Otherwise it looks good

TevinWang:
@AyushExel 
1. that would remove the explanations in the notebook though right? Since we only have a `js` file for the examples in the vectordb recipes repo.
2. I've tested it locally and it works, but not with the CI yet.

AyushExel (changes requested):
@TevinWang okay move mkdocs.yaml to top level, and don't move stuff during CI, we need to make sure users can build the docs using `mkdocs serve` to see the changes.
* Leave the existing js examples in the node/ folder
Otherwise this should be good. I'll take a look again once these changes are made.

PR #272, opened 2023-07-10
[Python] Schema to JSON 

eddyxu:


python/lancedb/schema.py
```diff
@@ -0,0 +1,253 @@
...
+
+"""Schema related utilities."""
+
+import json
+from typing import Any, Dict, Type
+
+import pyarrow as pa
+
+
+def vector(dimension: int, value_type: pa.DataType = pa.float32()) -> pa.DataType:
```
chebbyChefNEQ, line 22: nit: should we wrap vector value type to `lance.FloatType`? That would make supporting bf16 easier I suppose? @wjones127 wdyt?

wjones127: We'll have a `lance.arrow.BFloat16Type` that subclasses `pa.ExtensionType`, which should subclass `pa.DataType`. So I think this is fine as is.

python/lancedb/schema.py
```diff
@@ -0,0 +1,253 @@
...
+            "value_type": _type_to_dict(dt.value_type),
+        }
+    elif pa.types.is_struct(dt):
+        return {
+            "name": "struct",
+            "fields": [_field_to_dict(dt.field(i)) for i in range(dt.num_fields)],
+        }
+    # TODO: support extension types
+
+    raise TypeError(f"Unsupported type: {dt}")
```
wjones127, line 101: Should we add dictionary types?

python/lancedb/schema.py
```diff
@@ -0,0 +1,253 @@
...
+    ...     pa.field("id", pa.int64()),
+    ...     pa.field("vector", lancedb.vector(756)),
+    ... ])
+    """
+    return pa.list_(value_type, dimension)
+
+
+def _type_to_dict(dt: pa.DataType) -> Dict[str, Any]:
+    if pa.types.is_boolean(dt):
+        return {"name": "boolean"}
```
wjones127, line 50: nit: even if it looks redundant, I think `type` is more descriptive. Don't want it to collide with Field's `name`.
```suggestion
        return {"type": "boolean"}
```

python/lancedb/schema.py
```diff
@@ -0,0 +1,253 @@
...
+
+    raise TypeError(f"Unsupported type: {dt}")
+
+
+def _field_to_dict(field: pa.field) -> Dict[str, Any]:
+    return {
+        "name": field.name,
+        "type": _type_to_dict(field.type),
+        "nullable": field.nullable,
+        # "metadata": field.metadata,
```
wjones127, line 109: Any reason this is still commented out? (if so, provide a comment)

python/lancedb/schema.py
eddyxu: a field will look like

```json
{ "name": "blabl",  "type": {"type": "int32"}}
```

wdyt. Or are there other suggestion other than `name` and `type`?

python/lancedb/schema.py
eddyxu: Changed to be optional.

python/lancedb/schema.py
eddyxu: Oh yeah, definitely.

eddyxu: Done.

python/lancedb/schema.py
wjones127: I'm fine with that redundancy personally, but I admit it's a matter of taste. If we want to make them totally distinct we could use `datatype`.

chebbyChefNEQ (approved):

PR #197, opened 2023-06-15
feat(node): pull node binaries into separate packages (2)

gsilvestrin:
New Changes

* Changed the publish task to run after a release is published
* removed bump versions scripts since they are handled by another task
* removed organization (@lance) from packages

Previous Changes

* Refactors the Node module to load the shared library from a separate
package. When a user does `npm install vectordb`, the correct optional
dependency is automatically downloaded by npm.
* Brings Rust and Node versions in alignment at 0.1.2.
* Add scripts and instructions to build Linux and MacOS node artifacts
locally.
* Add instructions for publishing the npm module and crates.

PR #285, opened 2023-07-12
feat(node): pull node binaries into separate packages (3)

gsilvestrin:
This PR builds on top of the changes in https://github.com/lancedb/lancedb/pull/197. The problems we faced previously were with linux builds, the new changes are mostly in the `ci/build_linux_artifacts.sh` file:

- Build openssl statically
- Remove openblas / lapack dependencies
- Uses the rust Docker image for building the package instead of manylinux

I tested the generated packages in the following distros / versions:
- debian:bookworm (stable) - PASS
- fedora:38 (latest) - PASS
- ubuntu:22.04 (LTS) - FAIL, nodejs version is too old (12). Works after installing node 17 using nvm
- rockylinux:9 (latest) - red hat compatible - FAIL. ``GLIBC_2.35' not found`. It also fails with the current generated package, so there is no regression here.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -67,8 +67,12 @@ jobs:
     - name: Build
       run: |
         npm ci
-        npm run build
         npm run tsc
+        npm run build
+        npm run pack-build
+        npm install --no-save ./dist/vectordb-*.tgz
+        # Remove index.node to test with dependency installed
+        rm index.node
     - name: Test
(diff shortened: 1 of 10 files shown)


wjones127 (approved):
Looks like it will work. Thanks for testing on those platforms.

My only nit is that we might just want to cleanup the unused code paths in the GitHub action.

PR #291, opened 2023-07-12
[Python] Convert Pydantic Model to Arrow Schema

### LINKED ISSUE #256: [Pydantic integration] helper methods to convert a pydantic data model to LanceDB (arrow) schema

eddyxu:
Provide utility to automatically convert Pydantic model to Arrow Schema

Closes #256


python/lancedb/pydantic.py
```diff
@@ -0,0 +1,165 @@
...
+    >>> from lancedb.pydantic import vector
+    ...
+    >>> class MyModel(pydantic.BaseModel):
+    ...     vector: vector(756)
+    ...     id: int
+    ...     description: str
+    """
+
+    # TODO: make a public parameterized type.
+    class FixedSizeList(list, FixedSizeListMixin):
```
chebbyChefNEQ, line 59: nit: should we add a validation on length? making sure the underlying data is always the size of `dim`

eddyxu: Good catch. lemme add to it.

PR #298, opened 2023-07-13
feat(node): Fix npm publish task

gsilvestrin:
- npm understands subdirectories as scope names when publishing packages

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -132,6 +132,7 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.LANCEDB_NPM_REGISTRY_TOKEN }}
         run: |
-          for filename in */*.tgz; do
+          mv */*.tgz .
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -132,6 +132,7 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.LANCEDB_NPM_REGISTRY_TOKEN }}
         run: |
-          for filename in */*.tgz; do
+          mv */*.tgz .
```
wjones127, line 135: `*/*tgz` does seem a little sketchy. Shouldn't just copying `dist` be sufficient? or does the folder change for some reason?

```suggestion
          mv dist/*.tgz .
```

wjones127: Oh nevermind. Thanks for the explanation!

wjones127 (approved):

PR #305, opened 2023-07-14
manually install latest rust

changhiskhan:


changhiskhan:
Don't need this if we pin half

PR #313, opened 2023-07-15
[Python] Get table schema

eddyxu:


python/lancedb/remote/table.py
```diff
@@ -31,7 +32,12 @@ def __repr__(self) -> str:
         return f"RemoteTable({self._conn.db_name}.{self.name})"
 
     def schema(self) -> pa.Schema:
```
chebbyChefNEQ, line 34: let's cache this? we don't need one schema call per table version

python/lancedb/remote/db.py
```diff
@@ -75,4 +76,21 @@ def create_table(
...
+            if schema is None:
+                raise ValueError("Either data or schema must be provided")
+            data = pa.Table.from_pylist([], schema=schema)
+
+        payload = {
+            "name": name,
+            "schema": schema_to_json(data.schema),
+            "records": data.to_pydict(),
+        }
+        self._loop.run_until_complete(self._client.create_table("/table/", payload))
```
chebbyChefNEQ, line 95: nit: I think this breaks a notebook. Don't know how to fix it yet. Jupyter will complain about loop already running.

Maybe using a lazy cache_property would fix the problem?

chebbyChefNEQ (approved):

python/lancedb/remote/db.py
eddyxu: Lazy property of the loop object?

python/lancedb/remote/table.py
eddyxu: Done

python/lancedb/remote/db.py
chebbyChefNEQ: Nvm, just found the solution. https://stackoverflow.com/questions/46827007/runtimeerror-this-event-loop-is-already-running-in-python

PR #316, opened 2023-07-17
[Python] Create table with Iterator[RecordBatch] and add docs

eddyxu:


python/lancedb/db.py
```diff
@@ -38,8 +39,10 @@ def table_names(self) -> list[str]:
     def create_table(
         self,
         name: str,
-        data: DATA = None,
-        schema: pa.Schema = None,
+        data: Optional[
```
eddyxu, line 42: To make `mkdocs` looks good.

chebbyChefNEQ (approved):

PR #321, opened 2023-07-17
[Rust] Checkout a version of dataset.

eddyxu:
* `Table::open()` from absolute path, and gives the responsibility of organizing metadata out of Table object
* Fix Clippy warnings
* Add `Table::checkout(version)` API

### DIFF
--- a/rust/vectordb/src/index/vector.rs
+++ b/rust/vectordb/src/index/vector.rs
@@ -24,6 +24,7 @@ pub trait VectorIndexBuilder {
     fn get_replace(&self) -> bool;
 }
 
+#[derive(Default)]
 pub struct IvfPQIndexBuilder {
     column: Option<String>,
     index_name: Option<String>,
--- a/rust/vectordb/src/index/vector.rs
+++ b/rust/vectordb/src/index/vector.rs
@@ -24,7 +24,6 @@ pub trait VectorIndexBuilder {
     fn get_replace(&self) -> bool;
 }
 
-
 pub struct IvfPQIndexBuilder {
     column: Option<String>,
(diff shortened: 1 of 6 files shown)


rust/vectordb/src/index/vector.rs
```diff
@@ -24,6 +24,7 @@ pub trait VectorIndexBuilder {
     fn get_replace(&self) -> bool;
 }
 
+#[derive(Default)]
```
chebbyChefNEQ, line 27: I think the default of `bool` is `false` we might want to manually implement default here

rust/vectordb/src/table.rs
```diff
@@ -69,25 +63,42 @@ impl Table {
...
     /// # Returns
     ///
     /// * A [Table] object.
-    pub async fn open_with_params(
-        base_uri: &str,
-        name: &str,
-        params: OpenTableParams,
-    ) -> Result<Self> {
-        let path = Path::new(base_uri);
+    pub async fn open_with_params(uri: &str, name: &str, params: &ReadParams) -> Result<Self> {
```
chebbyChefNEQ, line 71: do we still need `name`?

eddyxu: Yea, this is for the table to have a better display name without leaking storage details. 

Might be able to deduct the name from uri if not provided, wdyt @chebbyChefNEQ ?

rust/vectordb/src/index/vector.rs
eddyxu: Fixed

chebbyChefNEQ (approved):

PR #326, opened 2023-07-18
[Bug] Fix dataset path check in Table::open

eddyxu:
It prevents to open remote dataset.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -534,7 +534,9 @@ class LanceTable(Table):
     @classmethod
     def open(cls, db, name):
         tbl = cls(db, name)
-        if not os.path.exists(tbl._dataset_uri):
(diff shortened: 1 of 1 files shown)


python/lancedb/table.py
```diff
@@ -534,7 +534,9 @@ def create(
     @classmethod
     def open(cls, db, name):
         tbl = cls(db, name)
-        if not os.path.exists(tbl._dataset_uri):
+        fs, path = pa.fs.FileSystem.from_uri(tbl._dataset_uri)
```
changhiskhan, line 537: Is this the only place we do this existence check? Should be extract a method and use it in all instances ?

eddyxu: This is the only place `path.exists()` is called. In current API, yes.

PR #339, opened 2023-07-19
fix: only use util function to build filesystem

wjones127:


eddyxu (approved):
LGTM.

Why we dont want to pass endpoint overload via https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow.fs.S3FileSystem  tho?

wjones127:
> Why we dont want to pass endpoint overload via https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow.fs.S3FileSystem tho?

We could pass it through there instead if we want. Do you think that would be better?

eddyxu:
My personal feeling is that pass via parameters is easy to reason, and we will have more user fine grainer control later. but up to you.

PR #353, opened 2023-07-23
dont print apikey in remote client toString, add hostoverride to python client

chebbyChefNEQ:


node/src/remote/client.ts
```diff
@@ -102,4 +102,11 @@ export class HttpLancedbClient {
     }
     return response
   }
+
+  toString (): string {
```
eddyxu, line 106: is this what `console.log(..)` uses?

python/lancedb/__init__.py
```diff
@@ -55,5 +60,9 @@ def connect(
     if isinstance(uri, str) and uri.startswith("db://"):
         if api_key is None:
             raise ValueError(f"api_key is required to connected LanceDB cloud: {uri}")
-        return RemoteDBConnection(uri, api_key, region)
+        return RemoteDBConnection(uri, api_key, region, host_override)
+    if host_override is not None:
+        warnings.warn(
+            "host_override has no effect when connecting to FS based DB", Warning
```
eddyxu, line 66: what is `fs based db`??  We don't need this warning, for example, api-key is not checked here as well.

eddyxu (approved):
Minor qq.

node/src/remote/client.ts
chebbyChefNEQ: I think so yeah

PR #358, opened 2023-07-24
[Docs]: Add badges, CTA and updates examples

AyushExel:
Screenshot 2023-07-24 at 6 13 00 PM
Is this a problem? Are we using metadata to track usage or something?


AyushExel:
@unkn-wn is there a way to exclude a particular integration doc test from CI? Fiftyone is pretty heavy so I think it doesn't make sense to test it? Or maybe we should. I'm not sure, let me know what would be better.

changhiskhan (approved):
nice thanks.

PR #365, opened 2023-07-25
bump lance version for vectordb, fix minor bugs in lancedb remote client

chebbyChefNEQ:


python/lancedb/remote/client.py
```diff
@@ -105,7 +105,7 @@ async def get(self, uri: str, params: Union[Dict[str, Any], BaseModel] = None):
     async def post(
         self,
         uri: str,
-        data: Union[Dict[str, Any], BaseModel, bytes],
+        data: Union[Dict[str, Any], BaseModel, bytes, None] = None,
```
eddyxu, line 108: Why do you need `Non` in the Union?

Can we just use `Optional[Union[...]]`?

python/lancedb/remote/db.py
```diff
@@ -103,7 +103,7 @@ def create_table(
 
         self._loop.run_until_complete(
             self._client.post(
-                f"/v1/table/{name}/create",
+                f"/v1/table/{name}/create/",
```
eddyxu, line 106: Would there be a  integration tests ?

eddyxu (approved):
Minor comments.

python/lancedb/remote/db.py
chebbyChefNEQ: yes, I caught in integration test.

PR #370, opened 2023-07-26
fix(node): Relax EmbeddingFunction type guard

### LINKED ISSUE #369: OpenAIEmbeddingFunction is not creating embedding
"vector column is missing" error when attempting to create a table using an OpenAI embedding function as exemplified here: https://github.com/lancedb/lancedb/blob/main/node/examples/js-openai/index.js

```bash
node index.js
/Users/eto/dev/lancedb/node/dist/index.js:112
            const tbl = yield tableCreate.call(...createArgs);
                                          ^

Error: column 'vector' is missing
    at LocalConnection.<anonymous> (/Users/eto/dev/lancedb/node/dist/index.js:112:43)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/eto/dev/lancedb/node/dist/index.js:18:58)

Node.js v20.3.1
```

gsilvestrin:
- It failed to match OpenAPIEmbeddings

Tested on top of main

```
OPENAI_API_KEY=sk-... node index.js
creating OpenAIEmbeddingFunction
[ 'vectors' ]
Leather Jacket
All done!
```

Closes #369

### DIFF
--- a/node/src/test/embedding/openai.ts
+++ b/node/src/test/embedding/openai.ts
@@ -16,6 +16,7 @@ import { describe } from 'mocha'
 import { assert } from 'chai'
 
 import { OpenAIEmbeddingFunction } from '../../embedding/openai'
+import { isEmbeddingFunction } from '../../embedding/embedding_function'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { OpenAIApi } = require('openai')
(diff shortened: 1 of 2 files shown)


node/src/test/embedding/openai.ts
```diff
@@ -47,4 +48,10 @@ describe('OpenAPIEmbeddings', function () {
       assert.deepEqual(vectors[1], stubValue.data.data[1].embedding)
     })
   })
+
+  describe('isEmbeddingFunction', function () {
+    it('should match the isEmbeddingFunction guard', function () {
+      assert.isTrue(isEmbeddingFunction(new OpenAIEmbeddingFunction('text', 'sk-key')))
```
eddyxu, line 54: How does this test the fix? is it still two attributes?

eddyxu: Is there a `isinstance()` type of call here? Can we rely on type system instead of check two arbitrary fields from the interface / type ?

wjones127 (approved):

PR #371, opened 2023-07-26
[WIP] Workflow to trigger vectordb-recipes workflow

unkn-wn:


unkn-wn:
The workflow works once the examples-test.yml is committed in vectordb-recipes

AyushExel (approved):

PR #372, opened 2023-07-26
fix(node): Handle overflows in the node bridge

### LINKED ISSUE #363: createIndex: concat requires input of at least one array
Improve the error message when creating an index with invalid parameters

```javascript
...
      await table.createIndex({ type: 'ivf_pq', column: 'vector', num_partitions: -1, max_iters: 2 })
```

```
Error: internal error in Neon module: called `Result::unwrap()` on an `Err` value: ComputeError("concat requires input of at least one array")
```

gsilvestrin:
- Fixes many numeric conversions that results in hard to reproduce issues
- JsObjectExt extends JsObject with safe methods to extract numeric values
- using the conv library for numeric conversions

Closes #363

### DIFF
--- a/rust/ffi/node/src/error.rs
+++ b/rust/ffi/node/src/error.rs
@@ -22,8 +22,15 @@ use snafu::Snafu;
 pub enum Error {
     #[snafu(display("column '{name}' is missing"))]
     MissingColumn { name: String },
+    #[snafu(display("{name}: {message}"))]
+    RangeError { name: String, message: String },
+    #[snafu(display("{index_type} is not a valid index type"))]
+    InvalidIndexType { index_type: String },
+
     #[snafu(display("{message}"))]
     LanceDB { message: String },
+    #[snafu(display("{message}"))]
+    Neon { message: String },
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
@@ -47,6 +54,14 @@ impl From<lance::Error> for Error {
 impl From<ArrowError> for Error {
     fn from(value: ArrowError) -> Self {
         Self::LanceDB {
+            message: value.to_string(),
+        }
+    }
+}
+
+impl From<neon::result::Throw> for Error {
+    fn from(value: neon::result::Throw) -> Self {
+        Self::Neon {
             message: value.to_string(),
         }
     }
--- a/rust/ffi/node/src/lib.rs
+++ b/rust/ffi/node/src/lib.rs
@@ -31,16 +31,17 @@ use once_cell::sync::OnceCell;
 use tokio::runtime::Runtime;
 
 use vectordb::database::Database;
-use vectordb::error::Error;
 use vectordb::table::{ReadParams, Table};
 
 use crate::arrow::{arrow_buffer_to_record_batch, record_batch_to_buffer};
 use crate::error::ResultExt;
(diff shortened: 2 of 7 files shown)


rust/ffi/node/src/neon_ext/js_object_ext.rs
```diff
@@ -0,0 +1,82 @@
...
+            message: "must be > 0".to_string(),
+        },
+        FloatError::PosOverflow(_) => Error::RangeError {
+            name: key.into(),
+            message: format!("must be < {}", u32::MAX),
+        },
+        FloatError::NotANumber(_) => Error::RangeError {
+            name: key.into(),
+            message: "not a valid number".to_string(),
+        },
```
wjones127, line 61: Does this mean a user can't store NaN values in the table? Or just that if the user inputs a number and it gets converted to NaN then that's an error?

Perhaps we should verify in a test that NaN, +Inf, and -Inf can be roundtripped through LanceDB node?

rust/ffi/node/src/error.rs
```diff
@@ -22,8 +22,15 @@ use snafu::Snafu;
 pub enum Error {
     #[snafu(display("column '{name}' is missing"))]
     MissingColumn { name: String },
+    #[snafu(display("{name}: {message}"))]
+    RangeError { name: String, message: String },
```
wjones127, line 26: Should range error convert to a Javascript RangeError? https://docs.rs/neon/latest/neon/context/trait.Context.html#method.throw_range_error

rust/ffi/node/src/lib.rs
```diff
@@ -245,12 +246,9 @@ fn table_search(mut cx: FunctionContext) -> JsResult<JsPromise> {
...
     let refine_factor = query_obj
-        .get_opt::<JsNumber, _, _>(&mut cx, "_refineFactor")?
-        .map(|s| s.value(&mut cx))
-        .map(|i| i as u32);
-    let nprobes = query_obj
-        .get::<JsNumber, _, _>(&mut cx, "_nprobes")?
-        .value(&mut cx) as usize;
+        .get_opt_u32(&mut cx, "_refineFactor")
+        .or_throw(&mut cx)?;
+    let nprobes = query_obj.get_usize(&mut cx, "_nprobes").or_throw(&mut cx)?;
```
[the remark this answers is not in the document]
wjones127: I just wish Neon did more of this conversion work for us, similar to PyO3 :(

wjones127 (changes requested):
Seems like this will help the conversions a lot, but want to double check that we are still letting special float values through.

rust/ffi/node/src/neon_ext/js_object_ext.rs
wjones127: Ah thanks for clarifying. 👍

rust/ffi/node/src/error.rs
wjones127: I see. That's because `ResultExt` is generic over `E`, so you can't match on that specific variant.

I guess the only thing you could do is write a special `or_throw` method just for `crate::Error`. But it might not be worth it; the error types in JS seems so limited I doubt people are really caring about the type.

wjones127 (approved):

PR #373, opened 2023-07-27
fix(python): Pin pylance and add pandas as test dependency

gsilvestrin:


wjones127 (approved):
Thanks for fixing this!

PR #374, opened 2023-07-28
wip: search: Make vector optional

gsilvestrin:
## Current State

Currently we only expose one API to extract data from LanceDB, search by vector:

```typescript
const results = await table.search([0.1, 0.1]).execute()
```

We would like to allow users to search by any attributes / metadata, leveraging the support Lance already implements for that.

## Proposed Solution

Create a new method `Table.query(..)` that servers as a starting point to all search types supported by LanceDB (table scan, vector search, full text search).

### Table Scan - Python

```python
results = table.query() \
  .select(['id', 'name', 'price', 'is_active']) \
  .where('id > 2 and id < 10 and is_active = true')
  .to_df()
```

### Table Scan - Node.js

```javascript
const results = await table.query()
  .select(['id', 'name', 'price', 'is_active'])
  .filter('id > 2 and id < 10 and is_active = true')
  .execute()
```

### Vector Search - Python

2) Create a more detailed API for vector search and  and keep the existing `search` as it is

```python
results = table.query(VectorSearchBuilder([0.1, 0.2])) \
    .select(['id', 'name', 'price', 'is_active'])\
    .where('id > 2 and id < 10 and is_active = true')\
    .to_df()

... or with more ann params...

vector_search = VectorSearchBuilder([0.1, 0.2])\
  .nprobes(10)\
  .refine_factor(10)

results = table.query(vector_search) \
        .select(['id', 'name', 'price', 'is_active']) \
        .where('id > 2 and id < 10 and is_active = true')
```

### Vector Search - Node.js - Option 1

We can follow the Python API here

```typescript
const vectorSearch = new VectorSearch([0.1, 0.2])
  .nprobes(10)
  .refineFactor(45)

const results = await table.query(vectorSearch)
  .select(['id', 'name', 'price', 'is_active'])
  .filter('id > 2 and id < 10 and is_active = true')
  .execute()
```

### Vector Search - Node.js - Option 2

Use Options object instead of a builder

```typescript
const results = await table.query()
  .vectorSearchOption2({ query: [0.1, 0.2] })
  .select(['id', 'name', 'price', 'is_active'])
  .filter('id > 2 and id < 10 and is_active = true')
  .execute()
```

## Other Alternatives

The simplest solution is to make vector optional. For instance this query returns all elements that match the `filter` predicate:

```typescript
const results = await table.search()  
  .select(['id', 'name', 'price', 'is_active'])  
  .filter('id > 2 and id < 10 and is_active = true')  
  .execute()
```

An empty `filter` predicate returns all elements of the Table. We can keep the existing API interface `search(vector)` so that the change is backwards compatible.

Making the search vector optional can make the API confusing because some parameters (`nprobes`,  `refine_factor`, ...)  only make sense when paired with a vector. The list of such parameters will grow as we support more index types.

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -152,7 +152,7 @@ export interface Table<T = number[]> {
    * Creates a search query to find the nearest neighbors of the given search term
    * @param query The query search term
    */
-  search: (query: T) => Query<T>
+  search: (query?: T) => Query<T>
 
   /**
    * Insert records into this Table.
@@ -338,7 +338,7 @@ export class LocalTable<T = number[]> implements Table<T> {
    * Creates a search query to find the nearest neighbors of the given search term
    * @param query The query search term
    */
-  search (query: T): Query<T> {
+  search (query?: T): Query<T> {
     return new Query(query, this._tbl, this._embeddings)
   }
 
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -152,7 +152,9 @@ export interface Table<T = number[]> {
    * Creates a search query to find the nearest neighbors of the given search term
    * @param query The query search term
    */
-  search: (query?: T) => Query<T>
+  search: (query: T) => Query<T>
+
+  query: (params?: undefined | VectorSearch<T>) => Query<T>
 
   /**
    * Insert records into this Table.
@@ -309,6 +311,46 @@ export class LocalConnection implements Connection {
   }
 }
 
+export class VectorSearch<T = number[]> {
+  private readonly _query?: T
+  _refineFactor?: number
+  _nprobes: number
+  _metricType?: MetricType
+
+  constructor (readonly query: T) {
+    this._nprobes = 20
+    this._refineFactor = undefined
+    this._metricType = undefined
+  }
+
+  /**
+   * Refine the results by reading extra elements and re-ranking them in memory.
+   * @param value refine factor to use in this query.
+   */
+  refineFactor (value: number): VectorSearch<T> {
+    this._refineFactor = value
+    return this
+  }
+
+  /**
+   * The number of probes used. A higher number makes search more accurate but also slower.
+   * @param value The number of probes used.
+   */
+  nprobes (value: number): VectorSearch<T> {
+    this._nprobes = value
+    return this
+  }
+
+  /**
+   * The MetricType used for this Query.
+   * @param value The metric to the. @see MetricType for the different options
+   */
+  metricType (value: MetricType): VectorSearch<T> {
+    this._metricType = value
+    return this
+  }
+}
+
 export class LocalTable<T = number[]> implements Table<T> {
   private readonly _tbl: any
   private readonly _name: string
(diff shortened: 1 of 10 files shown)


wjones127:
ANN/KNN search and table scans are two different starting points, so I think it would be reasonable to separate them into two different methods. However, once we add full text search this starts to feel a bit messier, so it might not be a good idea in the long term.

In the end, I think my preference would be having a generic `query()` method that when passed nothing means a table scan, when passed a `KNNSearch` object means KNN search, and when passed a `FullTextSearch` object means full text search. This maintains the idea that a scan, KNN search, and FTS search are all different starting points, but that we can still layer in the same methods after to complete the query.

For backwards compatibility, `table.search()` could dispatch to `table.query()`. I think the rename is necessary because "search" isn't the right verb for table scans IMO.

```js
// Full scan
const results = await table.query()  
  .select(['id', 'name', 'price', 'is_active'])  
  .filter('id > 2 and id < 10 and is_active = true')  
  .execute()

// KNN Search
const results = await table.query(new KNNSearch([0.1, 0.1]))  
  .select(['id', 'name', 'price', 'is_active'])  
  .filter('id > 2 and id < 10 and is_active = true')  
  .execute()


// Full text search
const results = await table.query(new FullTextSearch("donkey kong"))  
  .select(['id', 'name', 'price', 'is_active'])  
  .filter('id > 2 and id < 10 and is_active = true')  
  .execute()
```

What do you think of that?

wjones127:
> focus too much on how LanceDB works internally

I agree users don't need to know the internals of how LanceDB works, but they do need to have a mental model for what a query *means*. For example, someone writing SQL should know the difference between an inner join and a left join; this doesn't mean they need to know how a hash join works or anything like that, but they need to know what the expected results should look like in order to be able to write queries. So similarly, our API needs to make it easy for our users to reason about what a query will return.

I think it's helpful to have the methods ordered in the order that they are run. The search part happens at the start, so it should be put at the beginning of the query builder. I don't know what it would mean to put ANN search in the middle of a pipeline, and I think that would be confusing to people.

Hybrid search is an interesting point. I think conceptually that will be a new search type:

```js
// Full text search
const search = new HybridSearch(new KNNSearch([0.1, 0.1]), new FullTextSearch("donkey kong"));
const results = await table.query(search)  
  .select(['id', 'name', 'price', 'is_active'])  
  .filter('id > 2 and id < 10 and is_active = true')  
  .execute()
```

wjones127:
IMO, I think this would work well:

```js
const results = await table.query({
  search: lancedb.vector_search([0.1, 0.1]),
  select: ['id', 'name', 'price', 'is_active'],
  filter: 'id > 2 and id < 10 and is_active = true'
}).to_records()
```

```python
results = table.query(
  search=lancedb.vector_search([0.1, 0.1]),
  select=['id', 'name', 'price', 'is_active'],
  filter='id > 2 and id < 10 and is_active = true'
).to_df()
```

wjones127:
Actually, @eddyxu  you might like this better: separate the `vector_search` parameter and the `search_options`. Then in most cases, users can just pass the search value, assuming we can tune the search options well enough ourselves.

```js
const results = await table.query({
  vector_search: [0.1, 0.1],
  select: ['id', 'name', 'price', 'is_active'],
  filter: 'id > 2 and id < 10 and is_active = true',
  search_options: { nprobes: 10 }
}).to_records()
```

```python
results = table.query(
    vector_search=[0.1, 0.1],
    select=['id', 'name', 'price', 'is_active'],
    filter='id > 2 and id < 10 and is_active = true',
    search_options={'nprobes': 10},
).to_df()
```

PR #376, opened 2023-07-29
feat(node): Improve concurrency

gsilvestrin:
- Moved computation out of JS main thread by using a mpsc inspired by [this example](https://github.com/neon-bindings/examples/blob/main/examples/async-sqlite/src/lib.rs)
- Removes the Arc/Mutex since Table is owned by JsTable now
- Parallelism is not affected - JsTable process one message at a time, in the same way the Mutex locks the Table now.
   - We may need to look into a mpmc implementation such as crossbeam_channel in the future
- Moved table / query methods to their own files 
- Fixed js-transformers example

### DIFF
--- a/rust/ffi/node/src/table.rs
+++ b/rust/ffi/node/src/table.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
 use std::sync::mpsc;
 use std::thread;
 use arrow_array::RecordBatchIterator;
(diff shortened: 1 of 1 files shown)


unkn-wn:
Could you also change the transformersjs notebook in docs/src/examples file as well?

rust/ffi/node/src/table.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    pub(crate) fn js_delete(mut cx: FunctionContext) -> JsResult<JsPromise> {
+        let js_table = cx.this().downcast_or_throw::<JsBox<JsTable>, _>(&mut cx)?;
+        let rt = runtime(&mut cx)?;
+        let (deferred, promise) = cx.promise();
+        let predicate = cx.argument::<JsString>(0)?.value(&mut cx);
+
+        js_table
+            .send(deferred, move |table, channel, deferred| {
```
wjones127, line 210: Food for thought: Can we eliminate more of this boilerplate? It seems like all that is essential is the argument extraction and the call to `.delete()`. So it would be nice if the body was just something like:

```rust
let predicate: String = cx.argument::<JsString>(0)?.value(&mut cx);
extract_table(cx).call(move |table| table.delete(&predicate))
```

PR #381, opened 2023-07-30
[Docs] add Tables guide

AyushExel:
* Rename "Reference" -> "Guides" to create distinction b/w api reference and user facing docs
* Add all the various ways to create, add and delete from table

Related - https://github.com/lancedb/lancedb/pull/391

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -76,26 +76,26 @@ A Table is a collection of Records in a LanceDB Database.
     It is recommended to use RecordBatch itertator to add large datasets in batches when creating your table in one go. This does not create multiple versions of your dataset unlike manually adding batches using `table.add()`
 
     ```python
-      import pyarrow as pa
+    import pyarrow as pa
 
-      def make_batches():
-          for i in range(5):
-              yield pa.RecordBatch.from_arrays(
-                  [
-                      pa.array([[3.1, 4.1], [5.9, 26.5]]),
-                      pa.array(["foo", "bar"]),
-                      pa.array([10.0, 20.0]),
-                  ],
-                  ["vector", "item", "price"],
-              )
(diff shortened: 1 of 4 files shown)


eddyxu (approved):

docs/src/guides/tables.md
```diff
@@ -0,0 +1,352 @@
...
+    schema = pa.schema([
+        pa.field("vector", pa.list_(pa.float32())),
+        pa.field("item", pa.utf8()),
+        pa.field("price", pa.float32()),
+    ])
+
+    db.create_table("table4", make_batches(), schema=schema)
+    ```
+
+    You can also use Pandas dataframe directly in the above example by converting it to `RecordBatch` object
```
eddyxu, line 116: iirc, we can directly write pandas `DataFrame` ,and do conversion internally.

eddyxu (approved):
LGTM overall

docs/src/guides/tables.md
AyushExel: @eddyxu No it doesn't work right now. It should be simple to make it work though. I have a branch for that but I thought let's get this in first.. 
To repro, this should throw a malloc error:
```
import lancedb
import pyarrow as pa
import pandas as pd

db = lancedb.connect("/tmp")

def make_batches_df():
    for i in range(5):
        yield  pd.DataFrame({
            "vector": [[3.1, 4.1], [1, 1]],
            "item": ["foo", "bar"],
            "price": [10.0, 20.0],
        })
schema = pa.schema([
    pa.field("vector", pa.list_(pa.float32(), 2)),
    pa.field("item", pa.utf8()),
    pa.field("price", pa.float32()),
])
tbl = db.create_table("table", make_batches_df(), schema=schema, mode="overwrite")
```

PR #384, opened 2023-07-31
Improve pydantic integration

changhiskhan:
See the documentation changes for details

### DIFF
--- a/python/lancedb/pydantic.py
+++ b/python/lancedb/pydantic.py
@@ -252,7 +252,26 @@ def pydantic_to_schema(model: Type[pydantic.BaseModel]) -> pa.Schema:
 
 
 class LanceModel(pydantic.BaseModel):
-    """A Pydantic Model that can be converted to a LanceDB Table."""
+    """
+    A Pydantic Model base class that can be converted to a LanceDB Table.
+
+    Examples
+    --------
+    >>> import lancedb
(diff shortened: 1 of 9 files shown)


python/lancedb/pydantic.py
```diff
@@ -249,3 +249,17 @@ def pydantic_to_schema(model: Type[pydantic.BaseModel]) -> pa.Schema:
     """
     fields = _pydantic_model_to_fields(model)
     return pa.schema(fields)
+
+
+class LanceModel(pydantic.BaseModel):
+    """A Pydantic Model that can be converted to a LanceDB Table."""
```
eddyxu, line 255: So user exepct to inherent this class or take this as a Mixin?
Can we add an explicit example for that?

python/lancedb/query.py
```diff
@@ -230,6 +231,23 @@ def to_arrow(self) -> pa.Table:
         )
         return self._table._execute_query(query)
 
+    def to(self, model: Type[LanceModel]) -> List[LanceModel]:
```
eddyxu, line 234: can we rename this to `to_pydantic` , `to` is too wide to be tight with one specific library.

changhiskhan: done

eddyxu (approved):

PR #385, opened 2023-07-31
make pandas an optional dependency in lancedb as well

changhiskhan:
reduce installation footprint for size conscious applications

### DIFF
--- a/python/lancedb/embeddings.py
+++ b/python/lancedb/embeddings.py
@@ -16,15 +16,22 @@ import sys
 from typing import Callable, Union
 
 import numpy as np
-import pandas as pd
 import pyarrow as pa
 from lance.vector import vec_to_table
 from retry import retry
 
+from .util import safe_import_pandas
+
+pd = safe_import_pandas()
+if pd is not None:
+    DATA = Union[pa.Table, pd.DataFrame]
+else:
+    DATA = pa.Table
+
 
 def with_embeddings(
     func: Callable,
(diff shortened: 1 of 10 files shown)


python/lancedb/query.py
```diff
@@ -198,7 +200,7 @@ def refine_factor(self, refine_factor: int) -> LanceQueryBuilder:
         self._refine_factor = refine_factor
         return self
 
-    def to_df(self) -> pd.DataFrame:
+    def to_df(self):
```
wjones127, line 203: It would be nice to keep this type hint.
```suggestion
    def to_df(self) -> "pd.DataFrame":
```

python/lancedb/embeddings.py
```diff
@@ -16,15 +16,22 @@
...
 from lance.vector import vec_to_table
 from retry import retry
 
+from .util import safe_import_pandas
+
+pd = safe_import_pandas()
+if pd is not None:
+    DATA = Union[pa.Table, pd.DataFrame]
+else:
+    DATA = pa.Table
```
wjones127, line 29: Would this not work fine? Or would it failing type hint checks on downstream apps if they didn't have Pandas installed?
```suggestion
DATA = Union[pa.Table, "pd.DataFrame"]
```

wjones127 (commented):
Two questions on some type hints

wjones127 (approved):

PR #390, opened 2023-08-01
version to versioninfo question

bkodes:
is there an older version of lancedb that used a semver version between (>=2.10.2,<3.0.0) ? I am using launchdarkly and they have that constraint on their semver versioning


wjones127:
FYI we removed semver in May of last year: https://github.com/lancedb/lancedb/pull/1311

PR #396, opened 2023-08-02
fix(node) Unit tests hangs and don't exit

gsilvestrin:
Sometimes the tests hang because the table reference's were not garbage collect. This PR adds a method to force the Table's underlying thread to exit, like [this example](https://github.com/neon-bindings/examples/blob/main/examples/async-sqlite/src/lib.rs#L121)

It also adds `--exit` to the mocha cli so the node vm can exit immediately after the tests are completed: https://mochajs.org/#-exit)


wjones127 (approved):
This looks like a reasonable fix for now.

I posted a future idea in https://github.com/lancedb/lancedb/issues/397.

PR #399, opened 2023-08-05
feat(node): Create empty tables / Arrow Tables

gsilvestrin:
- Supports creating an empty table as long as an Arrow Schema is provided
- Supports creating a table from an Arrow Table (can be passed as data)
- Using an options object to pass createTable params, the plan is to eventually deprecate the positional parameters
- Simplified some Arrow code in the TS/FFI side
- removed createTableArrow method, it was never documented / tested.

### DIFF
--- a/node/src/remote/index.ts
+++ b/node/src/remote/index.ts
@@ -14,7 +14,7 @@
 
 import {
   type EmbeddingFunction, type Table, type VectorIndexParams, type Connection,
-  type ConnectionOptions
+  type ConnectionOptions, type CreateTableOptions, type WriteOptions
 } from '../index'
 import { Query } from '../query'
 
@@ -66,7 +66,7 @@ export class RemoteConnection implements Connection {
     }
   }
 
-  async createTable<T> (...args: any[]): Promise<Table<T>> {
+  async createTable<T> (name: string | CreateTableOptions<T>, data?: Array<Record<string, unknown>>, optsOrEmbedding?: WriteOptions | EmbeddingFunction<T>, opt?: WriteOptions): Promise<Table<T>> {
     throw new Error('Not implemented')
   }
 
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -118,7 +118,7 @@ export interface Connection {
    * Creates a new Table, optionally initializing it with new data.
    *
    * @param {string} name - The name of the table.
-   * @param data - Non-empty Array of Records to be inserted into the table
+   * @param data - Array of Records to be inserted into the table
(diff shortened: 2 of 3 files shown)


node/src/arrow.ts
```diff
@@ -70,16 +68,39 @@ export async function convertToTable<T> (data: Array<Record<string, unknown>>, e
...
-  const children = new Field<Float32>('item', new Float32())
-  const list = new List(children)
+function newVectorListBuilder (dim: number): FixedSizeListBuilder<Float32> {
   return makeBuilder({
-    type: list
+    type: newVectorListType(dim)
   })
 }
 
+function newVectorListType (dim: number): FixedSizeList<Float32> {
```
eddyxu, line 77: Just `newVectorType`?  vector = list

node/src/arrow.ts
```diff
@@ -70,16 +68,39 @@ export async function convertToTable<T> (data: Array<Record<string, unknown>>, e
...
+
+    const vectors = await embeddings.embed(source.toArray() as T[])
+    const column = vectorFromArray(vectors, newVectorListType(vectors[0].length))
+    table = table.assign(new ArrowTable({ vector: column }))
+  }
+  const writer = RecordBatchFileWriter.writeAll(table)
+  return Buffer.from(await writer.toUint8Array())
+}
+
+export function createEmptyTable (schema: Schema): Table {
```
eddyxu, line 104: This is in memory arrow table. not a lancedb table? 

Can you add some comments for the functions in this file.

node/src/remote/index.ts
```diff
@@ -66,13 +66,7 @@ export class RemoteConnection implements Connection {
...
   }
 
-  async createTable (name: string, data: Array<Record<string, unknown>>): Promise<Table>
-  async createTable<T> (name: string, data: Array<Record<string, unknown>>, embeddings: EmbeddingFunction<T>): Promise<Table<T>>
-  async createTable<T> (name: string, data: Array<Record<string, unknown>>, embeddings?: EmbeddingFunction<T>): Promise<Table<T>> {
-    throw new Error('Not implemented')
-  }
-
-  async createTableArrow (name: string, table: ArrowTable): Promise<Table> {
+  async createTable<T> (...args: any[]): Promise<Table<T>> {
```
eddyxu, line 69: This signature does not match with the interface?

eddyxu: You dont use `Object[]` as signature in java / scala right?

wjones127 (approved):
Looks good, although would be nice to support empty array if schema is passed.

Issue #397, opened 2023-08-03
[node] Create a scheduler handler


wjones127:
1. Create a struct `TaskScheduler`. A global instance is lazily created when the package loads.
2. To run an async task, pass a future and an `on_completion` callback. The future represents the work to do, while the callback handles how to attach the result to `deferred` object.
3. On the "exit" nodejs event, call `.close()` on the `TaskScheduler`

Here's a rough sketch of `TaskScheduler`
```rust
struct TaskScheduler {
    runtime: Runtime
}

impl TaskScheduler {
    pub fn submit<T>(
        &self,
        deferred: Deferred,
        fut: Future<Output = T>,
        on_completion: FnMut(T, Deferred) -> ()
    ) {
        let wrapped_fut = async move {
            /// Spawn the future separately so we can handle the crash through the join handle.
            let result = tokio::task::spawn(fut).await;
            match result {
                Ok(result) => on_completion(result, deferred);,
                Err(JoinError) => {
                    deferred.settle_with(&channel, move |mut cx| {
                        todo!("handle join error")
                    });
                }
            }
            on_completion(result, deferred);
        };
        self.runtime.spawn(wrapped_fut);
    }

    pub fn close(&self) {
        self.runtime.shutdown_background();
    }
}
```

PR #407, opened 2023-08-07
fix(node) Remove mpsc from JS SDK

### LINKED ISSUE #405: NodeJS SDK hangs on index creation
```js
(async () => {

     const vectordb = require('vectordb')
     const db = await vectordb.connect('data/sample-lancedb')

     let data = []
     for (let i = 0; i < 10_000; i++) {
         data.push({vector: Array(1536).fill(i), id: `${i}`, content: "", longId: `${i}`},)
     }
     const table = await db.createTable('my_vectors', data)
     await table.createIndex({ type: 'ivf_pq', column: 'vector', num_partitions: 256, num_sub_vectors: 96 })


     const results_1 = await table
         .search(Array(1536).fill(1.2))
         .limit(2)
         .nprobes(20)
         .refineFactor(10)
         .execute()


     const results_2 = await table
         .search(Array(1536).fill(1.2))
         .where("id != '1141'")
         .execute()


     const results_3 = await table
         .search(Array(1536).fill(1.2))
         .select(["id"])
         .execute()

})();
```
SDK version: Main branch
`await table.createIndex({ type: 'ivf_pq', column: 'vector', num_partitions: 256, num_sub_vectors: 96 })` hangs with no output

gsilvestrin:
- remove &mut operations from Table
- Callers / SDKs are responsible for keeping track of the last version of the Table
   - This matches what we are doing in the Python side
- This allow us to remove the mpsc from Table and make all Table operations non-blocking
- remove close function from js sdk, it was never included as part of a public release
- There is a different approach proposed [here](https://github.com/lancedb/lancedb/issues/397), but I think we need to evolve the design first and integrate in the main rust crate vs being a js-only solution

Close #405

### DIFF
--- a/node/examples/js/index.js
+++ b/node/examples/js/index.js
@@ -12,38 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-(async () => {
+'use strict'
 
-  const vectordb = require('vectordb')
-  const db = await vectordb.connect('data/sample-lancedb')
+async function example () {
+  const lancedb = require('vectordb')
+  const db = await lancedb.connect('data/sample-lancedb')
 
-  let data = []
-  for (let i = 0; i < 10_000; i++) {
-    data.push({vector: Array(1536).fill(i), id: `${i}`, content: "", longId: `${i}`},)
-  }
-  const table = await db.createTable('my_vectors', data)
+  const data = [
+    { id: 1, vector: [0.1, 0.2], price: 10 },
+    { id: 2, vector: [1.1, 1.2], price: 50 }
+  ]
 
-  console.log('Create Index')
-  await table.createIndex({ type: 'ivf_pq', column: 'vector', num_partitions: 256, num_sub_vectors: 96 })
+  const table = await db.createTable('vectors', data)
+  console.log(await db.tableNames())
 
-  console.log('results 1')
-  const results_1 = await table
-      .search(Array(1536).fill(1.2))
-      .limit(2)
-      .nprobes(20)
-      .refineFactor(10)
+  const results = await table
+      .search([0.1, 0.3])
+      .limit(20)
       .execute()
+  console.log(results)
+}
 
-  console.log('results 2')
-  const results_2 = await table
-      .search(Array(1536).fill(1.2))
-      .where("id != '1141'")
-      .execute()
-
-  console.log('results 3')
-  const results_3 = await table
-      .search(Array(1536).fill(1.2))
-      .select(["id"])
-      .execute()
-
-})();
+example()


node/src/index.ts
```diff
@@ -390,7 +384,7 @@ export class LocalTable<T = number[]> implements Table<T> {
    * @param indexParams The parameters of this Index, @see VectorIndexParams.
    */
   async createIndex (indexParams: VectorIndexParams): Promise<any> {
-    return tableCreateVectorIndex.call(this._tbl, indexParams)
+    return tableCreateVectorIndex.call(this._tbl, indexParams).then((newTable: any) => { this._tbl = newTable })
```
wjones127, line 387: Okay so IIUC, how this works is:

```ts
// Initializes table
const table = lance.table(...) 
// Clones the table to perform operation in another thread
const create_index_promise = table.create_index(...) 
// Another clone to perform this operation
const delete = table.delete(...) 
// Either create_index or delete will win and overwrite the table reference
await Promise.all([create_index, delete])
```

This is fine for now, since we'll disallow concurrent writes for the time being. But eventually we probably want a better model for updating the table reference after a mutation, since this is a race condition. I'll see if I can create a method soon to update a dataset to the latest version, which is probably the better way to solve this.

rust/vectordb/src/table.rs
```diff
@@ -187,7 +187,7 @@ impl Table {
     }
 
     /// Create index on the table.
-    pub async fn create_index(&mut self, index_builder: &impl VectorIndexBuilder) -> Result<()> {
+    pub async fn create_index(&self, index_builder: &impl VectorIndexBuilder) -> Result<Table> {
```
wjones127, line 190: If you make this take `&self` instead of `&mut self`, you are making the Rust and the Node API different. Node users get to write:

```ts
await table.delete("id = 42")
```

And their table is updated for them. Rust users would now have to remember to use the returned value:

```rust
let table = table.delete("id = 42").await?;
```

This seems unnecessary. I'd keep the Rust methods using `&mut` and change the `Arc<Table>` in Javascript to be `Table`. `Table` is cheap to clone as-is, so no need for it to be in an `Arc`.

rust/ffi/node/src/table.rs
```diff
@@ -12,82 +12,33 @@
...
-use crate::{get_aws_creds, JsDatabase, runtime};
-
-type TableCallback = Box<dyn FnOnce(&mut vectordb::Table, &Channel, Deferred) + Send>;
+use crate::error::ResultExt;
+use crate::{get_aws_creds, runtime, JsDatabase};
 
-// Wraps a LanceDB table into a channel, allowing concurrent access
 pub(crate) struct JsTable {
-    tx: mpsc::Sender<JsTableMessage>,
+    pub table: Arc<Table>,
```
wjones127, line 29: As suggested in my other comment, if you don't use an `Arc` here it should be easier to clone and get an owned reference that can be mutated in the Rust API.

```suggestion
    pub table: Table,
```

node/src/index.ts
wjones127: https://github.com/lancedb/lance/issues/1124

wjones127 (changes requested):
The Node API changes seem good but the ones to Rust seem unnecessary. IMO we should be consistent across APIs about mutating table.

wjones127 (approved):

PR #411, opened 2023-08-08
implement remote drop table call

chebbyChefNEQ:
Also moves `request_id` to header instead of request param

### DIFF
--- a/python/lancedb/remote/client.py
+++ b/python/lancedb/remote/client.py
@@ -97,7 +97,12 @@ class RestfulLanceDBClient:
         """Send a GET request and returns the deserialized response payload."""
         if isinstance(params, BaseModel):
             params: Dict[str, Any] = params.dict(exclude_none=True)
-        async with self.session.get(uri, params=params, headers=self.headers) as resp:
+        async with self.session.get(
+            uri,
+            params=params,
+            headers=self.headers,
+            timeout=aiohttp.ClientTimeout(total=30),
+        ) as resp:
             await self._check_status(resp)
             return await resp.json()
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-lance = "=0.5.9"
+lance = "=0.6.0"
```
eddyxu, line 9: 🔥

python/lancedb/remote/client.py
```diff
@@ -109,6 +109,7 @@ async def post(
         params: Optional[Dict[str, Any]] = None,
         content_type: Optional[str] = None,
         deserialize: Callable = lambda resp: resp.json(),
+        request_id: Optional[str] = None,
```
eddyxu, line 112: why is that

python/lancedb/remote/db.py
```diff
@@ -105,8 +105,22 @@ def create_table(
...
+
+    def drop_table(self, name: str):
+        """Drop a table from the database.
+
+        Parameters
+        ----------
+        name: str
+            The name of the table.
+        """
+        self._loop.run_until_complete(
```
eddyxu, line 122: There is no timeout?

Cargo.toml
chebbyChefNEQ: oh no, this is an accident. will remove

python/lancedb/remote/client.py
chebbyChefNEQ: we already have request id processing from header. Our middleware will use client provided request id instead of generating one if the client has set `x-request-id`

python/lancedb/remote/db.py
chebbyChefNEQ: added

eddyxu (approved):
LGTM. Thanks!

Do we need to do it on javascript side?

PR #414, opened 2023-08-08
[breaking change] make schema a property

chebbyChefNEQ:
Both existing implementations, `LanceTable` and `RemoteTable`, of the `Table` abstract class implements `schema` as a property. This PR aligns the interface and implementation.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -93,9 +93,10 @@ class Table(ABC):
     [Table.create_index][lancedb.table.Table.create_index].
     """
 
+    @property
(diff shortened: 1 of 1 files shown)


python/lancedb/table.py
```diff
@@ -93,6 +93,7 @@ class Table(ABC):
     [Table.create_index][lancedb.table.Table.create_index].
     """
 
+    @property
```
eddyxu, line 96: Do we need doc changes

eddyxu (approved):

Issue #416, opened 2023-08-09
Implementation of commands for deleting all tables / whole database at once


ashiskumarnaik:
methods like drop_database or delete_all_tables command can be implemented.

PR #418, opened 2023-08-09
implementation of drop_database

ashiskumarnaik:
#416 Fixed.

added drop_database() method . This deletes all the tables from the database with a single command.

### DIFF
--- a/python/lancedb/db.py
+++ b/python/lancedb/db.py
@@ -192,6 +192,13 @@ class DBConnection(ABC):
         """
         raise NotImplementedError
 
+    def drop_database(self):
+        """
+        Drop database
+        This is the same thing as dropping all the tables
+        """
+        raise NotImplementedError
+
 
 class LanceDBConnection(DBConnection):
     """
@@ -333,3 +340,7 @@ class LanceDBConnection(DBConnection):
         except FileNotFoundError:
             if not ignore_missing:
                 raise
+
+    def drop_database(self):
+        filesystem, path = fs_from_uri(self.uri)
+        filesystem.delete_dir(path)
--- a/python/tests/test_db.py
+++ b/python/tests/test_db.py
@@ -159,6 +159,47 @@ def test_delete_table(tmp_path):
     db.drop_table("does_not_exist", ignore_missing=True)
 
 
+def test_drop_database(tmp_path):
+    db = lancedb.connect(tmp_path)
+    data = pd.DataFrame(
+        {
+            "vector": [[3.1, 4.1], [5.9, 26.5]],
+            "item": ["foo", "bar"],
+            "price": [10.0, 20.0],
+        }
+    )
+    new_data = pd.DataFrame(
+        {
+            "vector": [[5.1, 4.1], [5.9, 10.5]],
+            "item": ["kiwi", "avocado"],
+            "price": [12.0, 17.0],
+        }
+    )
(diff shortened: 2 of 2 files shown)


python/lancedb/db.py
```diff
@@ -333,3 +341,11 @@ def drop_table(self, name: str, ignore_missing: bool = False):
         except FileNotFoundError:
             if not ignore_missing:
                 raise
+
+    def drop_database(self):
+        try:
+            filesystem, path = fs_from_uri(self.uri)
+            shutil.rmtree(path)
```
eddyxu, line 348: We would like to use `filesystem` to run deletion.  pyarrow filesystem can support cloud storages, like, S3 or GCS.

python/lancedb/db.py
```diff
@@ -333,3 +341,11 @@ def drop_table(self, name: str, ignore_missing: bool = False):
...
+            print(f"Deleted all tables of the database stored in {path}")
```
eddyxu, line 349: Also please dont use print in production code.

python/lancedb/db.py
```diff
@@ -333,3 +341,11 @@ def drop_table(self, name: str, ignore_missing: bool = False):
...
+        except Exception as e:
+            print("Error ", e)
```
eddyxu, line 351: Let's not catch exception here, and let the caller to handle exception.

eddyxu (changes requested):

python/tests/test_db.py
```diff
@@ -159,6 +159,43 @@ def test_delete_table(tmp_path):
...
+
+    db.drop_database()
+    assert db.table_names() == []
+
+    db.create_table("test", data=new_data)
+    db.drop_table("test")
+    assert db.table_names() == []
+
+    # dropping a empty database should pass
+    db.drop_database()
```
westonpace, line 195: Would you expect two consecutive calls to `drop_database` to pass?  I think that will fail because `delete_dir` expects the directory to exist.

python/lancedb/db.py
```diff
@@ -192,6 +192,13 @@ def drop_table(self, name: str):
         """
         raise NotImplementedError
 
+    def drop_database(self):
+        """
+        Drop all the tables at once
+        or delete the database at once
```
westonpace, line 198:
```suggestion
        """Drop all tables
```

The phrasing with the `or` is a little confusing to me.  I think you are trying to say "this does X or Y since both X and Y are really the same idea"

However, I read this as "this can do X or Y" and I am left wondering how to figure out what flag controls this (i.e. I'm looking for a "operation_type" enum or something)

If it's important to communicate that these are the same thing you could do something like...

```
"""Delete the database

This is the same thing as dropping all of the tables
```

The "at once" phrasing is redundant I think and a user might wonder if it means that the operation is atomic / thread safe.  This operation is not "atomic". If a user calls `table_names` while a `drop_database` call is in progress then they might see a partial result.

westonpace (approved):
A few minor suggestions but this looks right to me.

python/lancedb/db.py
ashiskumarnaik: No, I wanted to check if there are no tables in the directory . i.e. if the user has deleted all the tables but  he called drop_database.

eddyxu (approved):
Thanks @ashiskumarnaik !

PR #423, opened 2023-08-11
feat(node) Use fetch library for http requests

gsilvestrin:
- Added fetch bindings generated from OpenAPI spec (everything under `node/src/generated-sources` was auto-generated
- Refactored existing remote calls 
- Added simple remote test script, disabled by default

Some things will need to be improved
- There is no timeout support in fetch, will need to look into https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static
- Automatically apply `ErrorHandler()` to handle error calls


o-alexandrov:
Timeout can be added using this[ SO answer](https://stackoverflow.com/a/79263202/4122857):

```ts
 const response = await fetch(URL_HERE, {
    // ...rest
    signal: AbortSignal.timeout(60 * 1000), // <=== HERE
  });
  ```

PR #424, opened 2023-08-11
fix(node) Electron crashes when creating external buffer

### LINKED ISSUE #382: [Typescript] Vector search crashes in Electron environment without clear message
From discord

> When I run identical code in typescript and then the same code in electron, populating the database works perfectly as does getting length, adding, etc.
For some reason query—for identical code—does not work in the electron development environment and crashes with no error logging, whereas in a plain typescript file it works perfectly

```
const results = await this.tbl.search(query_vector).limit(limit).execute();
```

> query_vector is an embedding, limit is error-handled to never be more than the rowCount of the db. So those runtime errors are avoided

gsilvestrin:
Output of `https://github.com/TevinWang/electron-quick-start-typescript/`

```
> electron-quick-start-typescript@1.0.0 start
> npm run build && electron ./dist/main.js


> electron-quick-start-typescript@1.0.0 build
> tsc

[
  '18',      '20', '29',
  '32',      '35', '39',
  '45',      '48', '53',
  '58',      '59', '69',
  '88',      '93', '97',
  'vectors'
]
[
  {
    vector: Float32Array(2) [ 0.10000000149011612, 0.20000000298023224 ],
    id: 1,
    price: 10,
    score: 0.010000001639127731
  },
  {
    vector: Float32Array(2) [ 1.100000023841858, 1.2000000476837158 ],
    id: 2,
    price: 50,
    score: 1.8100000619888306
  }
]
All done!
```

Close #382

### DIFF
--- a/node/src/query.ts
+++ b/node/src/query.ts
@@ -112,7 +112,8 @@ export class Query<T = number[]> {
       this._queryVector = this._query as number[]
     }
 
-    const buffer = await tableSearch.call(this._tbl, this)
+    const isElectron = this.isElectron()
+    const buffer = await tableSearch.call(this._tbl, this, isElectron)
     const data = tableFromIPC(buffer)
 
     return data.toArray().map((entry: Record<string, unknown>) => {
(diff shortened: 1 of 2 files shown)


node/src/query.ts
```diff
@@ -127,4 +128,14 @@ export class Query<T = number[]> {
       return newObject as unknown as T
     })
   }
+
+  // See https://github.com/electron/electron/issues/2288
+  private isElectron (): boolean {
+    try {
+      // eslint-disable-next-line no-prototype-builtins
+      return (process?.versions?.hasOwnProperty('electron') || navigator?.userAgent?.toLowerCase()?.includes(' electron'))
```
TevinWang, line 136: Small question here - why do we need to specify two conditions? Seems like userAgent one is already working from the issue?

TevinWang (approved):
LGTM! Tested and works.

node/src/query.ts
TevinWang: Oh I see, LGTM

PR #441, opened 2023-08-22
expose awsRegion to be configurable

chebbyChefNEQ:


Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-lance = "=0.6.2"
+lance = { path = "../lance/rust" }
```
chebbyChefNEQ, line 9: wil fix after 0.6.3 releases on lance

rust/ffi/node/src/lib.rs
```diff
@@ -156,6 +159,19 @@ fn get_aws_creds<T>(
...
+/// Get AWS region arguments from the context
+fn get_aws_region<T>(cx: &mut FunctionContext, arg_location: i32) -> Result<Option<String>, NeonResult<T>> {
+    let region = cx
+        .argument_opt(arg_location)
+        .map(|arg| arg.downcast_or_throw::<JsString, FunctionContext>(cx));
+
+    match region {
+        Some(Ok(region)) => Ok(Some(region.value(cx))),
+        None => Ok(None),
+        Some(Err(e)) => Err(Err(e))
```
westonpace, line 171: You're returning `Err(Err(e))`.  What is this?  I would have thought the return would have been `Err(e)`.

rust/ffi/node/src/table.rs
```diff
@@ -66,10 +66,15 @@ impl JsTable {
             Ok(creds) => creds,
             Err(err) => return err,
         };
+        let aws_region = match get_aws_region(&mut cx, 6) {
```
westonpace, line 69: Is this not:

```
let aws_region = get_aws_region(&mut cx, 6)?;
```

node/src/index.ts
```diff
@@ -51,6 +53,24 @@ export interface ConnectionOptions {
...
 
+function appendAwsCreds (
+  callArgs: any[],
+  awsCredentials?: AwsCredentials,
+  awsRegion?: string
+): void {
+  if (awsCredentials !== undefined) {
+    callArgs.push(awsCredentials.accessKeyId)
+    callArgs.push(awsCredentials.secretKey)
+    callArgs.push(awsCredentials.sessionToken)
```
westonpace, line 64: Was the removal of the `if (awsCredentials.sessionToken !== undefined)` check intentional?

westonpace (commented):
A few minor questions but I don't see anything obviously wrong

node/src/index.ts
chebbyChefNEQ: yes, not pushing and pushing `undefined` gets us the same behavior rust side. We need to start pushing `undefined` because we need to use the position to identify the argument. We probably need to do some refactoring with arg passing across FFI as the current pattern is pretty messy.

rust/ffi/node/src/lib.rs
chebbyChefNEQ: Yeah, I think there might be ways to simplify this. Let me try a few things.

rust/ffi/node/src/table.rs
chebbyChefNEQ: yes, fixed

node/src/index.ts
```diff
@@ -258,14 +278,8 @@ export class LocalConnection implements Connection {
...
     const callArgs = [this._db, name]
-    const awsCredentials = this._options().awsCredentials
-    if (awsCredentials !== undefined) {
-      callArgs.push(awsCredentials.accessKeyId)
-      callArgs.push(awsCredentials.secretKey)
-      if (awsCredentials.sessionToken !== undefined) {
-        callArgs.push(awsCredentials.sessionToken)
-      }
-    }
+    appendAwsCreds(callArgs, this._options().awsCredentials, this._options().awsRegion)
```
changhiskhan: why have side-effects as part of the method as opposed to having the call site here concatenate the callArgs and the output?

rust/ffi/node/src/lib.rs
```diff
@@ -152,7 +154,21 @@ fn get_aws_creds<T>(
             }),
         ))),
         (None, None, None) => Ok(None),
-        _ => Err(cx.throw_error("Invalid credentials configuration")),
+        _ => cx.throw_error("Invalid credentials configuration"),
+    }
+}
+
+/// Get AWS region arguments from the context
+fn get_aws_region(cx: &mut FunctionContext, arg_location: i32) -> NeonResult<Option<String>> {
```
changhiskhan, line 162: why does this return NeonResult but get_aws_creds only return Option?

changhiskhan (commented):
minor questions but otherwise looks good

node/src/index.ts
chebbyChefNEQ: ah, already fixed this. Passing array back and expanding them now

changhiskhan (approved):

mclean-connor:
@changhiskhan these changes don't seem to work. what was your input to the connect() method?

mclean-connor:
looks like it actually works for other methods. the problem seems to be with the tableNames method

/**
     * Get the names of all tables in the database.
     */
    tableNames() {
        return __awaiter(this, void 0, void 0, function* () {
            return databaseTableNames.call(this._db);
        });
    }

I think the getAwsArgs should be passed

mclean-connor:
looks like the following rust needs to accept was creds are args

```rust
fn database_table_names(mut cx: FunctionContext) -> JsResult<JsPromise> {
    let db = cx
        .this()
        .downcast_or_throw::<JsBox<JsDatabase>, _>(&mut cx)?;

    let rt = runtime(&mut cx)?;
    let (deferred, promise) = cx.promise();
    let channel = cx.channel();
    let database = db.database.clone();

    rt.spawn(async move {
        let tables_rst = database.table_names().await;

        deferred.settle_with(&channel, move |mut cx| {
            let tables = tables_rst.or_throw(&mut cx)?;
            let table_names = convert::vec_str_to_array(&tables, &mut cx);
            table_names
        });
    });
    Ok(promise)
}

fn database_drop_table(mut cx: FunctionContext) -> JsResult<JsPromise> {
    let db = cx
        .this()
        .downcast_or_throw::<JsBox<JsDatabase>, _>(&mut cx)?;
    let table_name = cx.argument::<JsString>(0)?.value(&mut cx);

    let rt = runtime(&mut cx)?;
    let channel = cx.channel();
    let database = db.database.clone();

    let (deferred, promise) = cx.promise();
    rt.spawn(async move {
        let result = database.drop_table(&table_name).await;
        deferred.settle_with(&channel, move |mut cx| {
            result.or_throw(&mut cx)?;
            Ok(cx.null())
        });
    });
    Ok(promise)
}

PR #451, opened 2023-08-23
[python] improve restore functionality

changhiskhan:
Previously the temporary restore feature required copying data. The new feature in pylance does not.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -311,7 +311,7 @@ class LanceTable(Table):
 
         This allows viewing previous versions of the table. If you wish to
         keep writing to the dataset starting from an old version, then use
-        the `restore` function instead.
+        the `restore` function.
 
         Parameters
         ----------
@@ -341,16 +341,18 @@ class LanceTable(Table):
             raise ValueError(f"Invalid version {version}")
         self._reset_dataset(version=version)
 
-    def restore(self, version: int):
+    def restore(self, version: int = None):
         """Restore a version of the table. This is an in-place operation.
 
         This creates a new version where the data is equivalent to the
-        specified previous version. Note that this creates a new snapshot.
+        specified previous version. Data is not copied (as of python-v0.2.1).
 
         Parameters
         ----------
-        version : int
-            The version to restore.
+        version : int, default None
+            The version to restore. If unspecified then restores the currently
+            checkedout version. If the currently checkedout version is the latest
+            version then this is a no-op.
 
         Examples
         --------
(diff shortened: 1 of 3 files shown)


python/lancedb/table.py
```diff
@@ -341,16 +341,18 @@ def checkout(self, version: int):
             raise ValueError(f"Invalid version {version}")
         self._reset_dataset(version=version)
 
-    def restore(self, version: int):
+    def restore(self, version: int = None):
         """Restore a version of the table. This is an in-place operation.
 
         This creates a new version where the data is equivalent to the
-        specified previous version. Note that this creates a new snapshot.
+        specified previous version. Data is not copied (as of python-v0.2.1).
```
westonpace, line 348:
```suggestion
        specified previous version. Data is not copied (as of lancedb-v0.2.1).
```

python/lancedb/table.py
```diff
@@ -341,16 +341,18 @@ def checkout(self, version: int):
...
 
         Parameters
         ----------
-        version : int
-            The version to restore.
+        version : int, default None
+            The version to restore. If unspecified then restores the currently
+            checkedout version. If the currently checkedout version is the latest
+            version then this is a no-op.
```
westonpace, line 355:
```suggestion
            checked out version. If the currently checked out version is the
            latest version then this is a no-op.
```
I'm not used to seeing "checkedout" as one word but if its a convention feel free to ignore this.

python/lancedb/table.py
```diff
@@ -373,15 +375,21 @@ def restore(self, version: int):
         3
         """
         max_ver = max([v["version"] for v in self._dataset.versions()])
+        if version is None:
+            version = self.version
+
         if version < 1 or version >= max_ver:
             raise ValueError(f"Invalid version {version}")
+        if version is None:
+            version = self.version
```
westonpace, line 384: Looks like you have this check twice (here and on 378).  Is it needed in both spots?

python/tests/test_table.py
```diff
@@ -280,3 +280,9 @@ def test_restore(db):
     table.restore(1)
     assert len(table.list_versions()) == 3
     assert len(table) == 1
+
+    expected = table.to_arrow()
+    table.checkout(1)
+    table.restore()
+    assert len(table.list_versions()) == 4
+    assert table.to_arrow() == expected
```
westonpace, line 288: Maybe add some test cases for the no-op restore (`list_versions` should have the same length before and after) and the default restore?

python/lancedb/table.py
```diff
@@ -373,15 +375,21 @@ def restore(self, version: int):
         3
         """
         max_ver = max([v["version"] for v in self._dataset.versions()])
+        if version is None:
+            version = self.version
+
         if version < 1 or version >= max_ver:
```
westonpace, line 381: Do you intend for this to be `>=`?  Isn't `version == max_ver` supposed to be a no-op (and not an error)?

westonpace (changes requested):

python/lancedb/table.py
changhiskhan: oh that's the anticipated tag name. We prefix it with `python-` to distinguish it from nodejs package releases.

python/lancedb/table.py
changhiskhan: oh that's a total typo. thanks for flagging it

python/lancedb/table.py
changhiskhan: good catch let me add a unit test

python/lancedb/table.py
changhiskhan: forgive me father for i have sinned. On this day in the year 2023 of our lord, i committed the grave sin of over-reliance on github copilot.

python/tests/test_table.py
changhiskhan: will do

westonpace (approved):

PR #452, opened 2023-08-24
[python] Support schema evolution in local LanceDB

changhiskhan:
Previously if you needed to add a column to a table you'd have to rewrite the whole table. Instead,
we use the merge functionality from Lance format
to incrementally add columns from another table
or dataframe.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -17,13 +17,14 @@ import inspect
 import os
 from abc import ABC, abstractmethod
 from functools import cached_property
-from typing import Iterable, List, Union
+from typing import Iterable, List, Optional, Union
 
 import lance
 import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc
 from lance import LanceDataset
+from lance.dataset import ReaderLike
 from lance.vector import vec_to_table
 
 from .common import DATA, VEC, VECTOR_COLUMN_NAME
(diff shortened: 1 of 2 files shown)


python/lancedb/table.py
```diff
@@ -500,6 +501,65 @@ def add(
...
+        >>> table.merge(new_df, 'x')
+        >>> table.to_pandas()
+           x  y  z
+        0  1  a  d
+        1  2  b  e
+        2  3  c  f
+        """
+        if isinstance(schema, LanceModel):
+            schema = schema.to_arrow_schema()
+        if isinstance(other_table, LanceTable):
```
eddyxu, line 556: is that other type like pd.DataFrame handled by `dataset.merge()`?

python/lancedb/table.py
```diff
@@ -500,6 +501,65 @@ def add(
...
+    def merge(
+        self,
+        other_table: Union[LanceTable, ReaderLike],
+        left_on: str,
+        right_on: Optional[str] = None,
+        schema: Optional[pa.Schema, LanceModel] = None,
+    ):
+        """Merge another table into this table.
+
+        Performs a left join, where the dataset is the left side and data_obj
```
westonpace, line 513:
```suggestion
        Performs a left join, where the dataset is the left side and other_table
```

python/lancedb/table.py
```diff
@@ -500,6 +501,65 @@ def add(
...
+        is the right side. Rows existing in the dataset but not on the left will
+        be filled with null values, unless Lance doesn't support null values for
+        some types, in which case an error will be raised.
```
westonpace, line 516:
```suggestion
        is the right side. Rows existing in the dataset that do not match any value
        in other_table will be filled with null values, unless Lance doesn't support
        null values for some types, in which case an error will be raised.
```

Which types don't support null values?

python/lancedb/table.py
```diff
@@ -500,6 +501,65 @@ def add(
...
+        other_table: Union[LanceTable, ReaderLike],
+        left_on: str,
+        right_on: Optional[str] = None,
+        schema: Optional[pa.Schema, LanceModel] = None,
+    ):
+        """Merge another table into this table.
+
+        Performs a left join, where the dataset is the left side and data_obj
+        is the right side. Rows existing in the dataset but not on the left will
+        be filled with null values, unless Lance doesn't support null values for
```
westonpace, line 515: What happens if a row in the dataset matches multiple rows in `other_table`?  In a typical `left join` I would expect to get one output row per match but I assume that can't happen here (since we aren't actually rewriting the left side).

westonpace (commented):
I didn't realize we supported this.  I'll have to dig in at some point to see how this is implemented :)

python/lancedb/table.py
wjones127: Only String and List support nulls right now. All others don't support them.

python/lancedb/table.py
wjones127: We collect the right side into a hashmap, so the values on the RHS are de-depulicated.

python/lancedb/table.py
changhiskhan: yup. All of those types are grouped under `ReaderLike` (see the method type signature).

eddyxu (approved):

PR #457, opened 2023-08-29
[python] Temporary update feature

changhiskhan:
Combine delete and append to make a temporary update feature that is only enabled for the local python lancedb.

The reason why this is temporary is because it first has to load the data that matches the where clause into memory, which is technical unbounded.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -717,7 +717,6 @@ class LanceTable(Table):
         orig_data = self._dataset.to_table(filter=where).combine_chunks()
         if len(orig_data) == 0:
             return
-        self.delete(where)
         for col, val in values.items():
             i = orig_data.column_names.index(col)
             if i < 0:
@@ -725,6 +724,7 @@ class LanceTable(Table):
             orig_data = orig_data.set_column(
                 i, col, pa.array([val] * len(orig_data), type=orig_data[col].type)
             )
+        self.delete(where)
         self.add(orig_data, mode="append")
         self._reset_dataset()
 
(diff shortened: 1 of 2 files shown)


python/lancedb/table.py
```diff
@@ -678,6 +678,56 @@ def open(cls, db, name):
...
+        if len(orig_data) == 0:
+            return
+        self.delete(where)
+        for col, val in values.items():
+            i = orig_data.column_names.index(col)
+            if i < 0:
+                raise ValueError(f"Column {col} does not exist")
+            orig_data = orig_data.set_column(
+                i, col, pa.array([val] * len(orig_data), type=orig_data[col].type)
+            )
```
wjones127, line 727: Could we call `self.delete(where)` *after* we compute the columns, in case there is an error during the compute step? Otherwise the user won't be able to rerun this without first calling restore.

changhiskhan: done

wjones127 (approved):

westonpace (approved):
This will be helpful for prototyping.  There's potentially some weird concurrency / isolation effects here but given it's a temporary feature I think that is expected?

For example, if someone calls delete at the same time someone else is calling append you could get:

1. Delete Thread -> Read Data Where ... (would not include appended data)
2. Append Thread -> Write Data
3. Delete Thread -> Delete Data Where ... (would also delete appended data)
4. Delete Thread -> Write Data

This could lead to data loss.  Maybe just add a comment saying that update is experimental and should not be run at the same time as other operations?

PR #467, opened 2023-09-01
[python] Use pydantic for embedding function persistence

changhiskhan:
1. Support persistent embedding function so users can just search using query string
2. Add fixed size list conversion for multiple vector columns
3. Add support for empty query (just apply select/where/limit).
4. Refactor and simplify some of the data prep code


python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,192 @@
...
+            then the class name is used.
+        """
+
+        # This is a decorator for a class that inherits from BaseModel
+        # It adds the class to the registry
+        def decorator(cls):
+            if not issubclass(cls, EmbeddingFunctionModel):
+                raise TypeError("Must be a subclass of EmbeddingFunctionModel")
+            cls.__function_alias__ = alias or cls.__name__
+            self._functions[alias or cls.__name__] = cls
```
AyushExel, line 50: Should it warn or reject when overriding a key( if someone accidentally overrides it with another function, it'll break for the existing table).
Maybe either just warn
Or
Throw exception and ask for explicit "overwrite" if they actually want to overwrite?

changhiskhan: I may end up just removing the alias for now to simplify things. But good point. For now we can just raise an error on conflict.

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,183 @@
...
+        elif isinstance(texts, pa.ChunkedArray):
+            texts = texts.combine_chunks().to_pylist()
+        return self.embedding_model.encode(
+            list(texts),
+            convert_to_numpy=True,
+            normalize_embeddings=self.normalize,
+        ).tolist()
+
+    @classmethod
+    @cached(cache={})
```
westonpace, line 162: How big is a sentence transformer model?  Do we need to file a future issue to maybe control / monitor the size of this cache?

westonpace (commented):
A few thoughts from a quick pass

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,183 @@
...
+    @property
+    def embedding_model(self):
+        """
+        Get the sentence-transformers embedding model specified by the
+        name and device. This is cached so that the model is only loaded
+        once per process.
+        """
+        return self.__class__.get_embedding_model(self.name, self.device)
+
+    def __call__(self, texts: Union[str, List[str]]) -> List[np.array]:
```
wjones127, line 140: nit: the `texts` typehint doesn't match the type checks below.

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,183 @@
...
+    vector_column: str
+
+    @abstractmethod
+    def __call__(self, *args, **kwargs):
+        pass
+
+
+@REGISTRY.register()
+class SentenceTransformerEmbeddingFunction(EmbeddingFunctionModel):
+    name: str = "all-MiniLM-L6-v2"
```
wjones127, line 127: This is an example that is registered by default?

python/lancedb/query.py
```diff
@@ -54,46 +55,115 @@ class Query(pydantic.BaseModel):
...
-    ...       .where("b < 10")
-    ...       .select(["b"])
-    ...       .limit(2)
-    ...       .to_df())
-       b      vector  _distance
-    0  6  [0.4, 0.4]        0.0
-    """
+        if isinstance(query, str):
+            # fts
+            return LanceFtsQueryBuilder(table, query)
```
wjones127, line 76: The docstring said:

> If the query is a string, then the query type is "vector" if the
> table has embedding functions else the query type is "fts"

But I don't see that reflected here. Am I missing something?

wjones127 (changes requested):
A few questions

python/lancedb/query.py
changhiskhan: validate_query can change the query type . Should be renamed resolve_query type probably

python/lancedb/embeddings/functions.py
changhiskhan: This is the first of a set of "default" functions. The name variable is the model name ; sentence-transformers can be configured with a list of different models.

In a follow on PR we will add OpenAI and OpenClip embedding functions as well.

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,183 @@
...
+
+REGISTRY = EmbeddingFunctionRegistry()
+
+
+class EmbeddingFunctionModel(BaseModel, ABC):
+    """
+    A callable ABC for embedding functions
+    """
+
+    source_column: Optional[str]
```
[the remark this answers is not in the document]
changhiskhan: I can definitely come up with ways like you have a title column and a description column and you want to embed the concat of the two. But in practice I haven't seen it used. In addition if we try to handle multiple, we will also need to have some way to combine the columns which gets messy. I think if the use case is complicated, chances are the user will just handle a lot of the stuff themselves.

python/lancedb/conftest.py
```diff
@@ -14,3 +17,22 @@ def doctest_setup(monkeypatch, tmpdir):
...
+
+@registry.register()
+class MockEmbeddingFunction(EmbeddingFunctionModel):
+    def __call__(self, data):
+        if isinstance(data, str):
+            data = [data]
+        elif isinstance(data, pa.ChunkedArray):
+            data = data.combine_chunks().to_pylist()
+        elif isinstance(data, pa.Array):
+            data = data.to_pylist()
```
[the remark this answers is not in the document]
changhiskhan: Yeah I can add like a sanitize input method in the superclass that gets call automatically

python/lancedb/table.py
```diff
@@ -28,46 +28,78 @@
...
-        data = _sanitize_schema(
-            data, schema=schema, on_bad_vectors=on_bad_vectors, fill_value=fill_value
-        )
-    if isinstance(data, dict):
+    elif isinstance(data, dict):
         data = vec_to_table(data)
-    if pd is not None and isinstance(data, pd.DataFrame):
+    elif pd is not None and isinstance(data, pd.DataFrame):
         data = pa.Table.from_pandas(data, preserve_index=False)
+        # Do not serialize Pandas metadata
```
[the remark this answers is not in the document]
changhiskhan: I forgot the GitHub issue that precipitated this but because we do some type conversions in arrow before we write to Lance, the pandas metadata would cause exceptions when converting back to pandas, possibly due to conflicts between the retrieved arrow type and pandas metadata.

Conceptually LanceDB isn't a container for data frames so there's no guarantee of round trip.

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    @classmethod
+    def get_instance(cls):
+        return REGISTRY
+
+    def __init__(self):
+        self._functions = {}
+
+    def register(self):
```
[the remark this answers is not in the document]
changhiskhan: I was planning to add a separate docs focused PR to keep this one from getting too bloated. I can add it here too if you feel strongly about it

python/lancedb/table.py
```diff
@@ -325,14 +367,14 @@ def checkout(self, version: int):
...
+        2
         >>> table.to_pandas()
                vector    type
         0  [1.1, 0.9]  vector
         >>> table.add([{"vector": [0.5, 0.2], "type": "vector"}])
         >>> table.version
-        2
-        >>> table.checkout(1)
+        3
+        >>> table.checkout(2)
```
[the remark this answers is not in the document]
changhiskhan: Yeah I did a slight refactor if LanceTable.create so that the version history looks the same regardless of whether you added data in the beginning or not. Also this way the schema/metadata manipulation paths are more consolidated

python/lancedb/table.py
```diff
@@ -649,20 +722,52 @@ def create(
...
+                            "supported for generator data input"
+                        )
+                    )
+
+        if metadata:
+            schema = schema.with_metadata(metadata)
+
+        empty = pa.Table.from_pylist([], schema=schema)
+        lance.write_dataset(empty, tbl._dataset_uri, schema=schema, mode=mode)
+        table = LanceTable(db, name)
```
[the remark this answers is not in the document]
changhiskhan: And also that a directory is written as well

python/lancedb/table.py
```diff
@@ -196,17 +228,27 @@ def add(
...
     ) -> LanceQueryBuilder:
         """Create a search query to find the nearest neighbors
         of the given query vector.
 
         Parameters
         ----------
-        query: list, np.ndarray
-            The query vector.
-        vector_column: str, default "vector"
+        query: str, list, np.ndarray, default None
```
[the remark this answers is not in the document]
changhiskhan: partially. but also because full text search. Once we enable multi-modal embedding functions, we'll add to this as well (with Image for example).

python/lancedb/table.py
```diff
@@ -569,18 +615,52 @@ def merge(
...
+
+        Returns
+        -------
+        funcs: dict
+            A mapping of the vector column to the embedding function
+            or empty dict if not configured.
+        """
+        metadata = self.schema.metadata
+        if metadata is None or b"embedding_functions" not in metadata:
+            return {}
```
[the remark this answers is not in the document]
changhiskhan: done

python/lancedb/table.py
```diff
@@ -770,22 +875,36 @@ def _sanitize_schema(
             return data
         # cast the columns to the expected types
         data = data.combine_chunks()
-        data = _sanitize_vector_column(
+        for field in schema:
+            # TODO: we're making an assumption that fixed size list of 10 or more
+            # is a vector column. This is definitely a bit hacky.
+            is_fixed_list = (
```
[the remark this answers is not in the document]
changhiskhan: changed

python/lancedb/embeddings/functions.py
changhiskhan: the default is ~90mb in pickle format (https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/tree/main) and roughly 150mb when loaded into memory. I added a TODO already in the docstring "TODO: use lru_cache instead with a reasonable/configurable maxsize". I can file a Github issue as well if you like

python/lancedb/query.py
wjones127: Ah that makes sense. Getting so used to Rust where we get to know what parameters are mutated and which aren't :)

python/lancedb/embeddings/functions.py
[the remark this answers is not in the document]
wjones127: I guess my question is, if the user wants to add an embedding function that does take multiple columns, do this allow them? Or can they just always process the embedding themself regardless?

python/lancedb/embeddings/functions.py
changhiskhan: fixed

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        self._functions = {}
+
+    def load(self, name: str):
+        """
+        Fetch an embedding function class by name
+        """
+        return self._functions[name]
+
+    def parse_functions(self, metadata: Optional[dict]) -> dict:
```
eddyxu, line 66: Can the type `dict` work in python3.8>

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        Parse the metadata from an arrow table and
+        return a mapping of the vector column to the
```
eddyxu, line 69: So the key is input column? or the generated embedding (output) column?

python/lancedb/embeddings/functions.py
```diff
@@ -0,0 +1,224 @@
...
+class EmbeddingFunctionModel(BaseModel, ABC):
+    """
+    A callable ABC for embedding functions
+    """
+
+    source_column: Optional[str]
+    vector_column: str
+
+    @abstractmethod
+    def __call__(self, *args, **kwargs) -> List[np.array]:
```
eddyxu, line 130: Do they expect to be called in batch? or this will be called one row at a time?

python/lancedb/embeddings/functions.py
changhiskhan: The key is the vector column.
This is so we can have fast lookup when executing a search query

python/lancedb/embeddings/functions.py
changhiskhan: Can be both but for performance it's  best to be called in batches. We generally call it in batches (feed it a whole pyarrow table column for example)

python/lancedb/embeddings/functions.py
changhiskhan: The 3.8 CI seems to pass

python/lancedb/db.py
```diff
@@ -289,6 +290,7 @@ def create_table(
         mode: str = "create",
         on_bad_vectors: str = "error",
         fill_value: float = 0.0,
+        embedding_functions: List[EmbeddingFunctionModel] = None,
```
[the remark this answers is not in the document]
changhiskhan: done

wjones127 (approved):

PR #474, opened 2023-09-04
[Python] Pydantic vector field with default value

eddyxu:
Rename `lance.pydantic.vector` to `Vector` and deprecate `vector(dim)`

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -49,11 +49,11 @@ A Table is a collection of Records in a LanceDB Database. You can follow along o
 
     db.create_table("table2", data)
 
-    db["table2"].head() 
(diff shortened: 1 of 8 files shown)


catethos:
This works
```
class TestModel(LanceModel):
        vec: vector(128) = Field(default=[])
```
but not this

```
class TestModel(LanceModel):
        vector: vector(128) = Field(default=[])
```

where the field name is the same as the constructor. LanceDB needs the embedding field to be exactly `vector` right? I find no way to change it

changhiskhan (approved):
Do any of the examples in vectordb-recipes need to be updated?

eddyxu:
I can look into `vectordb-recepits`. It needs a release of this PR tho.

PR #476, opened 2023-09-06
[Rust] schema coerce and vector column inference

eddyxu:
Split the rust core from #466 for easy review and less merge conflicts.

### DIFF
--- a/rust/vectordb/src/data/inspect.rs
+++ b/rust/vectordb/src/data/inspect.rs
@@ -17,41 +17,31 @@ use std::collections::HashMap;
 use arrow::compute::kernels::{aggregate::bool_and, length::length};
 use arrow_array::{
     cast::AsArray,
-    types::{Int32Type, Int64Type},
+    types::{ArrowPrimitiveType, Int32Type, Int64Type},
     Array, GenericListArray, OffsetSizeTrait, RecordBatchReader,
 };
 use arrow_ord::comparison::eq_dyn_scalar;
 use arrow_schema::DataType;
-use num_traits::{AsPrimitive, Zero};
+use num_traits::{ToPrimitive, Zero};
 
 use crate::error::{Error, Result};
 
-pub(crate) fn infer_dimension<O: OffsetSizeTrait + Zero>(
-    list_arr: &GenericListArray<O>,
-) -> Result<Option<O>>
+pub(crate) fn infer_dimension<T: ArrowPrimitiveType>(
+    list_arr: &GenericListArray<T::Native>,
+) -> Result<Option<T::Native>>
 where
-    i32: AsPrimitive<O>,
-    i64: AsPrimitive<O>,
+    T::Native: OffsetSizeTrait + ToPrimitive,
 {
     let len_arr = length(list_arr)?;
     if len_arr.is_empty() {
         return Ok(Some(Zero::zero()));
     }
 
-    if O::IS_LARGE {
-        let dim = len_arr.as_primitive::<Int64Type>().value(0);
-        if bool_and(&eq_dyn_scalar(len_arr.as_primitive::<Int64Type>(), dim)?) != Some(true) {
-            Ok(None)
-        } else {
-            Ok(Some(dim.as_()))
-        }
+    let dim = len_arr.as_primitive::<T>().value(0);
+    if bool_and(&eq_dyn_scalar(len_arr.as_primitive::<T>(), dim)?) != Some(true) {
+        Ok(None)
     } else {
-        let dim = len_arr.as_primitive::<Int32Type>().value(0);
-        if bool_and(&eq_dyn_scalar(len_arr.as_primitive::<Int32Type>(), dim)?) != Some(true) {
-            Ok(None)
-        } else {
-            Ok(Some(dim.as_()))
-        }
+        Ok(Some(dim))
     }
 }
 
@@ -92,9 +82,9 @@ pub fn infer_vector_columns(
             })?;
             if let Some(dim) = match *col.data_type() {
                 DataType::List(_) => {
-                    infer_dimension::<i32>(col.as_list::<i32>())?.map(|d| d as i64)
+                    infer_dimension::<Int32Type>(col.as_list::<i32>())?.map(|d| d as i64)
                 }
-                DataType::LargeList(_) => infer_dimension::<i64>(col.as_list::<i64>())?,
+                DataType::LargeList(_) => infer_dimension::<Int64Type>(col.as_list::<i64>())?,
                 _ => {
                     return Err(Error::Schema {
                         message: format!("Column {} is not a list", col_name),
--- a/rust/vectordb/src/data/inspect.rs
+++ b/rust/vectordb/src/data/inspect.rs
@@ -15,11 +15,28 @@
 use std::collections::HashMap;
 
 use arrow::compute::kernels::{aggregate::bool_and, length::length};
-use arrow_array::{cast::AsArray, types::Int32Type, RecordBatchReader};
+use arrow_array::{cast::AsArray, types::Int32Type, Array, RecordBatchReader};
 use arrow_ord::comparison::eq_dyn_scalar;
 use arrow_schema::DataType;
 
 use crate::error::{Error, Result};
+
+pub(crate) fn infer_dimension(arr: &dyn Array) -> Result<Option<i32>> {
+    if !matches!(arr.data_type(), DataType::List(_)) {
+        return Ok(None);
+    };
+    let list_arr = arr.as_list_opt::<i32>().expect("must be a list now");
+    let len_arr = length(list_arr)?;
+    if len_arr.is_empty() {
+        return Ok(Some(0));
+    }
+
+    let dim = len_arr.as_primitive::<Int32Type>().value(0);
+    if bool_and(&eq_dyn_scalar(len_arr.as_primitive::<Int32Type>(), dim)?) != Some(true) {
+        return Ok(None);
+    }
+    return Ok(Some(dim));
+}
 
 /// Infer the vector columns from a dataset.
 ///
@@ -34,7 +51,7 @@ pub fn infer_vector_columns(
 ) -> Result<Vec<String>> {
     let mut columns = vec![];
 
-    let mut columns_map: HashMap<String, Option<u32>> = HashMap::new();
+    let mut columns_map: HashMap<String, Option<i32>> = HashMap::new();
     for field in reader.schema().fields() {
         match field.data_type() {
             DataType::FixedSizeList(sub_field, _) if sub_field.data_type().is_floating() => {
@@ -53,17 +70,16 @@ pub fn infer_vector_columns(
             let col = batch.column_by_name(&col_name).ok_or(Error::Schema {
                 message: format!("Column {} not found", col_name),
             })?;
-            let list_arr = col.as_list_opt::<i32>().expect("Must be a list array now");
-            let len_arr = length(list_arr)?;
-            if len_arr.is_empty() {
-                columns_map.remove(&col_name);
-            } else {
-                let len = len_arr.as_primitive::<Int32Type>().value(0);
-                if bool_and(&eq_dyn_scalar(len_arr.as_primitive::<Int32Type>(), len)?)
-                    != Some(true)
-                {
-                    columns_map.remove(&col_name);
+            if let Some(dim) = infer_dimension(col.as_ref())? {
+                if let Some(Some(prev_dim)) = columns_map.get(&col_name) {
+                    if prev_dim != &dim {
+                        columns_map.remove(&col_name);
+                    }
+                } else {
+                    columns_map.insert(col_name, Some(dim));
                 }
+            } else {
+                columns_map.remove(&col_name);
             }
         }
     }
--- a/rust/vectordb/src/data/inspect.rs
+++ b/rust/vectordb/src/data/inspect.rs
@@ -15,27 +15,42 @@
 use std::collections::HashMap;
 
 use arrow::compute::kernels::{aggregate::bool_and, length::length};
-use arrow_array::{cast::AsArray, types::Int32Type, Array, RecordBatchReader};
+use arrow_array::{
+    cast::AsArray,
+    types::{Int32Type, Int64Type},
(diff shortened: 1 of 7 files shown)


rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,263 @@
...
+    batch: RecordBatch,
+    schema: Arc<Schema>,
+) -> std::result::Result<RecordBatch, ArrowError> {
+    if batch.schema() == schema {
+        return Ok(batch);
+    }
+    let columns = schema
+        .fields()
+        .iter()
+        .flat_map(|field| {
```
wjones127, line 141: What is this `flat_map` and not `map`?

rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,263 @@
...
+) -> Result<Box<dyn RecordBatchReader + Send>> {
+    if reader.schema() == schema {
+        return Ok(Box::new(RecordBatchIterator::new(
+            reader.into_iter().collect::<Vec<_>>(),
+            schema,
+        )));
+    }
+    let batches = reader
+        .map(|batch| coerce_schema_batch(batch?, schema.clone()))
+        .collect::<Vec<_>>();
```
wjones127, line 167: Why collect? Can we just keep as an iterator?
```suggestion
    let batches = reader
        .map(|batch| coerce_schema_batch(batch?, schema.clone()));
```

rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,263 @@
...
+///
+pub fn coerce_schema(
+    reader: impl RecordBatchReader,
+    schema: Arc<Schema>,
+) -> Result<Box<dyn RecordBatchReader + Send>> {
+    if reader.schema() == schema {
+        return Ok(Box::new(RecordBatchIterator::new(
+            reader.into_iter().collect::<Vec<_>>(),
+            schema,
+        )));
```
wjones127, line 163: Similarly here, why collect and wrap it again:
```suggestion
        return Ok(Box::new(reader));
```

rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,263 @@
...
+                let actual_sub = array.as_fixed_size_list();
+                let values = coerce_array(actual_sub.values(), exp_field)?;
+                Ok(Arc::new(FixedSizeListArray::try_new_from_values(
+                    values.clone(),
+                    *dim,
+                )?) as Arc<dyn Array>)
+            }
+            DataType::List(_) => {
+                let list_arr = array.as_list::<i32>();
+                if list_arr.values().len() != (*exp_dim as usize) * list_arr.len() {
```
wjones127, line 101: hmm this isn't strictly sufficient. To be fully robust you'd need to check the length of each list. But maybe this is fine for now, given that if we reach this branch the user has already said they intend for a FSL of a certain length, right?

wjones127 (changes requested):
I think you can avoid collecting in the reader.

rust/vectordb/src/data/sanitize.rs
eddyxu: `cargo clippy` suggested this way.

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,144 @@
...
+use arrow_schema::DataType;
+
+use crate::error::{Error, Result};
+
+/// Infer the vector columns from a dataset.
+///
+/// Parameters
+/// ----------
+/// - reader: RecordBatchReader
+/// - strict: if set true, only fixed_size_list<float> is considered as vector column. If set to false,
```
changhiskhan, line 46: I'm wondering if we should have a minimum length requirement here. I guess UMAP/PCA results can be low dimensions, but I also worry about false positives?

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,144 @@
...
+pub fn infer_vector_columns(
+    reader: impl RecordBatchReader + Send,
+    strict: bool,
+) -> Result<Vec<String>> {
+    let mut columns = vec![];
+
+    let mut columns_map: HashMap<String, Option<u32>> = HashMap::new();
+    for field in reader.schema().fields() {
+        match field.data_type() {
+            DataType::FixedSizeList(sub_field, _) if sub_field.data_type().is_floating() => {
```
changhiskhan, line 57: i forgot does `is_floating` check only for f32 or any float?

rust/vectordb/src/data/sanitize.rs
eddyxu: Fixed. Need to fix lifetime as well.

rust/vectordb/src/data/sanitize.rs
eddyxu: Fixed.

rust/vectordb/src/data/sanitize.rs
eddyxu: Oh, right. This was a directly copy from python sanitizer.  Lemme fix it.

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,144 @@
...
+                columns.push(field.name().to_string());
+            }
+            DataType::List(sub_field) if sub_field.data_type().is_floating() && !strict => {
```
changhiskhan, line 60: what about LargeListArray? safe to ignore for now?

rust/vectordb/src/data/sanitize.rs
eddyxu: Fixed.

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+    for batch in reader {
+        let batch = batch?;
+        let col_names = columns_map.keys().cloned().collect::<Vec<_>>();
+        for col_name in col_names {
+            let col = batch.column_by_name(&col_name).ok_or(Error::Schema {
+                message: format!("Column {} not found", col_name),
+            })?;
+            if let Some(dim) = infer_dimension(col.as_ref())? {
```
changhiskhan, line 73: this is not in the query path right? Feels like the performance penalty might be big here?

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,160 @@
...
+        return Ok(None);
+    };
+    let list_arr = arr.as_list_opt::<i32>().expect("must be a list now");
+    let len_arr = length(list_arr)?;
+    if len_arr.is_empty() {
+        return Ok(Some(0));
+    }
+
+    let dim = len_arr.as_primitive::<Int32Type>().value(0);
+    if bool_and(&eq_dyn_scalar(len_arr.as_primitive::<Int32Type>(), dim)?) != Some(true) {
```
changhiskhan, line 35: i assume this short circuits right?

rust/vectordb/src/data/inspect.rs
eddyxu: `is_floating` is `f16 | f32 | f64`

https://github.com/apache/arrow-rs/blob/65edbb1702e25420aacebe656dcd789690f72c82/arrow-schema/src/datatype.rs#L361

We need later to extend it to bf16 tho.

rust/vectordb/src/data/inspect.rs
eddyxu: I can add it in this PR for safety.

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::collections::HashMap;
+
+use arrow::compute::kernels::{aggregate::bool_and, length::length};
+use arrow_array::{cast::AsArray, types::Int32Type, Array, RecordBatchReader};
+use arrow_ord::comparison::eq_dyn_scalar;
+use arrow_schema::DataType;
+
+use crate::error::{Error, Result};
+
+pub(crate) fn infer_dimension(arr: &dyn Array) -> Result<Option<i32>> {
```
chebbyChefNEQ, line 24: should we make this a generic function so we can customize the data width?

rust/vectordb/src/data/inspect.rs
eddyxu: No, this is on the ingestion path. This is not slower than what we have in python `table::create()` now.

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,160 @@
...
+    if !matches!(arr.data_type(), DataType::List(_)) {
+        return Ok(None);
+    };
+    let list_arr = arr.as_list_opt::<i32>().expect("must be a list now");
```
chebbyChefNEQ, line 28: nit: iiuc, this `i32` really just means 32-bit wide, doesn't mean the data is actually integer. Could we alias this to be more clear?

```
type BIT_WIDTH_16 = i16;
...
```

rust/vectordb/src/data/inspect.rs
eddyxu: 32? 64? I dont have a good sense where to cut the line now.

rust/vectordb/src/data/inspect.rs
eddyxu: i32 means it is i32 as offset (which is `List` not `LargeList`).

rust/vectordb/src/data/inspect.rs
eddyxu: I can just make it work with both list/large list? the data width of each element should be not relevant i think.

rust/vectordb/src/data/inspect.rs
```diff
@@ -0,0 +1,160 @@
...
+/// - reader: RecordBatchReader
+/// - strict: if set true, only fixed_size_list<float> is considered as vector column. If set to false,
+///           a list<float> column with same length is also considered as vector column.
+pub fn infer_vector_columns(
+    reader: impl RecordBatchReader + Send,
+    strict: bool,
+) -> Result<Vec<String>> {
+    let mut columns = vec![];
+
+    let mut columns_map: HashMap<String, Option<i32>> = HashMap::new();
```
chebbyChefNEQ, line 54: nit: should we call this `columns_to_infer`?

rust/vectordb/src/data/inspect.rs
eddyxu: https://docs.rs/arrow-array/46.0.0/arrow_array/array/struct.GenericListArray.html

rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,263 @@
...
+        .fields()
+        .iter()
+        .flat_map(|field| {
+            batch
+                .column_by_name(field.name())
+                .map(|c| coerce_array(c, field))
+                .ok_or(|| {
+                    ArrowError::SchemaError(format!("Column {} not found in batch", field.name()))
+                })
+        })
```
wjones127, line 158: Try this instead:
```suggestion
        .map(|field| {
            batch
                .column_by_name(field.name())
                .ok_or(|| {
                    ArrowError::SchemaError(format!("Column {} not found in batch", field.name()))
                })
                .and_then(|c| coerce_array(c, field))
        })
```

rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,271 @@
...
+) -> std::result::Result<Arc<dyn Array>, ArrowError> {
+    if array.data_type() == field.data_type() {
+        return Ok(array.clone());
+    }
+    match (array.data_type(), field.data_type()) {
+        // Normal cast-able types.
+        (adt, dt) if can_cast_types(adt, dt) => cast(&array, dt),
+        // Casting between f16/f32/f64 can be lossy.
+        (adt, dt) if (adt.is_floating() || dt.is_floating()) => {
+            if adt.byte_width() < dt.byte_width() {
```
chebbyChefNEQ, line 75: should this be flipped?

array (32-bit) < field (64-bit) should be lossless

rust/vectordb/src/data/sanitize.rs
wjones127: Oh I see it's a nested result. I don't think you should call `flatten()` on result; it ignores the errors.

rust/vectordb/src/data/inspect.rs
chebbyChefNEQ: Ah ok that makes sense

rust/vectordb/src/data/sanitize.rs
eddyxu: This results to a `Result<Vec<Result<_, _>>` ?

rust/vectordb/src/data/inspect.rs
eddyxu: Done.

rust/vectordb/src/data/sanitize.rs
eddyxu: Good catch.

rust/vectordb/src/data/sanitize.rs
eddyxu: ok, made it work.  Change to use `ok_or(ArrowError::...)` directly

rust/vectordb/src/data/sanitize.rs
```diff
@@ -0,0 +1,282 @@
...
+    let columns = schema
+        .fields()
+        .iter()
+        .map(|field| {
+            batch
+                .column_by_name(field.name())
+                .map(|c| coerce_array(c, field))
+                .ok_or(ArrowError::SchemaError("Column not found".to_string()))
+        })
+        .collect::<std::result::Result<std::result::Result<Vec<_>, ArrowError>, ArrowError>>()??;
```
wjones127, line 170: Sorry, I didn't mean for you to nest the `Result`. This time I've tested this locally:
```suggestion
        .map(|field| {
            batch
                .column_by_name(field.name())
                .ok_or_else(|| {
                    ArrowError::SchemaError(format!("Column {} not found", field.name()))
                })
                .and_then(|c| coerce_array(c, field))
        })
        .collect::<std::result::Result<Vec<_>, ArrowError>>()?;
```

wjones127 (approved):

PR #480, opened 2023-09-06
[Python][Rust] Expose rust code sanitization function via pyo3

eddyxu:


eddyxu:
Pending the delection vector perf regression fix before making this ready for review. Because it requires more work to setup manylinux / windows / mac native build for python release.

python/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[lib]
+name = "lancedb"
+crate-type = ["cdylib"]
+
+[dependencies]
+arrow = { version = "43.0.0", features = ["pyarrow"] }
+arrow-array = "43.0"
+arrow-schema = "43.0"
+pyo3 = { version = "0.19", features = ["extension-module", "abi3-py38"] }
+vectordb = { path = "../rust/vectordb" }
```
wjones127, line 19: The unfortunate thing here is we'll have two copies of the lance Rust library, so the install size of `lancedb` python is going to nearly double.

eddyxu: The plan is eventually we will just remove the pylance dependency from lancedb python package.

PR #484, opened 2023-09-08
multi-modal embedding-function

changhiskhan:


python/lancedb/embeddings/functions.py
```diff
@@ -28,25 +37,33 @@ class EmbeddingFunctionRegistry:
...
         """
 
         # This is a decorator for a class that inherits from BaseModel
         # It adds the class to the registry
         def decorator(cls):
-            if not issubclass(cls, EmbeddingFunctionModel):
-                raise TypeError("Must be a subclass of EmbeddingFunctionModel")
+            if not issubclass(cls, EmbeddingFunction):
+                raise TypeError("Must be a subclass of EmbeddingModel")
             if cls.__name__ in self._functions:
```
AyushExel, line 62: This should be `EmbeddingFunction` here I think

python/lancedb/embeddings/functions.py
```diff
@@ -28,25 +37,33 @@ class EmbeddingFunctionRegistry:
...
 
     def __init__(self):
         self._functions = {}
 
-    def register(self):
+    def register(self, alias: str = None):
         """
         This creates a decorator that can be used to register
-        an EmbeddingFunctionModel.
+        an EmbeddingModel.
```
AyushExel, line 48: This should also be `EmbeddingFunction` ?

python/lancedb/embeddings/functions.py
```diff
@@ -28,25 +37,33 @@ class EmbeddingFunctionRegistry:
...
         # It adds the class to the registry
         def decorator(cls):
-            if not issubclass(cls, EmbeddingFunctionModel):
-                raise TypeError("Must be a subclass of EmbeddingFunctionModel")
+            if not issubclass(cls, EmbeddingFunction):
+                raise TypeError("Must be a subclass of EmbeddingModel")
             if cls.__name__ in self._functions:
                 raise KeyError(f"{cls.__name__} was already registered")
-            self._functions[cls.__name__] = cls
+            key = alias or cls.__name__
```
AyushExel, line 64: Maybe it's better to check for `key` instead of `cls.__name__` in `self._functions` ?
I'm thinking of a case where user accidentally overrides an embedding function using an alias.

python/lancedb/embeddings/functions.py
```diff
@@ -220,9 +343,197 @@ def get_embedding_model(cls, name, device):
...
+        ----------
+        query : Union[str, PIL.Image.Image]
+            The query to embed. A query can be either text or an image.
+        """
+        if isinstance(query, str):
+            return [self.generate_text_embeddings(query)]
+        else:
+            PIL = self.safe_import("PIL", "pillow")
+            if isinstance(query, PIL.Image.Image):
+                return [self.generate_image_embedding(query)]
```
AyushExel, line 429: Possibly we can also accept `Path()` object and load it via PIL ourselves

python/lancedb/embeddings/functions.py
```diff
@@ -220,9 +343,197 @@ def get_embedding_model(cls, name, device):
...
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            futures = [
+                executor.submit(self.generate_image_embedding, image)
+                for image in images
+            ]
+            return [future.result() for future in futures]
+
+    def generate_image_embedding(
+        self, image: Union[str, bytes, "PIL.Image.Image"]
+    ) -> np.ndarray:
```
AyushExel, line 483: Okay so here we are handling str uri for Image but I think it needs to be handled in the `compute_embedding_function` too. If the input is of type `Path` it can be cast to str and treated as Img uri too?

python/lancedb/embeddings/functions.py
AyushExel, line 429: `generate_image_embeddings` accepts str, bytes and PIL images but here we only check for PIL instance. Maybe this can accept `Path` objects and we can handle loading in internally too?

AyushExel (commented):
Took a pass. A couple small questions.

python/lancedb/embeddings/functions.py
changhiskhan: good catch. incomplete rename refactor.

python/lancedb/embeddings/functions.py
changhiskhan: yeah we lookup by alias if it exists else just the plain class name. This is so you can name the class OpenAIEmbeddingFunction and just look it up as registry.get("openai")

python/lancedb/embeddings/functions.py
changhiskhan: The main thing here is that we wouldn't be able to tell whether `str` is user text query or an image uri, so we just ask the user to clarify by doing Image.open(uri). Similarly with `bytes` and `Path`, we wouldn't be able to tell whether it's an image, audio sample, or other things. So for now we just make it a more explicit API. Wdyt?

AyushExel: yeah okay makes sense. A good solution while the data types are in the works

python/lancedb/embeddings/functions.py
AyushExel: ok got it

AyushExel (approved):

PR #492, opened 2023-09-15
[Python]Sentry error reporting & Global config manager

AyushExel:
Conditions required to send errors (ALL conditions must be met or no errors will be reported):
        - sentry_sdk package is installed (Maybe we make it a dependency?)
        - sync=True in  settings
        - pytest is not running
        - running in a pip package installation
        - running in a non-git directory
        - online environment
 
- CLI config commands are needed to allow the user to be able to edit them, but maybe it'll be too large for this PR.

TODOs:
- [x] remove verbose sentry log msg (next PR)
- [x] needs lancedb.__version__ to tag the version of the error origin


python/lancedb/utils/config.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if not self.file.exists():
+            self.save()
+
+        self.load()
+        correct_keys = self.keys() == self.defaults.keys()
+        correct_types = all(
+            type(a) is type(b) for a, b in zip(self.values(), self.defaults.values())
+        )
+        if not (correct_keys and correct_types):
```
rok, line 82: Would this be an improvement or am I missing the point? :D
```suggestion
        config_created = False
        if not self.file.exists():
            self.save()
            config_created = True

        self.load()
        correct_keys = self.keys() == self.defaults.keys()
        correct_types = all(
            type(a) is type(b) for a, b in zip(self.values(), self.defaults.values())
        )
        if not (config_created and correct_keys and correct_types):
```

python/lancedb/utils/sentry_log.py
```diff
@@ -0,0 +1,92 @@
...
+            return
+
+        sentry_sdk.init(
+            dsn="https://2a892530e0c92d3a3194cddc897a9090@o4505873570791424.ingest.sentry.io/4505873580425216",
+            debug=False,
+            traces_sample_rate=1.0,
+            environment="production",  # 'dev' or 'production'
+            before_send=before_send,
+            ignore_errors=[KeyboardInterrupt, FileNotFoundError],
+        )
```
rok, line 84: I forget how tags work in sentry, but will be able to filter by lance/lancedb versions in sentry or should we add something here?

rok (commented):
Great stuff! Added some small comments.
I used sentry in the past and it was great to quickly triage python errors!

python/lancedb/utils/config.py
```diff
@@ -0,0 +1,108 @@
...
+import os
+import platform
+import uuid
+from pathlib import Path
+
+from .general import LOGGER, is_dir_writeable, yaml_load, yaml_save
+
+MACOS, LINUX, WINDOWS = (
+    platform.system() == x for x in ["Darwin", "Linux", "Windows"]
+)  # environment booleans
```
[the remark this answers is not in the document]
AyushExel: Yeah makes sense. I think they might be used for event tracking with posthog but for now eval on fly sounds better

python/lancedb/utils/config.py
AyushExel: I see. I think if `config_created` is true, the keys and types will always be correct. This condition is aimed at re-creating the config file if the default configs are ever changed in the package. So, I think if config_created == True, then theres no need to check for types and keys, making it a redundant check? Is there some other condition you can think of?

python/lancedb/utils/sentry_log.py
AyushExel: yeah version and other details are tracked here in `before_send` callback https://github.com/lancedb/lancedb/pull/492/files#diff-603e1842b90966a1a2eb9e41f61c62a6db0fd6b3d5a71cec2f0355387c3d7eb7R62

python/lancedb/utils/config.py
rok: That would be redundant indeed. Ok nevermind.

wjones127 (changes requested):
Noticed a few small issues.

Is there a link where we can see some examples of the telemetry data?

python/lancedb/utils/general.py
```diff
@@ -0,0 +1,309 @@
...
+    "Colab"
+    if is_colab()
+    else "Kaggle"
+    if is_kaggle()
+    else "Jupyter"
+    if is_jupyter()
+    else "Docker"
+    if is_docker()
+    else platform.system()
+)
```
rok, line 309: If multiple of these are true (given the checks it wouldn't be too hard) we will only get the first one that was checked reported. Is the order of these checks ok or would we rather check all and concatenate into one string to capture all?

AyushExel: I think these should be mutually exclusive no? Like you cannot open colab inside of a kaggle notebook or open a kaggle kernel in a jupyter env?

rok: Sure, but results of checks might not be, for instance if you run jupyter server inside a docker container what would then be the most relevant environment to report to sentry?

AyushExel: I see.. So ENVIRONMENTS capturing all of them would be best then?
Something like this is okay or comma separated would be better?
```
ENVIRONMENTS = platform.system()
if is_colab():
    ENVIRONMENTS += " Colab "
if is_kaggle():
    ENVIRONMENTS += " Kaggle "
if is_jupyter():
    ENVIRONMENTS += " Jupyter "
if is_docker():
    ENVIRONMENTS += " Docker "
```

AyushExel:
@wjones127 looks like this, I cannot invite anyone as it's not a paid account. I'll ask Chang to create a teams account.
Screenshot 2023-09-27 at 10 58 06 AM

AyushExel:
@wjones127 sent an invite. You can see the issues here https://lancedb-00.sentry.io/issues/?referrer=sidebar

python/lancedb/utils/general.py
```diff
@@ -0,0 +1,298 @@
...
+
+PLATFORMS = platform.system()
+if is_colab():
+    PLATFORMS += " Colab "
+if is_kaggle():
+    PLATFORMS += " Kaggle "
+if is_jupyter():
+    PLATFORMS += " Jupyter "
+if is_docker():
+    PLATFORMS += " Docker "
```
rok, line 298: How about:
```suggestion
PLATFORMS = [platform.system()]
if is_colab():
    PLATFORMS.append("Colab")
if is_kaggle():
    PLATFORMS.append("Kaggle")
if is_jupyter():
    PLATFORMS.append("Jupyter")
if is_docker():
    PLATFORMS.append("Docker")

PLATFORMS = "|".join(PLATFORMS)
```

rok: Sorry for  the bikeshedding :D

rok (approved):
LGTM!

AyushExel:
Merging this in a different branch. where other telemetry PRs will be merged as well. Once ready, we can add everything to main

PR #500, opened 2023-09-18
refactor connection string processing

chebbyChefNEQ:
in #486 `connect` started converting path into uri. However, the PR didn't handle relative path and appended `file://` to relative path.

This PR changes the parsing strat to be more rational. If a path is provided instead of url, we do not try anythinng special.

engine and engine params may only be specified when a url with schema is provided


changhiskhan (approved):
thanks!

PR #505, opened 2023-09-19
fix: upgrade lance to 0.7.5 and add tests for searching empty dataset

chebbyChefNEQ:
This PR upgrade lance to `0.7.5`, which include fixes for searching an empty dataset.

This PR also adds two tests in node SDK to make sure searching empty dataset do no throw

### DIFF
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -19,7 +19,7 @@ import * as chaiAsPromised from 'chai-as-promised'
 
 import * as lancedb from '../index'
 import { type AwsCredentials, type EmbeddingFunction, MetricType, Query, WriteMode, DefaultWriteOptions, isWriteOptions } from '../index'
-import { Field, Int32, makeVector, Schema, Utf8, Table as ArrowTable, vectorFromArray } from 'apache-arrow'
+import { FixedSizeList, Field, Int32, makeVector, Schema, Utf8, Table as ArrowTable, vectorFromArray, Float32 } from 'apache-arrow'
 
 const expect = chai.expect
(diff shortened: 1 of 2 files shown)


node/src/test/test.ts
```diff
@@ -258,6 +258,34 @@ describe('LanceDB client', function () {
...
+  describe('when searching an empty dataset', function () {
+    it('should not fail', async function () {
+      const dir = await track().mkdir('lancejs')
+      const con = await lancedb.connect(dir)
+
+      const schema = new Schema(
+        [new Field('vector', new FixedSizeList(128, new Field('float32', new Float32())))]
+      )
+      const table = await con.createTable({ name: 'vectors', schema })
+      await table.search(Array(128).fill(0.1)).execute()
```
changhiskhan, line 270: can you add an assertion on what it should return?

node/src/test/test.ts
```diff
@@ -258,6 +258,34 @@ describe('LanceDB client', function () {
...
+      const dir = await track().mkdir('lancejs')
+      const con = await lancedb.connect(dir)
+
+      const schema = new Schema(
+        [new Field('vector', new FixedSizeList(128, new Field('float32', new Float32())))]
+      )
+      const table = await con.createTable({ name: 'vectors', schema })
+      await table.add([{ vector: Array(128).fill(0.1) }])
+      await table.delete('vector IS NOT NULL')
+      await table.search(Array(128).fill(0.1)).execute()
```
changhiskhan, line 285: same

changhiskhan (approved):

PR #508, opened 2023-09-21
[Python] Fix attrs dependency

hynek:
The `attr` project is unrelated to `attrs` that also provides the `attr` namespace (see also <https://hynek.me/articles/import-attrs/>).

It used to _usually_ work, because attrs is a dependency of aiohttp and somehow took precedence over `attr`'s `attr`.

Yes, sorry, it's a mess.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -8,7 +8,7 @@ dependencies = [
     "tqdm",
     "aiohttp",
     "pydantic",
-    "attr",
+    "attrs",
     "semver>=3.0",
     "cachetools"
 ]


python/pyproject.toml
```diff
@@ -62,4 +62,4 @@ addopts = "--strict-markers"
 markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
     "asyncio"
-]
\ No newline at end of file
+]
```
hynek, line 65: Eh this was added by GitHub's web UI. LMK if you want to get rid of it.

rec:
Wow, that was fast! :-)

rec (approved):

chebbyChefNEQ (approved):

chebbyChefNEQ:
Thank you for fixing this!

PR #510, opened 2023-09-21
fix `import attr` to use `import attrs`

chebbyChefNEQ:
Thanks to #508, I used `attr` instead of the correct package `attrs`

s/attr/attrs

### DIFF
--- a/python/lancedb/remote/__init__.py
+++ b/python/lancedb/remote/__init__.py
@@ -14,7 +14,7 @@
 import abc
 from typing import List, Optional
 
-import attr
+import attrs
 import pyarrow as pa
 from pydantic import BaseModel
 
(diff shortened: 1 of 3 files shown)


python/lancedb/remote/__init__.py
```diff
@@ -14,7 +14,7 @@
 import abc
 from typing import List, Optional
 
-import attr
```
eddyxu, line 17: Do you need to update `pyproject.toml`?

chebbyChefNEQ: Already done in #508

chebbyChefNEQ: `attr` and `attrs` are functionality the same

rec (approved):
Such service!

This change is not strictly necessary as `attrs` imports itself as `attr`, but I think it's a very good idea to prevent confusion.

  👍

changhiskhan (approved):
Thanks 🙏

Issue #512, opened 2023-09-23
Installing LanceDB on DataBricks


jtrenkle:
Hey, Folks, I have a DataBricks cluster with this configuration: 

```
12.2 LTS ML (includes Apache Spark 3.3.2, Scala 2.12)
Single-node cluster
Driver Type: r5n.24xlarge
```

Spark Config:
```
spark.serializer org.apache.spark.serializer.KryoSerializer
spark.sql.autoBroadcastJoinThreshold -1
spark.driver.maxResultSize 30g
spark.kryoserializer.buffer.max 2047M
spark.databricks.delta.preview.enabled true
spark.master local[*]
spark.rpc.message.maxSize 512
```

And I am simply trying to install LanceDB via pip to test out the basic capability.

Due to various packages already attached and other issues, I ended up running these steps in a notebook  to install LanceDB without errors:

```
!python -m pip install --upgrade pip
!pip install --upgrade pandas
!pip install --upgrade pyarrow
!pip install lancedb
```

This leads to an error-free install (for the most part); however, when I do this: 

```
import lancedb
```

I get this error: 

```
AttributeError: module 'pydantic' has no attribute '__version__'
```

------

For completeness, here are the various logs: 

```
!pip install --upgrade pandas
```

Logs: 

```
Looking in indexes: https://pypi.org/simple, https://databricks:****@tubins.jfrog.io/tubins/api/pypi/pypi-release-local/simple
Requirement already satisfied: pandas in /databricks/python3/lib/python3.9/site-packages (1.4.2)
Collecting pandas
  Obtaining dependency information for pandas from https://files.pythonhosted.org/packages/bc/7e/a9e11bd272e3135108892b6230a115568f477864276181eada3a35d03237/pandas-2.1.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata
  Downloading pandas-2.1.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (18 kB)
Collecting numpy>=1.22.4 (from pandas)
  Obtaining dependency information for numpy>=1.22.4 from https://files.pythonhosted.org/packages/75/cd/7ae0f2cd3fc68aea6cfb2b7e523842e1fa953adb38efabc110d27ba6e423/numpy-1.26.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata
[...]

PR #515, opened 2023-09-25
fix: define minimum dependency versions

### LINKED ISSUE #513: Set minimal versions for dependencies
Based on questions in #512

We don't set any minimal versions of dependencies, even though we definetely aren't compatible with all versions. This forces users having to play wack-a-mole upgrading dependencies until lancedb successfully imports, which is frustrating.

https://github.com/lancedb/lancedb/blob/bc8670d7af5f3e55453f0461577a70dfc311c7cd/python/pyproject.toml#L4-L14

wjones127:
Closes #513

For each of these, I found the minimum version that would allow the unit tests to pass.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -3,12 +3,12 @@ name = "lancedb"
 version = "0.2.5"
 dependencies = [
     "pylance==0.7.4",
-    "ratelimiter",
-    "retry",
-    "tqdm",
+    "ratelimiter~=1.0",
+    "retry>=0.9.2",
+    "tqdm>=4.1.0",
     "aiohttp",
-    "pydantic",
-    "attrs",
+    "pydantic>=1.10",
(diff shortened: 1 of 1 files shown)


chebbyChefNEQ (approved):
Do we need version on aiohttp and cachetools?

changhiskhan:
I think people have complained about compat with databricks but maybe that was the attr vs attrs thing? How recent are these?

wjones127:
> Do we need version on aiohttp and cachetools?

These haven't come up recently, and I wasn't sure if there were any unit tests that exercised these anyways.

> I think people have complained about compat with databricks but maybe that was the attr vs attrs thing?
Partly yeah:

https://github.com/lancedb/lancedb/issues/512#issuecomment-1732180017

PyArrow and Pandas came up, but that's not even a strong dependency. Not sure what went on there TBH.

PR #516, opened 2023-09-26
[DOCS][PYTHON] Update embeddings API docs & Example

AyushExel:
This PR adds an overview of embeddings docs:
- 2 ways to vectorize your data using lancedb - explicit & implicit
- explicit - manually vectorize your data using `wit_embedding` function
- Implicit - automatically vectorize your data as it comes by ingesting your embedding function details as table metadata
- Multi-modal example w/ disappearing embedding function


prrao87:
Hi @AyushExel I made some edits to other docs in #531 (maintaining your structure as per this PR), so hope that works!

I presume you'd like to keep things clean with edits to the docs so I just thought I'd let you know.

docs/src/embeddings/default_embedding_functions.md
```diff
@@ -0,0 +1,162 @@
...
+registry = EmbeddingFunctionRegistry.get_instance()
+func = registry.get("sentence-transformers").create(device="cpu")
+
+class Words(LanceModel):
+    text: str = func.SourceField()
+    vector: Vector(func.ndims()) = func.VectorField()
+
+table = db.create_table("words", schema=Words)
+table.add(
+    pd.DataFrame(
```
changhiskhan, line 27: let's use list of dict instead for these default examples

docs/src/embeddings/embedding_functions.md
```diff
@@ -0,0 +1,82 @@
...
+table = db.create_table("pets", schema=Pets)
+
+```
+That's it! We have ingested all the information needed to embed source and query inputs. We can now forget about the model and dimension details and start to build or VectorDB
+
+### Step 4 - Ingest lots of data and run vector search!
+Now you can just add the data and it'll be vectorized automatically
+
+```python
+table.add(pd.DataFrame({"image_uri": uris}))
```
changhiskhan, line 43: i guess use list of dict instead? like `[{"image_uri": u} for u in uris]` or smth without the dataframe

changhiskhan (commented):
I forgot how much overlap there is between the two notebooks. Might make sense to either merge them or make them more different?

docs/src/embeddings/embedding_functions.md
AyushExel: done

AyushExel:
Yeah, the 2nd notebook is pretty much the same, it wasn't added in the navigation. Removed it. I think we need another non-multimodal notebook example but it can come later as well

changhiskhan (approved):
Please fix CI before merging

PR #517, opened 2023-09-29
Events manager & posthog telemetry 

AyushExel:
Depends on - https://github.com/lancedb/lancedb/pull/492
Current approach:
* All events are batched and sent once every time the `rate_limit` is crossed. All events past `max_events` limit are dropped within each `rate_limit` timeframe. This means that we're not capturing the exact usage and we'll need to compare relative usage
* Currently the `rate_limit` is set to 60seconds, meaning there will be 1 request made after each 60 seconds. `max_events` is set to 25 which means maximum 25 events will be captures at a time and will be dropped past that. These numbers will need to be tuned according to our needs.

Introduced Events class to track events without disrupting any workflow. Allows setting rate limits.
EDIT:
- [ ] Ohh need to turn on sentry integration too


python/lancedb/utils/events.py
```diff
@@ -0,0 +1,135 @@
...
+        metadata (dict): A dictionary containing metadata about the environment.
+        enabled (bool): A flag to enable or disable Events based on certain conditions.
+    """
+
+    url = "https://app.posthog.com/capture/"
+    headers = {"Content-Type": "application/json"}
+    api_key = "phc_oENDjGgHtmIDrV6puUiFem2RB4JA8gGWulfdulmMdZP"
+    # This api-key is write only and is safe to expose in the codebase.
+
+    def __new__(cls):
```
AyushExel, line 39: This might not be needed. Just used as a precaution if some module manually initializes the class

python/lancedb/table.py
```diff
@@ -490,6 +491,7 @@ def create_index(
             replace=replace,
         )
         self._reset_dataset()
+        EVENTS("create_index")
```
changhiskhan, line 494: Would it make more sense to have this as a decorator? Not a blocker for the first version though

python/lancedb/utils/events.py
```diff
@@ -0,0 +1,136 @@
...
+    is_online,
+    is_pip_package,
+    is_pytest_running,
+    threaded_request,
+)
+
+
+class Events:
+    """
+    A class for collecting anonymous event analytics. Event analytics are enabled when sync=True in config and
```
changhiskhan, line 24: Same comment on the naming of "sync"

python/lancedb/utils/general.py
```diff
@@ -298,3 +301,78 @@ def yaml_print(yaml_file: Union[str, Path, dict]) -> None:
...
+        return True
+
+
+def threaded_request(
+    method, url, retry=3, timeout=30, thread=True, code=-1, verbose=True, **kwargs
+):
+    """
+    Makes an HTTP request using the 'requests' library, with exponential backoff retries up to a specified timeout.
+
+    Args:
```
changhiskhan, line 330: Let's use the same docstring format throughout the codebase? (Ie numpy style doc strings).

python/lancedb/utils/general.py
```diff
@@ -298,3 +301,78 @@ def yaml_print(yaml_file: Union[str, Path, dict]) -> None:
...
+            try:
+                m = response.json().get("message", "No JSON message.")
+            except AttributeError:
+                m = "Unable to read JSON."
+            if i == 0:
+                if response.status_code in retry_codes:
+                    m += f" Retrying {retry}x for {timeout}s." if retry else ""
+                elif response.status_code == 429:  # rate limit
+                    m = f"Rate limit reached"
+                if verbose:
```
changhiskhan, line 365: Would it make sense to remove this parameter and just do logger.info? By default the python log level is warning and if the user sets log level to info they will start seeing these messages. It's possible we may want to put them at debug level even since these are for debugging

python/lancedb/utils/general.py
```diff
@@ -298,3 +301,78 @@ def yaml_print(yaml_file: Union[str, Path, dict]) -> None:
...
+                    LOGGER.warning(f"{response.status_code} #{code}")
+                if response.status_code not in retry_codes:
+                    return response
+            time.sleep(2**i)  # exponential standoff
```
changhiskhan, line 369: In the first loop i=0 so it would not sleep. Is that what you want?

python/lancedb/__init__.py
```diff
@@ -69,3 +69,6 @@ def connect(
             raise ValueError(f"api_key is required to connected LanceDB cloud: {uri}")
         return RemoteDBConnection(uri, api_key, region, host_override)
     return LanceDBConnection(uri)
+
+
+EVENTS = events.Events()
```
changhiskhan, line 74: Should this be wrapped in a try/except? The constructors has a bunch of initialization code. Not sure what's the probability of an exception occurring there?

python/lancedb/table.py
```diff
@@ -673,6 +676,7 @@ def search(
             and also the "_distance" column which is the distance between the query
             vector and the returned vector.
         """
+        EVENTS("search")
```
changhiskhan, line 679: For a serious user, wouldn't we expect them to always hit the rate limit since search is a very frequent event? Does this make the analytics less useful?

python/lancedb/table.py
```diff
@@ -776,6 +780,7 @@ def create(
         if data is not None:
             table.add(data)
 
+        EVENTS("create_table")
```
changhiskhan, line 783: Potential places to instrument:

1. LanceDB.Connect
2. Drop_table
3. Delete
4. Update
5. Merge
6. Restore

python/lancedb/utils/general.py
AyushExel: yeah so sleep time increases exponentially, right?:
2**0 = 1
2**1 = 2
2**2 = 4
...
and so on

python/lancedb/utils/general.py
AyushExel: Yeah it's up to us. The thinking here is that user would only get more confused if they somehow saw http errors/interference in their terminal along with the actual lancedb error making it confusing to debug + how much of the analytics stuff we'd want to keep silent. WDYT?

python/lancedb/table.py
AyushExel: Yes a serious user is meant to hit the limits or we'll start making large network requests. We'd definitely have to treat these as the representation of usage and not the exact usage. 

It's a choice between - we don't track event-X OR out of the registered events, which is an approximation of actual usage, event-X grew by Y%.
Does that make sense?

AyushExel: Or if we want to make things a little complex we can start setting event level rate limiting.. The tradeoff then would be that dashboarding will get complex as the scales will be different.

python/lancedb/__init__.py
AyushExel: I think exceptions should not occur here because more of the conditions in the constructor are simply calling os.environ.get() which should be exception safe.

python/lancedb/table.py
AyushExel: Yeah totally depends on out taste.

python/lancedb/__init__.py
AyushExel: This seems exception safe

python/lancedb/utils/events.py
AyushExel: Which comment are you referring to here ?

python/lancedb/table.py
changhiskhan: i think it's fine for now. my main concern is just that search is likely the only event to run into this issue.
but let's keep it simple in this version and let's see what the actual data looks like first.

python/lancedb/__init__.py
changhiskhan: main concern is just that this is in the critical path for `import lancedb` so any error means the user will not be able to use lancedb at all

AyushExel: SG!

python/lancedb/table.py
AyushExel: SG!

changhiskhan (approved):
lgtm. thanks!

Issue #488, opened 2023-09-11
Expose file compaction


wjones127:
We'll have file compaction soon in Lance. We should expose this in LanceDB.
We'll also want to expose clean up and document the caveats around the timestamps.

PR #518, opened 2023-09-29
feat: cleanup and compaction

wjones127:
#488

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -466,14 +466,21 @@ export class LocalTable<T = number[]> implements Table<T> {
    * @param olderThan The minimum age in minutes of the versions to delete. If not
    *                  provided, defaults to two weeks.
    * @param deleteUnverified Because they may be part of an in-progress
-   *                  transaction, files newer than 7 days old are not deleted
-   *                  by default. If you are sure that there are no in-progress
-   *                  transactions, then you can set this to `true` to delete all
-   *                  files older than `older_than`.
+   *                  transaction, uncommitted files newer than 7 days old are
+   *                  not deleted by default. This means that failed transactions
+   *                  can leave around data that takes up disk space for up to
+   *                  7 days. You can override this safety mechanism by setting
+   *                 this option to `true`, only if you promise there are no
+   *                 in progress writes while you run this operation. Failure to
+   *                 uphold this promise can lead to corrupted tables.
    * @returns
    */
-  async cleanup_old_versions (olderThan?: number, deleteUnverified?: boolean): Promise<CleanupStats> {
+  async cleanupOldVersions (olderThan?: number, deleteUnverified?: boolean): Promise<CleanupStats> {
     return tableCleanupOldVersions.call(this._tbl, olderThan, deleteUnverified)
+      .then((res: { newTable: any, metrics: CleanupStats }) => {
+        this._tbl = res.newTable
+        return res.metrics
+      })
   }
 
   /**
@@ -487,9 +494,13 @@ export class LocalTable<T = number[]> implements Table<T> {
    *               for most tables.
    * @returns Metrics about the compaction operation.
    */
-  async compact_files (options?: CompactionOptions): Promise<CompactionMetrics> {
+  async compactFiles (options?: CompactionOptions): Promise<CompactionMetrics> {
     const optionsArg = options ?? {}
     return tableCompactFiles.call(this._tbl, optionsArg)
+      .then((res: { newTable: any, metrics: CompactionMetrics }) => {
+        this._tbl = res.newTable
+        return res.metrics
+      })
   }
 }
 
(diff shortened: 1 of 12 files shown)


node/src/index.ts
```diff
@@ -459,6 +459,73 @@ export class LocalTable<T = number[]> implements Table<T> {
...
+   *                  transactions, then you can set this to `true` to delete all
+   *                  files older than `older_than`.
+   * @returns
+   */
+  async cleanup_old_versions (olderThan?: number, deleteUnverified?: boolean): Promise<CleanupStats> {
+    return tableCleanupOldVersions.call(this._tbl, olderThan, deleteUnverified)
+  }
+
+  async compact_files (options?: CompactionOptions): Promise<CompactionMetrics> {
+    const optionsArg = (options != null) || {}
```
westonpace, line 480:
```suggestion
    const optionsArg = options || {}
```
This will use `options` unless it is `null` or `undefined`.  The way it is currently written `optionsArg` will either be given the value `{}` or the value `(options != null)`, which is `true`.

node/src/index.ts
```diff
@@ -459,6 +459,84 @@ export class LocalTable<T = number[]> implements Table<T> {
...
+  /**
+   * Clean up old versions of the table, freeing disk space.
+   *
+   * @param olderThan The minimum age in minutes of the versions to delete. If not
+   *                  provided, defaults to two weeks.
+   * @param deleteUnverified Because they may be part of an in-progress
+   *                  transaction, files newer than 7 days old are not deleted
+   *                  by default. If you are sure that there are no in-progress
+   *                  transactions, then you can set this to `true` to delete all
+   *                  files older than `older_than`.
```
westonpace, line 472: Do you think this detail is something that a lancedb user is going to want to internalize.  We should still delete most files if this is false anyways.  The only things that really get missed are leftover files from aborted transactions which should (hopefully) be quite rare.

rust/ffi/node/src/index/vector.rs
```diff
@@ -78,9 +78,12 @@ fn get_index_params_builder(
 
             num_partitions.map(|np| {
                 let max_iters = max_iters.unwrap_or(50);
-                let mut ivf_params = IvfBuildParams::default();
-                ivf_params.num_partitions = np;
-                ivf_params.max_iters = max_iters;
+                let ivf_params = IvfBuildParams {
+                    num_partitions: np,
+                    max_iters,
+                    centroids: None,
```
westonpace, line 84: Isn't `centroids: None` part of the `Default` impl?  Why override?

python/lancedb/table.py
```diff
@@ -394,6 +395,17 @@ def checkout(self, version: int):
             raise ValueError(f"Invalid version {version}")
         self._reset_dataset(version=version)
 
+        try:
+            # Accessing the property updates the cached value
+            _ = self._dataset
+        except Exception as e:
+            if "not found" in str(e):
+                raise ValueError(
+                    f"Version {version} no longer exists. Was it cleaned up?"
```
rok, line 404: Maybe print the oldest version currently available?

node/src/index.ts
```diff
@@ -459,6 +459,84 @@ export class LocalTable<T = number[]> implements Table<T> {
...
+   * If true, fragments that have rows that are deleted may be compacted to
+   * remove the deleted rows. This can improve the performance of queries.
+   */
+  materializeDeletions?: boolean
+  /**
+   * A number between 0 and 1, representing the proportion of rows that must be
+   * marked deleted before a fragment is a candidate for compaction to remove
+   * the deleted rows.
+   */
+  materializeDeletionsThreshold?: number
```
westonpace, line 527: Do you want to mention defaults for these parameters?

node/src/index.ts
```diff
@@ -459,6 +459,84 @@ export class LocalTable<T = number[]> implements Table<T> {
...
+   * the number of cores on the machine.
+   */
+  numThreads?: number
+}
+
+export interface CompactionMetrics {
+  fragments_added: number
+  fragments_removed: number
+  files_removed: number
+  files_added: number
```
westonpace, line 539: These are pretty self explanatory but maybe go ahead and document anyways?  Also, a less advance user might not know the difference between "fragments" and "files"

node/src/test/test.ts
```diff
@@ -446,3 +446,34 @@ describe('WriteOptions', function () {
...
+    })
+    assert.equal(compactionMetrics.fragments_removed, 2)
+    assert.equal(compactionMetrics.fragments_added, 1)
+    assert.equal(await table.countRows(), 3)
+
+    await table.cleanup_old_versions()
+    assert.equal(await table.countRows(), 3)
+
+    await table.cleanup_old_versions(0, true)
+    assert.equal(await table.countRows(), 3)
```
westonpace, line 477: Maybe assert the result of `cleanup_old_versions`?

rust/vectordb/src/table.rs
```diff
@@ -305,6 +308,41 @@ impl Table {
         self.dataset = Arc::new(dataset);
         Ok(())
     }
+
+    /// Removed old versions of the dataset from disk.
```
westonpace, line 312:
```suggestion
    /// Remove old versions of the dataset from disk.
```

westonpace (approved):
A few suggestions, but none of them major

node/src/index.ts
wjones127: I think providing some way to cleanup aborted transactions is important. Basically the TLDR for this options is "If I choose True here, I promise there aren't any ongoing writes happening".

rust/ffi/node/src/index/vector.rs
wjones127: Merge conflict. i can simplify.

PR #519, opened 2023-09-29
feat: add an option for prefilter to vectordb, update to lance 0.8

westonpace:


westonpace:
I was able to add a unit test for the python implementation.  I didn't find much unit testing around the query capabilities in the rust bindings and so I left it alone.

westonpace:
Switched PR to only target the rust vectordb

rust/vectordb/src/table.rs
```diff
@@ -190,8 +190,8 @@ impl Table {
     pub async fn create_index(&mut self, index_builder: &impl VectorIndexBuilder) -> Result<()> {
         use lance::index::DatasetIndexExt;
 
-        let dataset = self
-            .dataset
+        let mut dataset = self.dataset.as_ref().clone();
```
westonpace, line 193: This is due to https://github.com/lancedb/lance/pull/1118

changhiskhan (commented):
i wonder if it would be better to have it as an option for now, like `.filter(..., prefilter=true)`. Mainly it's easier to change/remove a kwargs than a separate method. It also makes it more consistent with how the python API works.

PR #522, opened 2023-09-30
feat(python): expose prefilter to lancedb

changhiskhan:
We have experimental support for prefiltering (without ANN) in pylance. This means that we can now apply a filter BEFORE vector search is performed. This can be done via the `.where(filter_string, prefilter=True)` kwargs of the query.

Limitations:
- When connecting to LanceDB cloud, `prefilter=True` will raise NotImplemented
- When an ANN index is present, `prefilter=True` will raise NotImplemented
- This option is not available for full text search query
- This option is not available for empty search query (just filter/project)

Additional changes in this PR:
- Bump pylance version to v0.8.0 which supports the experimental prefiltering.

### DIFF
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -38,6 +38,9 @@ class Query(pydantic.BaseModel):
     # sql filter to refine the query with
     filter: Optional[str] = None
 
+    # if True then apply the filter before vector search
+    is_prefilter: bool = False
+
     # top k results to return
     k: int
 
@@ -162,7 +165,7 @@ class LanceQueryBuilder(ABC):
             for row in self.to_arrow().to_pylist()
         ]
 
-    def limit(self, limit: int) -> LanceVectorQueryBuilder:
+    def limit(self, limit: int) -> LanceQueryBuilder:
         """Set the maximum number of results to return.
 
         Parameters
@@ -172,13 +175,13 @@ class LanceQueryBuilder(ABC):
 
         Returns
         -------
-        LanceVectorQueryBuilder
+        LanceQueryBuilder
             The LanceQueryBuilder object.
         """
         self._limit = limit
         return self
 
-    def select(self, columns: list) -> LanceVectorQueryBuilder:
+    def select(self, columns: list) -> LanceQueryBuilder:
         """Set the columns to return.
 
         Parameters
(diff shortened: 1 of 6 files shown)


python/lancedb/query.py
```diff
@@ -329,6 +334,30 @@ def to_arrow(self) -> pa.Table:
         )
         return self._table._execute_query(query)
 
+    def where(self, where: str, prefilter: bool = False) -> LanceVectorQueryBuilder:
```
eddyxu, line 337: Why this one returns `LanceVectorQueryBuilder` while the other function returns `LanceQueryBUilder`?

python/lancedb/remote/table.py
```diff
@@ -98,6 +98,8 @@ def search(
         return LanceVectorQueryBuilder(self, query, vector_column_name)
 
     def _execute_query(self, query: Query) -> pa.Table:
+        if query.is_prefilter:
+            raise NotImplementedError("Cloud support for prefiltering is coming soon")
```
westonpace, line 102: Related, if you want to take a look: https://github.com/lancedb/lancedb/pull/519

python/tests/test_query.py
```diff
@@ -97,6 +98,25 @@ def test_query_builder_with_filter(table):
...
+    df = (
+        LanceVectorQueryBuilder(table, [0, 0], "vector")
+        .where("id = 2", prefilter=True)
+        .limit(1)
+        .to_df()
+    )
+    assert df["id"].values[0] == 2
+    assert all(df["vector"].values[0] == [3, 4])
+
+
```
westonpace, line 119: In the mock assert on line 156 you need to pass `prefilter=False` since the mock doesn't recognize the call otherwise.  Although the tests are passing so maybe not.  I just know I ended having to make that change.

python/lancedb/query.py
```diff
@@ -38,6 +38,9 @@ class Query(pydantic.BaseModel):
     # sql filter to refine the query with
     filter: Optional[str] = None
 
+    # if True then apply the filter before vector search
+    is_prefilter: bool = False
```
westonpace, line 42: Very minor nit: do you want to just name this `prefilter` since that is what is used everywhere else?

westonpace (approved):
I am also a little confused between `LanceVectorQueryBuilder` and `LanceQueryBuilder`.

python/lancedb/query.py
changhiskhan: this overrides the superclass method since prefilter is only a thing for vector search right now

python/lancedb/remote/table.py
changhiskhan: yeah, we can remove this once the new thing is deployed. Had feedback from users that the gap between Cloud and OSS in features is confusing.

PR #531, opened 2023-10-02
[DOCS][PYTHON]  Update docs for clarity

prrao87:
I only modified those docs pages that are untouched in existing unmerged PRs, so hopefully there are no merge conflicts!

1. The `tantivy-py` version specified in the docs doesn't work (pip install fails), but with the latest version of pip and wheel installed on my Mac, I was able to just `pip install tantivy` and FTS works great for me. I updated the docs page to include this in 7ca4b757ce2d56a17c0f1fa53df2bafcf5e45dc5 but can always modify to another specific version in case this breaks any tests.
2. The `.add()` method for Python should take in a list of dicts as the first option (to also align with the JS API), and additionally, users can pass an existing pandas DataFrame to add to a table. Hope this makes sense.
3. I've had multiple conversations with users who are unclear that the terms "exhaustive", "flat" and "KNN" are all the same kind of search, so I've updated the verbiage of this section to clarify this.
4. Fixed typos and improved clarity in the ANN indexes page.

### DIFF
--- a/docs/src/ann_indexes.md
+++ b/docs/src/ann_indexes.md
@@ -6,7 +6,7 @@ LanceDB provides many parameters to fine-tune the index's size, the speed of que
 
 Currently, LanceDB does *not* automatically create the ANN index.
 LanceDB has optimized code for KNN as well. For many use-cases, datasets under 100K vectors won't require index creation at all.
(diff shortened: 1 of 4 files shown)


AyushExel (approved):

AyushExel:
Looks good! Will merge once tests pass

PR #537, opened 2023-10-04
implement mirroring object store

chebbyChefNEQ:
This PR implements a mirroring object store and allows and table to be mirrored to a local path when param `mirroredStore` is set in the url

### DIFF
--- a/rust/vectordb/src/io/object_store.rs
+++ b/rust/vectordb/src/io/object_store.rs
@@ -1,0 +1,397 @@
+// Copyright 2023 Lance Developers.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A mirroring object store that mirror writes to a secondary object store
+
+
+use std::{
+    fmt::Formatter,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+use bytes::Bytes;
+use futures::{stream::BoxStream, FutureExt, StreamExt};
+use lance::io::object_store::WrappingObjectStore;
+use object_store::{
+    path::Path, GetOptions, GetResult, ListResult, MultipartId, ObjectMeta, ObjectStore, Result,
+};
+
+use async_trait::async_trait;
+use tokio::{
+    io::{AsyncWrite, AsyncWriteExt},
+    task::JoinHandle,
+};
+
+#[derive(Debug)]
+struct MirroringObjectStore {
+    primary: Arc<dyn ObjectStore>,
+    secondary: Arc<dyn ObjectStore>,
+}
+
+impl std::fmt::Display for MirroringObjectStore {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "MirrowingObjectStore")?;
+        writeln!(f, "primary:")?;
+        self.primary.fmt(f)?;
+        writeln!(f, "secondary:")?;
+        self.secondary.fmt(f)?;
+        Ok(())
+    }
+}
+
+trait PrimaryOnly {
+    fn primary_only(&self) -> bool;
+}
+
+impl PrimaryOnly for Path {
+    fn primary_only(&self) -> bool {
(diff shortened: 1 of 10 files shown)


rust/vectordb/src/io/object_store.rs
```diff
@@ -0,0 +1,317 @@
...
+}
+
+impl IsManifest for Path {
+    fn is_manifest(&self) -> bool {
+        self.to_string().contains("manifest")
+    }
+}
+
+/// An object store that mirrors write to secondsry object store first
+/// and than commit to primary object store
```
wjones127, line 49: Is the secondary to be used as the cache?

chebbyChefNEQ: Yeah, the secondary is expected to be some kind of less-durable but lower latency storage. I'll add a note

rust/vectordb/src/io/object_store.rs
```diff
@@ -0,0 +1,380 @@
+use std::{
```
eddyxu, line 1: Add license header.

Btw, why this one is in lancedb instead of lance

rust/vectordb/src/io/object_store.rs
```diff
@@ -0,0 +1,380 @@
...
+};
+
+use async_trait::async_trait;
+use tokio::{
+    io::{AsyncWrite, AsyncWriteExt},
+    task::JoinHandle,
+};
+
+#[derive(Debug)]
+struct MirrowingObjectStore {
```
eddyxu, line 22: Should we call this as pass through ObjectStore?

rust/vectordb/src/io/object_store.rs
```diff
@@ -0,0 +1,380 @@
...
+/// and than commit to primary object store.
+///
+/// This is meant to mirrow writes to a less-durable but lower-latency
+/// store. We have primary store that is durable but slow, and a secondary
+/// store that is fast but not asdurable
+///
+/// Note: this object store does not mirror writes to *.manifest files
+#[async_trait]
+impl ObjectStore for MirrowingObjectStore {
+    async fn put(&self, location: &Path, bytes: Bytes) -> Result<()> {
```
eddyxu, line 58: Lets make this more generic? Get a function to test whether we should write to primary / Secondary or not.

```rust

if should_write_to_slow_storage(&location) {
   self.secondary.put(..)
}

self.primary.put(..)?
```

rust/vectordb/src/io/object_store.rs
eddyxu: Btw, primary / secondary which one is fast / caching, which one is slow / persist storage?

rust/vectordb/src/io/object_store.rs
```diff
@@ -0,0 +1,380 @@
...
+        if location.is_manifest() {
+            self.primary.put(location, bytes).await
+        } else {
+            self.secondary.put(location, bytes.clone()).await?;
```
eddyxu, line 62: What is consistency mechansim here? if we only write secondary but not primary what will happen

chebbyChefNEQ: we filter out manifest writes to primary only, so if they two goes out of sync, the manifest won't show us and the out of sync object won't be used

rust/vectordb/src/io/object_store.rs
chebbyChefNEQ: primary is the slow one.

I added some comments here
https://github.com/lancedb/lancedb/pull/537/files#diff-801c5c16cbc0082436e711a6275900a60849b2b01e2c8589c9b579579f81abedR48-R55

rust/vectordb/src/io/object_store.rs
```diff
@@ -0,0 +1,380 @@
...
+
+    async fn abort_multipart(&self, location: &Path, multipart_id: &MultipartId) -> Result<()> {
+        if location.is_manifest() {
+            return self.primary.abort_multipart(location, multipart_id).await;
+        }
+
+        self.secondary.abort_multipart(location, multipart_id).await
+    }
+
+    // Reads are routed to primary only
```
wjones127, line 96: Why are reads only routed to primary? Isn't that the slow one? What's the point of mirroring if we always just read the slow one?

rust/vectordb/src/io/object_store.rs
chebbyChefNEQ: Made a `PrimaryOnly` trait

rust/vectordb/src/io/object_store.rs
chebbyChefNEQ: the secondary isn't garueeteed to have the file as it is expected to have garbage collection and/or less durable.

We have a implementation for reading the secondary store internally here: https://github.com/lancedb/sophon/blob/main/src/phalanx/src/cache/prewarmed.rs

wjones127 (approved):
Looks good now. Nice job on this :)

PR #538, opened 2023-10-05
feat: telemetry, error tracking, CLI & config manager

AyushExel:


python/lancedb/cli/cli.py
```diff
@@ -0,0 +1,33 @@
...
+@click.version_option(help="LanceDB command line interface entry point")
+def cli():
+    "LanceDB command line interface"
+
+
+diagnostics_help = """
+Enable or disable LanceDB diagnostics. When enabled, LanceDB will send anonymous events to help us improve LanceDB.
+These diagnostics are used only for error reporting and no data is collected. You can find more about diagnosis on
+our docs: https://lancedb.github.io/lancedb/diagnostics
+"""
```
AyushExel, line 16: TODO: this link is wrong.

docs/src/cli_config.md
```diff
@@ -0,0 +1,37 @@
...
+lancedb config
+```
+These config parameters can be tuned using the cli.
+```
+lancedb {config_name} --{argument}
+```
+
+## LanceDB Opt-in Diagnostics
+When enabled, LanceDB will send anonymous events to help us improve LanceDB. These diagnostics are used only for error reporting and no data is collected. Error & stats allow us to automates certain aspects of bug reporting, prioritization of fixes and feature requests.
+These diagnostics are opt-in and can be enabled or disabled using the `lancedb diagnostics` command. These are enabled by default.
```
AyushExel, line 25: "automates" -> "automate"

python/lancedb/utils/config.py
```diff
@@ -0,0 +1,103 @@
...
+        path = Path.home() / ".config" / sub_dir
+    else:
+        raise ValueError(f"Unsupported operating system: {platform.system()}")
+
+    # GCP and AWS lambda fix, only /tmp is writeable
+    if not is_dir_writeable(path.parent):
+        LOGGER.warning(
+            f"WARNING ⚠️ user config directory '{path}' is not writeable, defaulting to '/tmp' or CWD."
+            "Alternatively you can define a LANCEDB_CONFIG_DIR environment variable for this path."
+        )
```
AyushExel, line 36: Document this?

python/lancedb/utils/config.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Manages lancedb config stored in a YAML file.
+
+    Args:
+        file (str | Path): Path to the lancedb config YAML file. Default is USER_CONFIG_DIR / 'config.yaml'.
+    """
+
+    def __init__(self, file=CONFIG_FILE):
+        self.file = Path(file)
+        self.defaults = {  # Default global config values
```
AyushExel, line 63: possibly in another PR but make this a module level global variable

changhiskhan (approved):
just realized there are no tests for these.
could you add tests when you get a chance?

for click: https://click.palletsprojects.com/en/8.1.x/testing/
for sentry/posthog: use pytest-mock

you'll need to have some option to turn on diagnostics during CI but it should only send to the mock endpoint and then you can assert the event attributes is what you expect.

python/tests/test_telemetry.py
```diff
@@ -0,0 +1,47 @@
...
+
+def mock_request(
+    method, url, retry=3, timeout=30, thread=True, code=-1, verbose=True, **kwargs
+):
+    assert method.lower() == "post"
+    json = kwargs.get("json", {})
+    # TODO: don't hardcode these here. Instead create a module level json scehma in lancedb.utils.events for better evolvability
+    batch_keys = ["api_key", "distinct_id", "batch"]
+    event_keys = ["event", "properties", "timestamp", "distinct_id"]
+    property_keys = ["cli", "install", "platforms", "version", "session_id", "blud"]
```
AyushExel, line 17: Added loose tests for now here. Ideally we could hardcode the json schema as global var and directly assert here but I don't want to edit the reviewed files so will do in another PR

AyushExel:
addressed testing in above comment^

AyushExel:
Wanted to use [responses]( https://pypi.org/project/responses/) for requests testing but I don't think adding dependency for a single test would be the best idea. Take a look and merge if all looks good.

PR #543, opened 2023-10-08
Use query.limit(..) in README

ankrgyl:
If you run the README javascript example in typescript, it complains that the type of limit is a function and cannot be set to a number.


changhiskhan (commented):
I forgot does query.limit(20) modify in place or return a new instance. If the latter it should be query = table.search(...).limit(20) ?

ankrgyl:
Updated to add `limit(...)` after `.search(...)`, and made the limit `2` (to be consistent with python)

changhiskhan (approved):
thanks!

PR #550, opened 2023-10-10
Add cohere embedding function

AyushExel:


python/lancedb/embeddings/__init__.py
```diff
@@ -12,6 +12,7 @@
 #  limitations under the License.
 
 
+from .cohere import CohereEmbeddingFunction
```
changhiskhan, line 15: is it worth separating out the existing embedding functions into their own modules? e.g., `openai.py`, `openclip.py`, `sentence_transformers.py`?

python/lancedb/embeddings/cohere.py
```diff
@@ -0,0 +1,90 @@
...
+    ----------
+    name: str, default "embed-multilingual-v2.0"
+        The name of the model to use. See the Cohere documentation for a list of available models.
+
+    Examples
+    --------
+    import lancedb
+    from lancedb.pydantic import LanceModel, Vector
+    from lancedb.embeddings import EmbeddingFunctionRegistry
+    import pandas as pd
```
changhiskhan, line 40: can we just use list of dict here instead? just to make the default examples less pandas-centric.

python/lancedb/embeddings/cohere.py
```diff
@@ -0,0 +1,90 @@
...
+
+    cohere = EmbeddingFunctionRegistry.get_instance().get("cohere").create(name="embed-multilingual-v2.0")
+
+    class TextModel(LanceModel):
+        text: str = cohere.SourceField()
+        vector: Vector(cohere.ndims()) =  cohere.VectorField()
+
+    df = pd.DataFrame({"text": ["hello world", "goodbye world"]})
+    db = lancedb.connect("~/lancedb")
```
changhiskhan, line 49: `~/.lancedb`

python/lancedb/embeddings/cohere.py
```diff
@@ -0,0 +1,90 @@
...
+            The texts to embed
+        """
+        # TODO retry, rate limit, token limit
+        self._init_client()
+        rs = CohereEmbeddingFunction.client.embed(texts=texts, model=self.name)
+
+        return [emb for emb in rs.embeddings]
+
+    def _init_client(self):
+        cohere = self.safe_import("cohere")
```
changhiskhan, line 86: can you add this to the optional dependencies?

python/lancedb/embeddings/cohere.py
```diff
@@ -0,0 +1,90 @@
...
+    name: str = "embed-multilingual-v2.0"
+    client: ClassVar = None
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._ndims = None
+
+    def ndims(self):
+        if self._ndims is None:
+            self._ndims = len(self.generate_embeddings(["foo"])[0])
```
changhiskhan, line 65: this requires a remote api call which is a bit unfortunate. Can we just hard code it for now like we do with openai?

python/lancedb/embeddings/cohere.py
```diff
@@ -0,0 +1,90 @@
...
+    ) -> List[np.array]:
+        """
+        Get the embeddings for the given texts
+
+        Parameters
+        ----------
+        texts: list[str] or np.ndarray (of str)
+            The texts to embed
+        """
+        # TODO retry, rate limit, token limit
```
changhiskhan, line 79: yeah this can be done in a separate PR where we use the code in the `with_embeddings` function for both this and the openai functions

python/lancedb/embeddings/__init__.py
AyushExel: yeah I think so. The scope and size of the file functions.py will become too large otherwise.

python/lancedb/embeddings/cohere.py
AyushExel: Done. But what does this accomplish tho? If a user uses a different model, won't hardcoding error out?

changhiskhan (approved):

PR #553, opened 2023-10-11
[Docs] Improve visibility of table ops 

AyushExel:
A little verbose, but better than being non-discoverable 
Screenshot from 2023-10-11 16-26-02


changhiskhan (approved):
thanks!

Issue #551, opened 2023-10-10
Switch Python examples to be row-based, like Node


wjones127:

changhiskhan:
This has been addressed by #554

PR #554, opened 2023-10-11
docs: switch python examples to be row based

rok:
See #551


python/lancedb/db.py
```diff
@@ -106,6 +106,10 @@ def create_table(
         ...    "lat": [45.5, 40.1],
         ...    "long": [-122.7, -74.1]
         ... })
+        >>> data = pd.DataFrame([
```
changhiskhan, line 109: we assign to `data` twice here?

python/lancedb/table.py
```diff
@@ -292,7 +292,9 @@ def delete(self, where: str):
         --------
         >>> import lancedb
         >>> import pandas as pd
-        >>> data = pd.DataFrame({"x": [1, 2, 3], "vector": [[1, 2], [3, 4], [5, 6]]})
+        >>> data = pd.DataFrame([
```
changhiskhan, line 295: for delete and create, let's make the examples use list of dict instead of pandas dataframe

python/lancedb/db.py
rok: Ah, sorry bad change, fixed.

python/lancedb/table.py
rok: Changed.

changhiskhan (approved):
pending CI

wjones127 (changes requested):
I think this isn't quite what we want. The row-based examples shouldn't use Pandas at all, just list of dict. And when we do use Pandas, being column-oriented is fine.

docs/src/guides/tables.md
```diff
@@ -41,11 +41,10 @@ A Table is a collection of Records in a LanceDB Database. You can follow along o
...
 
-    data = pd.DataFrame({
-        "vector": [[1.1, 1.2, 1.3, 1.4], [0.2, 1.8, 0.4, 3.6]],
-        "lat": [45.5, 40.1],
-        "long": [-122.7, -74.1]
-    })
+    data = pd.DataFrame([
+        {"vector": [1.1, 1.2, 1.3, 1.4], "lat": 45.5, "long": -122.7},
+        {"vector": [0.2, 1.8, 0.4, 3.6], "lat": 40.1, "long": -74.1},
+    ])
```
wjones127, line 47: Let's leave this one as is. pandas dataframes make sense to be columnar.

docs/src/guides/tables.md
```diff
@@ -308,7 +304,9 @@ Use the `delete()` method on tables to delete rows from a table. To choose which
     import lancedb
     import pandas as pd
 
-    data = pd.DataFrame({"x": [1, 2, 3], "vector": [[1, 2], [3, 4], [5, 6]]})
+    data = pd.DataFrame([{"x": 1, "vector": [1, 2]},
+                         {"x": 2, "vector": [3, 4]},
+                         {"x": 3, "vector": [5, 6]}])
```
wjones127, line 309: My intention with being row-based is we would completely remove pandas from those examples. So something like:
```suggestion
    data = [{"x": 1, "vector": [1, 2]},
            {"x": 2, "vector": [3, 4]},
            {"x": 3, "vector": [5, 6]}]
```

python/lancedb/context.py
```diff
@@ -33,11 +33,20 @@ def contextualize(raw_df: "pd.DataFrame") -> Contextualizer:
...
+    ...    {'token': 'fox', 'document_id': 1},
+    ...    {'token': 'jumped', 'document_id': 1},
+    ...    {'token': 'over', 'document_id': 1},
+    ...    {'token': 'the', 'document_id': 1},
+    ...    {'token': 'lazy', 'document_id': 1},
+    ...    {'token': 'dog', 'document_id': 1},
+    ...    {'token': 'I', 'document_id': 2},
+    ...    {'token': 'love', 'document_id': 2},
+    ...    {'token': 'sandwiches', 'document_id': 2},
+    ... ])
```
wjones127, line 49: Let's leave this example as-is.

python/lancedb/table.py
```diff
@@ -292,7 +292,9 @@ def delete(self, where: str):
         --------
         >>> import lancedb
         >>> import pandas as pd
```
wjones127, line 294: I think we don't need the pandas import anymore.

python/lancedb/table.py
```diff
@@ -825,7 +829,9 @@ def update(self, where: str, values: dict):
...
-        >>> data = pd.DataFrame({"x": [1, 2, 3], "vector": [[1, 2], [3, 4], [5, 6]]})
+        >>> data = pd.DataFrame([
+        ...   {"x": 1, "vector": [1, 2]}, {"x": 2, "vector": [3, 4]}, {"x": 3, "vector": [5, 6]}
+        ... ])
```
wjones127, line 834: Can we remove the `pd.DataFrame` and pandas import here?

changhiskhan:
> I think this isn't quite what we want. The row-based examples shouldn't use Pandas at all, just list of dict. And when we do use Pandas, being column-oriented is fine.

@wjones127 i think some of these are specifically within the section on pandas integration so those are ok. But I agree that in general we should default to list-of-dict

wjones127:
> i think some of these are specifically within the section on pandas integration so those are ok. But I agree that in general we should default to list-of-dict

Yeah I agree many of these places should stay as Pandas and are fine as is. So my complaint there is this PR changes them to do row-wise construction with Pandas, which is weird. We should leave the Pandas-specific examples alone, and focus on removing Pandas from places where Pandas isn't relevant.

docs/src/guides/tables.md
rok: Reverted.

python/lancedb/context.py
rok: Reverted.

python/lancedb/table.py
rok: Removed.

rok:
Thanks for the feedback, fully agreed. I've pushed changes, please see if everything is addressed.

wjones127 (approved):

PR #560, opened 2023-10-13
feat: expose use_index in LanceDB OSS

### LINKED ISSUE #535: Add `bypass_vector_index` to Python sync API
We added this parameter to Node and Async Python in https://github.com/lancedb/lancedb/pull/1113.

* [ ] Add this to the Python sync API.
* [ ] Add a quick user guide on how to measure recall of a vector index using this.

This functionality is critical for debugging and measuring recall.

wjones127:
Closes #535

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -559,7 +559,8 @@ class LanceTable(Table):
             The data to insert into the table.
         mode: str
             The mode to use when writing the data. Valid values are
-            "append" and "overwrite".
+            "append", which inserts new rows, and "overwrite", which replaces
+            the entire content of the table with the new rows.
         on_bad_vectors: str, default "error"
             What to do if any of the vectors are not the same size or contains NaNs.
             One of "error", "drop", "fill".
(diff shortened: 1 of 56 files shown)


python/lancedb/table.py
```diff
@@ -559,7 +559,8 @@ def add(
             The data to insert into the table.
         mode: str
             The mode to use when writing the data. Valid values are
-            "append" and "overwrite".
+            "append", which inserts new rows, and "overwrite", which replaces
+            the entire content of the table with the new rows.
```
wjones127, line 563: Drive-by change based on user feedback in discord. In retrospect, `Table.add(mode="overwrite")` is a little confusing, since `add` implies append. Perhaps we should make a separate `Table.overwrite()` method?

wjones127:
It appears in Node that Cloud will just ignore the `use_index` parameter here:

https://github.com/lancedb/lancedb/blob/683824f1e9e08d666e590067d4c26d23ee74f954/node/src/remote/index.ts#L96-L104

Maybe I should do that in Python too?

PR #567, opened 2023-10-14
implement remote api calls for table mutation

chebbyChefNEQ:
Add more APIs to remote table for Node SDK
* `add` rows
* `overwrite` table with rows
* `create` table

This has been tested against dev stack

### DIFF
--- a/node/src/remote/index.ts
+++ b/node/src/remote/index.ts
@@ -18,8 +18,10 @@ import {
 } from '../index'
 import { Query } from '../query'
 
-import { Vector } from 'apache-arrow'
+import { Vector, Table as ArrowTable } from 'apache-arrow'
 import { HttpLancedbClient } from './client'
+import { isEmbeddingFunction } from '../embedding/embedding_function'
+import { createEmptyTable, fromRecordsToStreamBuffer, fromTableToStreamBuffer } from '../arrow'
 
 /**
  * Remote connection.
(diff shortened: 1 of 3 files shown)


node/src/remote/index.ts
```diff
@@ -66,8 +68,60 @@ export class RemoteConnection implements Connection {
     }
   }
 
-  async createTable<T> (name: string | CreateTableOptions<T>, data?: Array<Record<string, unknown>>, optsOrEmbedding?: WriteOptions | EmbeddingFunction<T>, opt?: WriteOptions): Promise<Table<T>> {
-    throw new Error('Not implemented')
+  async createTable<T> (nameOrOpts: string | CreateTableOptions<T>, data?: Array<Record<string, unknown>>, optsOrEmbedding?: WriteOptions | EmbeddingFunction<T>, opt?: WriteOptions): Promise<Table<T>> {
```
westonpace, line 71: It doesn't look like you handle `writeOptions` (either as `opt` or `optsOrEmbedding`) at all.  Is this intentional?

westonpace (approved):
I'm not super familiar with the JS stuff (either lance or arrow) so just the one question

node/src/remote/index.ts
chebbyChefNEQ: this is meant to support create by overwrite. Since we can overwrite by insert already, I just skipped this for now.

PR #574, opened 2023-10-17
list table pagination draft

QianZhu:


python/lancedb/remote/client.py
```diff
@@ -151,10 +151,13 @@ async def post(
...
 
     @_check_not_closed
-    async def list_tables(self):
+    async def list_tables(self, limit: int, page_token: str):
         """List all tables in the database."""
-        json = await self.get("/v1/table/", {})
-        return json["tables"]
+        try:
+            json = await self.get("/v1/table/", {"limit": limit, "page_token": page_token})
+            return False, json["tables"]
```
chebbyChefNEQ, line 158: we can just determine if we are done by an empty `json["table"]`. if the returned data is empty there is no more to list

python/lancedb/remote/db.py
```diff
@@ -52,10 +52,18 @@ def __init__(
...
+        page_token = ""
+        while True:
+            done, result = self._loop.run_until_complete(
+                self._client.list_tables(limit, page_token)
+            )
+            if len(result) > 0:
+                page_token = result[len(result) - 1]
+            if done or len(result) == 0:
+                break
+            yield result
```
chebbyChefNEQ, line 66: nit:
```
for item in result:
    yield result
```

python/lancedb/remote/db.py
```diff
@@ -52,10 +52,18 @@ def __init__(
     def __repr__(self) -> str:
         return f"RemoveConnect(name={self.db_name})"
 
-    def table_names(self) -> List[str]:
+    def table_names(self, limit: int):
```
chebbyChefNEQ, line 55: nit: type annotation `Iterator[str]`

python/lancedb/remote/client.py
```diff
@@ -151,10 +151,15 @@ async def post(
...
         """List all tables in the database."""
-        json = await self.get("/v1/table/", {})
-        return json["tables"]
+        try:
+            json = await self.get(
+                "/v1/table/", {"limit": limit, "page_token": page_token}
+            )
+            return json["tables"]
+        except StopAsyncIteration:
+            return None
```
chebbyChefNEQ, line 162: Should return [] or the caller might crash assuming this is a list

chebbyChefNEQ (approved):

PR #578, opened 2023-10-17
feat: add timezone handling for datetime in pydantic

changhiskhan:
If you add timezone information in the Field annotation for a datetime then that will now be passed to the pyarrow data type.

I'm not sure how pyarrow enforces timezones, right now, it silently coerces to the timezone given in the column regardless of whether the input had the matching timezone or not. This is probably not the right behavior. Though we could just make it so the user has to make the pydantic model do the validation instead of doing that at the pyarrow conversion layer.

### DIFF
--- a/python/tests/test_table.py
+++ b/python/tests/test_table.py
@@ -142,6 +142,7 @@ def test_add(db):
 
 
 def test_add_pydantic_model(db):
+<<<<<<< HEAD
     # https://github.com/lancedb/lancedb/issues/562
 
     class Metadata(BaseModel):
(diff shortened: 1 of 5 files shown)


python/tests/test_table.py
```diff
@@ -144,8 +146,15 @@ def test_add_pydantic_model(db):
...
-
-    data = TestModel(vector=list(range(16)), li=[1, 2, 3])
+        dt: datetime
+        dt_with_tz: datetime = Field(tz="Asia/Kolkata")
+
+    data = TestModel(
+        vector=list(range(16)),
+        li=[1, 2, 3],
+        dt=datetime.now(),
+        dt_with_tz=datetime.now(tz=pytz.timezone("Asia/Kolkata")),
```
eddyxu, line 156: do we want to check whether tz is appropriately serialized

changhiskhan: If you put in an invalid timezone pytz will complain. If you put in a tz that doesn't match the schema, then the current behavior is that it'll just get coerced into the timezone that matches the schema.
I was thinking we should ask the user to explicitly add validation to the pydantic model, rather than spreading the validation logic across a bunch of different places. That isn't ideal because if you're not careful you can certainly end up with bad data. But putting this adhoc validator in LanceDB also doesn't feel right 🤷‍♂️

eddyxu (approved):

PR #589, opened 2023-10-21
feat: include manifest files in mirrow store

chebbyChefNEQ:
we now only exclude `_latest.manifest` from replication

### DIFF
--- a/rust/vectordb/src/io/object_store.rs
+++ b/rust/vectordb/src/io/object_store.rs
@@ -57,7 +57,7 @@ trait PrimaryOnly {
 
 impl PrimaryOnly for Path {
     fn primary_only(&self) -> bool {
-        self.to_string().contains("manifest")
+        self.filename().unwrap_or("") == "_latest.manifest"
     }
 }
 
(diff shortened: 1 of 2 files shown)


rust/vectordb/src/io/object_store.rs
```diff
@@ -118,8 +118,10 @@ impl ObjectStore for MirroringObjectStore {
         self.primary.head(location).await
     }
 
-    // garbage collection on secondary will happen async from other means
     async fn delete(&self, location: &Path) -> Result<()> {
+        if !location.primary_only() {
+            self.secondary.delete(location).await?;
```
changhiskhan, line 123: if the secondary delete fails then we also don't do the primary delete? is that intentional?

rust/vectordb/src/io/object_store.rs
```diff
@@ -142,6 +144,9 @@ impl ObjectStore for MirroringObjectStore {
     }
 
     async fn copy_if_not_exists(&self, from: &Path, to: &Path) -> Result<()> {
+        if !to.primary_only() {
+            self.secondary.copy(from, to).await?;
```
changhiskhan, line 148: same question.

rust/vectordb/src/io/object_store.rs
chebbyChefNEQ: delete is usually just for cleaning up tmp files or garbage collect. This is okay if we fail on secondary

rust/vectordb/src/io/object_store.rs
chebbyChefNEQ: This is intentional, if we fail to write to secondary, we just fail the whole operation, because we can't guarantee the replication if secondary fails.

we write secondary first for consistency, as we list primary for discovering versions.

changhiskhan (approved):

PR #593, opened 2023-10-23
feat: add checkout method to table to reuse existing store and connections

chebbyChefNEQ:
Prior to this PR, to get a new version of a table, we need to re-open the table. This has a few downsides w.r.t. performance:
* Object store is recreated, which takes time and throws away existing warm connections
* Commit handler is thrown aways as well, which also may contain warm connections


changhiskhan (approved):
lgtm

PR #597, opened 2023-10-24
fix(python): fix multiple embedding functions bug

### LINKED ISSUE #594: when using 2 different embedding functions on two different columns the resultant embeddings are identical
i have two seperate embedding functions and am embedding a name and address one for each embedding function, but when i do a full text search on name , it works, but i noticed that the two embeddings for address and name are exactly the same, the same applies for if i use the same embedding function for the different columns

here is my data model:

```
class Users(LanceModel):
    CITY: str
    STATE: str
    ZIP: str 
    NAME: str = snt_xfmr.SourceField() #source col
    ADDRESS: str = snt_xfmr2.SourceField() #source col2
    NAME_VECTOR: Vector(snt_xfmr.ndims()) = snt_xfmr.VectorField() #vectorized
    ADDRESS_VECTOR: Vector(snt_xfmr2.ndims()) = snt_xfmr2.VectorField() #vectorized2
```

changhiskhan:
Closes #594

The embedding functions are pydantic models so multiple instances with the same parameters are considered ==, which means that if you have multiple embedding columns it's possible for the embeddings to get overwritten. Instead we use `is` instead of == to avoid this problem.

testing: modified unit test to include this case

### DIFF
--- a/python/lancedb/pydantic.py
+++ b/python/lancedb/pydantic.py
@@ -327,7 +327,12 @@ class LanceModel(pydantic.BaseModel):
         for vec, func in vec_and_function:
             for source, field_info in cls.safe_get_fields().items():
                 src_func = get_extras(field_info, "source_column_for")
-                if src_func == func:
+                if src_func is func:
+                    # note we can't use == here since the function is a pydantic
+                    # model so two instances of the same function are ==, so if you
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
Looks good. I think you just have to run formatting.

PR #602, opened 2023-10-25
feat: expose optimize index api

chebbyChefNEQ:
expose `optimize_index` api.


eddyxu (approved):
LGTM.
Should we consolidate optimization tasks into same API in the future ? Cc @wjones127

PR #603, opened 2023-10-25
expose remap index api

chebbyChefNEQ:
expose index remap options in `compact_files`


eddyxu (approved):

westonpace:
There are no index remapper options today right?  Why did we need to expose it?

chebbyChefNEQ:
> There are no index remapper options today right? Why did we need to expose it?

we use the rust version directly in our internal repo. This isn't exposed to JS but we do use it in our internal code

PR #611, opened 2023-10-30
doc: add doc to use GPU for indexing

eddyxu:


changhiskhan (commented):
Might be worth adding a quick note like:
1. If you see X it means you need pytorch installed
2. If you see Y it means you installed torch cpu but you want to use cuda etc

QianZhu (approved):

PR #612, opened 2023-10-31
Multi-task instructor model with quantization support & weak_lru cache for embedding function models

### LINKED ISSUE #608: Multitask Instructor embedding function support + example
- Instructor model -  https://github.com/xlang-ai/instructor-embedding

AyushExel:
resolves #608


python/tests/test_embeddings_slow.py
```diff
@@ -162,3 +162,19 @@ class TextModel(LanceModel):
...
+@pytest.mark.slow
+def test_instructor_embedding():
+    model = get_registry().get("instructor").create()
+
+    class TextModel(LanceModel):
+        text: str = model.SourceField()
+        vector: Vector(model.ndims()) = model.VectorField()
+
+    df = pd.DataFrame({"text": ["hello world", "goodbye world"]})
+    db = lancedb.connect("~/lancedb")
```
changhiskhan, line 176: please pass in `tmp_path` fixture and use that instead of home.

python/lancedb/embeddings/base.py
```diff
@@ -125,6 +137,12 @@ def VectorField(self, **kwargs):
         """
         return Field(json_schema_extra={"vector_column_for": self}, **kwargs)
 
+    def __eq__(self, __value: object) -> bool:
+        return vars(self) == vars(__value)
```
changhiskhan, line 141: if the input does not have a __dict__ attr then this raises a TypeError. e.g., try `vars("foo")`

python/lancedb/embeddings/base.py
```diff
@@ -125,6 +137,12 @@ def VectorField(self, **kwargs):
         """
         return Field(json_schema_extra={"vector_column_for": self}, **kwargs)
 
+    def __eq__(self, __value: object) -> bool:
```
changhiskhan, line 140: where are you using eq/hash btw?

python/lancedb/embeddings/instructor.py
```diff
@@ -0,0 +1,137 @@
...
+
+    name: str = "hkunlp/instructor-base"
+    batch_size: int = 32
+    device: str = "cpu"
+    show_progress_bar: bool = True
+    normalize_embeddings: bool = True
+    quantize: bool = False
+    # convert_to_numpy: bool = True # Hardcoding this as numpy can be ingested directly
+
+    source_instruction: str = "represent the docuement for retreival"
```
changhiskhan, line 91: "document" and "retrieval"

python/lancedb/embeddings/instructor.py
```diff
@@ -0,0 +1,137 @@
...
+    query_instruction: str = (
+        "represent the document for retreiving the most similar documents"
```
changhiskhan, line 93: "retrieving"

python/lancedb/embeddings/instructor.py
```diff
@@ -0,0 +1,137 @@
...
+
+from .base import TextEmbeddingFunction
+from .registry import register
+from .utils import TEXT, weak_lru
+
+
+@register("instructor")
+class InstructorEmbeddingFunction(TextEmbeddingFunction):
+    """
+    An embedding function that uses the InstructorEmbedding library. Instructor models support multi-task learning, and can be used for a
```
changhiskhan, line 25: Would be great for you to do a 5-10 min preso on instructor embeddings that can show how it's different from other embedding models and when people should be using it.

python/lancedb/embeddings/base.py
changhiskhan: oic you're using it with the weak references.

python/lancedb/embeddings/base.py
AyushExel: Yeah it should fail with any object that does not have a `__dict__` attr. All class instances should have a dict attribute. This is like overriding equlity operator so it's only supposed to be called with an object of same type.

```
class A

a = A()
b = A()
a == b
```
Used by hash function which is in turn used for creating weak references for caching

python/lancedb/embeddings/base.py
AyushExel: yea

changhiskhan (approved):
niiiiice

AyushExel:
Slow tests are failing in local CI. I'll take a look later and fix them before merging

PR #614, opened 2023-10-31
Exponential standoff retry support for handling rate limited embedding functions

### LINKED ISSUE #579: Support rate limiting in embeddings API
Most LLM APIs and their derivates have some form of rate limiting. The trail versions for testing have smaller limits. When using new new embeddings API, the calls to model APIs is made implicitly when the data is added to the tables. There are some ways to manually rate limit by sleeping for a few seconds when adding data in smaller batches/individual rows. But if when adding data in larger batches using the ingested `EmebddingFunction` instance, there is no way to prevent  hitting the rate limit. 

There are 2 types of rate limits that we could potentially support:
* **Requests level** - Allow user to set the RPM(request per minute) when initializing the `EmbeddingFunction` instance. It keeps the rolling count of requests made in the last 60 seconds and sleeps if the limit occurs. Each call to `EmbeddingFunction.generate_embeddings` can be assumed to be 1 batched request.
* **Token level** - We should simply provide an interface for the user to handle this case themselves as handling this on a lower level can be a bit tricky, as there are 2 cases. 1 - TRM occurs because the combines token of multiple texts is more than the limit. This can be handled same as the request limit by waiting out. but in case 2 - the TRL can be exceeded by a single text, so the solution can be to either chunk it or maybe just truncate everything after the token limit. And token limit can be applied in combination with request limit.

something like this
```
cohere = EmbeddingFunctionRegistry().get_instance("model_name", rate_limit=10, token_limit=1000000)
```

AyushExel:
Users ingesting data using rate limited apis don't need to manually make the process sleep for counter rate limits
resolves #579

### DIFF
--- a/python/lancedb/embeddings/utils.py
+++ b/python/lancedb/embeddings/utils.py
@@ -169,7 +169,7 @@ def retry_with_exponential_backoff(
     initial_delay: float = 1,
     exponential_base: float = 2,
     jitter: bool = True,
-    max_retries: int = 10,
+    max_retries: int = 7,
     # errors: tuple = (),
 ):
     """Retry a function with exponential backoff.
--- a/python/lancedb/embeddings/utils.py
+++ b/python/lancedb/embeddings/utils.py
@@ -14,8 +14,8 @@
 import math
 import random
 import socket
-import time
 import sys
+import time
 import urllib.error
 from typing import Callable, List, Union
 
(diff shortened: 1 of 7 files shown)


rok (approved):

python/lancedb/embeddings/utils.py
```diff
@@ -162,6 +164,55 @@ def _chunker(arr):
...
+            # should check the error message to be sure
+            except Exception as e:
+                num_retries += 1
+
+                if num_retries > max_retries:
+                    raise Exception(
+                        f"Maximum number of retries ({max_retries}) exceeded."
+                    )
+
+                delay *= exponential_base * (1 + jitter * random.random())
```
rok, line 209: Average exponent here will be 2.5, so last delay with 10 retries could easily be hours. How about setting default `max_retries` to 7 so max delay is in tens of minutes?

AyushExel: great point!

PR #617, opened 2023-10-31
added api docs for prefilter flag

albertlockett:
Added the prefilter flag argument to the `LanceQueryBuilder.where`.

This should make it display here:
https://lancedb.github.io/lancedb/python/python/#lancedb.query.LanceQueryBuilder.select

And also in intellisense like this:
image

Also adds some improved documentation about the `where` argument to this method.

### DIFF
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -226,13 +226,22 @@ class LanceQueryBuilder(ABC):
         self._columns = columns
         return self
 
-    def where(self, where) -> LanceQueryBuilder:
+    def where(self, where: str, prefilter: bool = False) -> LanceQueryBuilder:
         """Set the where clause.
 
         Parameters
         ----------
         where: str
-            The where clause.
+            The where clause which is a valid SQL where clause. See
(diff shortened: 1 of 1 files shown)


python/lancedb/query.py
```diff
@@ -226,20 +226,30 @@ def select(self, columns: list) -> LanceQueryBuilder:
...
+            The where clause which is a valid SQL where clause. See
+            `Lance filter pushdown <https://lancedb.github.io/lance/read_and_write.html#filter-push-down>`_
+            for valid SQL expressions.
+        prefilter: bool, default False
+            If True, apply the filter before vector search, otherwise the
+            filter is applied on the result of vector search.
+            This feature is **EXPERIMENTAL** and may be removed and modified
+            without warning in the future. Currently this is only supported
+            in OSS and can only be used with a table that does not have an ANN
+            index.
```
westonpace, line 244:
```suggestion
            without warning in the future.
```
This restriction is no longer there.  Can you fix the other comment too?

westonpace (approved):
One suggestion, but otherwise thank you for adding this.

PR #618, opened 2023-10-31
increment pylance

wjones127:


eddyxu (commented):
Do we need / want to upgrade rust / nodejs as well

wjones127:
@eddyxu should we be concerned about the illegal instruction on macos? https://github.com/lancedb/lancedb/actions/runs/6713644210/job/18245531807?pr=618

eddyxu (approved):

PR #619, opened 2023-11-01
fix!: sort table names

albertlockett:
https://github.com/lancedb/lance/issues/1385

### DIFF
--- a/python/lancedb/db.py
+++ b/python/lancedb/db.py
@@ -263,7 +263,7 @@ class LanceDBConnection(DBConnection):
         return self._uri
 
     def table_names(self) -> list[str]:
-        """Get the names of all tables in the database.
+        """Get the names of all tables in the database. The names are sorted.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


python/lancedb/db.py
```diff
@@ -287,6 +287,7 @@ def table_names(self) -> list[str]:
             for file_info in paths
             if file_info.extension == "lance"
         ]
+        tables.sort()
```
westonpace, line 290: Let's mention in the method description that table names are returned in sorted order.  Also, this is technically a breaking change.  Let's [label it as such](https://www.conventionalcommits.org/en/v1.0.0/).

westonpace (approved):
Thanks!  Just two minor suggestions and then this is good to go I think

python/lancedb/db.py
albertlockett: done & done. Thanks @westonpace

PR #634, opened 2023-11-07
[Python] External Integrations tests

AyushExel:


changhiskhan (commented):
Are these fast enough to run in GHA? Would it make sense to create a new GHA to always run these on every PR?

Also some merge conflicts need to be resolved?

PR #640, opened 2023-11-08
fix saas open_table and table_names issues

QianZhu:
- added check whether a table exists in SaaS open_table
- remove prefilter not supported warning in SaaS search
- fixed issues for SaaS table_names

### DIFF
--- a/python/lancedb/remote/client.py
+++ b/python/lancedb/remote/client.py
@@ -152,16 +152,11 @@ class RestfulLanceDBClient:
 
     @_check_not_closed
     async def list_tables(
-        self, limit: int, page_token: Optional[str] = None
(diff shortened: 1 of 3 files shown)


python/lancedb/remote/client.py
```diff
@@ -152,16 +152,11 @@ async def post(
 
     @_check_not_closed
     async def list_tables(
-        self, limit: int, page_token: Optional[str] = None
+        self, limit: int, page_token: Optional[str] = ""
```
eddyxu, line 155: Does `page_token=None` break anything? We should prob fix where it breaks.

`Optional[blabla]` does expect the default value as `None`.

eddyxu (approved):

PR #644, opened 2023-11-10
feat: add RemoteTable.version in Python

rok:
Please note: this is not tested as we don't have a server here and testing against a mock object wouldn't be that interesting.


chebbyChefNEQ (approved):
we can test using dev stage I'll post the code snippet in slack

PR #649, opened 2023-11-14
SaaS create_index API

aidangomar:


python/lancedb/remote/table.py
```diff
@@ -72,7 +72,17 @@ def create_index(
...
+
+        data = {
+            "column": vector_column_name,
+            "index_type": index_type,
+            "metric_type": metric,
+        }
+        resp = self._conn._loop.run_until_complete(
+            self._conn._client.post(f"/v1/table/{self._name}/create_index/", data=data)
+        )
+        return resp
```
QianZhu, line 86: can we add a test for this API ?

python/lancedb/remote/table.py
```diff
@@ -72,7 +72,17 @@ def create_index(
         replace: bool = True,
         accelerator: Optional[str] = None,
     ):
-        raise NotImplementedError
```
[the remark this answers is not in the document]
aidangomar: done!

python/lancedb/remote/table.py
aidangomar: Added the test, but for it to run properly on gha we might need to add the env variables to github secrets, which I don't think I have perms to do
Also attempted adding the test to the integration tests instead, but that didn't seem to want to use the non-published version of lancedb

albertlockett (approved):

python/lancedb/remote/table.py
albertlockett: I don't think we've generally tested the remote implementations in this test suite. This is a limitation of the node SDK as well.

albertlockett:
Should this be implemented on the node sdk as well?

https://github.com/lancedb/lancedb/blob/797514bcbf9dd4432a8685540f1bd29d25fac49d/node/src/remote/index.ts#L235-L237

PR #653, opened 2023-11-16
[broken] upgrade lance to 0.8.16

chebbyChefNEQ:
upgrade lance to 0.8.15

lance 0.8.15 removed `aws_region` configurability from `store_options`. This breaks aws auth on the write path for lancedb, as there is no way to pass in `aws_region` any more. We need to fix this before releasing lancedb.


wjones127:
Closing in favor of #656

PR #675, opened 2023-12-01
feat: enable prefilter in node js

chebbyChefNEQ:
enable prefiltering in node js, both native and remote

### DIFF
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -102,6 +102,20 @@ describe('LanceDB client', function () {
       assertResults(results)
     })
 
+    it('should correctly process prefilter/postfilter', async function () {
+      const uri = await createTestDB(16, 300)
+      const con = await lancedb.connect(uri)
+      const table = await con.openTable('vectors')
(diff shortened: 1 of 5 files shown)


node/src/test/test.ts
```diff
@@ -102,6 +102,20 @@ describe('LanceDB client', function () {
...
+    it('should correctly process prefilter/postfilter', async function () {
+      const uri = await createTestDB(16, 300)
+      const con = await lancedb.connect(uri)
+      const table = await con.openTable('vectors')
+      await table.createIndex({ type: 'ivf_pq', column: 'vector', num_partitions: 2, max_iters: 2, num_sub_vectors: 2 })
+      // post filter should return less than the limit
+      let results = await table.search(new Array(16).fill(0.1)).limit(10).filter('id >= 0 AND id <= 10').prefilter(false).execute()
+      assert.isTrue(results.length < 10)
+
+      // pre filter should return less than the limit
```
wjones127, line 114:
```suggestion
      // pre filter should return exactly the limit
```

wjones127 (approved):
Small nit, otherwise looks good.

PR #693, opened 2023-12-08
feat: LocalTable for vectordb now supports filters without vector search

changhiskhan:
Note this currently the filter/where is only implemented for LocalTable so that it requires an explicit cast to "enable" (see new unit test). The alternative is to add it to the Table interface, but since it's not available on RemoteTable this may cause some user experience issues.

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -489,6 +489,16 @@ export class LocalTable<T = number[]> implements Table<T> {
   }
 
   /**
+   * Creates a filter query to find all rows matching the specified criteria
+   * @param value The filter criteria (like SQL where clause syntax)
+   */
+  filter (value: string): Query<T> {
+    return new Query(undefined, this._tbl, this._embeddings).filter(value)
+  }
+
+  where = this.filter
+
+  /**
    * Insert records into this Table.
    *
    * @param data Records to be inserted into the Table
--- a/rust/vectordb/src/query.rs
+++ b/rust/vectordb/src/query.rs
@@ -24,7 +24,7 @@ use crate::error::Result;
 /// A builder for nearest neighbor queries for LanceDB.
 pub struct Query {
     pub dataset: Arc<Dataset>,
-    pub query_vector: Float32Array,
+    pub query_vector: Option<Float32Array>,
     pub limit: usize,
     pub filter: Option<String>,
     pub select: Option<Vec<String>>,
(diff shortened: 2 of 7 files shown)


node/src/index.ts
```diff
@@ -426,6 +426,16 @@ export class LocalTable<T = number[]> implements Table<T> {
     return new Query(query, this._tbl, this._embeddings)
   }
 
+  /**
+   * Creates a filter query to find all rows matching the specified criteria
+   * @param value The filter criteria (like SQL where clause syntax)
+   */
+  filter (value: string): Query<T> {
+    return new Query(undefined, this._tbl, this._embeddings).filter(value)
```
westonpace, line 434: Does `table.filter("id != 2")` only return 10 results?  This is creating a brand new `Query` and I think in the constructor  `_limit` defaults to 10.  As a user, this wouldn't be obvious from reading the description of this method.

westonpace (approved):
Just one question but this looks good otherwise.

node/src/index.ts
changhiskhan: nice catch. I removed the default limit if the vector query is not specified. This way users aren't surprised later on. Would you mind taking another look?

rust/vectordb/src/query.rs
```diff
@@ -69,11 +69,13 @@ impl Query {
...
-            &self.query_vector,
-            self.limit,
-        )?;
+        if let Some(query) = self.query_vector.as_ref() {
+            // If there is a vector query, default to limit=10 if unspecified
+            scanner.nearest(crate::table::VECTOR_COLUMN_NAME, query, self.limit.unwrap_or(10))?;
+        } else if let Some(limit) = self.limit {
+            // If there is no vector query, it's ok to not have a limit
+            scanner.limit(Some(limit as i64), None)?;
+        }
```
westonpace, line 78: Minor nit:  I think this can maybe be simplified to...

```
} else {
  scanner.limit(self.limit,None)?;
}
```

but I'm not entirely certain it's the same thing.

westonpace (approved):

rust/vectordb/src/query.rs
changhiskhan: i changed it to `scanner.limit(self.limit.map(|limit| limit as i64, None)?`

PR #696, opened 2023-12-08
Update in Node & Rust

albertlockett:


Cargo.toml
```diff
@@ -5,10 +5,10 @@ exclude = ["python"]
...
 
 [workspace.dependencies]
-lance = { "version" = "=0.8.17", "features" = ["dynamodb"] }
-lance-index = { "version" = "=0.8.17" }
-lance-linalg = { "version" = "=0.8.17" }
-lance-testing = { "version" = "=0.8.17" }
+lance = { "version" = "=0.8.19", "features" = ["dynamodb"] }
+lance-index = { "version" = "=0.8.19" }
+lance-linalg = { "version" = "=0.8.19" }
+lance-testing = { "version" = "=0.8.19" }
```
wjones127, line 11: We should skip v0.8.19 and go straight to 0.8.20. 0.8.19 has a serious bug that corrupts tables.

node/src/index.ts
```diff
@@ -261,6 +261,39 @@ export interface Table<T = number[]> {
    */
   delete: (filter: string) => Promise<void>
 
+  /**
+   * Update rows in this table.
+   *
+   * This can be used to update a single row, many rows, all rows, or
+   * sometimes no rows (if your predicate matches nothing).
+   *
+   * @param filter  A filter in the same format used by a sql WHERE clause. The
```
wjones127, line 270:
```suggestion
   * @param filter  A filter in the same format used by a SQL WHERE clause. The
```

node/src/index.ts
```diff
@@ -261,6 +261,39 @@ export interface Table<T = number[]> {
...
+   *                filter may be empty, in which case all rows will be updated.
```
wjones127, line 271: More specifically, `null` right?
```suggestion
   *                filter may be null, in which case all rows will be updated.
```

node/src/index.ts
```diff
@@ -481,6 +514,19 @@ export class LocalTable<T = number[]> implements Table<T> {
     return tableDelete.call(this._tbl, filter).then((newTable: any) => { this._tbl = newTable })
   }
 
+  /**
+   * Update rows in this table.
+   *
+   * @param filter A filter for the rows to update, in the same format used by a sql
+   *              WHERE clause.
```
wjones127, line 521:
```suggestion
   * @param filter A filter for the rows to update, in the same format used by a SQL
   *              WHERE clause.
```

node/src/index.ts
```diff
@@ -261,6 +261,39 @@ export interface Table<T = number[]> {
...
+   *
+   * let results = yield tbl.search([1, 1]).execute();
+   * // Returns [
+   * //   {id: 2, vector: [2, 2], name: 'Michael'}
+   * //   {id: 1, vector: [3, 3], name: 'Ye'}
+   * // ]
+   * ```
+   *
+   */
+  update: (filter: string | null, updates: Record<string, string>) => Promise<void>
```
wjones127, line 295: The signature in Python allows literal values by default, and optionally as SQL values. I think we should try to keep it as similar to Python as possible. (Chang already implemented the API this way in Python, so we can't modify it without a breaking change.)

https://github.com/lancedb/lancedb/blob/d7075bc952f67b2d6d820a79cef6b8a9935f0009/python/lancedb/table.py#L909-L915

What do you think of this signature? It uses restructuring to mimic the named parameters of Python.

```ts
    update: ({where, values, values_sql}: {where?: string, values?: Record<string, unknown>, values_sql?: Record<string, string>}) => Promise<void>
```

Also, should we be using `filter: string | null` or just `filter?: string` (which I think is equivalent to `filter: string | undefined`)?

wjones127 (changes requested):
Thanks for working on this. I think the most important thing we need to do here is match the Python API, which already has existing semantics.

node/src/index.ts
albertlockett: Ah OK -- I can go ahead and make a change to the signature so the caller has the option to pass updates as literals or SQL.

wrt to `filter: string | null` vs `filter?: string`.  I wanted to keep `filter` as the first argument as it matches the Python API and an optional argument is not allowed to come before a required arg. e.g.
```js
update: (filter: ?string, updates: Record<string, string>) => Promise<void>
                // ^ not allowed
```
I think it's probably OK to keep it as is. The other option is we could change filter? to be the second arg and make the type `?string`. WDYT?

wjones127: Well in Python we made `values` (the equivalent of `updates`) optional, since we need a separate parameter for `values_sql`. (Otherwise, we couldn't differentiate between a literal string and a SQL string.)

albertlockett: using destructuring for the `update` method also makes it inconsistent with other method on the Table interface. Are we OK with that? E.g. `delete` for exaple doesn't use destructuring for it's args:
```js
  delete: (filter: string) => Promise<void>
```

albertlockett: @wjones127 changed function signature as requested

wjones127 (approved):

PR #704, opened 2023-12-14
ci: fix broken npm publication

wjones127:
Most recent release failed because `release` depends on `node-macos`, but we renamed `node-macos` to `node-macos-{x86,arm64}`. This fixes that by consolidating them back to a single `node-macos` job, which also has the side effect of making the file shorter.


changhiskhan (approved):
Much better than what I had. Thanks!

PR #709, opened 2023-12-15
feat: allow custom column name in query

chebbyChefNEQ:
Allow user to select any column for vector query in rust

### DIFF
--- a/rust/vectordb/src/io/object_store.rs
+++ b/rust/vectordb/src/io/object_store.rs
@@ -359,7 +359,9 @@ mod test {
         assert_eq!(t.count_rows().await.unwrap(), 100);
 
         let q = t
-            .search(Some(PrimitiveArray::from_iter_values(vec![0.1, 0.1, 0.1, 0.1])))
+            .search(Some(PrimitiveArray::from_iter_values(vec![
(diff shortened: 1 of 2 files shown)


rust/vectordb/src/io/object_store.rs
```diff
@@ -359,7 +359,9 @@ mod test {
         assert_eq!(t.count_rows().await.unwrap(), 100);
 
         let q = t
-            .search(Some(PrimitiveArray::from_iter_values(vec![0.1, 0.1, 0.1, 0.1])))
+            .search(Some(PrimitiveArray::from_iter_values(vec![
```
chebbyChefNEQ, line 362: this came from `cargo fmt`

rust/vectordb/src/query.rs
```diff
@@ -25,6 +25,7 @@ use crate::error::Result;
 pub struct Query {
     pub dataset: Arc<Dataset>,
     pub query_vector: Option<Float32Array>,
+    pub column: String,
```
eddyxu, line 28: Would this break existing user ?

eddyxu (approved):

PR #713, opened 2023-12-17
[WIP] Hybrid search in python

changhiskhan:
Proposed v1 hybrid search:

API: table.search(<text>, type="hybrid").rerank(weight=0.5, normalize="rank").limit(10)

Constraints:

- The table to have an embedding function configured
- Raise exception if more than one vector column / more than one fts index is detected

Behavior:

1. Do vector search with limit
2. Do fts search with limit
3. Combine scores and rerank according to rerank.
- weight determines the weight of the vector search score (after normalization).
- normalize can be:
None - just raw scores
"auto" (default) - same as "rank"
"rank" - use the rank of the result as the reranking score
"score" - convert the score to standard normal before combining and reranking.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.3.10
+current_version = 0.3.11
 commit = True
 message = Bump version: {current_version} → {new_version}
 tag = True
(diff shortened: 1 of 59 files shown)


AyushExel:
going to work on top of this branch. I think this makes sense, the only problem is that there isn't a right answer for reranking that fits all-- so 2 maybe we provide:
* One fast but simple reranking method -- the current one
* One slow but more efficient -- Transformers based cross encoders
* Allow custom callable as a reranking function
I guess then it would cover all bases.

Symbolk:
Does Hybrid Search only support Python API now? I failed to find the corresponding JS/TS API.

PR #717, opened 2023-12-18
feat: Node Schema API

aidangomar:


aidangomar:
> How do we want to test methods like this?

there's an integration test suite that's currently in review here https://github.com/lancedb/sophon/pull/1588 which includes a test for schema

PR #720, opened 2023-12-19
feat: Node create index API

aidangomar:


node/src/remote/index.ts
```diff
@@ -236,7 +236,18 @@ export class RemoteTable<T = number[]> implements Table<T> {
   }
 
   async createIndex (indexParams: VectorIndexParams): Promise<any> {
-    throw new Error('Not implemented')
+    const data = {
+      column: indexParams.column,
+      index_type: 'vector',
+      metric_type: indexParams.metric_type
+    }
```
westonpace, line 243: Do we want to make `index_cache_size` configurable to match what we have on the python remote table?

node/src/remote/index.ts
```diff
@@ -236,7 +236,18 @@ export class RemoteTable<T = number[]> implements Table<T> {
   }
 
   async createIndex (indexParams: VectorIndexParams): Promise<any> {
-    throw new Error('Not implemented')
+    const data = {
```
westonpace, line 239: There are a lot of parameters in `indexParams` that we are ignoring here.  We might want to check and raise if they are not set to `undefined`.  Otherwise users might be confused when their parameters have no effect.

node/src/remote/index.ts
aidangomar: done!

node/src/remote/index.ts
aidangomar: done! also added a test for it unsupported params in the node integration test pr

westonpace (approved):

QianZhu (approved):

PR #722, opened 2023-12-20
feat(python): add option to flatten output in to_pandas

changhiskhan:
Closes https://github.com/lancedb/lance/issues/1738

We add a `flatten` parameter to the signature of `to_pandas`. By default this is None and does nothing.
If set to True or -1, then LanceDB will flatten structs before converting to a pandas dataframe. All nested structs are also flattened. If set to any positive integer, then LanceDB will flatten structs up to the specified level of nesting.

### DIFF
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -147,7 +147,7 @@ class LanceSchema(LanceModel):
     vector: Vector(1536)
     payload: Document
 
-# add works
+# Let's add 100 sample rows to our dataset
 data = [LanceSchema(
     id=f"id{i}",
     vector=np.random.randn(1536),
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -119,3 +119,96 @@ This is why it is often called **Approximate Nearest Neighbors (ANN)** search, w
 always returns 100% recall.
 
 See [ANN Index](ann_indexes.md) for more details.
+
+
+### Output formats
+
+LanceDB returns results in many different formats commonly used in python.
+Let's create a LanceDB table with a nested schema:
+
+```python
+import lancedb
+from lancedb.pydantic import LanceModel, Vector
+import numpy as np
+from pydantic import BaseModel
+uri = "data/sample-lancedb-nested"
+
+class Metadata(BaseModel):
+    source: str
+    timestamp: datetime
+
+class Document(BaseModel):
+    content: str
+    meta: Metadata
+
+class LanceSchema(LanceModel):
+    id: str
+    vector: Vector(1536)
+    payload: Document
+
+# add works
+data = [LanceSchema(
+    id=f"id{i}",
+    vector=np.random.randn(1536),
+    payload=Document(
+        content=f"document{i}", meta=Metadata(source=f"source{i%10}", timestamp=datetime.now())
+    ),
+) for i in range(100)]
+
+tbl = db.create_table("documents", data=data)
+```
+
+#### As a pyarrow table
+
+Using `to_arrow()` we can get the results back as a pyarrow Table.
+This result table has the same columns as the LanceDB table, with 
+the addition of an `_distance` column for vector search or a `score`
+column for full text search.
+
+```python
+tbl.search(np.random.randn(1536)).to_arrow()
+```
+
+#### As a pandas dataframe
+
+You can also get the results as a pandas dataframe.
+
+```python
+tbl.search(np.random.randn(1536)).to_pandas()
+```
+
+For a table with a nested schema, you can also tell LanceDB to flatten
+the schema when creating the pandas dataframe.
+
+```python
+tbl.search(np.random.randn(1536)).to_pandas(flatten=True)
+```
+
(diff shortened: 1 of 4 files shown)


prrao87 (approved):
Looks good! I tested the double-nested flatten as you have in the unit test and it works well.

I may have some merge conflicts on the `search.md` but the docs look good too -- will update my version with the changes from this one and modify further in the docs PR, if that works.

wjones127 (approved):

docs/src/search.md
```diff
@@ -118,4 +118,98 @@ However, fast vector search using indices often entails making a trade-off with
...
+class Document(BaseModel):
+    content: str
+    meta: Metadata
+
+class LanceSchema(LanceModel):
+    id: str
+    vector: Vector(1536)
+    payload: Document
+
+# add works
```
westonpace, line 150:
```suggestion
# Let's add 100 sample rows to our dataset
```
Since this is an example I think we can be more verbose in our comments, or we can get rid of this comment entirely.

docs/src/search.md
```diff
@@ -118,4 +118,98 @@ However, fast vector search using indices often entails making a trade-off with
...
+) for i in range(100)]
+
+tbl = db.create_table("documents", data=data)
+```
+
+#### As a pyarrow table
+
+Using `to_arrow()` we can get the results back as a pyarrow Table.
+This result table has the same columns as the LanceDB table, with 
+the addition of an `_distance` column for vector search or a `score`
```
westonpace, line 166:
```suggestion
the addition of a `_distance` column for vector search or a `score`
```

docs/src/search.md
```diff
@@ -118,4 +118,98 @@ However, fast vector search using indices often entails making a trade-off with
...
+
+```python
+tbl.search(np.random.randn(1536)).to_pandas(flatten=True)
+```
+
+If your table has a deeply nested struct, you can control how many levels
+of nesting to flatten by passing in a positive integer.
+
+```python
+tbl.search(np.random.randn(1536)).to_pandas(flatten=1)
```
westonpace, line 192: Does the output of these examples get generated as part of the build?  I don't think so, since this is just markdown and not rst/doctest but having the output of these commands would be very useful to understand the difference here I think.

docs/src/search.md
```diff
@@ -118,4 +118,98 @@ However, fast vector search using indices often entails making a trade-off with
...
+#### As a pandas dataframe
+
+You can also get the results as a pandas dataframe.
+
+```python
+tbl.search(np.random.randn(1536)).to_pandas()
+```
+
+For a table with a nested schema, you can also tell LanceDB to flatten
+the schema when creating the pandas dataframe.
```
westonpace, line 182: We don't need to address this now but a reader might be wondering...

Does pandas not have good support for working with nested columns?  Why doesn't this option exist for `to_arrow`/`to_list`? (it should be obvious that it wouldn't make sense for `to_pydantic`)

python/lancedb/query.py
```diff
@@ -185,14 +185,33 @@ def to_df(self) -> "pd.DataFrame":
...
+        ----------
+        flatten: Optional[Union[int, bool]]
+            If flatten is True, flatten all nested columns.
+            If flatten is an integer, flatten the nested columns up to the
+            specified depth.
+            If unspecified, do not flatten the nested columns.
         """
-        return self.to_arrow().to_pandas()
+        tbl = self.to_arrow()
+        if (flatten == -1) or (flatten is True):
```
westonpace, line 204: Minor nit: maybe `flatten < 0` instead of `flatten == -1`? (or just return an error if flatten is <= 0?, maybe suggest setting it to `True`)

westonpace (approved):
Minor nits but overall this is a good addition, thanks

docs/src/search.md
changhiskhan: i was reading it as "an underscore-distance" column. I'll admit I don't know the formal grammaticaly rules on this :)

docs/src/search.md
changhiskhan: i don't think we have the output anywhere. I'm not sure if mkdocs support running the code and rendering the output. I do agree it would be nice.

docs/src/search.md
changhiskhan: ok added a sentence.

docs/src/search.md
prrao87: Ah, the classic "an SQL query" vs. "a SQL query" 😂

docs/src/search.md
prrao87: I had the same thought about outputs, but I couldn't find an mkdocs feature that does that, so I planned on adding the output examples manually. Alternatively, we could look at having this code in notebook and render that via mkdocs too.

python/lancedb/query.py
changhiskhan: sg, i'll just raise ValueError if flatten is <= 0

PR #723, opened 2023-12-20
feat(python): support nested reference for fts

changhiskhan:
https://github.com/lancedb/lance/issues/1739

Support nested field reference in full text search

### DIFF
--- a/python/lancedb/fts.py
+++ b/python/lancedb/fts.py
@@ -75,8 +75,14 @@ def populate_index(index: tantivy.Index, table: LanceTable, fields: List[str]) -
         The number of rows indexed
     """
     # first check the fields exist and are string or large string type
+    nested = []
     for name in fields:
-        f = table.schema.field(name)  # raises KeyError if not found
+        try:
+            f = table.schema.field(name)  # raises KeyError if not found
+        except KeyError:
+            f = resolve_path(table.schema, name)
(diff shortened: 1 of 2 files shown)


python/lancedb/fts.py
```diff
@@ -98,6 +113,32 @@ def populate_index(index: tantivy.Index, table: LanceTable, fields: List[str]) -
...
+    path = field_name.split(".")
+    if len(path) == 1:
+        return False
+    tbl = schema.empty_table()
+    for _ in range(len(path) - 1):
+        prev = tbl
+        tbl = prev.flatten()
+        if prev == tbl:
+            break
+    return tbl.schema.field(field_name)
```
wjones127, line 139: Can we make this handle one-level and also not create temporary tables?

```suggestion
    path = field_name.split(".")
    field = schema.field(path.pop(0))
    for segment in path:
        if isinstance(field.data_type, pa.StructType):
            field = field.data_type.field(segment)
        else:
            raise KeyError(f"field {field_name} not found in schema {schema}")
    return field
```

wjones127 (approved):

python/lancedb/fts.py
changhiskhan: thanks. needs to be `field.type` and I changed it to use `pa.types.is_struct`.

Issue #673, opened 2023-12-01
`db.table_names()` comes up empty on GCS


wjones127:
From a user report:

```python
>>> import lancedb
>>> db = lancedb.connect("gs://xxxx/lance")
>>> db.table_names()
[]
>>> db.open_table("meta")
LanceTable(meta)
>>> t = db.open_table("meta")
>>> len(t.to_pandas())
11
>>>
```

We use PyArrow filesystems under the hood to introspect the bucket:

https://github.com/lancedb/lancedb/blob/b085d9aaa1ab95c0ba62b479a863bd850c9b0ae6/python/lancedb/db.py#L296-L298

However, PyArrow's `GCSFileSystem` is known to report common prefixes as empty if there isn't a "directory marker": https://github.com/apache/arrow/issues/32403 object-store-rs doesn't create these directory markers.

I suspect that is the underlying issue. We could either create the directory markers or we could use some other mechanism to list the bucket.

sergiocorreia:
FWIW, I get the same error on a Windows installation (Win10+Py3.12), and for a different reason

Given a path `C:/foo/bar.lancedb`,  `get_uri_location()` returns `/foo/bar.lancedb` which then will obviously fail when passed to `filesystem.get_file_info()` because it's not a valid path or file.

If instead of doing

```python
paths = filesystem.get_file_info( 
     fs.FileSelector(get_uri_location(self.uri)) 
 )
```

We do:

```python
paths = filesystem.get_file_info( 
     fs.FileSelector(self.uri)
 )
```

Then the code works correctly, **on Windows**.

westonpace:
The `table_names` feature should be fixed in https://github.com/lancedb/lancedb/pull/1059

fekstr:
`table_names` still returns an empty list for me on GCS

PR #724, opened 2023-12-20
bug(python): fix path handling in windows

### LINKED ISSUE #703: Changes to dataset URI creation for LanceTable
This is a bit of a niche problem that would matter primarily for Windows users, but dataset URI creation within `lancedb.table.LanceTable._dataset_uri` right now makes use of `os.path.join` which, on a Windows machine, means that a connection URI joined with a table name inserts backslashes instead of forward slashes. Would it be possible to either have this section modified to account for different filesystems? Apologies if this is already solvable / fixable via some method that I'm not aware of

changhiskhan:
Use pathlib for local paths so that pathlib
can handle the correct separator on windows.

Closes #703


sergiocorreia:
I also found this related path: https://github.com/lancedb/lancedb/issues/673#issuecomment-1850692449 might be worth it to deal with both in the same PR

python/lancedb/util.py
```diff
@@ -84,6 +85,18 @@ def fs_from_uri(uri: str) -> Tuple[pa_fs.FileSystem, str]:
     return pa_fs.FileSystem.from_uri(uri)
 
 
+def join_uri(base: Union[str, pathlib.Path], *parts: str) -> str:
+    """
+    Join a URI with multiple parts, handling extra environment variables.
```
wjones127, line 90: "extra environment variables"?

wjones127 (commented):
Is there a test we can write for windows to validate this?

python/lancedb/util.py
```diff
@@ -84,6 +85,18 @@ def fs_from_uri(uri: str) -> Tuple[pa_fs.FileSystem, str]:
...
+    """
+    if isinstance(base, pathlib.Path):
```
eddyxu, line 92: Can we add tests?

wjones127:
There seems to already be a test that fails on Windows due to path handling: https://github.com/lancedb/lancedb/actions/runs/7280056572/job/19837672999?pr=729

I have rebased your PR on my PR to add a Windows CI job.

python/lancedb/util.py
changhiskhan: oops that was generated by copilot and I forgot to change it >_<

changhiskhan:
> I also found this related bug: [#673 (comment)](https://github.com/lancedb/lancedb/issues/673#issuecomment-1850692449) might be worth it to deal with both in the same PR

good catch. I added a condition so that if the uri scheme is one letter then just return the uri instead of the netloc + path or path. Lmk if this addresses the cases you've encountered on windows.

wjones127 (approved):

PR #728, opened 2023-12-20
fix: prevent duplicate data in FTS index

### LINKED ISSUE #498: create_fts_index creates duplicates
Reported by catethos on discord

"call the 'create_fts_index' twice will double the index instead of replacing it, i.e. i would get twice of the same row when i do the search"

### LINKED ISSUE #726: Invalid string array when doing FTS
[Reported from user on discord](https://discord.com/channels/1030247538198061086/1101154357778591824/1186539229199749120):

> I'm getting this error when performing fts search with limit 10 
> ```
> {OSError}OSError('LanceError(Arrow): Invalid argument error: Last offset 360 of Utf8 is larger than values length 36, /Users/runner/work/lance/lance/rust/lance-core/src/encodings/binary.rs:397:26')
> ```

This sounds suspiciously like https://github.com/lancedb/lance/issues/1707.

wjones127:
This forces the user to replace the whole FTS directory when re-creating the index, prevent duplicate data from being created. Previously, the whole dataset was re-added to the existing index, duplicating existing rows in the index.

This (in combination with lancedb/lance#1707) caused #726, since the duplicate data emitted duplicate indices for `take()` and an upstream issue caused those queries to fail.

This solution isn't ideal, since it makes the FTS index temporarily unavailable while the index is built. In the future, we should have multiple FTS index directories, which would allow atomic commits of new indexes (as well as multiple indexes for different columns).

Fixes #498.
Fixes #726.

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -23,6 +23,7 @@ import lance
 import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc
+import pyarrow.fs as pa_fs
 from lance import LanceDataset
 from lance.vector import vec_to_table
 
@@ -577,7 +578,9 @@ class LanceTable(Table):
         self._reset_dataset()
         register_event("create_index")
 
-    def create_fts_index(self, field_names: Union[str, List[str]]):
+    def create_fts_index(
+        self, field_names: Union[str, List[str]], *, replace: bool = False
+    ):
         """Create a full-text search index on the table.
 
         Warning - this API is highly experimental and is highly likely to change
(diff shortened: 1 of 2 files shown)


python/lancedb/table.py
```diff
@@ -587,11 +590,23 @@ def create_fts_index(self, field_names: Union[str, List[str]]):
...
 
         if isinstance(field_names, str):
             field_names = [field_names]
+
+        fs, path = fs_from_uri(self._get_fts_index_path())
+        index_exists = fs.get_file_info(path).type != pa_fs.FileType.NotFound
+        if index_exists and not replace:
+            raise ValueError(f"Index already exists. Use replace=True to overwrite.")
+        elif index_exists:
+            fs.delete_dir(path)
```
changhiskhan, line 608: this is super nit-picky but wdyt about:

```suggestion
        if index_exists: 
            if not replace:
                raise ValueError(f"Index already exists. Use replace=True to overwrite.")
            fs.delete_dir(path)
```

changhiskhan (approved):
looks great thanks. just one minor nit.

PR #732, opened 2023-12-20
chore: add issue templates

wjones127:
This PR adds issue templates, which help two recurring issues:

* Users forget to tell us whether they are using the Node or Python SDK
* Issues don't get appropriate tags

This doesn't force the use of the templates. Because we set `blank_issues_enabled: true`, users can still create a custom issue.


changhiskhan (approved):
thanks

PR #735, opened 2023-12-20
docs: enhance Update user guide

### LINKED ISSUE #705: Provide more examples of update queries
In particular, show expression-based updates.

wjones127:
Closes #705


python/lancedb/table.py
wjones127, line 1: Needed to copy to the main class otherwise it doesn't show up in the docs.

AyushExel (approved):

Issue #676, opened 2023-12-01
[Feature request] uri for connect in memory 


ternaus:
I love that LanceDb is disk-based, but it would be great to have an option for an in-memory option.  

Use case: unit tests

Api could be similar to DeepLake that uses `mem://<path>`

https://docs.deeplake.ai/en/latest/deeplake.VectorStore.html#deeplake-vectorstore

PR #739, opened 2023-12-22
feat(python): add post filtering for full text search

### LINKED ISSUE #721: Full Text Seach: `where` clause ignored or unclear behavior
Assuming I've set up an fts_index... when doing this:
```
table.search(query).where(filter).limit(10).to_list()
```
I expect the FTS to be limited to records where `filter` is True. But it doesn't seem to be the case, 
as in the first query below -- the `filter` seems to be ignored, or I don't understand the behavior, or I'm missing some setting.

I expected the first query below to return NO results. The second one seems to work as expected.
```
In [1]: import lancedb
   ...:
   ...: uri = "data/sample-lancedb"
   ...: db = lancedb.connect(uri)
   ...:
   ...: table = db.create_table("my_table",
   ...:             data=[{"vector": [3.1, 4.1], "text": "Frodo was a happy puppy"},
   ...:                   {"vector": [5.9, 26.5], "text": "There are several kittens playing"}])

In [2]: table.create_fts_index("text")

In [3]: table.search("Frodo").where("text like '%kitten%'").limit(10).to_list()
Out[3]:
[{'vector': [3.0999999046325684, 4.099999904632568],
  'text': 'Frodo was a happy puppy',
  'score': 0.6931471824645996}]

In [4]: table.search(None).where("text like '%kitten%'").limit(10).to_list()
Out[4]:
[{'vector': [5.900000095367432, 26.5],
  'text': 'There are several kittens playing'}]
```

changhiskhan:
Closes #721 

fts will return results as a pyarrow table. Pyarrow tables has a `filter` method but it does not take sql filter strings (only pyarrow compute expressions). Instead, we do one of two things to support  `tbl.search("keywords").where("foo=5").limit(10).to_arrow()`:

Default path: If duckdb is available then use duckdb to execute the sql filter string on the pyarrow table.
Backup path: Otherwise, write the pyarrow table to a lance dataset and then do `to_table(filter=<filter>)`

Neither is ideal. 
Default path has two issues:
1. requires installing an extra library (duckdb)
2. duckdb mangles some fields (like fixed size list => list)

Backup path incurs a latency penalty (~20ms on ssd) to write the resultset to disk.

In the short term, once #676 is addressed, we can write the dataset to "memory://" instead of disk, this makes the post filter evaluate much quicker (ETA next week).

In the longer term, we'd like to be able to evaluate the filter string on the pyarrow Table directly, one possibility being that we use Substrait to generate pyarrow compute expressions from sql string. Or if there's enough progress on pyarrow, it could support Substrait expressions directly (no ETA)

### DIFF
--- a/docs/src/fts.md
+++ b/docs/src/fts.md
@@ -29,8 +29,9 @@ uri = "data/sample-lancedb"
 db = lancedb.connect(uri)
 
 table = db.create_table("my_table",
-            data=[{"vector": [3.1, 4.1], "text": "Frodo was a happy puppy"},
-                  {"vector": [5.9, 26.5], "text": "There are several kittens playing"}])
+            data=[{"vector": [3.1, 4.1], "text": "Frodo was a happy puppy", "meta": "foo"},
+                  {"vector": [5.9, 26.5], "text": "Sam was a loyal puppy", "meta": "bar"},
+                  {"vector": [15.9, 6.5], "text": "There are several kittens playing"}])
 
 ```
 
@@ -64,10 +65,23 @@ table.create_fts_index(["text1", "text2"])
 
 Note that the search API call does not change - you can search over all indexed columns at once.
 
+## Filtering
+
+Currently the LanceDB full text search feature supports *post-filtering*, meaning filters are
+applied on top of the full text search results. This can be invoked via the familiar
+`where` syntax:
+
+```python
+table.search("puppy").where("meta='foo'").limit(10).to_list()
+```
+
 ## Current limitations
 
 1. Currently we do not yet support incremental writes.
-If you add data after fts index creation, it won't be reflected
-in search results until you do a full reindex.
+   If you add data after fts index creation, it won't be reflected
+   in search results until you do a full reindex.
 
-2. We currently only support local filesystem paths for the fts index.
+2. We currently only support local filesystem paths for the fts index. 
+   This is a tantivy limitation. We've implemented an object store plugin
+   but there's no way in tantivy-py to specify to use it.
+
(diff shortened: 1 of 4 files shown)


docs/src/fts.md
```diff
@@ -64,10 +65,23 @@ table.create_fts_index(["text1", "text2"])
...
 Note that the search API call does not change - you can search over all indexed columns at once.
 
+## Filtering
+
+Currently the LanceDB full text search feature supports *post-filtering*, meaning filters are
+applied on top of the full text search results. This can be invoked via the familiar
+`where` syntax:
+
+```python
+table.search("puppy").where("meta='foo'").limit(10).to_list()
```
wjones127, line 75: I know it technically doesn't make a difference, but might be clearer to put the limit before the where to hint at the post-filtering.

```suggestion
table.search("puppy").limit(10).where("meta='foo'").to_list()
```

python/lancedb/query.py
```diff
@@ -488,6 +488,25 @@ def to_arrow(self) -> pa.Table:
...
+
+        if self._where is not None:
+            try:
+                # TODO would be great to have Substrait generate pyarrow compute expressions
+                # or conversely have pyarrow support SQL expressions using Substrait
+                import duckdb
+
+                output_tbl = duckdb.sql(
+                    f"SELECT * FROM output_tbl WHERE {self._where}"
+                ).to_arrow_table()
```
wjones127, line 500: It may be necessary to protect from SQL injection attacks:
```suggestion
                output_tbl = duckdb.sql(
                    f"SELECT * FROM output_tbl"
                ).filter(self._where).to_arrow_table()
```

wjones127 (changes requested):

wjones127 (approved):

PR #740, opened 2023-12-22
SaaS JS API sdk doc

QianZhu:


docs/src/javascript/saas-modules.md
```diff
@@ -0,0 +1,72 @@
...
+```javascript
+// connect to a remote DB
+const lancedb = require('vectordb');
+const db = await lancedb.connect({
+  uri: "db://your-project-name",
+  apiKey: "sk_...",
+  region: "us-east-1"
+});
+// create a new table
+tableName = "my_table"
```
aidangomar, line 48:
```suggestion
const tableName = "my_table"
```

docs/src/javascript/saas-modules.md
```diff
@@ -0,0 +1,72 @@
...
+const table = await db.createTable(tableName,
```
aidangomar, line 49:
```suggestion
const table = await db.createTable(tableName,
    [
        { id: 1, vector: [0.1, 1.0], item: "foo", price: 10.0 },
        { id: 2, vector: [3.9, 0.5], item: "bar", price: 20.0 }
    ]
)
```

docs/src/javascript/saas-modules.md
```diff
@@ -0,0 +1,72 @@
...
+      [{ id: 1, vector: [0.1, 1.0], item: "foo", price: 10.0 },
+      { id: 2, vector: [3.9, 0.5], item: "bar", price: 20.0 }])
+// list the table
+const tableNames_1 = await db.tableNames('')
+// add some data and search should be okay
+const data = [
```
aidangomar, line 55: should we include an example with schema?

changhiskhan:
btw you can take a look at the doc test github workflow to see how to run the doctests locally so the feedback loop is faster

docs/src/javascript/classes/RemoteTable.md
```diff
@@ -0,0 +1,355 @@
...
+
+#### Defined in
+
+[remote/index.ts:290](https://github.com/lancedb/lancedb/blob/main/node/src/remote/index.ts#L290)
+
+___
+
+### createIndex
+
+▸ **createIndex**(`metric_type`, `column`): `Promise`<`any`\>
```
aidangomar, line 159: we also expose `index_cache_size` for both python and js
```suggestion
**createIndex**(`metric_type`, `column`): `Promise`<`any`\>
```

aidangomar (approved):
lgtm once the schema stuff is sorted out & ci passes

Can we also add examples to each function similar to the python docs? We don't have to do it in this pr but it would definitely help ease of use, especially since I don't think we have it for oss js either

QianZhu:
the doc test failure is fixed here: https://github.com/lancedb/lancedb/pull/770

PR #755, opened 2023-12-30
feat(js): support list of string input

changhiskhan:
Add support for adding lists of string input (e.g., list of categorical labels)

Follow-up items: #757 #758

### DIFF
--- a/node/src/arrow.ts
+++ b/node/src/arrow.ts
@@ -20,7 +20,7 @@ import {
   Utf8,
   type Vector,
   FixedSizeList,
-  vectorFromArray, type Schema, Table as ArrowTable, RecordBatchStreamWriter, List
+  vectorFromArray, type Schema, Table as ArrowTable, RecordBatchStreamWriter, List, Float64
 } from 'apache-arrow'
 import { type EmbeddingFunction } from './index'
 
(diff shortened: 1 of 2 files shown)


eddyxu (approved):

node/src/arrow.ts
```diff
@@ -59,6 +59,14 @@ export async function convertToTable<T> (data: Array<Record<string, unknown>>, e
       if (typeof values[0] === 'string') {
         // `vectorFromArray` converts strings into dictionary vectors, forcing it back to a string column
         records[columnsKey] = vectorFromArray(values, new Utf8())
+      } else if (Array.isArray(values[0]) && typeof values[0][0] === 'string') {
```
westonpace, line 62: What if `values` contains a single empty array like `[[]]`?  You'll get an error trying `values[0][0]`

westonpace (approved):
Is this patching a hole in arrow?  Does `vectorFromArray` fail when given a list of strings?

changhiskhan:
> Is this patching a hole in arrow? Does `vectorFromArray` fail when given a list of strings?

yeah it throws `TypeError: Unable to infer Vector type from input values, explicit type declaration expected.`

node/src/arrow.ts
changhiskhan: i changed it to give a more helpful error message. Ideally we would pass in the schema here, or keep going to the next element til we get a non-empty observation otherwise.

PR #762, opened 2024-01-02
feat(python): Set heap size to get faster fts indexing performance

changhiskhan:
By default tantivy-py uses 128MB heapsize. We change the default to 1GB and we allow the user to customize this

locally this makes `test_fts.py` run 10x faster

### DIFF
--- a/python/lancedb/fts.py
+++ b/python/lancedb/fts.py
@@ -60,7 +60,7 @@ def populate_index(
     index: tantivy.Index,
     table: LanceTable,
     fields: List[str],
-    writer_heap_size: Optional[int] = 1024 * 1024 * 1024,
+    writer_heap_size: int = 1024 * 1024 * 1024,
 ) -> int:
     """
     Populate an index with data from a LanceTable
@@ -73,7 +73,7 @@ def populate_index(
         The table to index
     fields : List[str]
         List of fields to index
-    writer_heap_size : Optional[int]
+    writer_heap_size : int
         The writer heap size in bytes, defaults to 1GB
 
     Returns
--- a/python/lancedb/fts.py
+++ b/python/lancedb/fts.py
@@ -60,7 +60,7 @@ def populate_index(
     index: tantivy.Index,
     table: LanceTable,
     fields: List[str],
-    writer_heap_size: Optional[int] = None,
+    writer_heap_size: Optional[int] = 1024 * 1024 * 1024,
 ) -> int:
     """
     Populate an index with data from a LanceTable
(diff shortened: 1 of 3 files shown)


python/lancedb/fts.py
```diff
@@ -113,6 +118,14 @@ def populate_index(index: tantivy.Index, table: LanceTable, fields: List[str]) -
     return row_id
 
 
+def open_writer(index: tantivy.Index, writer_heap_size: int = None) -> "IndexWriter":
+    if writer_heap_size is not None:
+        writer = index.writer(heap_size=writer_heap_size)
+    else:
+        writer = index.writer(heap_size=1024 * 1024 * 1024)
+    return writer
```
wjones127, line 126: Why are we burying the default? It's less code and better documented if we just make the default `1024 * 1024 * 1024` instead of `None`.

changhiskhan: added as parameter defaults

python/lancedb/table.py
```diff
@@ -709,7 +709,11 @@ def create_scalar_index(self, column: str, *, replace: bool = True):
         self._dataset.create_scalar_index(column, index_type="BTREE", replace=replace)
 
     def create_fts_index(
-        self, field_names: Union[str, List[str]], *, replace: bool = False
+        self,
+        field_names: Union[str, List[str]],
+        *,
+        replace: bool = False,
+        writer_heap_size: Optional[int] = 1024 * 1024 * 1024,
```
chebbyChefNEQ, line 716: nit: we could do something like 50% of the system's memory and make this a fraction

chebbyChefNEQ (approved):

python/lancedb/fts.py
```diff
@@ -56,7 +56,12 @@ def create_index(index_path: str, text_fields: List[str]) -> tantivy.Index:
     return index
 
 
-def populate_index(index: tantivy.Index, table: LanceTable, fields: List[str]) -> int:
+def populate_index(
+    index: tantivy.Index,
+    table: LanceTable,
+    fields: List[str],
+    writer_heap_size: Optional[int] = 1024 * 1024 * 1024,
```
wjones127, line 63: `Optional` implies you can pass `None`, which I think would error in tantivy:

https://github.com/quickwit-oss/tantivy-py/blob/0cf1661fb371c8a9c12050ac378a187223ef50af/src/index.rs#L265-L270

```suggestion
    writer_heap_size: int = 1024 * 1024 * 1024,
```

PR #763, opened 2024-01-03
chore(python): handle NaN input in fts ingestion

changhiskhan:
If the input text is None, Tantivy raises an error
complaining it cannot add a NoneType. We handle this
upstream so None's are not added to the document.
If all of the indexed fields are None then we skip
this document.

### DIFF
--- a/python/lancedb/fts.py
+++ b/python/lancedb/fts.py
@@ -103,10 +103,13 @@ def populate_index(index: tantivy.Index, table: LanceTable, fields: List[str]) -
                 b = b.flatten()
         for i in range(b.num_rows):
             doc = tantivy.Document()
-            doc.add_integer("doc_id", row_id)
             for name in fields:
-                doc.add_text(name, b[name][i].as_py())
-            writer.add_document(doc)
+                value = b[name][i].as_py()
+                if value is not None:
+                    doc.add_text(name, value)
+            if not doc.is_empty:
+                doc.add_integer("doc_id", row_id)
+                writer.add_document(doc)
             row_id += 1
     # commit changes
     writer.commit()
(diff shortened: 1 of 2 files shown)


python/lancedb/fts.py
```diff
@@ -103,10 +103,13 @@ def populate_index(index: tantivy.Index, table: LanceTable, fields: List[str]) -
...
             doc = tantivy.Document()
-            doc.add_integer("doc_id", row_id)
             for name in fields:
-                doc.add_text(name, b[name][i].as_py())
-            writer.add_document(doc)
+                value = b[name][i].as_py()
+                if value is not None:
+                    doc.add_text(name, value)
+            if not doc.is_empty:
+                doc.add_integer("doc_id", row_id)
```
westonpace, line 111: This moves `doc_id` to the end of `doc` instead of the beginning but I don't know tantivy at all so I'm guessing this isn't a problem.

python/tests/test_fts.py
```diff
@@ -147,3 +147,18 @@ def import_mock(name, *args):
...
+
+def test_null_input(table):
+    table.add(
+        [
+            {
+                "vector": np.random.randn(128),
+                "id": 101,
+                "text": None,
+                "text2": None,
+                "nested": {"text": None},
```
westonpace, line 160: Is this column used at all?

python/tests/test_fts.py
```diff
@@ -147,3 +147,18 @@ def import_mock(name, *args):
...
+            }
+        ]
+    )
```
westonpace, line 163: Maybe have one test case that triggers the `doc.is_empty` case and one that does not?

chebbyChefNEQ (approved):

westonpace (commented):
Sorry, I see it's already merged, but looks good!

PR #764, opened 2024-01-03
Add relevant imports for each step

bengsoon:
I found that it was quite incoherent to have to read through the documentation and having to search which submodule that each class should be imported from. 

For example, it is cumbersome to have to navigate to another documentation page to find out that `EmbeddingFunctionRegistry` is from `lancedb.embeddings`


changhiskhan (approved):
lgtm. Thanks!

PR #772, opened 2024-01-05
bump lance to 0.9.3

albertlockett:


changhiskhan (approved):
is the python dependency also bumped? might as well do it in one go?

changhiskhan:
oh actually, 0.9.4 is out

changhiskhan:
this is no longer needed because of #790

PR #776, opened 2024-01-06
raise exception if fts index does not exist

lucasiscovici:
raise exception if fts index does not exist

### DIFF
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -485,7 +485,7 @@ class LanceFtsQueryBuilder(LanceQueryBuilder):
         if not Path(index_path).exists():
             raise FileNotFoundError(
                 "Fts index does not exist."
-                f"Please first call table.create_fts_index('<field_names>') to create the fts index."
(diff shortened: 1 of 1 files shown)


python/lancedb/query.py
```diff
@@ -480,6 +481,12 @@ def to_arrow(self) -> pa.Table:
 
         # get the index path
         index_path = self._table._get_fts_index_path()
+        # check if the index exist
+        if not Path(index_path).exists():
+            raise FileNotFoundError(
+                "Fts index does not exist."
+                f"Please first call table.create_fts_index('<field_names>') to create the fts index."
```
changhiskhan, line 488:
```suggestion
                f"Please first call table.create_fts_index(['<field_names>']) to create the fts index."
```
nit picky but just want to make it clear that when you have multiple fields, you pass in a list rather than concatenate field names in a string

changhiskhan (commented):
really minor comment. Thanks for adding this.

changhiskhan (approved):

PR #778, opened 2024-01-07
Make it work with new OpenAI Library versions

sudhirpatil:
Current example doesn't work, as openai api's changed after versions 1+


sudhirpatil:
will open new one with other changes

PR #789, opened 2024-01-08
feat(node): support table.schema for LocalTable

### LINKED ISSUE #773: Feature: Enable support for viewing table `Schema`
### SDK

Node

### Description

When using the NodeSDK it may be relevant to inspect a current table's schema. The `Table` class does not currently support reading of the table schema as it currently sits. This functionality does exist in the python library and it would be nice to have the same functionality within the NodeJS SDK

changhiskhan:
Close #773 

we pass an empty table over IPC so we don't need to manually deal with serde. Then we just return the schema attribute from the empty table.

### DIFF
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -498,6 +498,26 @@ describe('LanceDB client', function () {
       assert.equal(results.length, 2)
     })
   })
+
+  describe.only('when inspecting the schema', function () {
+    it('should return the schema', async function () {
+      const uri = await createTestDB()
+      const db = await lancedb.connect(uri)
+      assert.equal(db.uri, uri)
+      const table = await db.createTable({
+        name: 'some_table',
+        schema: new Schema(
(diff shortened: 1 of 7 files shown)


westonpace (approved):
+1 for this approach.  An empty table is what I've always used to transmit the schema.

node/src/test/test.ts
```diff
@@ -498,6 +498,28 @@ describe('LanceDB client', function () {
...
+          new Field('vector', new FixedSizeList(128, new Field('item', new Float32(), true))),
+          new Field('s', new Utf8())
+        ]
+      )
+      const table = await db.createTable({
+        name: 'some_table',
+        schema: expectedSchema
+      })
+      const schema = await table.schema
+      console.log({ schema })
```
albertlockett, line 519: nit: remove console log

albertlockett (approved):

Issue #797, opened 2024-01-09
Feature: option to turn off search interpretation


pchalasani:
### SDK

Python

### Description

Related to #769 -- a quoting workaround was suggested, but it won't work here.

I get a Tantivy search syntax error with FTS for a string of the form `who is "waldo"?`. 
Sample trace below.
The quoting work-around wouldn't work here. I inserted a space before the `?` and it works. 
It will be best if people did not have to be aware of Tantivy search syntax.

```
   def search_index(
        index: tantivy.Index, query: str, limit: int = 10
    ) -> Tuple[Tuple[int], Tuple[float]]:
        """
        Search an index for a query
    
        Parameters
        ----------
        index : tantivy.Index
            The index object
        query : str
            The query string
        limit : int
            The maximum number of results to return
    
        Returns
        -------
        ids_and_score: list[tuple[int], tuple[float]]
            A tuple of two tuples, the first containing the document ids
            and the second containing the scores
        """
        searcher = index.searcher()
>       query = index.parse_query(query)
E       ValueError: Syntax Error: Who directed the science fiction movie "The Vector"?
```

changhiskhan:
hmmm ok tantivy unfortunately does not expose an option for this. Mind taking a look at #798 and see if that would satisfy your use case?

pchalasani:
Neither single nor double fixes this. In my code I work around this by simply removing all quotes. 
The strings come from an LLM so I don’t have control on what they generate but I can do various post processing before passing to Lance search

PR #798, opened 2024-01-09
feat(python): add phrase query option for fts

changhiskhan:
addresses #797 

Problem: tantivy does not expose option to explicitly

Proposed solution here: 

1. Add a `.phrase_query()` option
2. Under the hood, LanceDB takes care of wrapping the input in quotes and replace nested double quotes with single quotes

I've also filed an upstream issue, if they support phrase queries natively then we can get rid of our manual custom processing here.

### DIFF
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -468,6 +468,24 @@ class LanceFtsQueryBuilder(LanceQueryBuilder):
     def __init__(self, table: "lancedb.table.Table", query: str):
         super().__init__(table)
         self._query = query
+        self._phrase_query = False
+
+    def phrase_query(self, phrase_query: bool = True) -> LanceFtsQueryBuilder:
+        """Set whether to use phrase query.
+
+        Parameters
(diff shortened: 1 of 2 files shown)


python/lancedb/query.py
```diff
@@ -490,7 +508,11 @@ def to_arrow(self) -> pa.Table:
         # open the index
         index = tantivy.Index.open(index_path)
         # get the scores and doc ids
-        row_ids, scores = search_index(index, self._query, self._limit)
+        query = self._query
+        if self._phrase_query:
+            query = query.replace('"', "'")
```
westonpace, line 513: Do we need to worry about escaped double quotes at all?

westonpace (approved):
Just one question

python/lancedb/query.py
changhiskhan: i added a test case with escaped double quotes. It does not cause a syntax error in this case

PR #800, opened 2024-01-09
chore(python): add docstring for limit behavior

### LINKED ISSUE #796: Document "no limit" behavior for LanceQueryBuilder
### Description

https://github.com/lancedb/lancedb/blob/121687231c2979844d3b3bc4662c848007cd1e02/python/lancedb/query.py#L263

The type annotation for `LanceQueryBuilder.limit()` says limit must be an int and there's no mention of what you should pass if you want "no limit" (it's `None` ... a user might think you need to pass in `limit=-1`?).  Type annotation should be changed to `Optional[int]` and `None` behavior should be documented accordingly.

Somewhat related - it is a little non-intuitive that the default behavior is `limit = 10` as opposed to no limit.

### Link

_No response_

changhiskhan:
Closes #796

### DIFF
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -272,7 +272,7 @@ class LanceQueryBuilder(ABC):
             or None to remove the limit.
             *WARNING* if you have a large dataset, removing
             the limit can potentially result in reading a
-            large amount of data into memory and cause 
+            large amount of data into memory and cause
             out of memory issues.
 
         Returns
--- a/python/lancedb/query.py
+++ b/python/lancedb/query.py
@@ -260,20 +260,30 @@ class LanceQueryBuilder(ABC):
             for row in self.to_arrow().to_pylist()
         ]
 
-    def limit(self, limit: int) -> LanceQueryBuilder:
+    def limit(self, limit: Union[int, None]) -> LanceQueryBuilder:
         """Set the maximum number of results to return.
 
         Parameters
         ----------
         limit: int
             The maximum number of results to return.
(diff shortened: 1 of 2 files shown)


python/lancedb/query.py
```diff
@@ -260,20 +260,30 @@ def to_pydantic(self, model: Type[LanceModel]) -> List[LanceModel]:
...
         ----------
         limit: int
             The maximum number of results to return.
+            By default the query is limited to the first 10.
+            Call this method and pass in any negative value
+            or None to remove the limit.
+            *WARNING* if you have a large dataset, removing
+            the limit can potentially result in reading a
+            large amount of data into memory and cause
+            out of memory issues.
```
westonpace, line 276: Even if I call `to_batches`?

python/lancedb/query.py
```diff
@@ -260,20 +260,30 @@ def to_pydantic(self, model: Type[LanceModel]) -> List[LanceModel]:
...
 
         Returns
         -------
         LanceQueryBuilder
             The LanceQueryBuilder object.
         """
-        self._limit = limit
+        if limit is None or limit < 0:
```
westonpace, line 283: What if `limit == 0`?

python/lancedb/query.py
changhiskhan: we don't have `to_batches` api yet. I'll create a follow-up issue

python/lancedb/query.py
changhiskhan: made limit == 0 also act as if there's no limit

PR #802, opened 2024-01-10
feat(node): align incoming data to table schema

### LINKED ISSUE #774: bug(node): Schema incompatibility due to Object order
### LanceDB version

0.4.2

### What happened?

When upserting embeddings with metadata using the `add` method on a table it appears the order of the object keys must be in the **exact** same order as the existing table schema.

Given object assembly in Javascript can be messy there is a possibility the metadata keys are all present, but not in the exact same order as the table schema.

Previous versions did not enforce this strict upsert rule and the only requirement was all keys in the schema be present.

1. Create table with embeddings with no present schema so that schema is inferred.
`client.createTable('testing', [{id: 1, vector: [....], propOne: 'a', propTwo: 'b'}]);`

2. Add another embedding with a different order of metadata keys
```javascript
const table = await client.openTable('testing');
table.add([{id: 1, vector: [....],propTwo: 'b',  propOne: 'a', }]);
// [Error: LanceDBError: Append with different schema:]
```

changhiskhan:
Closes #774

Caveat: only handles top level field order.
For nested subfields you still need to manually
ensure the order is correct (for now).

### DIFF
--- a/node/src/arrow.ts
+++ b/node/src/arrow.ts
@@ -17,8 +17,7 @@ import {
   Float32,
   makeBuilder,
   RecordBatchFileWriter,
-  Utf8,
-  type Vector,
+  Utf8, type Vector,
   FixedSizeList,
   vectorFromArray, type Schema, Table as ArrowTable, RecordBatchStreamWriter, List, Float64
 } from 'apache-arrow'
@@ -78,6 +77,7 @@ export async function convertToTable<T> (data: Array<Record<string, unknown>>, e
         }
         records[columnsKey] = listBuilder.finish().toVector()
       } else {
+        // TODO if this is a struct field then recursively align the subfields
         records[columnsKey] = vectorFromArray(values)
       }
     }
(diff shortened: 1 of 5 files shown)


node/src/arrow.ts
```diff
@@ -153,10 +162,26 @@ export async function fromTableToStreamBuffer<T> (table: ArrowTable, embeddings?
...
 
+function alignTable (table: ArrowTable, schema: Schema): ArrowTable {
+  const alignedColumns: Record<string, Vector> = {}
+  for (const field of schema.fields) {
+    const column = table.getChild(field.name)
+    if (column === null) {
+      throw new Error(`The column ${field.name} was not found in the Arrow Table`)
+    }
+    // TODO if this is a struct field then recursively align the subfields
+    alignedColumns[field.name] = column
```
westonpace, line 180: I thought the keys of a `Record` / `object` were unordered in TS?

westonpace (approved):

PR #803, opened 2024-01-10
use requests instead of aiohttp for underlying http client

SebbyLaw:
instead of starting and stopping the current thread's event loop on every http call, just make an http call.

### DIFF
--- a/python/lancedb/remote/client.py
+++ b/python/lancedb/remote/client.py
@@ -14,6 +14,7 @@
 
 import functools
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
+from urllib.parse import urljoin
 
 import requests
 import attrs
@@ -54,11 +55,14 @@ class RestfulLanceDBClient:
 
     @functools.cached_property
     def session(self) -> requests.Session:
-        self.url: str = (
+        return requests.Session()
+
+    @property
+    def url(self) -> str:
+        return (
             self.host_override
             or f"https://{self.db_name}.{self.region}.api.lancedb.com"
         )
-        return requests.Session()
 
     def close(self):
         self.session.close()
@@ -98,10 +102,10 @@ class RestfulLanceDBClient:
         if isinstance(params, BaseModel):
             params: Dict[str, Any] = params.dict(exclude_none=True)
         with self.session.get(
-            self.url + uri,
+            urljoin(self.url, uri),
             params=params,
             headers=self.headers,
-            timeout=30,
+            timeout=(5.0, 30.0),
         ) as resp:
             self._check_status(resp)
             return resp.json()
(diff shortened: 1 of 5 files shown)


python/lancedb/remote/client.py
```diff
@@ -76,38 +76,38 @@ def headers(self) -> Dict[str, str]:
...
     @_check_not_closed
-    async def get(self, uri: str, params: Union[Dict[str, Any], BaseModel] = None):
+    def get(self, uri: str, params: Union[Dict[str, Any], BaseModel] = None):
         """Send a GET request and returns the deserialized response payload."""
         if isinstance(params, BaseModel):
             params: Dict[str, Any] = params.dict(exclude_none=True)
-        async with self.session.get(
-            uri,
+        with self.session.get(
+            self.url + uri,
```
chebbyChefNEQ, line 101: I think we should use `urllib.parse.urljoin` here

python/lancedb/remote/client.py
```diff
@@ -76,38 +76,38 @@ def headers(self) -> Dict[str, str]:
...
             params=params,
             headers=self.headers,
-            timeout=aiohttp.ClientTimeout(total=30),
+            timeout=30,
```
chebbyChefNEQ, line 104: nit: maybe use `(5.0, 30.0)`. 5s for connection timeout and 30s for read timeout

python/lancedb/remote/client.py
```diff
@@ -139,31 +139,26 @@ async def post(
             headers["content-type"] = content_type
         if request_id is not None:
             headers["x-request-id"] = request_id
-        async with self.session.post(
-            uri,
+        with self.session.post(
+            self.url + uri,
```
chebbyChefNEQ, line 143: ditto: use `urljoin`

chebbyChefNEQ (commented):
Thank you for doing this!

I think we also need to update `pyproject.toml` to add `requests` as a dependency.

Otherwise looks good

python/lancedb/remote/client.py
```diff
@@ -53,15 +53,15 @@ class RestfulLanceDBClient:
     closed: bool = attrs.field(default=False, init=False)
 
     @functools.cached_property
-    def session(self) -> aiohttp.ClientSession:
-        url = (
+    def session(self) -> requests.Session:
+        self.url: str = (
             self.host_override
             or f"https://{self.db_name}.{self.region}.api.lancedb.com"
         )
```
chebbyChefNEQ, line 60: let's move this to it's own property

SebbyLaw:
> I think we also need to update `pyproject.toml` to add `requests` as a dependency.

requests is already listed as a dependency in the pyproject.toml

I've removed aiohttp from the module dependencies and replaced `requests` with it in the testing dependency extras since it is only used in tests now

chebbyChefNEQ (approved):
LGTM, barring CI

PR #806, opened 2024-01-10
feat(python): Add gemini text embedding function

AyushExel:
Named it Gemini-text for now. Not sure how complicated it will be to support both text and multimodal embeddings under the same class "gemini"..But its not something to worry about for now I guess.

### DIFF
--- a/docs/src/embeddings/default_embedding_functions.md
+++ b/docs/src/embeddings/default_embedding_functions.md
@@ -147,7 +147,7 @@ class TextModel(LanceModel):
     vector: Vector(model.ndims()) = model.VectorField()
 
 df = pd.DataFrame({"text": ["hello world", "goodbye world"]})
-db = lancedb.connect(tmp_path)
+db = lancedb.connect("~/.lancedb")
 tbl = db.create_table("test", schema=TextModel, mode="overwrite")
 
 tbl.add(df)
--- a/docs/src/embeddings/default_embedding_functions.md
+++ b/docs/src/embeddings/default_embedding_functions.md
@@ -139,7 +139,7 @@ import pandas as pd
 from lancedb.pydantic import LanceModel, Vector
 from lancedb.embeddings import get_registry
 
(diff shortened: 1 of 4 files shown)


python/lancedb/embeddings/gemini_text.py
```diff
@@ -0,0 +1,132 @@
...
+        Sets the task type for ingestion.
+
+    Examples
+    --------
+    import lancedb
+    import pandas as pd
+    from lancedb.pydantic import LanceModel, Vector
+    from lancedb.embeddings import get_registry
+
+    def test_instructor_embedding(tmp_path):
```
changhiskhan, line 61: again, copy-pasted unit test doesn't make sense here

python/lancedb/embeddings/gemini_text.py
```diff
@@ -0,0 +1,132 @@
...
+        ):  # Provide a title to use existing API design
+            title = "Embedding of a document"
+            kwargs["title"] = title
+
+        return [
+            genai.embed_content(model=self.name, content=text, **kwargs)["embedding"]
+            for text in texts
+        ]
+
+    def get_client(self):
```
changhiskhan, line 121: i think you can make this a cached property

docs/src/embeddings/default_embedding_functions.md
```diff
@@ -118,6 +118,42 @@ texts = [{"text": "Capitalism has been dominant in the Western world since the e
...
+
+Usage Example:
+
+```python
+import lancedb
+import pandas as pd
+from lancedb.pydantic import LanceModel, Vector
+from lancedb.embeddings import get_registry
+
+def test_instructor_embedding(tmp_path):
```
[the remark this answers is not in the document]
AyushExel: Moved everything outside the function

AyushExel:
@changhiskhan cached_property doesn't seem to work with pydantic 1.x, using lru_cache instead

changhiskhan (approved):

PR #811, opened 2024-01-12
feat(python): basic polars integration

changhiskhan:
We should now be able to directly ingest polars dataframes and return results as polars dataframes

image


prrao87 (approved):
I think this looks good for a starting point! Will refer to the tests for some docs examples too.

PR #814, opened 2024-01-14
feat(python): allow the entire table to be converted a polars dataframe

changhiskhan:
#811 allowed us to ingest polars dataframes directly and turn search results into polars dataframes.

This PR supports converting the entire LanceDB table into a lazy polars DataFrame (i.e., it does not read everything into memory immediately).

### DIFF
--- a/docs/src/basic.md
+++ b/docs/src/basic.md
@@ -67,7 +67,7 @@ We'll cover the basics of using LanceDB on your local machine in this section.
       !!! warning
 
             If the table already exists, LanceDB will raise an error by default.
-            If you want to overwrite the table, you can pass in `mode="overwrite"`
+            If you want to make sure you overwrite the table, pass in `mode="overwrite"`
             to the `createTable` function.
 
 === "Javascript"
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -73,7 +73,7 @@ def _sanitize_data(
         meta = data.schema.metadata if data.schema.metadata is not None else {}
         meta = {k: v for k, v in meta.items() if k != b"pandas"}
         data = data.replace_schema_metadata(meta)
-    elif pl is not None and isinstance(data, pl.DataFrame):
+    elif pl is not None and isinstance(data, (pl.DataFrame, pl.LazyFrame)):
         data = data.to_arrow()
 
     if isinstance(data, pa.Table):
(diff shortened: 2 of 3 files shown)


docs/src/basic.md
```diff
@@ -67,7 +67,9 @@ We'll cover the basics of using LanceDB on your local machine in this section.
       !!! warning
 
             If the table already exists, LanceDB will raise an error by default.
-            If you want to overwrite the table, you can pass in `mode="overwrite"`
+            If you want to just open the table if it exists, you can pass in
```
westonpace, line 70: What happens to the data I pass in if I call `createTable` with `exist_ok=True`?  Does it get dropped or does it get appended to the current dataset?

python/lancedb/table.py
```diff
@@ -697,6 +697,15 @@ def to_arrow(self) -> pa.Table:
         pa.Table"""
         return self._dataset.to_table()
 
+    def to_polars(self) -> "pl.DataFrame":
```
westonpace, line 700: It looks like `scan_pyarrow_dataset` returns a `LazyFrame` and not a `DataFrame`

python/lancedb/table.py
```diff
@@ -697,6 +697,15 @@ def to_arrow(self) -> pa.Table:
...
+        """Return the table as a polars DataFrame.
+
+        Returns
+        -------
+        pl.DataFrame
+        """
+        return pl.scan_pyarrow_dataset(self.to_lance())
```
westonpace, line 707: Should we expose `batch_size` as a parameter?  I think I'm ok with not doing so.  If users need fine grained control they can always call `scan_pyarrow_dataset` themselves.

python/tests/test_table.py
```diff
@@ -206,12 +207,17 @@ def test_polars(db):
...
 
+    # search results to polars dataframe
     q = [3.1, 4.1]
     result = table.search(q).limit(1).to_polars()
     assert np.allclose(result["vector"][0], q)
     assert result["item"][0] == "foo"
     assert np.allclose(result["price"][0], 10.0)
 
+    # enter table to polars dataframe
+    result = table.to_polars()
```
westonpace, line 218: The default scan_pyarrow_dataset will pushdown filters.  Do you want to add a test case with a filter to see if that breaks things?

westonpace (approved):

docs/src/basic.md
westonpace: I see now you have an entire other PR for this.  Did you mean for this line to be part of this PR?

python/lancedb/table.py
prrao87: What are we referring to as `batch_size` here? It's a bit of confusing term - is that the same as polars [`set_streaming_chunk_size`](https://docs.pola.rs/py-polars/html/reference/api/polars.Config.set_streaming_chunk_size.html#polars.Config.set_streaming_chunk_size)? It's good to drop a description here as to what we mean explicitly.

docs/src/basic.md
changhiskhan: yeah, i'll just modify this separately

python/lancedb/table.py
changhiskhan: image

prrao87: That description for `batch_size` makes sense. The type annotation for `LazyFrame` also makes more sense now!

python/tests/test_table.py
changhiskhan: hmmm...unfortunately pushdown doesn't work since it pushes down compute expressions.
So if you have an equality filter, the str repr of the compute expr parses so that works, but something `is_in` does not. I guess we need to hard code it to False for now?

PR #815, opened 2024-01-14
Add openai api key not found help

MrShevan:
This pull request adds check for the presence of an environment variable `OPENAI_API_KEY` and removes an unused parameter in `retry_with_exponential_backoff` function.

### DIFF
--- a/python/lancedb/embeddings/openai.py
+++ b/python/lancedb/embeddings/openai.py
@@ -10,6 +10,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import os
 from functools import cached_property
 from typing import List, Union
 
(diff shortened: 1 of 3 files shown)


python/lancedb/embeddings/openai.py
```diff
@@ -51,4 +53,7 @@ def generate_embeddings(
     @cached_property
     def _openai_client(self):
         openai = self.safe_import("openai")
+
+        if not os.environ.get("OPENAI_API_KEY"):
+            raise ValueError(api_key_not_found_help("openai"))
```
AyushExel, line 58: api_key_not_found_help should raise an error internally, you don't need to raise it here I think

MrShevan: Yes, indeed, it can be like this:
```
if not os.environ.get("OPENAI_API_KEY"):
     api_key_not_found_help("openai")
```

Initially, I used this [line](https://github.com/lancedb/lancedb/blob/main/python/lancedb/embeddings/gemini_text.py#L127) as a template. I can fix this in two places.

AyushExel: Thanks!

AyushExel (approved):

PR #826, opened 2024-01-19
bump lance to 0.9.7

albertlockett:


eddyxu (approved):
Do we need to bump python as well

chebbyChefNEQ (approved):

albertlockett:
fixing some compiler errors.. moving to draft while fixing

changhiskhan:
please bump python too while you're at it

changhiskhan (approved):
are there node and python tests for the index stats APIs? just want to make sure we're not breaking anything there

albertlockett:
> are there node and python tests for the index stats APIs? just want to make sure we're not breaking anything there

Looks like we only have coverage on the node apis. I don't think we ever exposed it through the python api.

https://github.com/lancedb/lancedb/blob/119b928a52040f64dc8ac80e56cd2bc3b12874a3/node/src/test/test.ts#L475-L491

Issue #812, opened 2024-01-13
bug(python): old table reference doesn't see that rows are deleted


dconathan:
### LanceDB version

0.4.4

### What happened?

If you have two separate references to a LanceTable and one of them deletes some rows, the other will not see that the rows got deleted (i.e., they will still appear in search results / table length / etc.). I'm guessing this is a caching issue?

### Are there known steps to reproduce?

```python
import tempfile

import lancedb

with tempfile.TemporaryDirectory() as uri:
    db = lancedb.connect(uri)

    db.create_table("foobar", data=[{"text": "hello world", "id": "1"}])

    table1 = db.open_table("foobar")
    table2 = db.open_table("foobar")

    assert len(table1.search().to_list()) == 1
    assert len(table2.search().to_list()) == 1

    table1.delete("id = '1'")

    # works
    assert len(table1.search().to_list()) == 0
    assert len(table1) == 0
    # error because table2 still returns the deleted row
    # same behavior with len(table2)
    assert len(table2.search().to_list()) == 0
```

dconathan:
Now that I know it’s expected behavior I’d be fine with it being the default for performance sake as long as it’s documented… having that parameter available helps explain what’s going on.

PR #828, opened 2024-01-19
feat(python): add `read_consistency_interval` argument

wjones127:
This PR refactors how we handle read consistency: does the `LanceTable` class always pick up modifications to the table made by other instance or processes. Users have three options they can set at the connection level:

1. (Default) `read_consistency_interval=None` means it will not check at all. Users can call `table.checkout_latest()` to manually check for updates.
2. `read_consistency_interval=timedelta(0)` means **always** check for updates, giving strong read consistency.
3. `read_consistency_interval=timedelta(seconds=20)` means check for updates every 20 seconds. This is eventual consistency, a compromise between the two options above.

## Table reference state

There is now an explicit difference between a `LanceTable` that tracks the current version and one that is fixed at a historical version. We now enforce that users cannot write if they have checked out an old version. They are instructed to call `checkout_latest()` before calling the write methods.

Since `conn.open_table()` doesn't have a parameter for version, users will only get fixed references if they call `table.checkout()`.

The difference between these two can be seen in the repr: Table that are fixed at a particular version will have a `version` displayed in the repr. Otherwise, the version will not be shown.

```python
>>> table
LanceTable(connection=..., name="my_table")
>>> table.checkout(1)
>>> table
LanceTable(connection=..., name="my_table", version=1)
```

I decided to not create different classes for these states, because I think we already have enough complexity with the Cloud vs OSS table references.

Based on #812

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -42,6 +42,12 @@ To run the unit tests:
 pytest
 ```
 
+To run the doc tests:
+
+```bash
+pytest --doctest-modules lancedb
+```
+
 To run linter and automatically fix all errors:
 
 ```bash
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import inspect
 import time
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from datetime import timedelta
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Tuple, Union
@@ -642,13 +643,33 @@ class Table(ABC):
         """
 
 
+class LanceTableMode:
+    pass
+
+
+@dataclass
+class LanceStandardMode(LanceTableMode):
+    last_consistency_check: Optional[float] = None
+
+
+@dataclass
+class LanceTimeTravelMode(LanceTableMode):
+    version: int
+
+
 class LanceTable(Table):
     """
     A table in a LanceDB database.
 
-    This can be opened in two modes: if `version` is None then the latest version
-    is opened and can be tracked. If `version` is an integer then that version is
-    opened and this instance is fixed to that version.
+    This can be opened in two modes: standard and time-travel.
+
+    Standard mode is the default. In this mode, the table is mutable and tracks
+    the latest version of the table. The level of read consistency is controlled
+    by the `read_consistency_interval` parameter on the connection.
+
+    Time-travel mode is activated by specifying a version number. In this mode,
+    the table is immutable and fixed to a specific version. This is useful for
+    querying historical versions of the table.
     """
 
     def __init__(
@@ -656,21 +677,17 @@ class LanceTable(Table):
         connection: "LanceDBConnection",
         name: str,
         version: Optional[int] = None,
-        /,
-        read_consistency_interval: Optional[timedelta] = None,
     ):
         self._conn = connection
         self.name = name
-        # To be lazy, we initially store the desired version as a LanceDataset
-        self._version: Optional[int] = version
-        self._ds: Optional[LanceDataset] = version
 
-        if self._version is not None and self.read_consistency_interval is not None:
-            raise ValueError(
-                "Cannot specify both version and read_consistency_interval."
-            )
-        self.read_consistency_interval = read_consistency_interval
-        self._last_consistency_check = None
+        if version is not None:
+            self._mode = LanceTimeTravelMode(version)
+        else:
+            self._mode = LanceStandardMode()
+
+        # Dataset is lazily loaded
+        self._ds: Optional[LanceDataset] = None
 
     @classmethod
     def open(cls, db, name, **kwargs):
@@ -695,23 +712,22 @@ class LanceTable(Table):
         # Returns the LanceDataset this wraps. This handles lazy loading (if
         # self._ds is None) and checking for new versions (if user has specified
         # a read_consistency_interval).
-        if not self._ds:
-            self._last_consistency_check = time.monotonic()
-            self._ds = lance.dataset(self._dataset_uri, version=self._version)
-        elif self._version is None and self.read_consistency_interval is not None:
+        if not self._ds and isinstance(self._mode, LanceTimeTravelMode):
+            self._ds = lance.dataset(self._dataset_uri, version=self._mode.version)
+        elif not self._ds and isinstance(self._mode, LanceStandardMode):
+            self._mode.last_consistency_check = time.monotonic()
+            self._ds = lance.dataset(self._dataset_uri)
+        elif (
+            isinstance(self._mode, LanceStandardMode)
+            and self._conn.read_consistency_interval is not None
+        ):
             now = time.monotonic()
-            diff = timedelta(seconds=now - self._last_consistency_check)
-            # print(diff, self.read_consistency_interval, self._ds.latest_version)
-            # breakpoint()
+            diff = timedelta(seconds=now - self._mode.last_consistency_check)
             if (
-                self._last_consistency_check is None
-                or diff > self.read_consistency_interval
+                self._mode.last_consistency_check is None
+                or diff > self._conn.read_consistency_interval
             ):
-                # TODO: use this method instead one available
-                # self._ds.checkout_version(self._ds.latest_version)
-                self._ds = lance.dataset(
-                    self._dataset_uri, version=self._ds.latest_version
-                )
+                self._ds = self._ds.checkout_version(self._ds.latest_version)
 
         return self._ds
 
@@ -719,10 +735,11 @@ class LanceTable(Table):
     def _dataset_mut(self) -> LanceDataset:
         # Returns the LanceDataset this wraps, but raises an error if the
         # dataset is fixed to a specific version.
-        if self._version is not None:
+        if isinstance(self._mode, LanceTimeTravelMode):
             raise ValueError(
-                f"Cannot mutate table reference fixed at version {self._version}. "
-                "Call checkout_latest() to get a mutable table reference."
+                "Cannot mutate table reference fixed at version "
+                f"{self._mode.version}. Call checkout_latest() to get a mutable "
+                "table reference."
             )
         return self._dataset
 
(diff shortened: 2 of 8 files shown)


python/lancedb/table.py
```diff
@@ -524,22 +523,90 @@ def update(
 class LanceTable(Table):
     """
     A table in a LanceDB database.
+
+    This can be opened in two modes: if `version` is None then the latest version
+    is opened and can be tracked. If `version` is an integer then that version is
```
westonpace, line 528: It's not clear to me what `can be tracked` means.

python/lancedb/table.py
```diff
@@ -524,22 +523,90 @@ def update(
...
+    def open(cls, db, name, **kwargs):
+        tbl = cls(db, name, **kwargs)
+        fs, path = fs_from_uri(tbl._dataset_uri)
+        file_info = fs.get_file_info(path)
+        if file_info.type != pa.fs.FileType.Directory:
+            raise FileNotFoundError(
+                f"Table {name} does not exist."
+                f"Please first call db.create_table({name}, data)"
+            )
+        register_event("open_table")
```
westonpace, line 563: What does `register_event` do?

python/lancedb/table.py
```diff
@@ -524,22 +523,90 @@ def update(
...
+        # self._ds is None) and checking for new versions (if user has specified
+        # a read_consistency_interval).
+        if not self._ds:
+            self._last_consistency_check = time.monotonic()
+            self._ds = lance.dataset(self._dataset_uri, version=self._version)
+        elif self._version is None and self.read_consistency_interval is not None:
+            now = time.monotonic()
+            diff = timedelta(seconds=now - self._last_consistency_check)
+            # print(diff, self.read_consistency_interval, self._ds.latest_version)
+            # breakpoint()
```
westonpace, line 583:
```suggestion
```

python/lancedb/table.py
```diff
@@ -524,22 +523,90 @@ def update(
...
+            if (
+                self._last_consistency_check is None
+                or diff > self.read_consistency_interval
+            ):
+                # TODO: use this method instead one available
```
westonpace, line 588: ?

python/lancedb/table.py
```diff
@@ -524,22 +523,90 @@ def update(
...
+
+    @property
+    def _dataset(self) -> LanceDataset:
+        # Returns the LanceDataset this wraps. This handles lazy loading (if
+        # self._ds is None) and checking for new versions (if user has specified
+        # a read_consistency_interval).
+        if not self._ds:
+            self._last_consistency_check = time.monotonic()
+            self._ds = lance.dataset(self._dataset_uri, version=self._version)
+        elif self._version is None and self.read_consistency_interval is not None:
```
westonpace, line 579: Isn't `self._version is None` implied by `self.read_consistency_interval is not None`?

python/lancedb/table.py
```diff
@@ -591,10 +658,20 @@ def checkout(self, version: int):
...
         0  [1.1, 0.9]  vector
         """
-        max_ver = max([v["version"] for v in self._dataset.versions()])
+        max_ver = self._dataset.latest_version
         if version < 1 or version > max_ver:
             raise ValueError(f"Invalid version {version}")
-        self._reset_dataset(version=version)
+
+        # TODO: use this method instead, once available.
+        # self._dataset.checkout_version(version)
```
westonpace, line 666: ?

python/lancedb/table.py
```diff
@@ -746,11 +843,12 @@ def create_index(
             accelerator=accelerator,
             index_cache_size=index_cache_size,
         )
-        self._reset_dataset()
```
westonpace, line 749: Was this meant to go away?

python/tests/test_table.py
```diff
@@ -657,3 +659,35 @@ def test_count_rows(db):
...
+    assert table_consistent.version == table.version
+    table_eventual = db.open_table(
+        "my_table", read_consistency_interval=timedelta(seconds=0.1)
+    )
+    assert table_eventual.version == table.version
+
+    table.add([{"id": 1}])
+    assert table_clone.version == table.version - 1
+    assert table_consistent.version == table.version
+    # assert table_eventual.version == table.version - 1
```
westonpace, line 681: Remove?

westonpace (commented):
This looks helpful to me.  Thanks for taking this on.

python/lancedb/table.py
wjones127: It's in the next release of Lance, and will allow us to update version without resetting the metadata cache.

python/lancedb/table.py
wjones127: Perhaps. I really miss Rust enums :(

python/lancedb/table.py
wjones127: Yes, I've removed reset_dataset intentionally. It's unnecessary.

python/tests/test_table.py
wjones127: Ah no that should be brought back. It was temporarily commented out while I was adding breakpoints.

python/lancedb/table.py
wjones127: That's the event tracking we setup to collect analytics on usage.

python/lancedb/table.py
wjones127: Okay, I've refactor this to be cleaner. There is now a `_mode` property which either holds a `version` field (if in time-travel mode) or a `last_consistency_check` field (if in standard mode).

python/lancedb/table.py
wjones127: I've rephrased this a bit. It now says the table can be in one of two modes: standard, which tracks the latest version, and time-travel, which is fixed at an older version. Hopefully this is clearer.

python/lancedb/table.py
```diff
@@ -523,23 +524,109 @@ def update(
...
+    version: int
+
+
 class LanceTable(Table):
     """
     A table in a LanceDB database.
+
+    This can be opened in two modes: standard and time-travel.
+
+    Standard mode is the default. In this mode, the table is mutable and tracks
```
changhiskhan, line 547: Should these docs be here or with the data classes?
The users don't actually set the mode, they just set the version number. So they might get confused? It might make more sense here to just say if you don't pass in a version number then the table is mutable and tracks the latest version etc?

python/lancedb/table.py
```diff
@@ -523,23 +524,109 @@ def update(
...
+    @property
+    def _dataset_uri(self) -> str:
+        return join_uri(self._conn.uri, f"{self.name}.lance")
+
+    @property
+    def _dataset(self) -> LanceDataset:
+        # Returns the LanceDataset this wraps. This handles lazy loading (if
+        # self._ds is None) and checking for new versions (if user has specified
+        # a read_consistency_interval).
+        if not self._ds and isinstance(self._mode, LanceTimeTravelMode):
```
changhiskhan, line 596: Instead of calling it a "mode", what if we called this `LanceDatasetRef` and you have immutable ("time travel mode") and mutable ("standard mode") variants. You construct a reference by passing in the dataset uri, the version (if specified), and the read_consistency_interval. The Ref holds on to the `._ds` dataset.

And the Ref class can have `get_dataset()` and `get_dataset_mut()` methods, and for the immutable one it just raises a TypeError.

This basically allows us to avoid all the different isinstance checks and don't need to keep all branches of the cascading if straight in our heads.

wdyt?

python/README.md
```diff
@@ -42,6 +42,12 @@ To run the unit tests:
 pytest
 ```
 
+To run the doc tests:
+
+```bash
+pytest --doctest-modules lancedb
```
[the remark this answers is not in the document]
wjones127: No, it validates the output values are correct.

python/lancedb/table.py
wjones127: > The users don't actually set the mode, they just set the version number

They do not. For the users, the modes are just naming concepts. That makes it easier to talk about them and explain.

>  So they might get confused?

I was trying to make it less confusing by given them a conceptual model for the two "modes" that it can run in. But I can leave the two modes unnamed instead, if you think that's clearer. We can just call it, "the mode where you passed an explicit version or called checkout()" and "the mode where you didn't pass an explicit version or call checkout()".

python/lancedb/table.py
wjones127: I like that idea. I can change to match that.

python/lancedb/table.py
```diff
@@ -523,23 +524,145 @@ def update(
...
+        )
+
+
 class LanceTable(Table):
     """
     A table in a LanceDB database.
+
+    This can be opened in two modes: standard and time-travel.
+
+    Standard mode is the default. In this mode, the table is mutable and tracks
```
changhiskhan, line 608: please update the comments here since Mode no longer exists

python/lancedb/table.py
```diff
@@ -523,23 +524,145 @@ def update(
...
-        self._version = version
 
-    def _reset_dataset(self, version=None):
-        try:
-            if "_dataset" in self.__dict__:
-                del self.__dict__["_dataset"]
-            self._version = version
-        except AttributeError:
-            pass
+        if version is not None:
```
changhiskhan, line 626: if a given version is the latest version, should we still treat it as time travel?

changhiskhan (approved):

python/lancedb/table.py
wjones127: Again, mode here was always conceptual so whether the mode attribute exists is irrelevant. But I can rephrase to not use the word mode if we don't like it.

python/lancedb/table.py
wjones127: Yes, I think it's much simpler to understand that way.

changhiskhan:
@wjones127 any more changes you're planning? Should we get this merged in?

wjones127:
@changhiskhan I think this is ready to go. I will merge this and then put up a PR to add Rust/Node changes as well as an entry in the docs about consistency.

PR #832, opened 2024-01-19
docs: document basics of configuring object storage

### LINKED ISSUE #681: Document how to configure object stores
There are various environment variables we accept. For example, for S3, we take these:

https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.from_env

We should document how to use these. A common use of them is credentials, but also how to connect to S3-compatible stores.

wjones127:
Created based on upstream PR https://github.com/lancedb/lance/pull/1849

Closes #681


prrao87:
@wjones127 the docs look great, but they specify which environment variables are required through the Rust crates for the different object stores. How would Python or JS users configure them? A couple of code snippets would be fabulous.

wjones127:
> How would Python or JS users configure them?

These cannot be configured within Python or Javascript; they are simply set as environment variables.

Code snippets for how to set them would look quite different depending on platform you are working in. Are you using Bash, Windows Powershell, jupyter notebooks, Github Actions, AWS Lambda, and or something else?

Maybe I can just show bash and that's good enough for 90% of users? But even then, the values I provide are going to be garbage, since these are mostly credentials or else private endpoints. So no user is going to be able to copy and paste an example and have it work on their machine. 🤷

docs/src/guides/storage.md
```diff
@@ -0,0 +1,37 @@
+# Configuring storage
+
+When using LanceDB OSS, you can choose where to store your data. The tradeoffs between different storage options are discussed in the [storage concepts guide](../concepts/storage.md). This guide shows how to configure LanceDB to use different storage options.
+
+## Object Stores
+
+LanceDB OSS supports object stores such as AWS S3 (and compatible stores), Azure Blob Store, and Google Cloud Storage. Which object store to use is determined by the URI scheme of the dataset path. For example, `s3://bucket/path` will use S3, `az://bucket/path` will use Azure, and `gs://bucket/path` will use GCS. In most cases, when running in the respective cloud and permissions are set up correctly, no additional configuration is required.
```
prrao87, line 7: The object store would have credentials specified by environment variables, no doubt. But for a new user, how would they go about configuring the database once they `pip install lancedb` or `npm install vectordb`? Is the only action on their part to specify `db = lancedb.connect("s3://bucket/path")`?

prrao87: It would still help to have the snippet of code showing how the DB connection is different from a local one.

prrao87 (commented):
From the perspective of a new user coming from Python or JavaScript, I think pointing them to a Rust crate might make it seem intimidating. Sometimes, people are just looking for code to get going :)

wjones127:
> Also, considering that this document's focus is only cloud storage, I'm leaning more towards Chang's comment in https://github.com/lancedb/lancedb/issues/681 to naming it "Cloud storage" because that makes it immediately clear.

My expectation is that we will add guides for networked file stores or even local files stores to this same page. But I can rename it "cloud storage" for now, and we can switch the name back once we add other types.

wjones127:
> From the perspective of a new user coming from Python or JavaScript, I think pointing them to a Rust crate might make it seem intimidating. Sometimes, people are just looking for code to get going :)

I've added code examples showing how to use the LanceDB `connect` function. I've also expanded the paragraph explaining that in most cases no additional configuration is required beyond setting credentials (which work just like any other cloud SDK).

Also added a couple sentences re-assuring them that the Rust docs are not so scary.

docs/src/guides/storage.md
```diff
@@ -0,0 +1,91 @@
...
+    ```
+
+In most cases, when running in the respective cloud and permissions are set up correctly, no additional configuration is required. When running outside of the respective cloud, authentication credentials must be provided using environment variables. In general, these environment variables are the same as those used by the respective cloud SDKs. The sections below describe the environment variables that can be used to configure each object store.
+
+LanceDB OSS uses the [object-store](https://docs.rs/object_store/latest/object_store/) Rust crate for object store access. There are general environment variables that can be used to configure the object store, such as the request timeout and proxy configuration. See the [object_store ClientConfigKey](https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html) doc for available configuration options. The environment variables that can be set are the snake-cased versions of these variable names. For example, to set `ProxyUrl` use the environment variable `PROXY_URL`. (Don't let the Rust docs intimidate you! We show you them so you can see an up-to-date list of the available options.)
```
prrao87, line 60:
```suggestion
LanceDB OSS uses the [object-store](https://docs.rs/object_store/latest/object_store/) Rust crate for object store access. There are general environment variables that can be used to configure the object store, such as the request timeout and proxy configuration. See the [object_store ClientConfigKey](https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html) doc for available configuration options. The environment variables that can be set are the snake-cased versions of these variable names. For example, to set `ProxyUrl` use the environment variable `PROXY_URL`. (Don't let the Rust docs intimidate you! We link to them so you can see an up-to-date list of the available options.)
```

prrao87 (approved):
Looks good, I think this is good to be merged. For a future PR that adds more information on managing DB settings for local or any of the other options in the storage concepts, we could just add them as subheaders to this one 👍🏽

PR #841, opened 2024-01-22
allow passing api key as env var

albertlockett:
Allow passing API key as env var:
```shell
export LANCEDB_API_KEY=sh_123...
```

with this set, apiKey argument can omitted from `connect`
```js
    const db = await vectordb.connect({
        uri: "db://test-proj-01-ae8343",
        region: "us-east-1",
  })
```
```py
    db = lancedb.connect(
        uri="db://test-proj-01-ae8343",
        region="us-east-1",
    )
```

### DIFF
--- a/node/src/remote/index.ts
+++ b/node/src/remote/index.ts
@@ -38,6 +38,9 @@ export class RemoteConnection implements Connection {
     if (!opts.uri.startsWith('db://')) {
       throw new Error(`Invalid remote DB URI: ${opts.uri}`)
     }
+    if (opts.apiKey == null || opts.apiKey === '') {
+      opts = Object.assign({}, opts, { apiKey: process.env.LANCEDB_API_KEY })
+    }
     if (opts.apiKey === undefined || opts.region === undefined) {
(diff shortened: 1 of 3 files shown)


node/src/remote/index.ts
```diff
@@ -38,6 +38,9 @@ export class RemoteConnection implements Connection {
     if (!opts.uri.startsWith('db://')) {
       throw new Error(`Invalid remote DB URI: ${opts.uri}`)
     }
+    if (opts.apiKey == null || opts.apiKey === '') {
+      opts = Object.assign({}, opts, { apiKey: process.env.LANCEDB_API_KEY })
+    }
     if (opts.apiKey === undefined || opts.region === undefined) {
       throw new Error('API key and region are not supported for remote connections')
```
QianZhu, line 45: Is this the right error message?

QianZhu: can we also add something like `you can set api through LANCEDB_API_KEY env variable or remote db url`?

albertlockett: updated

QianZhu (approved):

PR #845, opened 2024-01-22
feat: change create table to accept Arrow table

eddyxu:


node/src/query.ts
```diff
@@ -129,7 +129,7 @@ export class Query<T = number[]> {
       const newObject: Record<string, unknown> = {}
       Object.keys(entry).forEach((key: string) => {
         if (entry[key] instanceof Vector) {
-          newObject[key] = (entry[key] as Vector).toArray()
+          newObject[key] = (entry[key] as Vector).toJSON()
```
wjones127, line 132: Why this change?

wjones127 (commented):
So many formatting changes 😢

node/src/query.ts
eddyxu: Arrow `toArray()` returns f16 array as uint16

wjones127: That seems worth a comment.

eddyxu: done

PR #860, opened 2024-01-25
feat: improve the rust table query API and documents

eddyxu:
* Easy to type
* Handle `String, &str, [String] and [&str]` well without manual conversion
* Fix function name to be verb
* Improve docstring of Rust.
* Promote `query` and `search()` to public `Table` trait

### DIFF
--- a/rust/vectordb/src/index.rs
+++ b/rust/vectordb/src/index.rs
@@ -14,13 +14,12 @@
 
 use std::{cmp::max, sync::Arc};
 
-use arrow_schema::Schema;
 use lance_index::{DatasetIndexExt, IndexType};
 pub use lance_linalg::distance::MetricType;
 
 pub mod vector;
 
-use crate::{Error, Result, Table};
+use crate::{utils::default_vector_column, Error, Result, Table};
 
 /// Index Parameters.
 pub enum IndexParams {
@@ -110,6 +109,7 @@ impl IndexBuilder {
         self
     }
 
+    /// The columns to build index on.
     pub fn columns(&mut self, cols: &[&str]) -> &mut Self {
         self.columns = cols.iter().map(|s| s.to_string()).collect();
         self
@@ -174,7 +174,7 @@ impl IndexBuilder {
         let columns = if self.columns.is_empty() {
             // By default we create vector index.
             index_type = &IndexType::Vector;
-            vec![default_column_for_index(&schema)?]
+            vec![default_vector_column(&schema, None)?]
         } else {
             self.columns.clone()
         };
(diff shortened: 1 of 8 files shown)


rust/vectordb/src/index.rs
```diff
@@ -290,83 +290,3 @@ fn suggested_num_sub_vectors(dim: u32) -> u32 {
         1
     }
 }
-
-/// Find one default column to create index.
-fn default_column_for_index(schema: &Schema) -> Result<String> {
```
eddyxu, line 295: Code and tests are moved to utils.rs

rust/vectordb/src/query.rs
```diff
@@ -64,17 +76,24 @@ impl Query {
...
     }
 
-    /// Execute the queries and return its results.
+    /// Convert the query plan to a [`DatasetRecordBatchStream`]
     ///
     /// # Returns
     ///
     /// * A [DatasetRecordBatchStream] with the query's results.
-    pub async fn execute(&self) -> Result<DatasetRecordBatchStream> {
+    pub async fn into_stream(&self) -> Result<DatasetRecordBatchStream> {
```
wjones127, line 84: nit: `into_` methods will take (owned) `self` not `&self`. (see: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv)

What's wrong with `.execute()`? IMO that seemed like a fine name.

rust/vectordb/src/query.rs
```diff
@@ -64,17 +76,24 @@ impl Query {
...
 
         if let Some(query) = self.query_vector.as_ref() {
             // If there is a vector query, default to limit=10 if unspecified
-            scanner.nearest(&self.column, query, self.limit.unwrap_or(10))?;
+            let column = if let Some(col) = self.column.as_ref() {
+                col.clone()
+            } else {
+                // Infer a vector column with the same dimension of the query vector.
+                let arrow_schema = Schema::from(self.dataset.schema());
+                default_vector_column(&arrow_schema, Some(query.len() as i32))?
```
wjones127, line 94: Fancy! I hope there are no unintended consequences.

wjones127 (commented):
Just one nit

rust/vectordb/src/query.rs
eddyxu: `execute` means the materialization happens within the function itself (as it is done already after the call finishes).

`something_become_a_stream()` means that it returns a lazy evaluated stream that users should take care buffered / try_collect and etc. ?

eddyxu: what about `to_stream()` or `stream()`? Any other suggestions?

wjones127: But `to` / `into` mean converting from one type to some similar type. I think someone holding a `Query` object is looking for the word `execute`, because they want to do that action (execute the query). Whether or not the full results are materialized within the functions seems not so relevant IMO.

In DataFusion, they call the method `execute_stream()` [^1]

[^1]: https://docs.rs/datafusion/latest/datafusion/dataframe/struct.DataFrame.html#method.execute_stream

eddyxu: ok, changed to execute_stream()

wjones127 (approved):

PR #869, opened 2024-01-26
chore(rust): simplified version of optimize

eddyxu:
Consolidate various optimize() into one method, similar to postgres VACCUM in the process of preparing Rust API for public use

### DIFF
--- a/rust/vectordb/src/table.rs
+++ b/rust/vectordb/src/table.rs
@@ -70,6 +70,15 @@ impl Default for OptimizeAction {
     }
 }
 
+/// Statistics about the optimization.
+pub struct OptimizeStats {
+    /// Stats of the file compaction.
+    pub compaction: Option<CompactionMetrics>,
+
+    /// Stats of the version pruning
+    pub prune: Option<RemovalStats>,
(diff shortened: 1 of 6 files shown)


rust/vectordb/src/table.rs
```diff
@@ -396,17 +443,8 @@ impl NativeTable {
...
-        let params = ScalarIndexParams::default();
-        dataset
-            .create_index(&[column], IndexType::Scalar, None, &params, replace)
-            .await?;
-        Ok(())
-    }
-
-    pub async fn optimize_indices(&mut self, options: &OptimizeOptions) -> Result<()> {
+    async fn optimize_indices(&self, options: &OptimizeOptions) -> Result<()> {
+        info!("LanceDB: optimizing indices: {:?}", options);
```
changhiskhan, line 447: this should be debug level?

eddyxu: I was meant to make this show progress of optimization tasks, as it is a slow process.

I am fine either way.

albertlockett (approved):

PR #871, opened 2024-01-26
feat(rust): provide connect and connect_with_options in Rust SDK

eddyxu:
* Bring the feature parity of Rust connect methods.
* A global connect method that can connect to local and remote / cloud table, as the same as in js/python today.

### DIFF
--- a/rust/vectordb/src/connection.rs
+++ b/rust/vectordb/src/connection.rs
@@ -68,7 +68,7 @@ pub trait Connection: Send + Sync {
     async fn drop_table(&self, name: &str) -> Result<()>;
 }
 
-
+#[derive(Debug)]
 pub struct ConnectOptions {
     /// Database URI
     ///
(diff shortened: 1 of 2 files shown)


rust/vectordb/src/connection.rs
```diff
@@ -68,6 +68,87 @@ pub trait Connection: Send + Sync {
...
+    /// - `s3://bucket/path/to/database` or `gs://bucket/path/to/database` - database on cloud object store
+    /// - `db://dbname` - Lance Cloud
+    pub uri: String,
+
+    /// Lance Cloud API key
+    pub api_key: Option<String>,
+    /// Lance Cloud region
+    pub region: Option<String>,
+    /// Lance Cloud host override
+    pub host_override: Option<String>,
```
eddyxu, line 87: Can add AwsCredntials and other parameters we need later

albertlockett (approved):

PR #872, opened 2024-01-26
doc: update quick start for full rust example

eddyxu:


docs/src/basic.md
```diff
@@ -11,43 +11,59 @@
 ## Installation
 
 === "Python"
+
```
eddyxu, line 14: Seems to make vscode happy without crazy reformatting

docs/src/basic.md
```diff
@@ -88,6 +149,7 @@ Sometimes you may not have the data to insert into the table at creation time.
 In this case, you can create an empty table and specify the schema.
 
 === "Python"
+
```
QianZhu, line 152: nit: should we also add js and rust code

eddyxu:
Nodejs document tests failed on main as well.

docs/src/basic.md
eddyxu: Oh i thought javascript / rust can not do it yet. I can double check

QianZhu: also, should we add to our workflow to check the example rust code?

QianZhu: javascript should be able to create an empty table

eddyxu: Yea that'd be great. https://github.com/lancedb/lancedb/issues/875

QianZhu (approved):

PR #883, opened 2024-01-28
fix: fix doc build to include the source snippet correctly

eddyxu:


.github/workflows/docs.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: Set up node
         uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
```
eddyxu, line 45: there is no matrix defined in this job

changhiskhan (approved):

PR #898, opened 2024-01-29
Update JS/TS example in README

AbrahamLopez10:
- The JS/TS library actually expects named parameters via an object in `.createTable()` rather than individual arguments
- Added example on how to search rows by criteria without a vector search. TS type of `.search()` currently has the `query` parameter as non-optional so we have to pass undefined for now.


changhiskhan (approved):

changhiskhan:
Thanks!

PR #900, opened 2024-01-30
docs: add cleanup_old_versions and compact_files to `Table` for documentation purposes

### LINKED ISSUE #819: Cleanup and compaction are missing from the API docs
### Description

I believe its because these methods are only on LanceTable (and not Table). I would expect to see them here:

image

### Link

_No response_

westonpace:
Closes #819

### DIFF
--- a/python/lancedb/table.py
+++ b/python/lancedb/table.py
@@ -1279,8 +1279,9 @@ class LanceTable(Table):
         This can be run after making several small appends to optimize the table
         for faster reads.
 
-        Arguments are passed onto :meth:`lance.dataset.DatasetOptimizer.compact_files`.
-        For most cases, the default should be fine.
+        Arguments are passed onto `lance.dataset.DatasetOptimizer.compact_files`.
+         (see Lance documentation for more details) For most cases, the default
+        should be fine.
         """
         return self.to_lance().optimize.compact_files(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


python/lancedb/table.py
```diff
@@ -521,6 +521,52 @@ def update(
...
+        """
+        Run the compaction process on the table.
+
+        Note: This function is not available in LanceDb Cloud (since LanceDb
+        Cloud manages compaction for you automatically)
+
+        This can be run after making several small appends to optimize the table
+        for faster reads.
+
+        Arguments are passed onto :meth:`lance.dataset.DatasetOptimizer.compact_files`.
```
wjones127, line 566: Should check me on this, but I think since we use mkdocs not Sphinx, method references look like:
```suggestion
        Arguments are passed onto [lance.dataset.DatasetOptimizer.compact_files][].
```

westonpace: Yep.  Good catch.

westonpace: I've removed the link entirely (since it is now an external link from LanceDB docs -> Lance docs and wasn't sure if we wanted a potentially unstable `https://...` link)

westonpace: I'm not sure if Github is just being slow / showing me stale files or if I'm missing something obvious: https://github.com/lancedb/lancedb/pull/900/commits/16aafe9427f5d53332462a05ef4f36ed22dd3ba6

wjones127 (approved):

PR #904, opened 2024-01-31
feat(python): Reranker DX improvements

AyushExel:
- Most users might not know how to use `QueryBuilder` object. Instead we should just pass the string query. 
- Add new rerankers: Colbert, openai

### DIFF
--- a/docs/src/hybrid_search.md
+++ b/docs/src/hybrid_search.md
@@ -186,7 +186,7 @@ class MyReranker(Reranker):
         self.param1 = param1
         self.param2 = param2
     
-    def rerank_hybrid(self, vector_results: pa.Table, fts_results: pa.Table):
+    def rerank_hybrid(self, query: str, vector_results: pa.Table, fts_results: pa.Table):
         # Use the built-in merging function
         combined_result = self.merge_results(vector_results, fts_results)
         
@@ -208,7 +208,7 @@ import pyarrow as pa
 class MyReranker(Reranker):
     ...
     
-    def rerank_hybrid(self, vector_results: pa.Table, fts_results: pa.Table, filter: str):
+    def rerank_hybrid(self, query: str, vector_results: pa.Table, fts_results: pa.Table, filter: str):
         # Use the built-in merging function
         combined_result = self.merge_results(vector_results, fts_results)
         
--- a/docs/src/hybrid_search.md
+++ b/docs/src/hybrid_search.md
@@ -121,6 +121,55 @@ Arguments
     Only returns `_relevance_score`. Does not support `return_score = "all"`.
 
 
+### ColBERT Reranker
+This reranker uses the ColBERT model to combine the results of semantic and full-text search. You can use it by passing `ColbertrReranker()` to the `rerank()` method. 
+
+```python
+from lancedb.rerankers import ColbertrReranker
+
+reranker = ColbertrReranker()
+
+results = table.search("harmony hall", query_type="hybrid").rerank(reranker=reranker).to_pandas()
+```
+
+Arguments
+----------------
+* `model_name` : `str`, default `"colbert-ir/colbertv2.0"`
+        The name of the cross encoder model to use.
+* `column` : `str`, default `"text"`
+        The name of the column to use as input to the cross encoder model.
+* `return_score` : `str`, default `"relevance"`
+        options are `"relevance"` or `"all"`. Only `"relevance"` is supported for now.
+
+!!! Note
+    Only returns `_relevance_score`. Does not support `return_score = "all"`.
+
+### OpenAI Reranker
+This reranker uses the OpenAI API to combine the results of semantic and full-text search. You can use it by passing `OpenaiReranker()` to the `rerank()` method.
+
+!!! Note
+    This prompts chat model to rerank results. This is not a dedicated reranker model.
+
+```python
+from lancedb.rerankers import OpenaiReranker
+
+reranker = OpenaiReranker()
+
+results = table.search("harmony hall", query_type="hybrid").rerank(reranker=reranker).to_pandas()
+```
+
+Arguments
+----------------
+`model_name` : `str`, default `"gpt-3.5-turbo-1106"`
+    The name of the cross encoder model to use.
+`column` : `str`, default `"text"`
+    The name of the column to use as input to the cross encoder model.
+`return_score` : `str`, default `"relevance"`
+    options are "relevance" or "all". Only "relevance" is supported for now.
+`api_key` : `str`, default `None`
+    The API key to use. If None, will use the OPENAI_API_KEY environment variable.
+
+
 ## Building Custom Rerankers
 You can build your own custom reranker by subclassing the `Reranker` class and implementing the `rerank_hybrid()` method. Here's an example of a custom reranker that combines the results of semantic and full-text search using a linear combination of the scores.
 
--- a/docs/src/hybrid_search.md
+++ b/docs/src/hybrid_search.md
@@ -133,6 +133,8 @@ Arguments
 ### ColBERT Reranker
 This reranker uses the ColBERT model to combine the results of semantic and full-text search. You can use it by passing `ColbertrReranker()` to the `rerank()` method. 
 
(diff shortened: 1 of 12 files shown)


python/tests/test_rerankers.py
```diff
@@ -150,13 +158,16 @@ def test_cross_encoder_reranker(tmp_path):
...
         .to_pydantic(schema)
     )
     assert result1 == result2
 
+    # test explicit hybrid query
+    query = "Our father who art in heaven"
+    query_vector = table.to_pandas()["vector"][0]
     result = (
-        table.search("Our father who art in heaven", query_type="hybrid")
+        table.search((query_vector, query), query_type="hybrid")
```
AyushExel, line 170: @changhiskhan I'm not sure if we should document this usage or come up with another design for accepting text,vector pairs for running hybrid search in tables without ingested embedding functions

python/lancedb/rerankers/colbert.py
```diff
@@ -0,0 +1,106 @@
...
+        docs = combined_results[self.column].to_pylist()
+
+        tokenizer, model = self._model
+
+        # Encode the query
+        query_encoding = tokenizer(query, return_tensors="pt")
+        query_embedding = model(**query_encoding).last_hidden_state.mean(dim=1)
+        scores = []
+        # Get score for each document
+        for document in docs:
```
AyushExel, line 49: This could be vectorized further but rn the 2nd dim of doc embeddings is dependent on the the doc length - [1, len, 768]..if we force them to be the same by padding the accuract will suffer..It's pretty fast as it is

python/lancedb/query.py
```diff
@@ -679,7 +678,13 @@ def to_arrow(self) -> pa.Table:
         # rerankers might need to preserve this score to support `return_score="all"`
         fts_results = self._normalize_scores(fts_results, "score")
 
-        results = self._reranker.rerank_hybrid(self, vector_results, fts_results)
+        results = self._reranker.rerank_hybrid(
+            self._fts_query._query, vector_results, fts_results
+        )
+
+        # apply limit after reranking
+        results = results.slice(length=self._limit)
```
AyushExel: @prrao87 @changhiskhan This is what we settled for right?

python/lancedb/query.py
```diff
@@ -679,12 +678,18 @@ def to_arrow(self) -> pa.Table:
...
+        )
+
         if not isinstance(results, pa.Table):  # Enforce type
             raise TypeError(
                 f"rerank_hybrid must return a pyarrow.Table, got {type(results)}"
             )
 
+        # apply limit after reranking
+        results = results.slice(length=self._limit)
+
```
AyushExel, line 692: @prrao87 @changhiskhan this is what we settled for right?

prrao87: Yup, lgtm

docs/src/hybrid_search.md
```diff
@@ -130,6 +130,55 @@ Arguments
     Only returns `_relevance_score`. Does not support `return_score = "all"`.
 
 
+### ColBERT Reranker
+This reranker uses the ColBERT model to combine the results of semantic and full-text search. You can use it by passing `ColbertrReranker()` to the `rerank()` method.
```
changhiskhan, line 134: Somewhere here you should a bit more explanation that:
1. The ColbertReranker needs the actual document rather than purely based on scores 
2. By default it uses the "text" column but if it's different in your table you can specify it in the ColbertReranker constructor

python/lancedb/rerankers/colbert.py
```diff
@@ -0,0 +1,107 @@
...
+        combined_results = combined_results.append_column(
+            self.column, pa.array(docs, type=pa.string())
+        )
+        # add the scores
+        combined_results = combined_results.append_column(
+            "_relevance_score", pa.array(scores, type=pa.float32())
+        )
+        if self.score == "relevance":
+            combined_results = combined_results.drop_columns(["score", "_distance"])
+        elif self.score == "all":
```
changhiskhan, line 70: Why isn't this supported?

python/lancedb/rerankers/openai.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def rerank_hybrid(
+        self,
+        query: str,
+        vector_results: pa.Table,
+        fts_results: pa.Table,
+    ):
+        combined_results = self.merge_results(vector_results, fts_results)
+        docs = combined_results[self.column].to_pylist()
+        response = self._client.chat.completions.create(
```
changhiskhan, line 50: What if the docs exceed the token limit?

docs/src/hybrid_search.md
```diff
@@ -130,6 +130,55 @@ Arguments
...
+        options are `"relevance"` or `"all"`. Only `"relevance"` is supported for now.
+
+!!! Note
+    Only returns `_relevance_score`. Does not support `return_score = "all"`.
+
+### OpenAI Reranker
+This reranker uses the OpenAI API to combine the results of semantic and full-text search. You can use it by passing `OpenaiReranker()` to the `rerank()` method.
+
+!!! Note
+    This prompts chat model to rerank results. This is not a dedicated reranker model.
```
changhiskhan, line 160: Does Instructor have something for this so we don't have to manage the actual prompt etc?

python/lancedb/query.py
```diff
@@ -679,12 +678,18 @@ def to_arrow(self) -> pa.Table:
...
+            self._fts_query._query, vector_results, fts_results
+        )
+
         if not isinstance(results, pa.Table):  # Enforce type
             raise TypeError(
                 f"rerank_hybrid must return a pyarrow.Table, got {type(results)}"
             )
 
+        # apply limit after reranking
+        results = results.slice(length=self._limit)
```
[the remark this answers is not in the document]
AyushExel: yeah..All default rerankers sort in descending order of relevancy.

python/lancedb/rerankers/colbert.py
AyushExel: These reranker functions (except linear combination) don't take the existing scores into account. So, we can use a a faster merging algo which just merges fts+vector and keeps the first instance of rowid if repeated. This means we just discard the existing scores. 

We can use the slower merging algo like linear combinator if we want to support scores="all" but I feel like the tradeoff won't make sense as these rerankers don't really use the existing scores to rerank

python/lancedb/rerankers/openai.py
AyushExel: errors out..should we handle it? I guess this could also happen in embedding function..So maybe we can just treat it as a user error?

docs/src/hybrid_search.md
AyushExel: You mean instructor by jason?

python/lancedb/rerankers/colbert.py
changhiskhan: is `score="all"` the default option? If so then people will immediately get an error using defaults.
maybe it's not a blocker for this PR but this option is still useful for debugging because it allows you to see whether the vector search is returning reasonable answers.

python/lancedb/rerankers/openai.py
changhiskhan: yeah i guess this depends on the model/api, some will raise and others will just truncate. 

So what is the user's recourse if they hit this? The annoying thing for them is that when they get here, they've already spent the time/effort to embed the table, so if it means they have re-chunk and re-embed it's going to be really annoying.

Probably not a blocker but definitely worth addressing

docs/src/hybrid_search.md
changhiskhan: yeah. we probably shouldn't depend on instructor given how fast it's moving, just curious.

changhiskhan:
lmk once you're ready, please make another DX improvement issue, i think we've collected a bunch of stuff at this point that's worth addressing next.

python/lancedb/rerankers/colbert.py
AyushExel: No the default is `score="relevance"`

changhiskhan (approved):
pending CI

AyushExel:
dunno whats happening.. will fight ci tmrw

PR #907, opened 2024-01-31
arrow table/f16 example

QianZhu:


docs/src/guides/tables.md
```diff
@@ -284,6 +294,35 @@ This guide will show how to create tables, insert data into them, and update the
         "price": 20.0
     }]);
     ```
+    ### From an Arrow Table
```
eddyxu, line 297: These are two different sections?
Can we put them into the same Section "From an Arrow Table", but two different tabs, as what happens in the quickstart.

Also we can include the code in real ts file and make sure they compile / tests

For example, https://github.com/lancedb/lancedb/blob/main/docs/src/basic_legacy.ts

docs/src/guides/tables.md
```diff
@@ -125,7 +158,38 @@ This guide will show how to create tables, insert data into them, and update the
     tbl = db.create_table("f16_tbl", data, schema=schema)
     ```
 
+=== "Javascript"
+    You can also create LanceDB tables directly from Arrow tables. 
+    LanceDB supports Float16 data type!
+
+    ```javascript
+    import { FixedSizeList,Field,Int32,Schema,Float16, Table as ArrowTable } from 'apache-arrow'
```
eddyxu, line 166: can we use snippet import of this code example? 
Mostly, i'd like to make sure that we can get this piece of code compiled and tested.

eddyxu (approved):

PR #908, opened 2024-01-31
fix(node): pass AWS credentials to db level operations

eddyxu:
Passed the following tests

```ts
const keyId = process.env.AWS_ACCESS_KEY_ID;
const secretKey = process.env.AWS_SECRET_ACCESS_KEY;
const sessionToken = process.env.AWS_SESSION_TOKEN;
const region = process.env.AWS_REGION;

  uri: "s3://bucket/path",
  awsCredentials: {
    accessKeyId: keyId,
    secretKey: secretKey,
    sessionToken: sessionToken,
  },
  awsRegion: region,
} as lancedb.ConnectionOptions);

  console.log(await db.createTable("test", [{ vector: [1, 2, 3] }]));
  console.log(await db.tableNames());
  console.log(await db.dropTable("test"))
```

### DIFF
--- a/rust/ffi/node/src/lib.rs
+++ b/rust/ffi/node/src/lib.rs
@@ -17,14 +17,14 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use lance::io::ObjectStoreParams;
 use neon::prelude::*;
-use object_store::aws::{AwsCredential, AwsCredentialProvider};
+use object_store::aws::{self, AwsCredential, AwsCredentialProvider};
 use object_store::CredentialProvider;
 use once_cell::sync::OnceCell;
 use tokio::runtime::Runtime;
 
 use vectordb::connection::Database;
 use vectordb::table::ReadParams;
-use vectordb::Connection;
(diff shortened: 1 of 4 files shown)


rust/ffi/node/src/lib.rs
```diff
@@ -82,13 +82,26 @@ fn runtime<'a, C: Context<'a>>(cx: &mut C) -> NeonResult<&'static Runtime> {
 
 fn database_new(mut cx: FunctionContext) -> JsResult<JsPromise> {
     let path = cx.argument::<JsString>(0)?.value(&mut cx);
+    let aws_creds = get_aws_creds(&mut cx, 1)?;
```
westonpace, line 85: It looks like you're getting a single argument here at index `1` but you're actually passing in 3 different arguments at index `1`, `2`, and `3`.  How does that work?

eddyxu: This `get_aws_cred` get three parameters starting from position 1. 

I reused the logic. But we will get ride of it soon in favor of the new napi based API.

PR #909, opened 2024-01-31
ci: bump to new version of python action to use node 20 gIthub action runtime

eddyxu:
Github action is deprecating old node-16 runtime.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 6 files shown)


wjones127 (commented):
Where is the node change? Also should use use 18 not 20 since our minimum version is 18?

https://github.com/lancedb/lancedb/blob/34e10caad24638f3f35d1407a4db01da5043036a/nodejs/package.json#L35C12-L35C20

eddyxu:
this is `python actions` that uses nodejs20 to run the GHA . Github actions are nodejs scripts. 

https://github.com/actions/setup-python/releases/tag/v5.0.0

It does not impact which version of node we run our code

wjones127:
Ohhhh I see. Thanks for explaining.

wjones127 (approved):

PR #912, opened 2024-02-01
feat(python): add support new openai embedding functions

AyushExel:
@PrashantDixit0

### DIFF
--- a/python/lancedb/embeddings/openai.py
+++ b/python/lancedb/embeddings/openai.py
@@ -12,7 +12,7 @@
 #  limitations under the License.
 import os
 from functools import cached_property
-from typing import List, Union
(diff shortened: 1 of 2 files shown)


PrashantDixit0:
I'll look into it

python/lancedb/embeddings/openai.py
```diff
@@ -30,10 +30,17 @@ class OpenAIEmbeddings(TextEmbeddingFunction):
     """
 
     name: str = "text-embedding-ada-002"
+    dim: Optional[int] = None
 
     def ndims(self):
```
changhiskhan, line 35: let's make this a cached property?

changhiskhan (approved):
1 minor change

PR #916, opened 2024-02-02
make it explicit about the vector column data type

QianZhu:
Screenshot 2024-02-01 at 4 23 34 PM
[
Screenshot 2024-02-01 at 4 26 03 PM
](url)

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -102,7 +102,7 @@ This guide will show how to create tables, insert data into them, and update the
     !!! info "Note"
     Data is converted to Arrow before being written to disk. For maximum control over how data is saved, either provide the PyArrow schema to convert to or else provide a PyArrow Table directly.
 
-    The **`vector`** column needs to be a [Vector](../python/pydantic.md#vector-field) (defined as [pyarrow.FixedSizeList](https://arrow.apache.org/docs/python/ generated/pyarrow.list_.html)) type. 
+    The **`vector`** column needs to be a [Vector](../python/pydantic.md#vector-field) (defined as [pyarrow.FixedSizeList](https://arrow.apache.org/docs/python/generated/pyarrow.list_.html)) type. 
 
     ```python
     custom_schema = pa.schema([
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -84,7 +84,7 @@ This guide will show how to create tables, insert data into them, and update the
     const table = await con.createTable(tableName, data, { writeMode: WriteMode.Overwrite })
     ```
 
-### From a Pandas DataFrame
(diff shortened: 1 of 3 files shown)


docs/src/guides/tables.md
```diff
@@ -100,7 +100,9 @@ This guide will show how to create tables, insert data into them, and update the
     db["my_table"].head()
     ```
     !!! info "Note"
-        Data is converted to Arrow before being written to disk. For maximum control over how data is saved, either provide the PyArrow schema to convert to or else provide a PyArrow Table directly.
+    Data is converted to Arrow before being written to disk. For maximum control over how data is saved, either provide the PyArrow schema to convert to or else provide a PyArrow Table directly.
+
+    The **`vector`** column needs to be a [Vector](../python/pydantic.md#vector-field) (defined as [pyarrow.FixedSizeList](https://arrow.apache.org/docs/python/ generated/pyarrow.list_.html)) type.
```
westonpace, line 105:
```suggestion
    The **`vector`** column needs to be a [Vector](../python/pydantic.md#vector-field) (defined as [pyarrow.FixedSizeList](https://arrow.apache.org/docs/python/generated/pyarrow.list_.html)) type. 
```
Is this space supposed to be in the URL?

docs/src/python/pydantic.md
```diff
@@ -13,7 +13,7 @@ via [pydantic_to_schema()](python.md#lancedb.pydantic.pydantic_to_schema) method
 
 ## Vector Field
 
-LanceDB provides a [`Vector(dim)`](python.md#lancedb.pydantic.Vector) method to define a
+LanceDB provides a [`Vector(dim)`](python.md#utilities) method to define a
```
westonpace, line 16: It seems you are changing the link target from `lancedb.pydantic.Vector` to `lancedb.vector`.  Should you change the link text from `Vector(dim)` to`vector(dim)`?

python/lancedb/table.py
```diff
@@ -438,6 +438,8 @@ def search(
             the table
         vector_column_name: str
             The name of the vector column to search.
+
+            The vector column needs to be a "pyarrow.FixedSizeList" type
```
westonpace, line 442: Hmm, this wording makes it seem like `pyarrow.FixedSizeList` is some kind of python class but it isn't.  How about just pyarrow fixed size list type (without the double quotes)?

westonpace (commented):
A few suggestions but this is definitely an improvement.  Thanks!

docs/src/python/pydantic.md
QianZhu: let me revert the change. vector(dim) will be removed in the future release

westonpace (approved):

PR #917, opened 2024-02-02
fix: add request retry to python client

### LINKED ISSUE #911: Feature: Retries on http 500 or 502 status.
### SDK

Python

### Description

Add retries logic in the event of network congestion, infra network issues, or recoverable service failures, to improve overall reliability of remote client

albertlockett:
Adds capability to the remote python SDK to retry requests (fixes #911)

This can be configured through environment:
- `LANCE_CLIENT_MAX_RETRIES`= total number of retries. Set to 0 to disable retries. default = 3
- `LANCE_CLIENT_CONNECT_RETRIES` = number of times to retry request in case of TCP connect failure. default = 3
- `LANCE_CLIENT_READ_RETRIES` = number of times to retry request in case of HTTP request failure. default = 3
- `LANCE_CLIENT_RETRY_STATUSES` = http statuses for which the request will be retried. passed as comma separated list of ints. default `500, 502, 503`
- `LANCE_CLIENT_RETRY_BACKOFF_FACTOR` = controls time between retry requests. see [here](https://github.com/urllib3/urllib3/blob/23f2287eb526d9384dddeedb6f6345e263bb9b86/src/urllib3/util/retry.py#L141-L146). default = 0.25

Only read requests will be retried:
- list table names
- query
- describe table
- list table indices

This does not add retry capabilities for writes as it could possibly cause issues in the case where the retried write isn't idempotent. For example, in the case where the LB times-out the request but the server completes the request anyway, we might not want to blindly retry an insert request.

### DIFF
--- a/python/lancedb/remote/client.py
+++ b/python/lancedb/remote/client.py
@@ -197,6 +197,7 @@ def retry_adapter_options(methods = ["GET"]) -> Dict[str, Any]:
         "connect_retries": int(os.environ.get("LANCE_CLIENT_CONNECT_RETRIES", "3")),
         "read_retries": int(os.environ.get("LANCE_CLIENT_READ_RETRIES", "3")),
         "backoff_factor": float(os.environ.get("LANCE_CLIENT_RETRY_BACKOFF_FACTOR", "0.25")),
+        "backoff_jitter": float(os.environ.get("LANCE_CLIENT_RETRY_BACKOFF_JITTER", "0.25")),
         "statuses": [
             int(i.strip()) for i in os.environ.get(
                 "LANCE_CLIENT_RETRY_STATUSES", "429, 500, 502, 503"
@@ -210,6 +211,7 @@ def retry_adapter(options: Dict[str, Any]) -> HTTPAdapter:
     connect_retries = options["connect_retries"]
     read_retries = options["read_retries"]
     backoff_factor = options["backoff_factor"]
+    backoff_jitter = options["backoff_jitter"]
     statuses = options["statuses"]
     methods = frozenset(options["methods"])
     logging.debug(f"Setting up retry adapter with {total_retries} retries," + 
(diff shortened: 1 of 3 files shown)


python/lancedb/remote/client.py
```diff
@@ -170,3 +178,50 @@ def query(self, table_name: str, query: VectorQuery) -> VectorQueryResult:
...
+
+def retry_adapter_options(methods = ["GET"]) -> Dict[str, Any]:
+    return {
+        "retries": int(os.environ.get("LANCE_CLIENT_MAX_RETRIES", "3")),
+        "connect_retries": int(os.environ.get("LANCE_CLIENT_CONNECT_RETRIES", "3")),
+        "read_retries": int(os.environ.get("LANCE_CLIENT_READ_RETRIES", "3")),
+        "backoff_factor": float(os.environ.get("LANCE_CLIENT_RETRY_BACKOFF_FACTOR", "0.25")),
+        "statuses": [
+            int(i.strip()) for i in os.environ.get(
+                "LANCE_CLIENT_RETRY_STATUSES", "500, 502, 503"
```
chebbyChefNEQ, line 202: also 429

python/lancedb/remote/client.py
```diff
@@ -170,3 +178,50 @@ def query(self, table_name: str, query: VectorQuery) -> VectorQueryResult:
...
+    read_retries = options["read_retries"]
+    backoff_factor = options["backoff_factor"]
+    statuses = options["statuses"]
+    methods = frozenset(options["methods"])
+    logging.debug(f"Setting up retry adapter with {total_retries} retries," + 
+                  f"connect retries {connect_retries}, read retries {read_retries}," +
+                  f"backoff factor {backoff_factor}, statuses {statuses}, " +
+                  f"methods {methods}")
+
+    return HTTPAdapter(max_retries=Retry(
```
chebbyChefNEQ, line 220: is this exponential retry? Also, anyway we can add jitter to the back off?

chebbyChefNEQ (approved):

python/lancedb/remote/client.py
albertlockett: It is exponential:
```
sleep = backoff_factor * (2 ** retry_num)
```
So by default (backoff_factor = 0.25 * total retries = 3), we sleep for 0.5, 1, and 2 seconds

I will add jitter as well

PR #923, opened 2024-02-03
chore(python): Change all instances of safe_import to attempt_import 

AyushExel:
safe_import is not safe, it raises an exception. Would be better to rename it to something different as similar functions safe_import_pandas and safe_import_polars don't raise


changhiskhan (approved):
Maybe just call it "import_or_msg" or smth?

AyushExel:
kk let's use import_or_raise..let's do the reranker PR before this as it adds a couple more instances of safe_import.

PR #942, opened 2024-02-07
feat: make it easier to create empty tables

westonpace:
This PR also reworks the table creation utilities significantly so that they are more consistent, built on top of each other, and thoroughly documented.

### DIFF
--- a/node/src/arrow.ts
+++ b/node/src/arrow.ts
@@ -13,10 +13,7 @@
 // limitations under the License.
 
 import {
-  Int64,
   Field,
-  type FixedSizeListBuilder,
-  Float32,
   makeBuilder,
   RecordBatchFileWriter,
   Utf8,
@@ -27,15 +24,19 @@ import {
   Table as ArrowTable,
   RecordBatchStreamWriter,
   List,
-  Float64,
   RecordBatch,
   makeData,
   Struct,
   type Float,
-  type DataType
+  DataType,
+  Binary,
+  Float32
 } from 'apache-arrow'
 import { type EmbeddingFunction } from './index'
 
+/*
+ * Options to control how a column should be converted to a vector array
+ */
 export class VectorColumnOptions {
   /** Vector column type. */
   type: Float = new Float32()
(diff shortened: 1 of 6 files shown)


node/src/arrow.ts
```diff
@@ -47,14 +48,50 @@ export class VectorColumnOptions {
...
+   * vector column.
+   */
   vectorColumns: Record<string, VectorColumnOptions> = {
     vector: new VectorColumnOptions()
   }
 
+  /**
+   * If true then string columns will be encoded with dictionary encoding
+   *
+   * Set this to true if your string columns tend to repeat the same values
```
eddyxu, line 87: Do we need to specify which column to apply dictionary encoding, in case there are multiple string columns?

For example, for a LLM training dataset, there will be prompt (which can not be encoded as dictionary), and tags / filter (which might be encoded as such).

node/src/arrow.ts
```diff
@@ -136,214 +196,304 @@ export function makeArrowTable (
...
-          throw new Error(`Invalid vector size, expected ${vectorSize}`)
-        }
+// Helper function to convert an Array of JS values to an Arrow Vector
+function makeVector (values: any[], type?: DataType, stringAsDictionary?: boolean): Vector<any> {
+  if (type !== undefined) {
+    // No need for inference, let Arrow create it
+    return vectorFromArray(values, type)
+  }
+  if (values.length === 0) {
+    throw TypeError('makeVector requires at least one value or the type must be specfied')
```
eddyxu, line 306: I am not familiar with js, but are there other errors? this supposes to be a `ValueError` in python, which is not TypeError

node/src/arrow.ts
```diff
@@ -136,214 +196,304 @@ export function makeArrowTable (
...
+  }
+  const sampleValue = values.find(val => val !== null && val !== undefined)
```
eddyxu, line 308: This is fancy :)

node/src/arrow.ts
```diff
@@ -136,214 +196,304 @@ export function makeArrowTable (
...
+  if (sampleValue === undefined) {
+    throw TypeError('makeVector cannot infer the type if all values are null or undefined')
+  }
+  if (Array.isArray(sampleValue)) {
+    // Default Arrow inference doesn't handle list types
+    return makeListVector(values)
```
eddyxu, line 314: Where do we handle list vs fixed size list?

node/src/arrow.ts
westonpace: If users care about that level of detail then they can always supply a schema.  `arrow-js` *always* uses dictionary encoding.  The old vectordb *never* used dictionary encoding.  I figured it might be nice to have this as an option for users that want compatibility with `arrow-js` but I can just remove this flag too.

node/src/arrow.ts
westonpace: It's a smaller set in node: `Error, AssertionError, RangeError, ReferenceError, SyntaxError, SystemError, TypeError`.  I'll change this to just be `Error`.

node/src/arrow.ts
westonpace: If the user wants fixed size list then either they supply a schema or they use `vectorColumns`.  Otherwise we will always interpret an `Array<Array<...>>`as a variable sized list.

eddyxu: in our user guide, the default behavior is

```
data = [
  "vec": [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]],
]
```
will it be interpreted as List<f32> instead of `FixedSizeList<f32>`?

westonpace: The examples I see in the user guide are all:

```
data = [
  { "vector": [1.0, 2.0, 3.0] },
  { "vector": [4.0, 5.0, 6.0] }
]
```

As long as the column name is `vector` then we are ok.  The default value of `vectorColumns` is `["vector"]` and so any column named `vector` will be made into a fixed size list.

Also, this change should not be changing any existing behavior so as long as those examples were working before they should still be working after this change.

eddyxu (approved):

PR #962, opened 2024-02-12
fix: wrap in BigInt to avoid upstream bug

### LINKED ISSUE #960: Feature: cast numbers to BigInt if type is Int64
### SDK

Node

### Description

We need to workaround this bug https://github.com/apache/arrow/issues/40051

Can handle it here: https://github.com/lancedb/lancedb/blob/7e50c239eb0e311393dd82e516d428d9c2bd2ff4/node/src/arrow.ts#L148

wjones127:
Closes #960

### DIFF
--- a/node/src/arrow.ts
+++ b/node/src/arrow.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import {
+  Int64,
   Field,
   type FixedSizeListBuilder,
   Float32,
@@ -30,7 +31,8 @@ import {
   RecordBatch,
   makeData,
   Struct,
-  type Float
+  type Float,
+  type DataType
 } from 'apache-arrow'
 import { type EmbeddingFunction } from './index'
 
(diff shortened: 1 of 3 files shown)


node/src/arrow.ts
```diff
@@ -142,15 +144,18 @@ export function makeArrowTable (
...
       // Explicit schema is provided, highest priority
-      vector = vectorFromArray(
-        values,
-        opt.schema?.fields.filter((f) => f.name === colName)[0]?.type
-      )
+      const fieldType: DataType | undefined = opt.schema.fields.filter((f) => f.name === colName)[0]?.type as DataType
+      if (fieldType instanceof Int64) {
+        // wrap in BigInt to avoid bug: https://github.com/apache/arrow/issues/40051
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+        values = values.map((v) => BigInt(v)) as any[]
```
westonpace, line 156: This is a minor nit, don't stress much over it.

Why is the `as any` needed here?  I would think `values` is already `any[]` and so you should be able to assign to it.  If it isn't, then maybe it would be more clear to change the above line?

```
let values = data.map((datum) => datum[colName]) as any[]
```

westonpace (approved):
This fix would need to go in `convertToTable` too but I've rewritten all of this in https://github.com/lancedb/lancedb/pull/942 so I can just take care of it when I rebase (care to review?)

node/src/arrow.ts
wjones127: I can try this. I really didn't understand the errors eslint was giving me...

wjones127: Okay turns out it wasn't necessary in either spot after all. 🤷

PR #966, opened 2024-02-13
[docs]: Fix typos and clarity in hybrid search docs

prrao87:
- Fixed typos and added some clarity to the hybrid search docs
- Changed "Airbnb" case to be as per the [official company name](https://en.wikipedia.org/wiki/Airbnb) (the "bnb" shouldn't be capitalized", and the text in the document aligns with this
- Fixed headers in nav bar


AyushExel (approved):
thnx

PR #971, opened 2024-02-14
ci: fix Node ARM release build

wjones127:
When we turned on fat LTO builds, we made the release build job **much** more compute and memory intensive. The ARM runners have particularly low memory per core, which makes them susceptible to OOM errors. To avoid issues, I have enabled memory swap on ARM and bumped the side of the runner.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -98,16 +98,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


eddyxu (approved):

wjones127:
I have confirmed the release job will now work based on this test run here: https://github.com/lancedb/lancedb/actions/runs/7905694879/job/21578881645

PR #977, opened 2024-02-15
docs: Minimal reranking evaluation benchmarks

AyushExel:
These are some minimal evaluation benchmarks. The Reranker docs need another iteration of overhaul which should put the latency and evaluation (hit-rate) metrics in the same page. But this is needed for a talk later today

### DIFF
--- a/docs/src/hybrid_search/eval.md
+++ b/docs/src/hybrid_search/eval.md
@@ -25,8 +25,9 @@ Vector Search baseline - `0.64`
 | Cohere | `0.81` | `0.81` | `0.85` |
 | ColBERT | `0.68` | `0.68` | `0.73` |
 
-<img width="1112" alt="Screenshot 2024-02-15 at 9 37 01 PM" src="https://github.com/AyushExel/assets/assets/15766192/d57b1780-ef27-414c-a5c3-73bee7808a45">
-
+<p>
+<img src="https://github.com/AyushExel/assets/assets/15766192/d57b1780-ef27-414c-a5c3-73bee7808a45">
+</p>
 
 <b> With OpenAI embedding-v3-small </b>
 
@@ -39,7 +40,9 @@ Vector Search baseline - `0.59`
 | Cohere | `0.79` | `0.79` | `0.84` |
 | ColBERT | `0.70` | `0.70` | `0.76` |
 
-<img width="1109" alt="Screenshot 2024-02-15 at 9 34 17 PM" src="https://github.com/AyushExel/assets/assets/15766192/259adfd2-6ec6-4df6-a77d-1456598970dd">
+<p>
+<img src="https://github.com/AyushExel/assets/assets/15766192/259adfd2-6ec6-4df6-a77d-1456598970dd">
+</p>
 
 ### Conclusion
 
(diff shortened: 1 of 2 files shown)


docs/src/hybrid_search/eval.md
```diff
@@ -0,0 +1,49 @@
...
+
+## The challenge of (re)ranking search results
+Once you have a group of the most relevant search results from multiple search sources, you'd likely standardize the score and rank them accordingly. This process can also be seen as another independent step - reranking.
+There are two approaches for reranking search results from multiple sources.
+* <b>Score-based</b>: Calculate final relevance scores based on a weighted linear combination of individual search algorithm scores. Example - Weighted linear combination of semantic search & keyword-based search results.
+* <b>Relevance-based</b>: Discards the existing scores and calculates the relevance of each search result - query pair. Example - Cross Encoder models
+
+Even though there are many strategies for reranking search results, none works for all cases. Moreover, evaluating them itself is a challenge. Also, reranking can be dataset, application specific so it's hard to generalize.
```
prrao87, line 11:
```suggestion
Even though there are many strategies for reranking search results, no single method works well for all cases. Evaluating the methods is itself a challenge. Additionally, reranking can be dataset or application-specific, so it's hard to devise a generalized strategy.
```

docs/src/hybrid_search/eval.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Example evaluation of hybrid search with Reranking
+
+Here's some evaluation numbers from experiment comparing these re-rankers on about 800 queries. It is modified version of an evaluation script from [llama-index](https://github.com/run-llama/finetune-embedding/blob/main/evaluate.ipynb) that measures hit-rate at top-k.
```
prrao87, line 15:
```suggestion
Below are some evaluation numbers from experiments comparing these rerankers on about 800 queries. It is a modified version of an evaluation script from [llama-index](https://github.com/run-llama/finetune-embedding/blob/main/evaluate.ipynb) that measures hit-rate at top-k.
```

docs/src/hybrid_search/eval.md
```diff
@@ -0,0 +1,49 @@
...
+| Cohere | `0.79` | `0.79` | `0.84` |
+| ColBERT | `0.70` | `0.70` | `0.76` |
+
+<p>
+<img src="https://github.com/AyushExel/assets/assets/15766192/259adfd2-6ec6-4df6-a77d-1456598970dd">
+</p>
+
+### Conclusion
+
+The results show that the reranking methods are able to improve the search results. However, the improvement is not consistent across all rerankers. The choice of reranker depends on the dataset and the application. It is also important to note that the reranking methods are not a replacement for the search methods. They are complementary and should be used together to get the best results. The speed to recall tradeoff is also an important factor to consider when choosing the reranker.
```
prrao87, line 49:
```suggestion
The results clearly show that the reranking methods are able to improve the search results from the baseline. However, no one particular reranker performs the best in all cases. The choice of reranker depends on the dataset and the application. It is also important to note that the reranking methods are **not** a replacement for the search methods. They are complementary, and should be used downstream of the search stage to get the best results. The latency-to-recall tradeoff is also an important factor to consider when choosing the reranker.
```

prrao87 (approved):
LGTM, just some clarity fixes from my end.

I'm pleasantly surprised that in this case, `LinearReranker` did _that_ well and is on par with Cohere. Perhaps this is specific to OpenAI embeddings? I'd be surprised if SentenceTransformers models had this sort of performance. Worth studying in more detail for a future experiment, I guess!

PR #978, opened 2024-02-15
Revert "docs: Minimal reranking evaluation benchmarks"

prrao87:
Reverts lancedb/lancedb#977 to add suggested fixes.


AyushExel:
will make another one with more details

Issue #769, opened 2024-01-04
Feature: Option to turn off search syntax in FTS


pchalasani:
### SDK

Python

### Description

Currently when doing 
```
tbl.search(query).to_pandas()
```
or similar, if `query` happens to contain a string that looks like a search syntax and the syntax is wrong, we get an error. E.g. I could be trying to match a query from the user that just happens to look like `"they could have been dogs OR cats"`, and this causes a syntax error. If we change "OR" to lower-case "or", there's no error. 

Maybe this search syntax error can be turned off entirely or an option given to disable syntax checking.

The FTS looks like it's serving a dual purpose: (a) deliberate search queries composed carefully adhering to syntax, and (b) blindly using  a text fragment to get match relevant docs.
In case of (a) the syntax check makes sense, but not in case (b).
We need an option to distinguish (a) from (b) or just default to support (b).

changhiskhan:
Yeah I think there should be query options. Also have you tried quoting the input?

pchalasani:
Haven't but my current workaround is to lower-case all OR AND NOT, and then it's not interpreted as a search syntax.

changhiskhan:
Lmk what you think of the solution presented in #788. Basically if you quote the input, e.g., `table.search('"they could have been dogs OR cats"')`, it should convert it into a phrase query so it skips the syntax checking.

pchalasani:
@changhiskhan do we have to worry about what happens if there's a single quote in the string already, e.g. if the string is `"the cats OR dogs were not really 'pets' at all"` -- will escaping be done automatically? If this is not an issue then this is a good fix!

changhiskhan:
let me add an additional test case

changhiskhan:
nested quoting should work fine (see the new unit tests in that PR).

pchalasani:
Another search syntax issue arose, which can't be fixed via quoting, please see #797

PR #982, opened 2024-02-17
doc(python): document the  method in fts

changhiskhan:


prrao87 (changes requested):
@changhiskhan the description as it exists is very confusing, so I rephrased it as per your comments and discussions in #797 and #769. The "Syntax" subsection is renamed to "Phrase queries vs. structured queries" for clarity.

I feel like this is easier to follow now, so it would be great if you and @wjones127 could carefully give that section a read and see if my understanding is correct.

I still think there are a few parts that aren't very clear when it comes to syntax, and looking through the code didn't help me much.

- How exactly is the syntax `"they could have been dogs OR cats"` being marked as invalid by LanceDB? Is it simply checking for the presence of logical operators like `AND` or symbolic operators like `-` in every query that's enclosed in double quotes?
- If the above is true, then how is a query like `"(Old AND Man) AND Sea"` marked as valid? How does the parser know what constitutes a phrase, and what constitutes a keyword?
- The `phrase_query` argument is true by default in the query builder. By this logic, does this mean that LanceDB will _always_ try to parse the query first as a phrase query under all conditions? If so, how does the user enforce structured query execution if they want it?

There also aren't any [tests](https://github.com/lancedb/lancedb/blob/e0277383a54c798a7e3597473a9cee94fda8104b/python/tests/test_fts.py#L167) that check success/failure of structured queries. Could some of those be added in as well?

prrao87:
Ah, I see the reason for part of my confusion: [This docstring](https://github.com/lancedb/lancedb/blob/e0277383a54c798a7e3597473a9cee94fda8104b/python/lancedb/query.py#L540) says True, but the code above it says False. That needs to be fixed.

prrao87:
Ok so I now see that the [code](https://github.com/lancedb/lancedb/blob/e0277383a54c798a7e3597473a9cee94fda8104b/python/lancedb/query.py#L533) shows that phrase query is False by default. But some tests and examples on structured queries would also be nice :)

changhiskhan:
> @changhiskhan the syntax description as it exists is very confusing, so I rephrased it as per your comments and discussions in #797 and #769. The "Syntax" subsection is renamed to "Phrase queries vs. structured queries" for clarity.
> 
> I feel like my modified verrsion is easier to follow (at least in my head), so it would be great if you and @wjones127 could carefully give that section a read and see if my understanding is correct.
> 
> I still think there are a few parts that aren't very clear when it comes to syntax, and looking through the code didn't help me much.
> 
> * How exactly is the syntax `"they could have been dogs OR cats"` being marked as invalid by LanceDB? Is it simply checking for the presence of logical operators like `AND` or symbolic operators like `-` in every query that's enclosed in double quotes?
> * If the above is true, then how is a query like `"(Old AND Man) AND Sea"` marked as valid? How does the parser know what constitutes a phrase, and what constitutes a keyword?
> * The `phrase_query` argument is true by default in the query builder. By this logic, does this mean that LanceDB will _always_ try to parse the query first as a phrase query under all conditions? If so, how does the user enforce structured query execution if they want it?

The parsing is all controlled by tantivy. I was going to propose just linking to tantivy docs but it's also lacking in detail:

"""
// QueryParser may fail if the query is not in the right
// format. For user facing applications, this can be a problem.
// A ticket has been opened regarding this problem.
"""

Maybe we could link to Lucene documentation? https://lucene.apache.org/core/2_9_4/queryparsersyntax.html

I'm not sure whether Tantivy follows this exactly (I implicitly assume it does but it's not explicitly stated anywhere in their docs)

changhiskhan:
> Ok so I now see that the [code](https://github.com/lancedb/lancedb/blob/e0277383a54c798a7e3597473a9cee94fda8104b/python/lancedb/query.py#L533) shows that phrase query is False by default. But some tests and examples on structured queries would also be nice :)

i added a couple of examples

docs/src/fts.md
```diff
@@ -75,21 +75,41 @@ applied on top of the full text search results. This can be invoked via the fami
...
 ```
 
-## Syntax
+## Phrase queries vs. terms queries
 
-For full-text search you can perform either a phrase query like "the old man and the sea", 
-or a structured search query like "(Old AND Man) AND Sea".
-Double quotes are used to disambiguate.
+For full-text search you can specify either a **phrase** query like `"the old man and the sea"`, 
+or a **terms** search query like `"(Old AND Man) AND Sea"`. For more details on the structured
```
prrao87, line 81:
```suggestion
or a **terms** search query like `"(Old AND Man) AND Sea"`. For more details on the terms
```

python/lancedb/query.py
```diff
@@ -106,8 +106,8 @@ class Query(pydantic.BaseModel):
 
 
 class LanceQueryBuilder(ABC):
-    """Build LanceDB query based on specific query type:
-    vector or full text search.
+    """An abstract query builder. Subclasses are defined for vector search,
+    full text search, hybrid, and plain sql filtering.
```
prrao87, line 110:
```suggestion
    full text search, hybrid, and plain SQL filtering.
```

prrao87 (changes requested):
Makes sense now, thanks! Just two small nits to update.

docs/src/fts.md
```diff
@@ -75,21 +75,41 @@ applied on top of the full text search results. This can be invoked via the fami
...
+query syntax, see Tantivy's [query parser rules](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html).
 
-For example: 
+!!! tip "Note"
+    As a protection against unintended use, LanceDB will raise a syntax error if your query contains
+    recognized logical or symbolic operators. The following raises a syntax error since `OR` is a recognized operator.
```
changhiskhan, line 86:
```suggestion
!!! tip "Example"
    The query parser will raise an exception on `they could have been dogs OR cats` since the operator `OR` is all caps. However, `they could have been dogs or cats` works fine, and `"they could have been dogs" OR cats` also works (though the semantic meaning of what you're searching for is different).
```

docs/src/fts.md
```diff
@@ -75,21 +75,40 @@ applied on top of the full text search results. This can be invoked via the fami
...
+
+It can be cumbersome to have to remember what will cause a syntax error depending on the type of
+query you want to perform. To make this simpler, when you want to perform a phrase query, you can
+enforce it in one of two ways:
+
+1. Place the double-quoted query inside single quotes. For example, `table.search('"they could have been dogs OR cats"')` is treated as
+a phrase query.
+2. Explicitly declare the `phrase_query()` method. This is useful when you have a phrase query that
+itself contains double quotes. For example, `table.search('the cats OR dogs were not really "pets" at all').phrase_query()`
+is treated as a phrase query.
```
changhiskhan, line 108:
```suggestion
is treated as a phrase query. This is done by wrapping the whole input in double quotes `"` and replacing any nested double quotes with a single quote `'`.
```

docs/src/fts.md
```diff
@@ -75,21 +75,40 @@ applied on top of the full text search results. This can be invoked via the fami
...
+
+It can be cumbersome to have to remember what will cause a syntax error depending on the type of
+query you want to perform. To make this simpler, when you want to perform a phrase query, you can
+enforce it in one of two ways:
+
+1. Place the double-quoted query inside single quotes. For example, `table.search('"they could have been dogs OR cats"')` is treated as
+a phrase query.
+2. Explicitly declare the `phrase_query()` method. This is useful when you have a phrase query that
+itself contains double quotes. For example, `table.search('the cats OR dogs were not really "pets" at all').phrase_query()`
+is treated as a phrase query. This is done by wrapping the whole input in double quotes `"` and replacing any nested double quotes with a single quote `'`.
```
changhiskhan, line 108:
```suggestion
is treated as a phrase query.
```

docs/src/fts.md
```diff
@@ -75,21 +75,40 @@ applied on top of the full text search results. This can be invoked via the fami
...
-For full-text search you can perform either a phrase query like "the old man and the sea", 
-or a structured search query like "(Old AND Man) AND Sea".
-Double quotes are used to disambiguate.
+For full-text search you can specify either a **phrase** query like `"the old man and the sea"`, 
+or a **terms** search query like `"(Old AND Man) AND Sea"`. For more details on the terms
+query syntax, see Tantivy's [query parser rules](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html).
 
-For example: 
+!!! tip "Example"
+    The query parser will raise an exception on `they could have been dogs OR cats` since the operator `OR` is all caps. However, `they could have been dogs or cats` works fine, and `"they could have been dogs" OR cats` also works (though the semantic meaning of what you're searching for is different).
```
changhiskhan, line 85:
```suggestion
!!! tip "Note"
    The query parser will raise an exception on queries that is ambiguous. For example, in the query `they could have been dogs OR cats`, `OR` is all caps so it is considered a keyword. But it's ambiguous how the left part should be treated. So if you search for that as is, you'll get `Syntax Error: they could have been dogs OR cats`
```

changhiskhan:
> Makes sense now, thanks! Just two small nits to update.

done. Thanks for the review!

changhiskhan:
@prrao87 lmk if everything is good now. thanks!

docs/src/fts.md
```diff
@@ -75,21 +75,40 @@ applied on top of the full text search results. This can be invoked via the fami
...
-For full-text search you can perform either a phrase query like "the old man and the sea", 
-or a structured search query like "(Old AND Man) AND Sea".
-Double quotes are used to disambiguate.
+For full-text search you can specify either a **phrase** query like `"the old man and the sea"`, 
+or a **terms** search query like `"(Old AND Man) AND Sea"`. For more details on the terms
+query syntax, see Tantivy's [query parser rules](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html).
 
-For example: 
+!!! tip "Note"
+    The query parser will raise an exception on queries that is ambiguous. For example, in the query `they could have been dogs OR cats`, `OR` is all caps so it is considered a keyword. But it's ambiguous how the left part should be treated. So if you search for that as is, you'll get `Syntax Error: they could have been dogs OR cats`
```
prrao87, line 85:
```suggestion
    The query parser will raise an exception on queries that are ambiguous. For example, in the query `they could have been dogs OR cats`, `OR` is capitalized so it's considered a keyword query operator. But it's ambiguous how the left part should be treated. So if you submit this search query as is, you'll get `Syntax Error: they could have been dogs OR cats`.
```

prrao87 (changes requested):
Forgot to get around to this! One last change and it should be good to merge.

prrao87 (approved):
LGTM. I wasn't able to build the docs locally due to errors (presumably because the branches are out of sync with main). But the verbiage looks good.

I'll wait for some guidance from @AyushExel on how to fix the mkdocs build errors.

PR #983, opened 2024-02-17
doc: improve embedding functions documentation

changhiskhan:
Got some user feedback that the `implicit` / `explicit` distinction is confusing.
Instead I was thinking we would just deprecate the `with_embeddings` API and then organize working with embeddings into 3 buckets:

1. manually generate embeddings
2. use a provided embedding function
3. define your own custom embedding function


AyushExel (approved):
Sounds good!

PR #989, opened 2024-02-19
apply fixes for notebook

johko:
Hey, 
I tried the DisappearingEmbeddingFunction notebook and found some problems that I fixed locally and would love to merge, to improve the experience for others as well.

I also thought about loading the dataset from the Hugging Face Hub, rather than require the user to download it, that would keep the user more in the flow of the notebook IMO. Feel free to let me know what you guys think.


AyushExel (approved):
thanks

PR #993, opened 2024-02-19
refactor: rust vectordb API stabilization of the Connection trait

westonpace:
This is the start of a more comprehensive refactor and stabilization of the Rust API.  The `Connection` trait is cleaned up to not require `lance` and to match the `Connection` trait in other APIs.  In addition, the concrete implementation `Database` is hidden.

BREAKING CHANGE: The struct `crate::connection::Database` is now gone.  Several examples opened a connection using `Database::connect` or `Database::connect_with_params`.  Users should now use `vectordb::connect`.

BREAKING CHANGE: The `connect`, `create_table`, and `open_table` methods now all return a builder object.  This means that a call like `conn.open_table(..., opt1, opt2)` will now become `conn.open_table(...).opt1(opt1).opt2(opt2).execute()`  In addition, the structure of options has changed slightly.  However, no options capability has been removed.

### DIFF
--- a/rust/vectordb/src/connection.rs
+++ b/rust/vectordb/src/connection.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 //! LanceDB Database
-//!
 
 use std::fs::create_dir_all;
 use std::path::Path;
@@ -56,11 +55,13 @@ impl Default for CreateTableMode {
 /// Describes what happens when a vector either contains NaN or
 /// does not have enough values
 #[derive(Clone, Debug)]
-pub enum BadVectorHandling {
+enum BadVectorHandling {
     /// An error is returned
     Error,
+    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
     /// The offending row is droppped
     Drop,
+    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
     /// The invalid/missing items are replaced by fill_value
     Fill(f32),
 }
@@ -85,9 +86,13 @@ pub struct OpenTableOptions {
     ///
     /// The default value is 256
     ///
-    /// The exact meaning will depend on the type of index:
+    /// The exact meaning of an "entry" will depend on the type of index:
     /// * IVF - there is one entry for each IVF partition
     /// * BTREE - there is one entry for the entire index
+    ///
+    /// This cache applies to the entire opened table, across all indices.
+    /// Setting this value higher will increase performance on larger datasets
+    /// at the expense of more RAM
     pub index_cache_size: u64,
     /// Advanced parameters that can be used to customize table reads
     ///
@@ -216,7 +221,6 @@ impl ConnectOptions {
     }
 
     /// [`AwsCredential`] to use when connecting to S3.
-    ///
     pub fn aws_creds(mut self, aws_creds: AwsCredential) -> Self {
         self.aws_creds = Some(aws_creds);
         self
--- a/rust/vectordb/src/connection.rs
+++ b/rust/vectordb/src/connection.rs
@@ -52,7 +52,7 @@ impl CreateTableMode {
     pub fn exist_ok(
         callback: impl FnOnce(OpenTableBuilder) -> OpenTableBuilder + Send + 'static,
     ) -> Self {
-        CreateTableMode::ExistOk(Box::new(callback))
+        Self::ExistOk(Box::new(callback))
     }
 }
 
@@ -101,7 +101,7 @@ impl CreateTableBuilder<true> {
     ) -> Self {
         Self {
             parent,
-            name: name,
+            name,
             data: Some(data),
             schema: None,
             mode: CreateTableMode::default(),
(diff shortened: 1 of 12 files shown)


westonpace:
Some questions for reviewers about preferred style:

 * Most operations take some kind of optional options object.  In `lance`, and previous iterations of `vectordb` we have used `Option<...>` or overloads (e.g. `connect` vs `connect_with_params`).  With the new APIs I am favoring a single method that always requires an options object that has a `Default` impl.  In other words:

```
// Yes
let table = conn.open_table("foo", OpenTableOptions::default());
// No
let table = conn.open_table("foo", None);
let table = conn.open_table("foo");
```

* Should options objects be all-pub structs?  This is simpler to start with but potentially less future-proof as things change.  For example:

```
// Current style (simple)
let opts = OpenTableOptions {
  index_cache_size: 64,
  ..Default::default()
};
// Builder style (more future proof)
let opts = OpenTableOptions::new().index_cache_size(64);
```

* Should options objects be passed in by value or reference?

Currently, I'm always having them passed by value (but be easily `Clone`d).  My thinking is that options eventually end up needing something that can't be freely cloned (e.g. a `Vec` / `Arc`) and taking by value can avoid an extra `Clone`.

* Use `&str/String` or `AsRef<str>/Into<String>`?

Most APIs are using `&str/String` right now.  Should we try and use `AsRef/Into` where easily done?

* "params" vs. "options"

There seems to be some places we use "params" and some places we use "options".  Should we consolidate on one word choice?

wjones127:
> Most operations take some kind of optional options object.

I'd prefer either a method with 1 options object or a fluent builder API ([like in object-store](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html)).

> Should options objects be all-pub structs?

I guess the future-proof-ness of builders is we can deprecate certain methods? Or adjust the behavior / defaults?

> Should options objects be passed in by value or reference?

Value is good. 

> Most APIs are using &str/String right now. Should we try and use AsRef/Into where easily done?

We can really only do the latter with builders, right?

I'm kind of thinking maybe we should make the API use builders, given the advantages.

westonpace:
> I guess the future-proof-ness of builders is we can deprecate certain methods? Or adjust the behavior / defaults?

Adjusting the behavior / defaults is what I was thinking.  For example, if we decide that `index_cache_size` is a tuple `(u32, u64)` of num_entries / num_bytes in the future then we can keep the existing builder method which would not just set the first part of the tuple.

> I'm kind of thinking maybe we should make the API use builders, given the advantages.

Switched to builders.

rust/vectordb/src/connection.rs
```diff
@@ -29,73 +29,288 @@ use snafu::prelude::*;
...
+    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
+    /// The invalid/missing items are replaced by fill_value
+    Fill(f32),
+}
+
+impl Default for BadVectorHandling {
+    fn default() -> Self {
+        Self::Error
+    }
+}
```
wjones127, line 83: TIL: you can derive `Default` for enums if the default value is a unit variant:
```suggestion
#[derive(Clone, Debug, Default)]
enum BadVectorHandling {
    /// An error is returned
    #[default]
    Error,
    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
    /// The offending row is droppped
    Drop,
    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
    /// The invalid/missing items are replaced by fill_value
    Fill(f32),
}
```

rust/vectordb/src/table.rs
```diff
@@ -85,6 +84,40 @@ pub struct OptimizeStats {
...
+    Append,
+    /// The existing table will be overwritten with the new data
+    Overwrite,
+}
+
+impl Default for AddDataMode {
+    fn default() -> Self {
+        Self::Append
+    }
+}
```
wjones127, line 111:
```suggestion
#[derive(Debug, Clone, Default)]
pub enum AddDataMode {
    /// Rows will be appended to the table (the default)
    #[default]
    Append,
    /// The existing table will be overwritten with the new data
    Overwrite,
}
```

rust/vectordb/src/table.rs
```diff
@@ -214,32 +264,44 @@ pub trait Table: std::fmt::Display + Send + Sync {
     ///
     /// ```no_run
     /// # use std::sync::Arc;
-    /// # use vectordb::connection::{Database, Connection};
```
wjones127, line 217: Don't have to do this in this PR, but: If we feel like merge_insert is sufficiently tested, we could use [the technique described here to shorten this and other examples](https://rust-unofficial.github.io/patterns/idioms/rustdoc-init.html).

wjones127 (approved):
Some minor suggestions, but otherwise looks good!

rust/vectordb/src/table.rs
westonpace: Oh man, I had been looking for something like this a while ago.  This is a very cool idea.  I will wait for a follow-up though.  I want to revisit the examples in a little more detail once the refactor is done (and I need to update the rst).

PR #1002, opened 2024-02-20
feat(node): add `read_consistency_interval` to Node and Rust

### LINKED ISSUE #998: Feature: add `read_consistency_interval` and document consistency behavior
### SDK

Node

### Description

Based on discussion in #812

* [x] Implement in Python https://github.com/lancedb/lancedb/pull/828
* [ ] Implement in Node 
* [ ] Write consistency documentation

wjones127:
This PR adds the same consistency semantics as was added in #828. It *does not* add the same lazy-loading of tables, since that breaks some existing tests.

This closes #998.

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -668,11 +668,13 @@ There are three possible settings for `read_consistency_interval`:
     table = db.open_table("my_table")
     ```
 
-    To manually check for updates, use `checkout_latest`:
+    By default, a `Table` will never check for updates from other writers. To manually check for updates you can use `checkout_latest`:
 
     ```python
     db = lancedb.connect("./.lancedb")
     table = db.open_table("my_table")
+    
+    # (Other writes happen to my_table from another process)
 
     # Check for updates
     table.checkout_latest()
@@ -687,10 +689,10 @@ There are three possible settings for `read_consistency_interval`:
     const table = await db.openTable("my_table");
     ```
 
-    For eventual consistency, use a positive value:
+    For eventual consistency, specify the update interval as seconds:
 
     ```javascript
-    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 5000 });
+    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 5 });
     const table = await db.openTable("my_table");
     ```
 
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -636,6 +636,68 @@ The `values` parameter is used to provide the new values for the columns as lite
 
     When rows are updated, they are moved out of the index. The row will still show up in ANN queries, but the query will not be as fast as it would be if the row was in the index. If you update a large proportion of rows, consider rebuilding the index afterwards.
 
+## Consistency
+
+In LanceDB OSS, users can set the `read_consistency_interval` parameter on connections to achieve different levels of read consistency. This parameter determines how frequently the database synchronizes with the underlying storage system to check for updates made by other processes. If another process updates a table, the database will not see the changes until the next synchronization.
+
+There are three possible settings for `read_consistency_interval`:
+
+1. **Unset (default)**: The database does not check for updates to tables made by other processes. This provides the best query performance, but means that clients may not see the most up-to-date data. This setting is suitable for applications where the data does not change during the lifetime of the table reference.
+2. **Zero seconds (Strong consistency)**: The database checks for updates on every read. This provides the strongest consistency guarantees, ensuring that all clients see the latest committed data. However, it has the most overhead. This setting is suitable when consistency matters more than having high QPS.
+3. **Custom interval (Eventual consistency)**: The database checks for updates at a custom interval, such as every 5 seconds. This provides eventual consistency, allowing for some lag between write and read operations. Performance wise, this is a middle ground between strong consistency and no consistency check. This setting is suitable for applications where immediate consistency is not critical, but clients should see updated data eventually.
+
+!!! tip "Consistency in LanceDB Cloud"
+
+    This is only tune-able in LanceDB OSS. In LanceDB Cloud, readers are always eventually consistent.
+
+=== "Python"
+    
+    To set strong consistency, use `timedelta(0)`:
+
+    ```python
+    from datetime import timedelta
+    db = lancedb.connect("./.lancedb",. read_consistency_interval=timedelta(0))
+    table = db.open_table("my_table")
+    ```
+
+    For eventual consistency, use a custom `timedelta`:
+
+    ```python
+    from datetime import timedelta
+    db = lancedb.connect("./.lancedb", read_consistency_interval=timedelta(seconds=5))
+    table = db.open_table("my_table")
+    ```
+
+    To manually check for updates, use `checkout_latest`:
+
+    ```python
+    db = lancedb.connect("./.lancedb")
+    table = db.open_table("my_table")
+
+    # Check for updates
+    table.checkout_latest()
+    ```
+
+=== "JavaScript/Typescript"
+
+    To set strong consistency, use `0`:
+
+    ```javascript
+    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 0 });
+    const table = await db.openTable("my_table");
+    ```
+
+    For eventual consistency, use a positive value:
+
+    ```javascript
+    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 5000 });
+    const table = await db.openTable("my_table");
+    ```
+
+<!-- Node doesn't yet support the version time travel: https://github.com/lancedb/lancedb/issues/1007 
+    Once it does, we can show manual consistency check for Node as well.
+-->
+
 ## What's next?
 
 Learn the best practices on creating an ANN index and getting the most out of it.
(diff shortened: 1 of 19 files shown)


rust/vectordb/src/table/dataset.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// This will track versions independently of the original reference and
+    /// will be tied to a different RwLock.
+    pub async fn duplicate(&self) -> Self {
+        let ds_ref = self.0.read().await;
+        Self(Arc::new(RwLock::new((*ds_ref).clone())))
+    }
+
+    /// Get an immutable reference to the dataset.
+    pub async fn get(&self) -> Result<DatasetReadGuard<'_>> {
+        self.ensure_up_to_date().await?;
```
wjones127, line 128: Right now, we will block on reloading the dataset if it is out-of-date. In the future, we could change this to be an asynchronous task that runs in the background, so it doesn't block reads. If we did that, I think it would make sense to switch the default to eventual consistency, as we would have eliminated most of the performance downside. IMO no / manual consistency can be a confusing default for a database.

rust/vectordb/src/table.rs
```diff
@@ -470,44 +482,35 @@ impl NativeTable {
...
-        let dataset = self.clone_inner_dataset();
-        let latest_version_id = dataset.latest_version_id().await?;
-        let dataset = if latest_version_id == dataset.version().version {
-            dataset
-        } else {
-            dataset.checkout_version(latest_version_id).await?
-        };
-
+        let mut dataset = self.dataset.duplicate().await;
+        dataset.as_latest(self.read_consistency_interval).await?;
```
westonpace, line 510: Maybe, instead of reusing `self.read_consistency_interval`, we could have this parameter accept `read_consistency_interval` as an argument?

I think users should generally only call `checkout_latest` if `self.read_consistency_interval` is `None` anyways.

As a perk, then we don't have to store it on this object.

rust/vectordb/src/table.rs
```diff
@@ -665,24 +676,27 @@ impl NativeTable {
...
-        let metrics = compact_files(&mut dataset, options, remap_options).await?;
-        self.reset_dataset(dataset);
+        let mut dataset_mut = self.dataset.get_mut().await?;
+        let metrics = compact_files(&mut dataset_mut, options, remap_options).await?;
         Ok(metrics)
     }
 
-    pub fn count_fragments(&self) -> usize {
-        self.dataset.lock().expect("lock poison").count_fragments()
+    // TODO: why are these individual methods and not some single "get_stats" method?
```
westonpace, line 684: +1, they also don't exist on the other APIs so they will need to be reconciled at some point.  Plus, they are only part of `NativeTable` and not `Table`.  I think these kind of get added piecemeal / ad-hoc as needed for internal use cases maybe?

rust/vectordb/src/table/dataset.rs
```diff
@@ -0,0 +1,234 @@
...
+/// A wrapper around a [Dataset] that provides lazy-loading and consistency checks.
+///
+/// This can be cloned cheaply. It supports concurrent reads or exclusive writes.
+#[derive(Debug, Clone)]
+pub struct DatasetConsistencyWrapper(Arc<RwLock<DatasetRef>>);
+
+/// A wrapper around a [Dataset] that provides consistency checks.
+///
+/// The dataset is lazily loaded, and starts off as None. On the first access,
+/// the dataset is loaded.
```
westonpace, line 35: This struct is not public (not even `pub(crate)`).  Don't document or use `//`.

rust/vectordb/src/table/dataset.rs
```diff
@@ -0,0 +1,234 @@
...
+        dataset: Dataset,
+        read_consistency_interval: Option<Duration>,
+        last_consistency_check: Option<time::Instant>,
+    },
+    /// In this mode, the dataset is a specific version. It cannot be mutated.
+    TimeTravel { dataset: Dataset, version: u64 },
+}
+
+impl DatasetRef {
+    /// Reload the dataset to the appropriate version.
```
westonpace, line 49:
```suggestion
    // Reload the dataset to the appropriate version.
```

rust/vectordb/src/table/dataset.rs
```diff
@@ -0,0 +1,234 @@
...
+        Self(Arc::new(RwLock::new(DatasetRef::TimeTravel {
+            dataset,
+            version,
+        })))
+    }
+
+    /// Create a independent copy of the dataset reference.
+    ///
+    /// This will track versions independently of the original reference and
+    /// will be tied to a different RwLock.
```
westonpace, line 120:
```suggestion
    /// Create an independent copy of self.
    ///
    /// Unlike Clone, this will track versions independently of the original wrapper and
    /// will be tied to a different RwLock.
```

rust/vectordb/src/table/dataset.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    /// Get a mutable reference to the dataset.
+    pub async fn get_mut(&self) -> Result<DatasetWriteGuard<'_>> {
+        self.ensure_up_to_date().await?;
+        Ok(DatasetWriteGuard {
+            guard: self.0.write().await,
+        })
+    }
+
+    /// Convert into a reference to the latest version of the dataset.
```
westonpace, line 142:
```suggestion
    /// Convert into a wrapper in latest version mode
```

westonpace (commented):
Just some minor documentation suggestions mostly.  Thanks for doing this.

docs/src/guides/tables.md
```diff
@@ -636,6 +636,68 @@ The `values` parameter is used to provide the new values for the columns as lite
...
+
+    For eventual consistency, use a custom `timedelta`:
+
+    ```python
+    from datetime import timedelta
+    db = lancedb.connect("./.lancedb", read_consistency_interval=timedelta(seconds=5))
+    table = db.open_table("my_table")
+    ```
+
+    To manually check for updates, use `checkout_latest`:
```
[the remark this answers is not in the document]
wjones127:
```suggestion
    By default, a `Table` will never check for updates from other writers. To manually check for updates you can use `checkout_latest`:
```

docs/src/guides/tables.md
```diff
@@ -636,6 +636,68 @@ The `values` parameter is used to provide the new values for the columns as lite
...
+
+    ```python
+    db = lancedb.connect("./.lancedb")
+    table = db.open_table("my_table")
+
+    # Check for updates
```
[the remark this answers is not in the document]
wjones127:
```suggestion
    table = db.open_table("my_table")
    
    # (Other writes happen to my_table from another process)

    # Check for updates
```

docs/src/guides/tables.md
```diff
@@ -636,6 +636,68 @@ The `values` parameter is used to provide the new values for the columns as lite
...
+
+    ```javascript
+    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 0 });
+    const table = await db.openTable("my_table");
+    ```
+
+    For eventual consistency, use a positive value:
+
+    ```javascript
+    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 5000 });
```
wjones127, line 693:
```suggestion
    const db = await lancedb.connect({ uri: "./.lancedb", readConsistencyInterval: 5 });
```

nodejs/src/lib.rs
```diff
@@ -21,11 +21,23 @@ mod iterator;
 mod query;
 mod table;
 
+// Because these map from a JS object, we use camelCase names.
```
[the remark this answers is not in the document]
wjones127: This was outdated. Turned out `napi` will map them to snake case for us (very nice!)

```suggestion
```

rust/vectordb/src/table.rs
```diff
@@ -116,6 +118,9 @@ pub struct AddDataOptions {
 /// A Table is a collection of strong typed Rows.
 ///
 /// The type of the each row is defined in Apache Arrow [Schema].
+///
+/// Because the table data is lazily loaded, almost all methods are async and
+/// fallible.
```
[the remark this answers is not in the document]
wjones127: Yes, we did.

rust/vectordb/src/table.rs
wjones127: There are two uses of `checkout_latest`:
1. To force an update check if `self.read_consistency_interval` is `None`
2. To switch from "time travel mode" into "latest mode".

The latter could happen when `read_consistency_interval` was not `None`. I like keeping it out of the parameters here because it is meant to be a connection-level parameter.

rust/vectordb/src/table/dataset.rs
wjones127: Are you worried that this will show up in the reference docs?

Or do you think it's confusing when private APIs have documentation?

This entire file is private. Anything labelled `pub` here is only exposed to the crate, as the module is `pub(crate)`. Would it be sufficient to state that up front at the top of the file?

westonpace: > Or do you think it's confusing when private APIs have documentation?

Mostly this, and confusing might not be the right word.  Maybe it's just my personal preference too.  I try to be more sparse when documenting things that are internal but I don't know if that's a good thing or not.

westonpace (approved):

rust/vectordb/src/table/dataset.rs
wjones127: Let's talk about this more later, and we can decide some norms we want. For now, I'll merge this as-is.

PR #1003, opened 2024-02-20
python(feat): Imagebind embedding fn support 

raghavdixit99:
Added imagebind fn support , steps to install mentioned in docstring. 
pytest slow checks done locally


AyushExel:
Tested the slow test locally?

raghavdixit99:
Yes , are there any changes to be done or can this be merged?

python/lancedb/embeddings/imagebind.py
```diff
@@ -0,0 +1,170 @@
...
+        Parameters
+        ----------
+        query : Union[str]
+            The query to embed. A query can be either text, image paths or audio paths.
+        """
+        query = self.sanitize_input(query)
+        if query[0].endswith((".mp3", ".wav", ".flac", ".ogg", ".aac")):
+            return [self.generate_audio_embeddings(query)]
+        elif query[0].endswith((".jpg", ".jpeg", ".png", ".gif", ".bmp")):
+            return [self.generate_image_embeddings(query)]
```
AyushExel, line 87: Because this logic gets repeated in the `compute source embeddings`, we should Abstract this instance vars like:
```
query = self.sanitize_input(query)
if query[0].endswith(self._audio_extensions)
```
and so on. This way if we add support for specific type it'll only needed updating once

AyushExel:
Just one comment, otherwise looks good. Thnx

AyushExel (approved):

PR #1012, opened 2024-02-22
refactor: rename the rust crate from vectordb to lancedb

westonpace:
This also renames the new experimental node package to lancedb.  The classic node package remains named vectordb.

The goal here is to avoid introducing piecemeal breaking changes to the vectordb crate.  Instead, once the new API is stabilized, we will officially release the lancedb crate and deprecate the vectordb crate.  The same pattern will eventually happen with the npm package vectordb.


eddyxu (approved):

prrao87:
Much awaited!!! Bring on the LanceDB brand 😈

PR #1013, opened 2024-02-23
feat(python):Better API for handling hybrid search queries 

AyushExel:
This isn't the cleanest way to do it but the other option of accepting vector and text directly in BaseQueryBuilder didn't make sense as it is only limited to hybrid search use case.

Usage:
- Works the same as before when using embedding API

Without embedding API:
```
   # automatically deduce the query type
    result = (
        table.search(vector=query_vector, text=query)
        .limit(30)
        .rerank(normalize="score")
        .to_arrow()
    )

    # wrong query type raises an error
    with pytest.raises(ValueError):
        table.search(vector=query_vector, text=query, query_type="vector").rerank(
            normalize="score"
        )

    # raise an error if only vector or text is provided
    with pytest.raises(ValueError):
        table.search(vector=query_vector).to_arrow()

    with pytest.raises(ValueError):
        table.search(text=query).to_arrow()
```


AyushExel:
Discord user convo - https://discord.com/channels/1030247538198061086/1030247538667827251/1210386944509485088

python/python/tests/test_rerankers.py
```diff
@@ -118,6 +118,32 @@ def test_linear_combination(tmp_path):
...
+            normalize="score"
+        )
+
+    # raise an error if only vector or text is provided
+    with pytest.raises(ValueError):
+        table.search(vector=query_vector).to_arrow()
+
+    with pytest.raises(ValueError):
+        table.search(text=query).to_arrow()
+
```
changhiskhan, line 146: If i'm just looking at this as a user, i wouldn't understand why that should fail unless i read the documentation pretty carefully.

changhiskhan (commented):
overall i'm wary of having `query`, `vector` and `text` parameters.

how about this. For hybrid search, if you have an embedding function setup, then you can do it like this:

```
table.search("user query", query_type="hybrid")
    .rerank(...)
    .limit(10)
    .to_arrow()
```

if you don't have an embedding function setup, you can do this:

```
    .vector(<vec>)
    .text(<str>)
    .rerank(...)
    .limit(10)
    .to_arrow()
```

if you have embedding function setup but you want to parse your own terms from the user query:

```
table.search(query="user question", query_type="hybrid")
    .text(<str>)
    .rerank(...)
    .limit(10)
    .to_arrow()
```

python/python/tests/test_rerankers.py
AyushExel: Makes sense I'll redo this PR

PR #1014, opened 2024-02-23
feat: add a basic async python client starting point

westonpace:
This changes `lancedb` from a "pure python" setuptools project to a maturin project and adds a rust lancedb dependency.

The async python client is extremely minimal (only `connect` and `Connection.table_names` are supported).  The purpose of this PR is to get the infrastructure in place for building out the rest of the async client.

Although this is not technically a breaking change (no APIs are changing) it is still a considerable change in the way the wheels are built because they now include the native shared library.


eddyxu (approved):
PR is too big to not approve :)

python/python/lancedb/_lancedb.pyi
```diff
@@ -0,0 +1,12 @@
...
+class Connection(object):
+    pass
+
+def connect(
+    uri: str,
+    api_key: Optional[str],
+    region: Optional[str],
+    host_override: Optional[str],
+    read_consistency_interval: Optional[float],
+) -> Connection: ...
```
wjones127, line 12: Doesn't this return a Future?

wjones127 (commented):
Have a few minor nits. Otherwise looks like a good foundation.

🤔 Since we are starting from scratch, it would be nice to enable `mypy` from the beginning. I might make a PR for this later this week.

.github/workflows/pypi-publish.yml
```diff
@@ -1,31 +1,92 @@
-name: PyPI Publish
+name: Build and upload python wheels
```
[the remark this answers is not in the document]
westonpace: I've reverted this change so the workflow will still be the same.  Do GA names matter for workflows?  Or jobs?  The job name has to change because its splitting one job into three.

python/python/lancedb/_lancedb.pyi
westonpace: I changed it to `async def` which should imply the future.  I'm on the fence about these having these pyi files at all and curious for your opinion.  These methods aren't exposed to the user so this is only for our own internal understanding.

wjones127: For internal only things, I don't think we have to write them. Sometimes we might publicly expose them though, if the Python wrapper is redundant.

If I add actual mypy checks to the repo, my opinion on this might change.

.github/workflows/pypi-publish.yml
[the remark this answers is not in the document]
wjones127: I think workflows. I'm mostly thinking of this left sidebar on https://github.com/lancedb/lancedb/actions

There some some dead names there, which I find annoying.

PR #1015, opened 2024-02-23
feat: `{add|alter|drop}_columns` APIs

wjones127:


node/src/index.ts
```diff
@@ -500,6 +503,59 @@ export interface Table<T = number[]> {
...
+   */
+  dropColumns(columnNames: string[]): Promise<void>
+}
+
+/**
+ * A definition of a column alteration. The alteration changes the column at
+ * `path` to have the new name `name`, to be nullable if `nullable` is true,
+ * and to have the data type `data_type`. At least one of `name`, `nullable`,
+ * or `data_type` must be provided.
+ */
```
westonpace, line 545: This talks about `name` but I don't see it in the interface below.  Should it be `rename`?  Also, it mentions `data_type` which I don't see.

nodejs/lancedb/native.d.ts
```diff
@@ -12,6 +12,32 @@ export const enum MetricType {
   Cosine = 1,
   Dot = 2
 }
+/**
+ *  A definition of a column alteration. The alteration changes the column at
+ * `path` to have the new name `name`, to be nullable if `nullable` is true,
+ * and to have the data type `data_type`. At least one of `name`, `nullable`,
+ * or `data_type` must be provided.
+ */
```
westonpace, line 20: Same comments as above

nodejs/lancedb/native.d.ts
```diff
@@ -12,6 +12,32 @@ export const enum MetricType {
...
+   * This must be distinct from the names of all other columns in the table.
+   */
+  rename?: string
+  /** Set the new nullability. Note that a nullable column cannot be made non-nullable. */
+  nullable?: boolean
+}
+export interface AddColumnsSql {
+  name: string
+  valueSql: string
+}
```
westonpace, line 40: Document this?

nodejs/src/table.rs
```diff
@@ -93,4 +96,100 @@ impl Table {
...
+    pub nullable: Option<bool>,
+}
+
+impl From<ColumnAlteration> for LanceColumnAlteration {
+    fn from(js: ColumnAlteration) -> Self {
+        let ColumnAlteration {
+            path,
+            rename,
+            nullable,
+        } = js;
```
westonpace, line 182: What is this line doing?

rust/lancedb/src/table.rs
```diff
@@ -376,6 +379,19 @@ pub trait Table: std::fmt::Display + Send + Sync {
...
     async fn optimize(&self, action: OptimizeAction) -> Result<OptimizeStats>;
+
+    /// Add new columns to the table, providing values to fill in.
+    async fn add_columns(
+        &self,
+        transforms: NewColumnTransform,
+        read_columns: Option<Vec<String>>,
+    ) -> Result<()>;
+
+    /// Change a column's data type, nullability, or name.
```
westonpace, line 390: Can't change the data type yet I think

python/lancedb/table.py
```diff
@@ -660,6 +660,56 @@ def compact_files(self, *args, **kwargs):
...
+        Parameters
+        ----------
+        transforms: Dict[str, str]
+            A map of column name to a SQL expression to use to calculate the
+            value of the new column. These expressions will be evaluated for
+            each row in the table, and can reference existing columns.
+        """
+
+    @abstractmethod
+    def alter_columns(self, alterations: Iterable[Dict[str, str]]):
```
westonpace, line 679:
```suggestion
    def alter_columns(self, alterations: Iterable[Dict[str, Any]]):
```

That being said, why use a dictionary here instead of a `ColumnAlteration` data class?  That way it would be consistent with the `nodejs` lib

node/src/index.ts
```diff
@@ -1028,6 +1084,18 @@ export class LocalTable<T = number[]> implements Table<T> {
       return false
     }
   }
+
+  async addColumns (newColumnTransforms: Array<{ name: string, valueSql: string }>): Promise<void> {
```
westonpace, line 1088: Why an array of key/values instead of `Map<string, string>`?

nodejs/lancedb/native.d.ts
```diff
@@ -89,4 +115,7 @@ export class Table {
   delete(predicate: string): Promise<void>
   createIndex(): IndexBuilder
   query(): Query
+  addColumns(transforms: Array<AddColumnsSql>): Promise<void>
```
westonpace, line 118: Same question as the `node` package.  Why not use `Map<string, string>`?

westonpace (approved):
Suggestions, but no concerns.  Looks good.

python/lancedb/table.py
wjones127: In `nodejs`, the `ColumnAlteration` is just a typescript interface. So I think the invocations will look similar:

```python
table.alter_columns(
   { 'path': 'id', 'rename': 'new_id' },
   { 'path': 'price', 'nullable': true },
)
```
versus:
```js
await table.alterColumns([
  { path: 'id', rename: 'new_id' },
  { path: 'price', nullable: true }
])
```

Lei already expressed he didn't like passing in explicit options classes in Python. "Feels too Java-like", apparently.

Issue #884, opened 2024-01-28
Feature: drop non exist table should return meaningful error message


eddyxu:
### SDK

Rust

### Description

How to reproduce
```
db.create_table("my_table", data,...);
db.drop_table("MyTable").await.unwrap();
```

```
thread 'main' panicked at rust/vectordb/examples/simple.rs:50:36:
called `Result::unwrap()` on an `Err` value: Lance { message: "LanceError(IO): No such file or directory (os error 2), /Users/lei/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lance-io-0.9.10/src/local.rs:50:5" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

changhiskhan:
i think trying to open a non-existent table has the same issue

BubbleCal:
- [ ] map `ErrorKind::NotFound` to `Error;:NotFound`
- [ ] Handle the `Error::NotFound` and map it to `TableNotFound` while dropping a table not exists
- [ ] Upgrade the lance lib for lancedb

BubbleCal:
> i think trying to open a non-existent table has the same issue

I tried this and it works fine (the error is `TableNotFound`), will add tests for both of the cases

BubbleCal:
#1038 has added a test for this and it passed

PR #1022, opened 2024-02-26
chore(rust): report the TableNotFound error while dropping non-exist table

BubbleCal:
this will work after upgrading lance with https://github.com/lancedb/lance/pull/1995 merged
see #884 for details


BubbleCal:
I will enable the tests after upgrading lance, for now it's blocked by some changes

rust/lancedb/src/connection.rs
```diff
@@ -691,13 +699,49 @@ mod tests {
...
+    #[tokio::test]
+    #[ignore = "this can't pass due to https://github.com/lancedb/lancedb/issues/1019, enable it after the bug fixed"]
+    async fn test_open_table() {
+        let tmp_dir = tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let db = connect(uri).execute().await.unwrap();
+
+        assert_eq!(db.table_names().await.unwrap().len(), 0);
+
+        db.open_table("invalid_table").execute().await.unwrap();
```
westonpace, line 711: Won't this `unwrap` cause a panic since the table doesn't exist?  Should this line be removed?

westonpace (approved):
We should wait for the lance fixes to merge and the tests to pass before merging but the fix looks good.

rust/lancedb/src/connection.rs
BubbleCal: you are right, will fix this in the enabling PR

PR #1023, opened 2024-02-26
feat(python): Set up for fine tuning: basic QADataset and chat completion model

AyushExel:
Very WIP. I'm thinking of merging individual pieces in this feature branch instead of main so we can still review code in pieces and avoid polluting main.

* Adds support for creating corpus from llama-index text-node object (aim to remove this dependency)
* Adds very basic support for LLM api for chat completion, will expand as need arises.
* Add basic universal evaluator
* Add Sentence transformer finetuning support

Known problems:
- [ ] W&B experiment tracking is not working for sentence transformers


python/lancedb/embeddings/fine_tuner/dataset.py
```diff
@@ -0,0 +1,98 @@
...
+        queries (Dict[str, str]): Dict id -> query.
+        corpus (Dict[str, str]): Dict id -> string.
+        relevant_docs (Dict[str, List[str]]): Dict query id -> list of doc ids.
+
+    """
+
+    queries: Dict[str, str]  # dict id -> query
+    corpus: Dict[str, str]  # dict id -> string
+    relevant_docs: Dict[str, List[str]]  # query id -> list of doc ids
+    mode: str = "text"
```
AyushExel, line 41: Here `mode` is similar to llama-index QADataset which I think is meant to be used when supporting multi-modal/vision embedding model fine-tuning

python/lancedb/embeddings/fine_tuner/dataset.py
```diff
@@ -0,0 +1,129 @@
...
+
+        return QADataset(
+            queries=dict(zip(queries["id"], queries["query"])),
+            corpus=dict(zip(corpus["id"], corpus["text"])),
+            relevant_docs=dict(zip(relevant_docs["query_id"], relevant_docs["doc_id"])),
+        )
+    
+    # generate queries as a convenience function
+    @classmethod
+    def generate_qa_embedding_pairs(
```
changhiskhan, line 93: this is for generating synthetic labels?

changhiskhan (commented):
let's have this in a separate directory (i.e., a peer of the python subdir) or perhaps a different repo even for now so we can move quicker?

python/lancedb/embeddings/fine_tuner/dataset.py
AyushExel: yes

python/lancedb/embeddings/fine_tuner/dataset.py
```diff
@@ -0,0 +1,129 @@
...
+
+You are a Teacher/ Professor. Your task is to setup \
+{num_questions_per_chunk} questions for an upcoming \
+quiz/examination. The questions should be diverse in nature \
+across the document. Restrict the questions to the \
+context information provided."
+"""
+
+
+class QADataset(BaseModel):
```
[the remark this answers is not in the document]
AyushExel: Yes all datasets for fine-tuning Embedding models will have this structure

python/lancedb/embeddings/fine_tuner/dataset.py
AyushExel: yes..haven't thought deeply about it but I think they should also be multi-modal..where string would represent the path of images

python/lancedb/embeddings/fine_tuner/dataset.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @property
+    def query_docid_pairs(self) -> List[Tuple[str, List[str]]]:
+        """Get query, relevant doc ids."""
+        return [
+            (query, self.relevant_docs[query_id])
+            for query_id, query in self.queries.items()
+        ]
+
+    def save(self, path: str, mode: str = "overwrite") -> None:
```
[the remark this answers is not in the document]
AyushExel: Ummmm so the thought was that user can essentially keep appending/overwriting the query <> doc pairs, keeping the same corpus.. But now that I think about it, it's possible to do that with one lance dataset as well....

AyushExel:
> let's have this in a separate directory (i.e., a peer of the python subdir) or perhaps a different repo even for now so we can move quicker?

So the target branch for this is not main, which should be similar to having a separate repo...Or did you want to develop this as a plugin that lives outside the the main repo/

python/lancedb/embeddings/fine_tuner/dataset.py
```diff
@@ -0,0 +1,171 @@
...
+from .llm import Openai, BaseLLM
+
+DEFAULT_PROMPT_TMPL = """\
+Context information is below.
+
+---------------------
+{context_str}
+---------------------
+
+Given the context information and not prior knowledge.
```
tanaymeh, line 19:
```suggestion
Given the context information and no prior knowledge, \
```

python/lancedb/embeddings/base.py
```diff
@@ -159,3 +174,53 @@ def generate_embeddings(
         Generate the embeddings for the given texts
         """
         pass
+
+    def evaluate(self, dataset: QADataset, path=None, *args, **kwargs):
+        """
+        Evaluate the embedding function on a dataset. This calculates the hit-rate for the
+        top-k retrieved documents for each query in the dataset. Assumes that the first
```
tanaymeh, line 181: Maybe we could use a better metric like Precision @ K?

tanaymeh (changes requested):

python/lancedb/embeddings/fine_tuner/dataset.py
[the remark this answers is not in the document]
AyushExel: Actually, let's discuss this a bit, the 3 dicts might be independent (of different length).

python/lancedb/embeddings/base.py
AyushExel: So now it should be precision at k with default to 5

AyushExel:
This branch is fked after rebasing. Will open a new PR

Issue #926, opened 2024-02-05
LanceDB Async API epic


changhiskhan:

skyfallsin:
excited to see this in the queue, `asyncio` support in Python would go far!

changhiskhan:
Release plan:

TENTATIVE:

1. Monday 3/18 - `lancedb` rust crate should be released.
2. Friday 3/22 TypeScript - `lancedb` npm package to be released 
3. Friday 3/22 Python - `async` API to be released

PR #1024, opened 2024-02-26
feat: Initial remote table implementation for rust

westonpace:
This will eventually replace the remote table implementations in python and node.

### DIFF
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -46,7 +46,7 @@ impl RestfulLanceDbClient {
             headers.insert(
                 "Host",
                 HeaderValue::from_str(&host).map_err(|_| Error::Http {
-                    message: format!("non-ascii table name '{}' provided", db_name),
+                    message: format!("non-ascii database name '{}' provided", db_name),
                 })?,
             );
         }
@@ -54,7 +54,7 @@ impl RestfulLanceDbClient {
             headers.insert(
                 "x-lancedb-database",
                 HeaderValue::from_str(db_name).map_err(|_| Error::Http {
-                    message: format!("non-ascii table name '{}' provided", db_name),
+                    message: format!("non-ascii database name '{}' provided", db_name),
                 })?,
             );
         }
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -69,6 +69,7 @@ impl RestfulLanceDbClient {
         host_override: Option<String>,
     ) -> Result<Self> {
         let parsed_url = url::Url::parse(db_url)?;
+        debug_assert_eq!(parsed_url.scheme(), "db");
         if !parsed_url.has_host() {
             return Err(Error::Http {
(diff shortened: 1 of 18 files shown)


rust/lancedb/src/remote/db.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+#[async_trait]
+impl ConnectionInternal for RemoteDatabase {
+    async fn table_names(&self) -> Result<Vec<String>> {
+        let rsp = self
+            .client
+            .get("/v1/table/")
+            .query(&[("limit", 10)])
+            .query(&[("page_token", "")])
```
wjones127, line 53: I guess pagination is TODO?

rust/lancedb/src/remote/client.rs
```diff
@@ -0,0 +1,118 @@
...
+    client: reqwest::Client,
+    host: String,
+}
+
+impl RestfulLanceDbClient {
+    fn default_headers(
+        api_key: &str,
+        region: &str,
+        db_name: &str,
+        has_host_override: bool,
```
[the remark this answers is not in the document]
westonpace: If the user has overridden the host then we set the `x-lancedb-database` header.  I copied this from the existing python client.  I believe the rationale is that we normally get the db header from the URI.  However, when testing locally, we don't have DNS setup and we need to fallback to some alternate method.

It would probably be simpler, and I think it would be safe, if we just always set this header too.  CC @albertlockett @chebbyChefNEQ for second opinion.

rust/lancedb/src/remote/db.rs
westonpace: Yeah.  It wasn't part of the `Table` abstract class, so it wasn't documented at all and there was no implementation for the local table.  I figure I can add it back in, with support for both tables, at a later date.  I'll make a PR.

wjones127: I was just referring to the fact we only request the first page here (first 10).

westonpace: Yeah, that's pretty broken.  Still, I think I'd rather defer.  I've added this to https://github.com/lancedb/lancedb/issues/926 and I'm starting to track all the cleanup tasks as part of https://github.com/lancedb/lancedb/issues/926

rust/lancedb/src/remote/client.rs
```diff
@@ -0,0 +1,118 @@
...
+            HeaderValue::from_str(api_key).map_err(|_| Error::Http {
+                message: "non-ascii api key provided".to_string(),
+            })?,
+        );
+        if region == "local" {
+            let host = format!("{}.local.api.lancedb.com", db_name);
+            headers.insert(
+                "Host",
+                HeaderValue::from_str(&host).map_err(|_| Error::Http {
+                    message: format!("non-ascii table name '{}' provided", db_name),
```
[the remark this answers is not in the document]
wjones127: @westonpace you marked this as resolved, but looks like it's still "table name". Is it supposed to be "table name"?

wjones127 (approved):
Looks good, but have a question about the earlier comment I made.

PR #1025, opened 2024-02-26
feat: more accessible errors

wjones127:
The fact that we convert errors to strings makes them really hard to work with. For example, in SaaS we want to know whether the underlying `lance::Error` was the `InvalidInput` variant, so we can return a 400 instead of a 500.


chebbyChefNEQ (approved):

chebbyChefNEQ:
we should add `location!` here soon

PR #1034, opened 2024-02-28
chore: upgrade to lance 0.10.1

chebbyChefNEQ:
upgrade to lance 0.10.1 and update doc string to reflect dynamic projection options

### DIFF
--- a/rust/lancedb/src/query.rs
+++ b/rust/lancedb/src/query.rs
@@ -24,6 +24,13 @@ use crate::Error;
 
 const DEFAULT_TOP_K: usize = 10;
 
+#[derive(Debug, Clone)]
+pub enum Select {
+    All,
+    Simple(Vec<String>),
+    Projection(Vec<(String, String)>),
+}
+
 /// A builder for nearest neighbor queries for LanceDB.
 #[derive(Clone)]
 pub struct Query {
@@ -44,7 +51,7 @@ pub struct Query {
     /// Apply filter to the returned rows.
     filter: Option<String>,
     /// Select column projection.
-    select: Option<Vec<String>>,
+    select: Select,
 
     /// Default is true. Set to false to enforce a brute force search.
     use_index: bool,
@@ -70,7 +77,7 @@ impl Query {
             metric_type: None,
             use_index: true,
             filter: None,
-            select: None,
+            select: Select::All,
             prefilter: false,
         }
     }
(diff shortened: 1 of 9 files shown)


rust/lancedb/src/query.rs
```diff
@@ -71,6 +73,7 @@ impl Query {
             use_index: true,
             filter: None,
             select: None,
+            select_with_transform: None,
```
westonpace, line 76: Minor nit: if the word "transform" is already used elsewhere then maybe this ship has sailed but I'd prefer this to be called `select_with_projection` if possible.

westonpace, line 76: Given that only one of these can be set is it worth making this an enum?

```
enum Selection {
    All,
    Simple(Vec<String>),
    Projection(Vec<(String, String)>),
}
```

Though you'd have to move the error handling from `execute_stream` to the `select` / `select_with_transform` methods (e.g. raise an error in `select` if self.select != Select::All` and vice versa)

rust/lancedb/src/query.rs
```diff
@@ -294,6 +324,51 @@ mod tests {
...
+        let ds = DatasetConsistencyWrapper::new_latest(ds, None);
+
+        let query = Query::new(ds.clone()).nearest_to(&[0.1; 4]);
+        let result = query.limit(10).filter("id % 2 == 0").execute_stream().await;
+        let mut stream = result.expect("should have result");
+        // should only have one batch
+        while let Some(batch) = stream.next().await {
+            // post filter should have removed some rows
+            assert!(batch.expect("should be Ok").num_rows() < 10);
+        }
```
westonpace, line 341: Is this query / assertion related to `select_with_transform` at all?  Should we remove it or move it into its own test?

rust/lancedb/src/query.rs
```diff
@@ -294,6 +324,51 @@ mod tests {
...
+
+        let query = Query::new(ds).nearest_to(&[0.1; 4]);
```
westonpace, line 343: Any particular reason to do `nearest_to`/`limit` instead of a plain query?  This is just a nitpick to keep the tests as simple as possible.

westonpace (approved):
Approved, though I have some requests if possible.

chebbyChefNEQ:
I promise the tests passes locally. CI is failing on disk being full. Going to land

PR #1036, opened 2024-02-28
A simple base usage that install the dependencies necessary to use FT…

natcharacter:
A simple base usage that install the dependencies necessary to use FTS and Hybrid search

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,0 +1,27 @@
+#Simple base dockerfile that supports basic dependencies required to run lance with FTS and Hybrid Search
+#Usage docker build -t lancedb:latest -f Dockerfile .
+FROM python:3.10-slim-buster
+
+# Install Rust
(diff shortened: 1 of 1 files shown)


dockerfiles/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+  apt-get -y upgrade && \
+  apt-get -y install git
+
+
+# Verify installations
+RUN python --version && \
+  rustc --version && \
+  protoc --version
+
+RUN pip install tantivy@git+https://github.com/quickwit-oss/tantivy-py#164adc87e1a033117001cf70e38c82a53014d985 lancedb
```
changhiskhan, line 27:
```suggestion
RUN pip install tantivy lancedb
```
they've published the wheels so i think the bare pip install should work now. We'll fix this up in our docs

changhiskhan (approved):
thanks! just the one minor suggestion to simplify / speed up the tantivy install

PR #1037, opened 2024-02-28
feat: add support for add to async python API

westonpace:
In order to add support for `add` we needed to migrate the rust `Table` trait to a `Table` struct and `TableInternal` trait (similar to the way the connection is designed).

While doing this we also cleaned up some inconsistencies between the SDKs:

 * Python and Node are garbage collected languages and it can be difficult to trigger something to be freed.  The convention for these languages is to have some kind of close method.  I added a close method to both the table and connection which will drop the underlying rust object.
 * We made significant improvements to table creation in https://github.com/lancedb/lancedb/commit/cc5f2136a61234a9cdbe20e19c56459db8ca1051 for the `node` SDK.  I copied these changes to the `nodejs` SDK.
 * The nodejs tables were using fs to create tmp directories and these were not getting cleaned up.  This is mostly harmless but annoying and so I changed it up a bit to ensure we cleanup tmp directories.
 * ~~countRows in the node SDK was returning `bigint`.  I changed it to return `number`~~ (this actually happened in a previous PR)
 * Tables and connections now implement `std::fmt::Display` which is hooked into python's `__repr__`.  Node has no concept of a regular "to string" function and so I added a `display` method.
 * Python method signatures are changing so that optional parameters are always `Optional[foo] = None` instead of something like `foo = False`.  This is because we want those defaults to be in rust whenever possible (though we still need to mention the default in documentation).
 * I changed the python `AsyncConnection/AsyncTable` classes from abstract classes with a single implementation to just classes because we no longer have the remote implementation in python.

Note: this does NOT add the `add` function to the remote table.  This PR was already large enough, and the remote implementation is unique enough, that I am going to do all the remote stuff at a later date (we should have the structure in place and correct so there shouldn't be any refactor concerns)


westonpace:
For simplicity I've disabled `nodejs` lint in this PR.  It was reporting a bunch of violations on the arrow/embedding functions/openai code that was copied over from the `node` folder.  In the next PR I am reworking nodejs eslint/prettier formatting significantly and will re-enable it.

nodejs/src/connection.rs
```diff
@@ -95,10 +122,30 @@ impl Connection {
...
 
+    #[napi]
+    pub async fn create_empty_table(
+        &self,
+        name: String,
+        schema_buf: Buffer,
+        mode: String,
+    ) -> napi::Result<Table> {
+        let schema = ipc_file_to_schema(schema_buf.to_vec())
+            .map_err(|e| napi::Error::from_reason(format!("Failed to read IPC file: {}", e)))?;
```
wjones127, line 133: I think the user would be confused by the idea of an "IPC file" involved here. This is really about the schema, right?
```suggestion
            .map_err(|e| napi::Error::from_reason(format!("Failed to parse schema: {}", e)))?;
```

python/python/lancedb/db.py
```diff
@@ -427,12 +427,64 @@ def drop_database(self):
...
-    """An active LanceDB connection interface."""
+class AsyncConnection(object):
+    """An active LanceDB connection
+
+    To obtain a connection you can use the [connect] function.
+
+    This could be a native connection (using lance) or a remote connection (e.g. for
+    connecting to LanceDb Cloud)
+
+    Local connections do currently hold any open resources but they may do so in the
```
wjones127, line 438:
```suggestion
    Local connections do not currently hold any open resources but they may do so in the
```

python/python/lancedb/table.py
```diff
@@ -1784,9 +1793,23 @@ def _sanitize_nans(data, fill_value, on_bad_vectors, vec_arr, vector_column_name
     return data
 
 
-class AsyncTable(ABC):
+class AsyncTable(object):
```
wjones127, line 1796: If it's just object, I don't think the parent class is necessary.
```suggestion
class AsyncTable:
```

wjones127 (approved):
Looks good. A few minor comments.

nodejs/src/connection.rs
westonpace: Yes, we use IPC to marshal schemas to and from Rust (we should ideally come up with something based on the C API, I'll make a ticket) Good idea.  I went with something slightly different `Failed to marshal schema from JS to Rust` just because the most likely cause is an internal bug of some kind something and not an invalid schema.

PR #1039, opened 2024-02-29
fix(python): Few fts patches

AyushExel:
1. filtering with fts mutated the schema, which caused schema mistmatch problems with hybrid search as it combines fts and vector search tables.
2. fts with filter failed with `with_row_id`. This was because row_id was calculated before filtering which caused size mismatch on attaching it after. 
3. The fix for 1 meant that now row_id is attached before filtering but passing a filter to `to_lance` on a dataset that already contains `_rowid` raises a panic from lance. So temporarily, in case where fts is used with a filter AND `with_row_id`, we just force user to using the duckdb pathway.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -587,27 +587,25 @@ class LanceFtsQueryBuilder(LanceQueryBuilder):
         scores = pa.array(scores)
         output_tbl = self._table.to_lance().take(row_ids, columns=self._columns)
         output_tbl = output_tbl.append_column("score", scores)
-
-        if self._with_row_id:
-            # Need to set this to uint explicitly as vector results are in uint64
-            row_ids = pa.array(row_ids, type=pa.uint64())
-            output_tbl = output_tbl.append_column("_rowid", row_ids)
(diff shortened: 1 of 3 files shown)


python/python/lancedb/query.py
```diff
@@ -590,32 +590,44 @@ def to_arrow(self) -> pa.Table:
...
                 import duckdb
 
+                schema = output_tbl.schema
                 output_tbl = (
                     duckdb.sql("SELECT * FROM output_tbl")
                     .filter(self._where)
                     .to_arrow_table()
                 )
+                # output_tbl's fixed_size_list is now mutated to list
+                output_tbl = output_tbl.cast(schema)
```
AyushExel, line 612: I'm not sure if this is the safest way to cast schema for pyarrow. I'm not sure about the implications of passing `safe=False` 
(https://arrow.apache.org/docs/1.0/python/generated/pyarrow.Table.html#pyarrow.Table.cast)

AyushExel:
CI system running OO space, local tests passing

eddyxu (approved):

PR #1042, opened 2024-02-29
feat: reconfigure typescript linter / formatter for nodejs

westonpace:
The eslint rules specify some formatting requirements that are rather strict and conflict with vscode's default formatter.  I was unable to get auto-formatting to setup correctly.  Also, eslint has quite recently [given up on formatting](https://eslint.org/blog/2023/10/deprecating-formatting-rules/) and recommends using a 3rd party formatter.

This PR adds prettier as the formatter.  It restores the eslint rules to their defaults.  This does mean we now have the "no explicit any" check back on.  I know that rule is pedantic but it did help me catch a few corner cases in type testing that weren't covered in the current code.  Leaving in draft as this is dependent on other PRs.


albertlockett (approved):
YES GREAT

PR #1043, opened 2024-02-29
Fix default_embedding_functions.md

louisguitton:
typo and broken table


changhiskhan (approved):
thanks. the docs CI failure is due to recent refactor and has been fixed on main

PR #1047, opened 2024-03-01
[docs]: Fix issues with Rust code snippets in "quick start"

prrao87:
The renaming of `vectordb` to `lancedb` broke the [quick start docs](https://lancedb.github.io/lancedb/basic/#__tabbed_5_3) (it's pointing to a non-existent directory). This PR fixes the code snippets and the paths in the docs page.

Additionally, more fixes related to indexing docs below 👇🏽.

### DIFF
--- a/docs/src/ann_indexes.md
+++ b/docs/src/ann_indexes.md
@@ -7,20 +7,11 @@ for brute-force scanning of the entire vector space.
 A vector index is faster but less accurate than exhaustive search (kNN or flat search).
 LanceDB provides many parameters to fine-tune the index's size, the speed of queries, and the accuracy of results.
 
-Currently, LanceDB does _not_ automatically create the ANN index.
-LanceDB has optimized code for kNN as well. For many use-cases, datasets under 100K vectors won't require index creation at all.
-If you can live with <100ms latency, skipping index creation is a simpler workflow while guaranteeing 100% recall.
+## Disk-based Index
 
-In the future we will look to automatically create and configure the ANN index as data comes in.
-
-## Types of Index
-
-Lance can support multiple index types, the most widely used one is `IVF_PQ`.
-
-- `IVF_PQ`: use **Inverted File Index (IVF)** to first divide the dataset into `N` partitions,
-  and then use **Product Quantization** to compress vectors in each partition.
-- `DiskANN` (**Experimental**): organize the vector as a on-disk graph, where the vertices approximately
-  represent the nearest neighbors of each vector.
+Lance provides an `IVF_PQ` disk-based index. It uses **Inverted File Index (IVF)** to first divide
+the dataset into `N` partitions, and then applies **Product Quantization** to compress vectors in each partition.
+See the [indexing](concepts/index_ivfpq.md) concepts guide for more information on how this works.
 
 ## Creating an IVF_PQ Index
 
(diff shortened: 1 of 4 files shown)


prrao87:
Apologies for combining two tasks in this one PR, but I forgot to switch branches 😅. These second set of commits address the manual indexing requirements in the docs.

## Indexing docs

To clarify to users why we need to index manually, I'd word it carefully in three places:
- Quick start
- Concepts
- Guides


### Remedies
- In the quick start section, we add an admonition box stating why we need to index manually: first, that we don't always need an ANN index because LanceDB is 🔥 for datasets of <1M vectors, and second, because there's no one-size-fits-all when it comes to indexing (most other vector DB vendors have sane defaults, and HNSW is more forgiving, but that doesn't mean that those settings are optimal)
- In the "ANN indexes" section of the guides, we add a subsection titled "Why do I need to manually create an index?" clarifying the same things
- In the "Concepts" section covering the `IVF_PQ` index, I've removed the FAQ section (it's repetitive as it already appears twice in the FAQ and the Guides sections), but I link to the guides section as that's where the most detailed explanation on how/why we index sits.

All in all, I think this helps a new user gain a sequential understanding on how/why indexing is done in LanceDB. Because `IVF_PQ` is fundamentally different from HNSW, some manual tuning is desirable in most cases and it's best to let the user be aware of that through each of these sections. Because we cannot predict which section a user might chance upon when they first encounter LanceDB, it's best to drop these hints/explanations in all these sections.

@changhiskhan @AyushExel and @raghavdixit99, if you have more additions or suggestions on top of these, am happy to go through and offer my inputs. Hope this makes sense!

prrao87:
@AyushExel No idea about why CI is failing here, seems like a lot has changed since my last PR!

changhiskhan:
@prrao87 thanks! We've done some major refactoring so that we can support an Async API and a much better Rust API in general. I'm fixing the CI in a separate PR. Once that's merge I'll update this one so CI can pass.

docs/src/ann_indexes.md
```diff
@@ -187,13 +178,21 @@ You can select the columns returned by the query using a select clause.
...
-`LanceDB` has manually-tuned SIMD code for computing vector distances.
-In our benchmarks, computing 100K pairs of 1K dimension vectors takes **less than 20ms**.
-For small datasets (< 100K rows) or applications that can accept 100ms latency, vector indices are usually not necessary.
+`LanceDB` comes out-of-the-box with highly optimized SIMD code for computing vector similarity.
+In our benchmarks, computing distances for 100K pairs of 1K dimension vectors takes **less than 20ms**.
+We observe that for small datasets (~100K rows) or for applications that can accept 100ms latency,
+vector indices are usually not necessary.
 
-For large-scale or higher dimension vectors, it is beneficial to create vector index.
+For large-scale or higher dimension vectors, it can beneficial to create vector index for performance.
```
changhiskhan, line 195: Another reason to create an ANN index is if you need <10ms latency. In which case you'll also want to tune the index parameters to achieve optimal recall vs latency profile

changhiskhan (approved):
this lgtm. Just one small nit. Thanks!

cc @westonpace in case there's additional places where we need to update vectordb references to lancedb instead

Issue #805, opened 2024-01-10
Feature: add `to_batches` API


changhiskhan:
### SDK

Python

### Description

esp relevant for when there is no vector search query. If the dataset is large, the filtered result can also be larger than memory.

ighose:
Hi @changhiskhan, I would like to work on this. Could you please assign this to me?

It seems like we want to expose this API through an alternative to the function - to_arrow() -> pa.Table in lancedb/python/lancedb/table.py

PR #1048, opened 2024-03-01
feat: add to_batches API #805

ighose:
SDK
Python

Description
Exposes pyarrow batch api during query execution - relevant when there is no vector search query, dataset is large and the filtered result is larger than memory.


changhiskhan:
> SDK Python
> 
> Description Exposes pyarrow batch api during query execution - relevant when there is no vector search query, dataset is large and the filtered result is larger than memory.

Thanks! for lint, you'll want to run `ruff` in the python directory. Thanks!

ighose:
Thanks @changhiskhan, I've run `ruff`.

python/python/lancedb/query.py
```diff
@@ -267,6 +267,16 @@ def to_list(self) -> List[dict]:
         """
         return self.to_arrow().to_pylist()
 
+    def to_batches(self, max_chunksize) -> list:
```
wjones127, line 270: We should name the item type here, which is a `RecordBatch`.

Also, we will likely change this in the future to return an iterator, rather than a list. I think many users will want to read batch-by-batch without having to load all of the data into memory at once. Therefore, we should just say `Iterable` in the type.
```suggestion
    def to_batches(self, max_chunksize) -> Iterable[pa.RecordBatch]:
```

python/python/lancedb/query.py
```diff
@@ -267,6 +267,16 @@ def to_list(self) -> List[dict]:
...
+        """
+        Execute the query and return the results as a list of dictionaries.
+
+        Each list entry is a dictionary with the selected column names as keys,
+        or all table columns if `select` is not called. The vector and the "_distance"
+        fields are returned whether or not they're explicitly selected.
```
wjones127, line 276: I don't know why this says the list items are dictionaries. `self.to_arrow` returns a PyArrow [Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_batches), and the `to_batches` method returns a list of `RecordBatch`. Could you update this to be more accurate?

> The vector and the "_distance" fields are returned whether or not they're explicitly selected.

I think we plan on changing the behaviour of this soon. See https://discord.com/channels/1030247538198061086/1030247538667827251/1212101602324914186

python/python/tests/test_query.py
```diff
@@ -94,6 +94,34 @@ def test_query_builder(table):
...
 
 
+def test_query_builder_batches(table):
+    rs = (
+        LanceVectorQueryBuilder(table, [0, 0], "vector")
+        .limit(2)
+        .select(["id", "vector"])
+        .to_batches(1)
+    )
+    assert len(rs) == 2
```
wjones127, line 104: Could we assert the type of the returned value?
```suggestion
    assert len(rs) == 2
    assert all(isinstance(item, pa.RecordBatch) for item in rs)
```

wjones127 (changes requested):

python/python/tests/test_query.py
ighose: Thanks, done.

python/python/lancedb/query.py
```diff
@@ -267,6 +267,16 @@ def to_list(self) -> List[dict]:
...
+        """
+        return self.to_arrow().to_batches(max_chunksize)
```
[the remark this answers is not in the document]
changhiskhan: in case you're looking for more help:

The `to_arrow` method will end up calling the `_execute_query` method here: https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/table.py#L1517, which calls `LanceDataset.to_table`. This pulls the entire result set into memory.

Instead, i think you can do the following:

1. Change `_execute_query` to call `LanceDataset.to_batches()`
2. This new `to_batches` method will call `_execute_query` directly
3. Change `to_arrow()` to be `pa.Table.from_batches(self.to_batches())`


cc @wjones127 does that make sense?

changhiskhan (changes requested):

python/python/lancedb/query.py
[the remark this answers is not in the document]
ighose: Thanks @changhiskhan I've tried `changing _execute_query` and `to_batches` to call `LanceDataset.to_batches() `and `_execute_query` respectively and rerunning the unit tests.

While executing `test_query_builder_batches` (with minor modifications) setting `batch_size = 1`, `number of rows = 2` and `limit = 2`, I expected two `RecordBatch` objects in the iterator, each containing 1 row but  I'm finding that the generator object of type `LanceScanner.to_batches` returned contains a single `RecordBatch` object with all (both) the rows.

This is the unit test passing currently:
```
def test_query_builder_batches(table):
    rs = (
        LanceVectorQueryBuilder(table, [0, 0], "vector")
        .limit(2)
        .select(["id", "vector"])
        .to_batches(1)
    )
    rs_list = []
    for item in rs:
        rs_list.append(item)
        assert isinstance(item, pa.RecordBatch)
    assert len(rs_list) == 1
    assert len(rs_list[0]["id"]) == 2
    assert all(rs_list[0].to_pandas()["vector"][0] == [1.0, 2.0])
    assert rs_list[0].to_pandas()["id"][0] == 1
    assert all(rs_list[0].to_pandas()["vector"][1] == [3.0, 4.0])
    assert rs_list[0].to_pandas()["id"][1] == 2
```
    
I was also able to reproduce this by calling the lance API directly on a table of 20 rows using
`res1 = ds.to_batches(columns=["vector", "id"], batch_size=1)`
here ds is a `LanceDataset`

Do we update the `LanceScanner` separately and make this change anyway?

ighose:
Thanks @changhiskhan and @wjones127 

Made these changes:
1. Changed `_execute_query` to call `[LanceDataset.to](http://lancedataset.to/)_batches()`.
2. New `to_batches` method calls `_execute_query` directly (this is not working as expected - issue described in the comment above).
3. Changed `to_arrow()` to be `pa.Table.from_batches([self.to](http://self.to/)_batches(), projected_schema)`.
5. Added `query(self) -> Query` in `python/python/lancedb/query.py` to extract the common `Query` initialization code in `to_arrow` and `to_batches`.
6. Added `_get_query_projected_schema(self, query: Query) -> pa.Schema` in `python/python/lancedb/table.py `- wrapper around `LanceDataset.scanner.projected_schema`. We need to pass in the schema to `from_batches` in to_arrow() if result from query execution is empty.

ighose:
`pylance` version was `0.10.1`, updating the `test_query_builder_batches` unit test

python/python/lancedb/query.py
```diff
@@ -518,7 +520,34 @@ def to_arrow(self) -> pa.Table:
...
+
+        In addition to the selected columns, LanceDB also returns a vector
+        and also the "_distance" column which is the distance between the query
+        vector and the returned vectors.
+        """
+        return pa.Table.from_batches(
+            self.to_batches(), self._table._get_query_projected_schema(self.query())
+        )
+
+    def to_batches(self, batch_size: Optional[int] = None) -> Iterable[pa.RecordBatch]:
```
wjones127, line 537: Matter of taste, maybe, but I'd prefer `batch_size` is keyword only.

```suggestion
    def to_batches(self, /, batch_size: Optional[int] = None) -> Iterable[pa.RecordBatch]:
```

python/python/lancedb/query.py
```diff
@@ -518,7 +520,34 @@ def to_arrow(self) -> pa.Table:
...
+        Execute the query and return the results as an
+        [Apache Arrow Table](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table).
+
+        In addition to the selected columns, LanceDB also returns a vector
+        and also the "_distance" column which is the distance between the query
+        vector and the returned vectors.
+        """
+        return pa.Table.from_batches(
+            self.to_batches(), self._table._get_query_projected_schema(self.query())
+        )
```
wjones127, line 535:
```suggestion
        return self.to_batches().read_all()
```

python/python/tests/test_query.py
```diff
@@ -46,6 +64,7 @@ def _execute_query(self, query):
                 "nprobes": query.nprobes,
                 "refine_factor": query.refine_factor,
             },
+            batch_size=batch_size,
         )
```
wjones127, line 68:
```suggestion
            batch_size=batch_size,
        ).to_reader()
```

python/python/tests/test_query.py
```diff
@@ -32,9 +34,25 @@ def __init__(self, tmp_path):
...
+                "metric": query.metric,
+                "nprobes": query.nprobes,
+                "refine_factor": query.refine_factor,
+            },
+        ).projected_schema
+
+    def _execute_query(self, query, batch_size: Optional[int] = None):
         ds = self.to_lance()
-        return ds.to_table(
+        return ds.to_batches(
```
wjones127, line 55:
```suggestion
        return ds.scanner(
```

python/python/lancedb/query.py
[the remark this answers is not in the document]
wjones127: I don't think batch size is expected to work with vector search. There's no way to actually do the search without loading all results in memory. Batch size is more for scanning the whole table in chunks. So I think this is fine as is.

wjones127 (changes requested):
I have a few more suggestions, and then I think this is ready to go.

I think we are better off calling `ds.scanner(...).to_reader()`, which returns a [RecordBatchReader](https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatchReader.html#pyarrow.RecordBatchReader). That object itself is an `Iterable[RecordBatch]`, but it also has a bunch of other useful methods. This includes the `schema`, so it makes `_get_query_projected_schema` unnecessary.

ighose:
Thanks for the suggestions and explanation, @wjones127!

Made these changes:

1. Addressed comments and made the suggested changes
2. Updated the result returned by `_execute_query` in `python/python/lancedb/remote/table.py` from pa.Table to pa. RecordBatchReader

changhiskhan (approved):
hey thanks for making the changes. We had a big refactor recently so I took the liberty to update and resolve merge conflicts / fix tests. 

cc @wjones127 do you have any other concerns here?

wjones127 (approved):
Looks good!

PR #1052, opened 2024-03-02
feat: add create_index to the async python API

westonpace:
This also refactors the rust lancedb index builder API (and, correspondingly, the nodejs API)


python/python/tests/test_index.py
```diff
@@ -0,0 +1,85 @@
...
+        data,
+    )
+
+
+@pytest.mark.asyncio
+async def test_create_scalar_index(some_table: AsyncTable):
+    # Can create
+    await some_table.create_index(column="id").scalar().btree()
+    # Can recreate if replace=True
+    await some_table.create_index(column="id", replace=True).scalar().btree()
```
wjones127, line 57: This API looks odd to me. I would expect something like:

```python
await some_table.create_index(on="id").btree().execute()
```

The last method call being the final action that is `await`'ed. And I don't see the benefit of having `scalar()` in the call chain.

rust/lancedb/src/index.rs
```diff
@@ -37,137 +36,276 @@ pub enum IndexParams {
...
-    pub(crate) name: Option<String>,
-    /// Replace the existing index.
+    pub(crate) columns: Option<Vec<String>>,
     pub(crate) replace: bool,
+}
+
+/// Builder for creating some kind of vector index.
+///
+/// See [IndexBuilder::vector] for more details.
+pub struct VectorIndexBuilder {
```
wjones127, line 66: Why are there separate vector vs scalar indices? Given that there are specific builders for BTree and IVF_PQ?

rust/lancedb/src/index.rs
```diff
@@ -37,137 +36,276 @@ pub enum IndexParams {
...
+    /// will likely result in poor performance.
+    pub fn num_sub_vectors(mut self, num_sub_vectors: u32) -> Self {
+        self.num_sub_vectors = Some(num_sub_vectors);
         self
     }
 
-    /// Whether to replace the existing index, default is `true`.
-    pub fn replace(mut self, v: bool) -> Self {
-        self.replace = v;
+    /// The rate used to calculate the number of training vectors for kmeans.
```
wjones127, line 199: "Rate" is a confusing name for integer parameter. Could we say something along the lines of

```suggestion
    /// The number of vectors per partition to sample when training IVF K Means.
```

rust/lancedb/src/index.rs
```diff
@@ -37,137 +36,276 @@ pub enum IndexParams {
...
-    /// Set the index name.
-    pub fn name(mut self, name: &str) -> Self {
-        self.name = Some(name.to_string());
+    /// Max iteration to train kmeans.
+    ///
+    /// When training an IVF PQ index we use kmeans to calculate the partitions.  This parameter
+    /// controls how many iterations of kmeans to run.
+    ///
+    /// Increasing this might improve the quality of the index but in most cases these extra
+    /// iterations have diminishing returns.
```
wjones127, line 223: IIRC, increasing this only has an effect if KMeans doesn't converge. If it does converge, then it will stop before it even hits the max. This is based on the `tolerance` parameter in Lance.

wjones127 (commented):
Most of this looks great. Have some questions on the API design for parts of it though.

rust/lancedb/src/index.rs
westonpace: I've switched the comment to something like this.

rust/lancedb/src/index.rs
westonpace: I changed the wording slightly to mention this.  I didn't mention `tolerance` since it isn't exposed.

rust/lancedb/src/index.rs
westonpace: These vector/scalar builders have gone away

westonpace:
@wjones127 I've reworked the API so that `create_index` is now an async method and python/nodejs no longer use a builder pattern.

Cargo.toml
```diff
@@ -28,13 +28,14 @@ arrow-schema = "50.0"
 arrow-arith = "50.0"
 arrow-cast = "50.0"
 async-trait = "0"
-chrono = "0.4.23"
+chrono = "0.4.35"
```
wjones127, line 31: 🙌

nodejs/lancedb/indices.ts
```diff
@@ -0,0 +1,198 @@
...
+   * the correct row ids.
+   *
+   * This index is good for scalar columns with mostly distinct values and does best when
+   * the query is highly selective.
+   *
+   * The btree index does not currently have any parameters though parameters such as the
+   * block size may be added in the future.
+   *
+   * Note that building a btree index on a large dataset may require a large amount
+   * of RAM.
```
wjones127, line 170: Did we fix this?

wjones127 (approved):

nodejs/lancedb/indices.ts
westonpace: Yes, thanks.  I've removed these comments.

PR #1054, opened 2024-03-03
feat(python): allow user to override api url

changhiskhan:


Sciumo:
API key should be optional.
May *not* actually be going to Open AI

python/python/lancedb/embeddings/openai.py
```diff
@@ -79,6 +83,16 @@ def generate_embeddings(
     def _openai_client(self):
         openai = attempt_import_or_raise("openai")
 
-        if not os.environ.get("OPENAI_API_KEY"):
+        if self.api_key is None and not os.environ.get("OPENAI_API_KEY"):
```
Sciumo, line 86: why not just let the API fail if missing?

AyushExel, line 86: when using self hosted version, api key might not be required:
```
is_local = self.base_url is not None
if self.api_key is None and not os.environ.get("OPENAI_API_KEY") and not is_local:
    api_key_not_found_help("openai")
```

AyushExel: Its just a minor UX thing. To provide a better error message that tells user what to change exactly :)

AyushExel (approved):

PR #1058, opened 2024-03-04
fix: sanitize foreign schemas

westonpace:
Arrow-js uses brittle `instanceof` checks throughout the code base.  These fail unless the library instance that produced the object matches exactly the same instance the vectordb is using.  At a minimum, this means that a user using arrow version 15 (or any version that doesn't match exactly the version that vectordb is using) will get strange errors when they try and use vectordb.

However, there are even cases where the versions can be perfectly identical, and the instanceof check still fails.  One such example is when using `vite` (e.g. https://github.com/vitejs/vite/issues/3910)

This PR solves the problem in a rather brute force, but workable, fashion.  If we encounter a schema that does not pass the `instanceof` check then we will attempt to sanitize that schema by traversing the object and, if it has all the correct properties, constructing an appropriate `Schema` instance via deep cloning.


westonpace:
Note: I disabled some lint checks because we are (hopefully) moving to use prettier in nodejs in https://github.com/lancedb/lancedb/pull/1042 and these checks disagreed with prettier's formatting.

node/src/arrow.ts
wjones127, line 1: What do you think of putting the `sanitize*` functions in their own module / file?

wjones127 (approved):

eddyxu (approved):

node/src/arrow.ts
westonpace: Yeah, that's a good idea.  I've put these into a `sanitize.ts` file.

wjones127 (approved):

Issue #485, opened 2023-09-08
lancedb connector functionality does not work with azure url path


EmilijaL:
create_table() works with azure url path to blob storage, but other functionality like table_names()  or open_table()  throws error

changhiskhan:
What's the error?

EmilijaL:
It works creating table:
![create_table](https://github.com/lancedb/lancedb/assets/32243027/652144e9-4817-4edc-b462-f1091b483b2f)

But other methods throw error:
![open_database](https://github.com/lancedb/lancedb/assets/32243027/a4bef0b9-cf43-4f6f-97f6-a244387dfdd8)
![table_names](https://github.com/lancedb/lancedb/assets/32243027/107dac00-d7af-4699-8b7f-9da1694507d1)

westonpace:
This should have been fixed by https://github.com/lancedb/lancedb/pull/1102

PR #1059, opened 2024-03-04
feat: page_token / limit to native table_names function.  Use async table_names function from sync table_names function

westonpace:
The synchronous table_names function in python lancedb relies on arrow's filesystem which behaves slightly differently than object_store.  As a result, the function would not work properly in GCS.

However, the async table_names function uses object_store directly and thus is accurate.  In most cases we can fallback to using the async table_names function and so this PR does so.  The one case we cannot is if the user is already in an async context (we can't start a new async event loop).  Soon, we can just redirect those users to use the async API instead of the sync API and so that case will eventually go away.  For now, we fallback to the old behavior.

### DIFF
--- a/python/src/connection.rs
+++ b/python/src/connection.rs
@@ -69,11 +69,20 @@ impl Connection {
         self.inner.take();
     }
 
-    pub fn table_names(self_: PyRef<'_, Self>) -> PyResult<&PyAny> {
-        let inner = self_.get_inner()?.clone();
-        future_into_py(self_.py(), async move {
-            inner.table_names().await.infer_error()
-        })
+    pub fn table_names(
+        self_: PyRef<'_, Self>,
+        page_token: Option<String>,
+        limit: Option<u32>,
+    ) -> PyResult<&PyAny> {
+        let inner = self_.inner.clone();
+        let mut op = inner.table_names();
+        if let Some(page_token) = page_token {
+            op = op.page_token(page_token);
+        }
+        if let Some(limit) = limit {
+            op = op.limit(limit);
+        }
+        future_into_py(self_.py(), async move { op.execute().await.infer_error() })
     }
 
     pub fn create_table<'a>(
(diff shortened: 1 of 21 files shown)


westonpace:
Partially fixes #485 (fixes table_names but does not fix open_table).  The latter will need to wait until the async API is fully ready (~1-2 weeks)

python/src/connection.rs
```diff
@@ -41,11 +41,20 @@ impl Connection {
 
 #[pymethods]
 impl Connection {
-    pub fn table_names(self_: PyRef<'_, Self>) -> PyResult<&PyAny> {
+    pub fn table_names(
+        self_: PyRef<'_, Self>,
+        page_token: Option<String>,
+        limit: Option<u32>,
+    ) -> PyResult<&PyAny> {
```
wjones127, line 48: Hmmm. I wonder if we would consider changing the opaque `page_token` parameter to `startOn` or `startAfter`. Since we are guaranteeing an alphabetical list of strings, it doesn't seem like we need the opaqueness, and having a more concrete parameter would be more useful.

For example, it makes it easier to get tables with a particular prefix, or to check for existence of a single table.

westonpace: I've renamed `page_token` to `start_after`/`startAfter` (I also added a lint rule for naming conventions since `page_token` in a typescript file was definitely wrong).

At the risk of bike shedding, I don't this really solves all problems.

 * `start_on` doesn't work well for paging because it forces you to grab the previous result
 * `start_after` doesn't work well for single table since it will skip the given item (it can work "ok" for prefix but its still a bit weird)

I've stuck with `start_after` because paging was the original intention.  Maybe we want to add a `filter` option in the future?

wjones127: `startAfter` sounds good to me. Yeah `filter` might be a better solution.

wjones127 (approved):

PR #1070, opened 2024-03-07
feat: add time travel operations to the async API

westonpace:


westonpace:
Leaving in draft until #1052 is merged

wjones127:
Looks like windows has a bad clock and is failing on my test. I guess we could bump it to 150ms wait and that should be fixed?

westonpace:
> Looks like windows has a bad clock and is failing on my test. I guess we could bump it to 150ms wait and that should be fixed?

The clock / sleep isn't really the problem.  The problem is that the CI machine hangs between the `add` and the `count_rows`.  So, by the time it runs `count_rows`, 100ms has already passed and so it grabs the latest version.  I can try bumping the interval in a future PR.  In my experience the Windows CI machines can hang for seconds at a time.

PR #1071, opened 2024-03-07
feat: add py.typed for mypy library stubs

eduardjbotha:
With `lancedb` as a dependency, running mypy on a project complains with the following error:
```
error: Skipping analyzing "lancedb": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```


wjones127:
@eduardjbotha  I would like to support this. However, we do not check LanceDB Python's annotations in CI yet, so I don't think many of them are event correct at the moment. Therefore, I'm reticent to turn this on until we start validating our codebase with mypy. I've filed https://github.com/lancedb/lancedb/issues/1117 to do that first. If you would like, you are welcome to open a PR for that.

wjones127:
We ended up starting to use `pyright` to check types in this repo: https://github.com/lancedb/lancedb/pull/2167

PR #1085, opened 2024-03-09
doc updates

### LINKED ISSUE #1084: update Imagebind integration docs
### Description

- [ ] imagebind demo addition
- [ ] add llamaindex video RAG example 

### Link

_No response_

raghavdixit99:
closes #1084


docs/src/embeddings/default_embedding_functions.md
```diff
@@ -221,7 +221,7 @@ We support CLIP model embeddings using the open source alternative, [open-clip](
 This embedding function supports ingesting images as both bytes and urls. You can query them using both test and other images.
 
 !!! info
-    LanceDB supports ingesting images directly from accessible links.
+    LanceDB supports ingesting images,audio and video directly from accessible links.
```
AyushExel, line 224: I think here ingesting images directly referred to Lance Having image extension type and not just that these embedding functions can generate embeddings directly from links. So, I'd be a little hesitant to say we support "ingesting" anything other than images directly. But I'd defer this part to @changhiskhan

docs/src/embeddings/default_embedding_functions.md
```diff
@@ -289,4 +289,65 @@ print(actual.label)
 
 ```
 
+### Imagebind embeddings
+We have support for [imagebind](https://github.com/facebookresearch/ImageBind) model embeddings. It is registered as `imagebind` and supports Audio, Video and Text modalities(extending to Thermal,Depth,IMU data):
```
AyushExel, line 293: Users still need to install this package from your fork right? Can you just publish a pypi package under some name like "imagebind-packaged" etc. and add an instruction here?

AyushExel (commented):
Thanks. Looks great overall. I've left 2 smol comments.
Just one more important this -- Can you just check the size of the notebook file locally? I think with the saved graphics, it should be at least a few MBs? I think we need to avoid adding such large notebook files as it'll just increase the size of the repo and even on being deleted, it'll stay in the branch history. I'm guilty of doing this too. Let's avoid this from now.
You can do either of these:
* Either just save the notebook without outputs.
* Or if you need to use the pics, replace the actual assets by assets links by dragging/dropping images to an .md file on github (It can be any personal github repo, like I have one here -https://github.com/AyushExel/assets )

raghavdixit99:
> Thanks. Looks great overall. I've left 2 smol comments. Just one more important this -- Can you just check the size of the notebook file locally? I think with the saved graphics, it should be at least a few MBs? I think we need to avoid adding large notebook files as it'll just increase the size of the repo and even on being deleted, it'll stay in the branch history. I'm guilty of doing this too. Let's avoid this from now. If the size is actually greater than a megabyte, You can do either of these:
> 
> * Either just save the notebook without outputs.
> * Or if you need to use the pics, replace the actual assets by assets links by dragging/dropping images to an .md file on github (It can be any personal github repo, like I have one here -https://github.com/AyushExel/assets )


@AyushExel I had already used hyperlinks from my personal assets repo both in .md and ipynb files. You can verify in pr

docs/src/embeddings/default_embedding_functions.md
raghavdixit99: yeah makes sense , I will specify how users can add local paths instead of this

docs/src/embeddings/default_embedding_functions.md
raghavdixit99: yeah will be publishing that soon

AyushExel: @raghavdixit99 any updates on this?

AyushExel (approved):

AyushExel:
Thanks looks good

PR #1090, opened 2024-03-11
feat: configurable timeout for LanceDB Cloud queries

chebbyChefNEQ:


wjones127 (approved):

albertlockett:
Maybe it would make sense to expose the retry params in the same way. Currently it's all configured through env vars:
https://github.com/lancedb/lancedb/blob/9fd06358035faa139c5bb513300fa86e711b2803/python/python/lancedb/remote/client.py#L206-L220
Doesn't have to be as part of this PR - can tackle that down the line.

albertlockett (approved):

PR #1091, opened 2024-03-11
fix: handle uri in object

### LINKED ISSUE #1078: bug(node): error when using local storage
### LanceDB version

v0.4.11

### What happened?

In short, I called

```
lancedb.connect({
  "uri": "/var/folders/j5/zqtb_vtx6p166jkfgpkcrfd80000gp/T/lance-db-test-hoMBti"
})
```

Then I got an exception:

```
Error: Invalid credentials configuration
    at LocalConnection.createTableImpl (...node_modules/vectordb/src/index.ts:715:35)
    at LocalConnection.createTable (...node_modules/vectordb/src/index.ts:676:12)
```

things seem to work if I do this:

```
lancedb.connect({
  awsCredentials: undefined,
  awsRegion: undefined,
  "uri": "/var/folders/j5/zqtb_vtx6p166jkfgpkcrfd80000gp/T/lance-db-test-hoMBti"
})
```

_No response_

wjones127:
Fixes #1078

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -176,16 +176,21 @@ export async function connect (
     opts = { uri: arg }
   } else {
     // opts = { uri: arg.uri, awsCredentials = arg.awsCredentials }
-    opts = Object.assign(
-      {
-        uri: '',
-        awsCredentials: undefined,
-        awsRegion: defaultAwsRegion,
-        apiKey: undefined,
-        region: defaultAwsRegion
-      },
-      arg
-    )
+    const keys = Object.keys(arg)
+    if (keys.length === 1 && keys[0] === 'uri' && typeof arg.uri === 'string') {
+      opts = { uri: arg.uri }
+    } else {
+      opts = Object.assign(
+        {
+          uri: '',
+          awsCredentials: undefined,
+          awsRegion: defaultAwsRegion,
+          apiKey: undefined,
+          region: defaultAwsRegion
+        },
+        arg
+      )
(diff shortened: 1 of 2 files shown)


node/src/index.ts
```diff
@@ -176,16 +176,21 @@ export async function connect (
...
+    } else {
+      opts = Object.assign(
+        {
+          uri: '',
+          awsCredentials: undefined,
+          awsRegion: defaultAwsRegion,
+          apiKey: undefined,
+          region: defaultAwsRegion
+        },
+        arg
```
flaviut, line 191: you're obviously more familiar with this code than I am, but I would have thought that the best solution here would be to explicitly check the URL type & follow two different code paths, like `RemoteConnection` below.

flaviut (approved):
this is extremely similar to in effect to my personal workaround -- should work great.

node/src/index.ts
flaviut: (or [perhaps edit the rust part to not try and validate aws creds](https://github.com/lancedb/lancedb/blob/main/rust/ffi/node/src/lib.rs#L205-L240) when it doesn't receive an S3 url)

wjones127: TBH we are a few weeks away from stopping development of this SDK in favor of the new Node one, so I'm not inclined to put too much effort here. This seemed like the easiest quick fix.

westonpace (approved):

PR #1092, opened 2024-03-11
fix: propagate filter validation errors

### LINKED ISSUE #1081: bug(node): filtering doesn't work, no errors thrown
### LanceDB version

0.4.12

### What happened?

```javascript
import * as lancedb from 'vectordb'
import { Field, Schema as ArrowSchema } from 'apache-arrow'
import * as types from 'apache-arrow/type'
import { WriteMode } from 'vectordb'

  // work around https://github.com/lancedb/lancedb/issues/1078
  awsCredentials: undefined,
  awsRegion: undefined,
  uri: './scratch',
})

const table = await db.createTable({
  name: 'table',
  schema: new ArrowSchema([
    new Field(
      'vector',
      new types.FixedSizeList(4, new Field('item', new types.Float32()))
    ),
    new Field('allowedGroups', new types.Utf8()),
  ]),
  writeOptions: {
    writeMode: WriteMode.Overwrite,
  }
})

await table.add([
  {
    vector: [1, 2, 3, 4],
    allowedGroups: '[[admin]]\t[[user]]'
  },
  {
    vector: [5, 6, 7, 8],
    allowedGroups: '[[user]]',
  },
])

console.log(await table.filter("allowedGroups = 'none'").execute())
console.log(await table.filter("allowedGroups LIKE '%[[admin]]%'").execute())
console.log(await table.filter("'123t3").execute())
```
```
[
  { vector: [ 1, 2, 3, 4 ], allowedGroups: '[[admin]]\t[[user]]' },
  { vector: [ 5, 6, 7, 8 ], allowedGroups: '[[user]]' }
]
[
  { vector: [ 1, 2, 3, 4 ], allowedGroups: '[[admin]]\t[[user]]' },
  { vector: [ 5, 6, 7, 8 ], allowedGroups: '[[user]]' }
]
[
  { vector: [ 1, 2, 3, 4 ], allowedGroups: '[[admin]]\t[[user]]' },
  { vector: [ 5, 6, 7, 8 ], allowedGroups: '[[user]]' }
]
```

I would expect this to have instead returned:

```
[ ]
[
  { vector: [ 1, 2, 3, 4 ], allowedGroups: '[[admin]]\t[[user]]' }
]
Error: Invalid filter expression
    at file:///...
```

this issue is particularly concerning, because this filtering may be used for access control. Applications that depend on that behavior may leak data to unauthorized users.

_No response_

wjones127:
In Rust and Node, we have been swallowing filter validation errors. If there was an error in parsing the filter, then the filter was silently ignored, returning unfiltered results.

Fixes #1081

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -978,9 +978,17 @@ impl TableInternal for NativeTable {
             Select::All => { /* Do nothing */ }
         }
 
-        query.filter.as_ref().map(|f| scanner.filter(f));
-        query.refine_factor.map(|rf| scanner.refine(rf));
-        query.metric_type.map(|mt| scanner.distance_metric(mt));
+        if let Some(filter) = &query.filter {
+            scanner.filter(filter)?;
+        }
+
+        if let Some(refine_factor) = query.refine_factor {
+            scanner.refine(refine_factor);
+        }
+
+        if let Some(metric_type) = query.metric_type {
+            scanner.distance_metric(metric_type);
+        }
         Ok(scanner.try_into_stream().await?)
     }
 
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -128,6 +128,11 @@ describe('LanceDB client', function () {
       assertResults(results)
       results = await table.where('id % 2 = 0').execute()
       assertResults(results)
+
+      // Should reject a bad filter
+      await expect(table.filter('id % 2 = 0 AND').execute()).to.be.rejectedWith(
+        /.*sql parser error: Expected an expression:, found: EOF.*/
+      )
     })
 
     it('uses a filter / where clause', async function () {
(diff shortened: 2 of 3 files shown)


eddyxu (approved):

rust/lancedb/src/table.rs
```diff
@@ -978,9 +978,17 @@ impl TableInternal for NativeTable {
...
+            scanner.filter(filter)?;
+        }
+
+        if let Some(refine_factor) = query.refine_factor {
+            scanner.refine(refine_factor);
+        }
+
+        if let Some(metric_type) = query.metric_type {
+            scanner.distance_metric(metric_type);
+        }
```
wjones127, line 991: I found that by changing to `if let Some(...) = ...`, then clippy will warn us about unused results. So to be defensive, I've changed these other parameters as well.

PR #1110, opened 2024-03-14
feat(python): support writing huggingface dataset and dataset dict

changhiskhan:
HuggingFace Dataset is written as arrow batches.
For DatasetDict, all splits are written with a "split" column appended.

- [x] what if the dataset schema already has a `split` column
- [x] add unit tests

### DIFF
--- a/python/python/tests/test_huggingface.py
+++ b/python/python/tests/test_huggingface.py
@@ -1,4 +1,4 @@
-#  Copyright 2023 Lance Developers
+#  Copyright 2024 Lance Developers
 #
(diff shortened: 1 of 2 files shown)


python/python/tests/test_huggingface.py
```diff
@@ -0,0 +1,126 @@
+#  Copyright 2023 Lance Developers
```
eddyxu, line 1: 2024

eddyxu (approved):
One minor copyright issue.

PR #1113, opened 2024-03-15
feat: refactor the query API and add query support to the python async API

westonpace:
In addition, there are also a number of changes in nodejs to the docstrings of existing methods because this PR adds a jsdoc linter.


python/python/lancedb/query.py
```diff
@@ -921,3 +924,337 @@ def refine_factor(self, refine_factor: int) -> LanceHybridQueryBuilder:
...
+        Note: if there is a vector index then the distance type used MUST match the
+        distance type used to train the vector index.  If this is not done then the
+        results will be invalid.
+
+        By default "l2" is used.
+        """
+        self._inner.distance_type(distance_type)
+        return self
+
+    def postfilter(self) -> AsyncVectorQuery:
```
wjones127, line 1225: Currently the default is pre-filter, right? I agree this should be default now. Just curious.

python/python/tests/test_query.py
```diff
@@ -184,3 +205,95 @@ def test_query_builder_with_different_vector_column():
...
+    async for batch in results:
+        if expected_columns is not None:
+            assert batch.schema.names == expected_columns
+        num_rows += batch.num_rows
+    if expected_num_rows is not None:
+        assert num_rows == expected_num_rows
+
+
+@pytest.mark.asyncio
+async def test_query_async(table_async: AsyncTable):
```
wjones127, line 224: Try a test with zero results (filter out everything)?

wjones127 (commented):
I have a bit more to look through soon, but here are my initial comments.

python/python/lancedb/arrow.py
```diff
@@ -0,0 +1,20 @@
+import pyarrow as pa
+
+from ._lancedb import RecordBatchStream
+
+
+class AsyncRecordBatchReader(object):
```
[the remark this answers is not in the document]
westonpace: Old habits...

I'll fix this.

python/python/lancedb/query.py
westonpace: The default currently (in the sync lancedb and node's vectordb) is postfilter.  This **is** a change to the API.

python/python/tests/test_query.py
westonpace: Added tests with zero results.

python/python/lancedb/query.py
```diff
@@ -921,3 +924,339 @@ def refine_factor(self, refine_factor: int) -> LanceHybridQueryBuilder:
...
+        Choose which columns to return
+
+        This is a shortcut for [select_with_projection][] when no
+        dynamic columns need to be calculated.  See that method for more
+        details on the performance impact of selecting the right columns.
+        """
+        self._inner.select(columns)
+        return self
+
+    def select_with_projection(self, columns: dict[str, str]) -> AsyncQuery:
```
eddyxu, line 966: Can we use same `select()` api for both function?
`projection` is a very database-eng term that application / ML developers are unfamiliar with.

python/src/util.rs
```diff
@@ -33,3 +37,15 @@ impl<T> BuilderWrapper<T> {
         Ok(result)
     }
 }
+
+pub fn parse_distance_type(distance_type: impl AsRef<str>) -> PyResult<DistanceType> {
+    match distance_type.as_ref() {
```
eddyxu, line 42: do we want to make distance type lower case ?

python/python/lancedb/query.py
westonpace: Ok.  I changed them both to `select`.  I agree that's simpler.

python/src/util.rs
westonpace: I did.  Distance type comparison is now case insensitive.

eddyxu (approved):

Issue #1112, opened 2024-03-15
Clarification on vonoroi seed values


koaning:
### Description

I'm a happy user of LanceDB but I'd like to understand the internals a bit better, in particular how the index is created. When I check [the docs page](https://lancedb.github.io/lancedb/concepts/index_ivfpq/#product-quantization) it mentions that the vector space is divided into Vonoroi cells, it's the part with this image. 

![image](https://github.com/lancedb/lancedb/assets/1019791/20748547-0623-4680-8566-212600b8aaab)

The quantization bit before is super clear to me, but I'm not sure how these Vonoroi cells are created. I understand how a collection of seed points can determine the boundaries of each cell, but how are these seed points chosen? Are these chosen randomly or do the internals do something clever to pick seed points? My gut is assuming that you can do something clever if you have knowledge of the distribution of your vectors, but I couldn't find the resources on how lancedb currently does this. 

I'll gladly contribute to the docs if somebody could help me understand. 

### Link

https://lancedb.github.io/lancedb/concepts/index_ivfpq/#product-quantization

koaning:
Ah, ok that's not unreasonable. Mind if I make a PR to add that little detail to the docs though? I can imagine it would give just that extra bit of context to folks.

koaning:
Made one over here, thanks, https://github.com/lancedb/lancedb/pull/1114

PR #1122, opened 2024-03-15
Unhide Pydantic guides in Docs

koaning:
@wjones127 after fixing https://github.com/lancedb/lancedb/issues/1112 I noticed something else on the docs. There's an odd chunk of the docs missing [here](https://lancedb.github.io/lancedb/guides/tables/#from-a-polars-dataframe). I can see the heading, but after clicking it the contents don't show. 

CleanShot 2024-03-15 at 23 40 17@2x

Apon inspection it was a markdown issue, one tab too many on a whole segment.

This PR fixes it. It looks like this now and the sections appear again:

CleanShot 2024-03-15 at 23 42 32@2x

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -168,151 +168,151 @@ This guide will show how to create tables, insert data into them, and update the
     --8<-- "docs/src/basic_legacy.ts:create_f16_table"
     ```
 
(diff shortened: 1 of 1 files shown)


changhiskhan (approved):
thank you!

PR #1130, opened 2024-03-18
Node SDK Client middleware for HTTP Requests

albertlockett:
Adds client-side middleware to LanceDB Node SDK to instrument HTTP Requests

Example - adding `x-request-id` request header:
```js
class HttpMiddleware {
    constructor({ requestId }) {
        this.requestId = requestId
    }

    onRemoteRequest(req, next) {
        req.headers['x-request-id'] = this.requestId
        return next(req)
    }
}

  uri: 'db://remote-123',
  apiKey: 'sk_...',
})

let tables = await db.withMiddleware(new HttpMiddleware({ requestId: '123' })).tableNames();

```

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -309,7 +309,7 @@ export interface Connection {
    *
    * The middleware will be called in the order they are added.
    *
-   *  Currently this functionality is only supported for remote Connections.
+   * Currently this functionality is only supported for remote Connections.
    *
    * @param {HttpMiddleware} - Middleware which will instrument the Connection.
    * @returns - this Connection instrumented by the passed middleware
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -556,7 +556,7 @@ export interface Table<T = number[]> {
   /**
    * Instrument the behavior of this Table with middleware.
    *
-   * The middlewares callback implementations will be called in the order they are added.
+   * The middleware will be called in the order they are added.
    *
    * @param {ConnectionMiddleware} - Middleware which will instrument the .
    * @returns - this Table instrumented by the passed middleware
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -24,7 +24,7 @@ import { RemoteConnection } from './remote'
 import { Query } from './query'
 import { isEmbeddingFunction } from './embedding/embedding_function'
 import { type Literal, toSQL } from './util'
-import { type ConnectionMiddleware } from './middleware'
+import { type HttpMiddleware } from './middleware'
 
 const {
   databaseNew,
@@ -305,16 +305,14 @@ export interface Connection {
   dropTable(name: string): Promise<void>
 
   /**
-   * Instrument the behavior of this connection with middleware.
+   * Instrument the behavior of this Connection with middleware.
    *
-   * The middlewares will be added in a chain and their respective instrumentation
-   * callback implementations will be called in the order they are passed into
-   * invocations of this method.
+   * The middlewares callback implementations will be called in the order they are added.
    *
-   * @param {ConnectionMiddleware} - Middleware which will instrument the connection.
-   * @returns - this connection instrumented by the passed middleware
+   * @param {HttpMiddleware} - Middleware which will instrument the Connection.
+   * @returns - this Connection instrumented by the passed middleware
    */
-  withMiddleware(middleware: ConnectionMiddleware): Connection
+  withMiddleware(middleware: HttpMiddleware): Connection
 }
 
 /**
@@ -555,7 +553,15 @@ export interface Table<T = number[]> {
    */
   dropColumns(columnNames: string[]): Promise<void>
 
-  // TODO add methods for middleware
+  /**
+   * Instrument the behavior of this Table with middleware.
+   *
+   * The middlewares callback implementations will be called in the order they are added.
+   *
+   * @param {ConnectionMiddleware} - Middleware which will instrument the .
+   * @returns - this Table instrumented by the passed middleware
+   */
+  withMiddleware(middleware: HttpMiddleware): Table<T>
 }
 
 /**
@@ -811,8 +817,7 @@ export class LocalConnection implements Connection {
     await databaseDropTable.call(this._db, name)
   }
 
-  withMiddleware (middleware: ConnectionMiddleware): Connection {
-    // noop
+  withMiddleware (middleware: HttpMiddleware): Connection {
     return this
   }
 }
@@ -1124,6 +1129,10 @@ export class LocalTable<T = number[]> implements Table<T> {
 
   async dropColumns (columnNames: string[]): Promise<void> {
     return tableDropColumns.call(this._tbl, columnNames)
+  }
+
+  withMiddleware (middleware: HttpMiddleware): Table<T> {
+    return this
   }
 }
 
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -24,7 +24,7 @@ import { RemoteConnection } from './remote'
 import { Query } from './query'
 import { isEmbeddingFunction } from './embedding/embedding_function'
 import { type Literal, toSQL } from './util'
-import { type ConnectionMiddleware, type MiddlewareContext } from './middleware'
+import { type ConnectionMiddleware } from './middleware'
 
 const {
   databaseNew,
@@ -315,15 +315,6 @@ export interface Connection {
    * @returns - this connection instrumented by the passed middleware
    */
   withMiddleware(middleware: ConnectionMiddleware): Connection
-
-  /**
-   * Set the context that will be passed to any middlewares that instrument this
-   * connection
-   *
-   * @param {MiddlewareContext} - Context that will be passed to middlewares
-   * @returns - this connection configured to pass the context to its middlewares
-   */
-  withMiddlewareContext(ctx: MiddlewareContext): Connection
 }
 
 /**
@@ -821,11 +812,6 @@ export class LocalConnection implements Connection {
   }
 
   withMiddleware (middleware: ConnectionMiddleware): Connection {
-    // noop
-    return this
-  }
-
-  withMiddlewareContext (ctx: MiddlewareContext): Connection {
     // noop
     return this
   }
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -307,7 +307,7 @@ export interface Connection {
   /**
    * Instrument the behavior of this Connection with middleware.
    *
-   * The middlewares callback implementations will be called in the order they are added.
+   * The middleware will be called in the order they are added.
    *
    * @param {HttpMiddleware} - Middleware which will instrument the Connection.
    * @returns - this Connection instrumented by the passed middleware
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -562,7 +562,7 @@ export interface Table<T = number[]> {
    *
    * Currently this functionality is only supported for remote tables.
    *
-   * @param {ConnectionMiddleware} - Middleware which will instrument the .
+   * @param {HttpMiddleware} - Middleware which will instrument the Table.
    * @returns - this Table instrumented by the passed middleware
    */
   withMiddleware(middleware: HttpMiddleware): Table<T>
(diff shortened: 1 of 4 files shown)


node/src/index.ts
```diff
@@ -302,6 +303,16 @@ export interface Connection {
    * @param name The name of the table to drop.
    */
   dropTable(name: string): Promise<void>
+
+  /**
+   * Instrument the behavior of this Connection with middleware.
+   *
+   * The middlewares callback implementations will be called in the order they are added.
```
westonpace, line 310:
```suggestion
   * The middleware will be called in the order they are added.
```
Or maybe `middlewares'`?

node/src/index.ts
```diff
@@ -541,6 +552,16 @@ export interface Table<T = number[]> {
    *                    names (e.g. "a").
    */
   dropColumns(columnNames: string[]): Promise<void>
+
+  /**
+   * Instrument the behavior of this Table with middleware.
+   *
+   * The middlewares callback implementations will be called in the order they are added.
```
westonpace, line 559:
```suggestion
   * The middleware will be called in the order they are added.
```

node/src/index.ts
```diff
@@ -795,6 +816,10 @@ export class LocalConnection implements Connection {
   async dropTable (name: string): Promise<void> {
     await databaseDropTable.call(this._db, name)
   }
+
+  withMiddleware (middleware: HttpMiddleware): Connection {
+    return this
```
westonpace, line 821: We should mention in the docs for `withMiddleware` that it only applies to remote connections.

node/src/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Middleware for Remote LanceDB Connection or Table
+ */
+export interface HttpMiddleware {
+  /**
+   * A callback that can be used to instrument the behavior of http requests to remote
+   * tables. It can be used to add headers, modify the request, or even short-circuit
+   * the request and return a response without making the request to the remote endpoint.
+   * It can also be used to modify the response from the remote endpoint.
+   *
+   * @param {RemoteResponse} res - Request ot the remote endpoint
```
westonpace, line 25:
```suggestion
   * @param {RemoteResponse} res - Request to the remote endpoint
```

node/src/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+   * @param {onRemoteRequestNext} next - Callback to advance the middleware chain
+   * @param {MiddlewareContext} ctx - Local context for ths invocation of the middleware
```
westonpace, line 27:
```suggestion
   * @param {MiddlewareContext} ctx - Local context for this invocation of the middleware
```

node/src/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * A LanceDB Remote HTTP Request
+ */
+export interface RemoteRequest {
+  uri: string
+  method: Method
+  headers: Record<string, string>
+  params?: Record<string, string | number>
```
westonpace, line 47: Providing these as `Map` instead of `Record` might be slightly nicer.

westonpace, line 47: Why `string | number`?  Are these partially parsed query params?

node/src/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+  data?: any
```
westonpace, line 48: What if the data is streaming?

node/src/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * A LanceDB Remote HTTP Response
+ */
+export interface RemoteResponse {
+  status: number
+  statusText: string
+  headers: Record<string, string>
+  body: () => Promise<any>
```
westonpace, line 58: What if the response is streaming?

node/src/remote/client.ts
```diff
@@ -12,13 +12,98 @@
...
+        res = await axios.post(
+          req.uri,
+          req.data,
+          {
+            headers: { ...req.headers },
+            params: req.params,
+            responseType: opts?.responseType,
+            timeout: 10000
+          }
+        )
```
westonpace, line 53: Is this missing `req.body`?

node/src/remote/client.ts
```diff
@@ -12,13 +12,98 @@
...
+      }
+
+      return toLanceRes(res)
+    }
+
+    // call next middleware in chain
+    return await middlewares[i - 1].onRemoteRequest(
+      req,
+      async (req) => {
+        return await call(i + 1, req)
```
westonpace, line 72: The indexing here seems off.  Should this be...

```
async func call(...) {
  if (i >= middlewares.length) {
    ...
  } else {
    middlewares[i].onRemoteRequest(..., async (req) => { return await call(i + 1, req); }
  }
return await call(0, req);
```

node/src/remote/client.ts
```diff
@@ -43,74 +129,61 @@ export class HttpLancedbClient {
...
+      return response
+    } catch (err: any) {
       console.error('error: ', err)
       if (err.response === undefined) {
         throw new Error(`Network Error: ${err.message as string}`)
       }
-      return err.response
-    })
+
+      response = toLanceRes(err.response)
```
westonpace, line 176: Can a middleware catch and handle an error?

westonpace (commented):
A few comments but overall this looks like something we should be able to support

Note: this is currently adding to the `node` impl which is on its way out.  Most of this code will need to move to rust soon.

node/src/middleware.ts
albertlockett: I prefer a Record. I think it'd be easier/more familiar to be able to pass an object instead of having to construct a map.
```js
req.data = { lmit, page_token: pageToken }
// vs
req.data = new Map().set('limit', limit).set('page_token', pageToken)
```

We could have the interface allow either?

albertlockett: These are parsed query string params (technically they're query string params before they're serialized). 

I hadn't thought much about the type here .. this is `string | number` b/c that's also what we're passing into client.get
https://github.com/lancedb/lancedb/blob/8f7ef26f5f0644e2cd39b8dd4c21617c6bc99a56/node/src/remote/client.ts#L88

If I were to guess why `number` is allowed here, it's probably to make it easier to pass numberic arguments w/out having to stringify them first.
```js
const limit = 10
client.get("/tables", { limit })
// vs.
client.get("/tables", { limit: `${limit}` })
```
I don't think this is so bad, but I can disallow it being a number if you feel strongly

node/src/remote/client.ts
albertlockett: This actually does work. For example if there are 2 middlewares [A, B]

When `call` is invoked for the first time (on line 77): `i = 1`, so we invoke the `i-1 = 0th = A` middleware's `onRemoteRequest` callback on line 69. Then when that middleware calls `next`, it invokes `call` for the second time on 72 with `i = i+1 = 2`

When `call` is invoked for the 2nd time, `i = 2` so we invoke the `i-1 = 1st = B` middleware's `onRemoteRequest` callback, and then when B middleware calls `next` it invokes `call` for the 3rd time with `i = i+1 = 3`

When `call` is invoked for the 3rd time, we get `i > middlewares.length` on line 41 and we make the request.

node/src/middleware.ts
albertlockett: ah OK good call. I hadn't thought about that ..  I was thinking about this more from the perspective of constructing the `RemoteRequest` object than having to modify it after. Yeah we can make these a Map

node/src/remote/client.ts
albertlockett: ah I see the confusion here -- I've changed `req.data` to be `req.body` to make it more clear.

albertlockett: The 2nd argument is the body

node/src/remote/client.ts
albertlockett: Yes it can. This is intentional. For example if a middleware wanted to catch an error and retry we could do:
```js
    class RetryMiddleware {
        constructor(attempts) {
            this.attempts = attempts
        }

        async onRemoteRequest(req, next) {
            let tries = 0;
            while(true) {
                try {
                    return await next(req)
                } catch(e) {
                    if (tries++ >= this.attempts) {
                        throw e
                    }
                    console.log('retrying')
                }
            }
        }
    }

    let tables = await db.withMiddleware(new RetryMiddleware(3)).tableNames()
```

node/src/middleware.ts
westonpace: I don't feel strongly.  I guess, in most cases, middleware authors are going to be adding to `req.params` and not reading it.  Should we add `boolean` then so we can handle the full set of JSON primitives?

node/src/middleware.ts
albertlockett: I don't think this API precludes us from doing that.. it's just not something we expose right now because we don't do any streaming responses. 

The  way I'd want to support this to have another method or property on RemoteResponse that returns a `ReadableStream`. This is similar to what `fetch` does:
https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams#consuming_a_fetch_as_a_stream

WYDT? Does that sound OK?

node/src/middleware.ts
albertlockett: Similar comment to here (https://github.com/lancedb/lancedb/pull/1130#discussion_r1533879989)

Not something we currently are doing, but I think to handle this functionality we could add another method/property to the `RemoteRequest` object to expose the body as a stream

node/src/index.ts
```diff
@@ -302,6 +303,18 @@ export interface Connection {
    * @param name The name of the table to drop.
    */
   dropTable(name: string): Promise<void>
+
+  /**
+   * Instrument the behavior of this Connection with middleware.
+   *
+   * The middleware will be called in the order they are added.
+   *
+   *  Currently this functionality is only supported for remote Connections.
```
westonpace, line 312:
```suggestion
   * Currently this functionality is only supported for remote Connections.
```

node/src/index.ts
```diff
@@ -541,6 +554,18 @@ export interface Table<T = number[]> {
...
   dropColumns(columnNames: string[]): Promise<void>
+
+  /**
+   * Instrument the behavior of this Table with middleware.
+   *
+   * The middleware will be called in the order they are added.
+   *
+   * Currently this functionality is only supported for remote tables.
+   *
+   * @param {ConnectionMiddleware} - Middleware which will instrument the .
```
westonpace, line 565:
```suggestion
   * @param {ConnectionMiddleware} - Middleware which will instrument the Table.
```

node/src/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface HttpMiddleware {
+  /**
+   * A callback that can be used to instrument the behavior of http requests to remote
+   * tables. It can be used to add headers, modify the request, or even short-circuit
+   * the request and return a response without making the request to the remote endpoint.
+   * It can also be used to modify the response from the remote endpoint.
+   *
+   * @param {RemoteResponse} res - Request to the remote endpoint
+   * @param {onRemoteRequestNext} next - Callback to advance the middleware chain
+   * @param {MiddlewareContext} ctx - Local context for this invocation of the middleware
```
westonpace, line 27: This parameter no longer exists?
```suggestion
```

node/src/middleware.ts
westonpace: Sounds ok to me.

node/src/remote/client.ts
```diff
@@ -12,13 +12,101 @@
...
+      let res
+      if (req.method === Method.POST) {
+        res = await axios.post(
+          req.uri,
+          req.body,
+          {
+            headers,
+            params,
+            timeout,
+            responseType: opts?.responseType
```
westonpace, line 54: This is fine for now but I wonder if we will need to make `MiddlewareInvocationOptions` visible to the middlware at some point?  E.g. wouldn't a middleware want to see the `responseType`?

westonpace (approved):
Just some minor nits

node/src/remote/client.ts
albertlockett: Probably yes -- the reason I didn't expose this for now was because this was an axios specific parameter and eventually we'll be probably be moving the client impl down into rust/reqwest

PR #1132, opened 2024-03-19
Order by field support FTS 

natcharacter:
This PR adds support for passing through a set of ordering fields at index time (unsigned ints that tantivity can use as fast_fields) that at query time you can sort your results on. This is useful for cases where you want to get related hits, i.e by keyword, but order those hits by some other score, such as popularity. 

I.e search for songs descriptions that match on "sad AND jazz AND 1920" and then order those by number of times played. Example usage can be seen in the fts tests.


python/python/lancedb/query.py
```diff
@@ -136,7 +137,9 @@ def create(
 
         if isinstance(query, str):
             # fts
-            return LanceFtsQueryBuilder(table, query)
+            return LanceFtsQueryBuilder(
```
natcharacter, line 140: this is only place we use it, we could assert if we hit another path and the ordering_field_name is not defined or log a warning. Or just ignore as we do here (kinda like the vector column name field). Not sure what you all prefer as a standard, so left like this.

python/python/lancedb/table.py
```diff
@@ -1103,21 +1104,37 @@ def create_fts_index(
...
         if index_exists:
             if not replace:
                 raise ValueError("Index already exists. Use replace=True to overwrite.")
             fs.delete_dir(path)
 
-        index = create_index(self._get_fts_index_path(), field_names)
-        populate_index(index, self, field_names, writer_heap_size=writer_heap_size)
+        index = create_index(
+            self._get_fts_index_path(),
+            field_names,
```
natcharacter, line 1128: you can have a multiple fields at index time and then dynamically choose which to sort on at query time.

python/python/tests/test_fts.py
```diff
@@ -79,6 +81,39 @@ def test_search_index(tmp_path, table):
     assert len(results[1]) == 10  # _distance
 
 
+def test_search_ordering_field_index_table(tmp_path, table):
+    table.create_fts_index("text", ordering_field_names="count")
+    rows = (
+        table.search("puppy", ordering_field_name="count")
+        .limit(20)
```
natcharacter, line 88: get a big number, chances of randomly sampling 12-20 rows in order are basically 0.

changhiskhan (approved):
lgtm thanks!

PR #1140, opened 2024-03-20
better error msg for query vector with wrong dim

QianZhu:


rust/lancedb/src/table.rs
```diff
@@ -1187,11 +1187,19 @@ impl NativeTable {
             })?;
             if !matches!(field.data_type(), arrow_schema::DataType::FixedSizeList(f, dim) if f.data_type().is_floating() && dim == query_vector.len() as i32)
             {
-                return Err(Error::Schema {
+                let schema_vec_dim =
+                    if let arrow_schema::DataType::FixedSizeList(_, dim) = field.data_type() {
+                        dim
+                    } else {
+                        unreachable!()
+                    };
```
westonpace, line 1195: This is a little hard to follow.  Maybe it would be easier now to just do..

```
if let arrow_schema::DataType::FixedSizeList(f, dim) = field.data_type()) {
  if !f.data_type.is_floating() {
    // Error, not a float
  } else if dim != query_vector.len() {
    // Error, not correct dimension
  }
} else {
  // Error, not a fixed-size-list type
}
```

westonpace (approved):
A minor readability suggestion but feel free to go on without it.  Thanks for working on this!

PR #1146, opened 2024-03-21
chore: validate table name

### LINKED ISSUE #1129: bug(python): Table name cannot have /
### LanceDB version

_No response_

### What happened?

I acidentally created a table name that has a `/` in it (huggingface standard). When you do that you get no error whatsoever when your inserting embeddings, but at query time the table anymore. That's because an extra folder is created. 

Probably best to catch this early, if a table name contains a slash maybe erroring out when the `db.create_table` is called?

_No response_

eddyxu:
Closes #1129


rust/lancedb/src/utils.rs
```diff
@@ -64,6 +82,21 @@ impl PatchReadParam for ReadParams {
...
+pub fn validate_table_name(name: &str) -> Result<()> {
+    if name.is_empty() {
+        return Err(Error::InvalidTableName {
+            name: name.to_string(),
+        });
+    }
+    if !TABLE_NAME_REGEX.is_match(name) {
+        return Err(Error::InvalidTableName {
+            name: name.to_string(),
+        });
```
wjones127, line 95: Maybe should give a reason here, so users know how to name their tables?
```suggestion
        return Err(Error::InvalidTableName {
            name: name.to_string(),
            reason: "Table names can only include letters, numbers, periods, and underscores.",
        });
```

rust/lancedb/src/utils.rs
```diff
@@ -64,6 +82,21 @@ impl PatchReadParam for ReadParams {
     }
 }
 
+/// Validate table name.
+pub fn validate_table_name(name: &str) -> Result<()> {
+    if name.is_empty() {
+        return Err(Error::InvalidTableName {
+            name: name.to_string(),
```
wjones127, line 89:
```suggestion
            name: name.to_string(),
            reason: "Table names cannot be empty strings",
```

rust/lancedb/src/utils.rs
```diff
@@ -1,12 +1,30 @@
...
 use arrow_schema::Schema;
-
 use lance::dataset::{ReadParams, WriteParams};
 use lance::io::{ObjectStoreParams, WrappingObjectStore};
+use lazy_static::lazy_static;
 
 use crate::error::{Error, Result};
 
+lazy_static! {
+    static ref TABLE_NAME_REGEX: regex::Regex = regex::Regex::new(r"^[a-zA-Z0-9_\.]+$").unwrap();
```
westonpace, line 25: No hyphen?

westonpace (approved):
This looks good.

rust/lancedb/src/utils.rs
eddyxu: Ok lemme add hyphen here

PR #1148, opened 2024-03-21
feat: ship fp16kernels in Python wheels

wjones127:
Same deal as https://github.com/lancedb/lance/pull/2098


westonpace (approved):
Thanks for figuring this out

PR #1149, opened 2024-03-22
Embedding_function: Colbert support added

raghavdixit99:
Added colbert embedding function, made a small change in testing script for easier invocation of newly added registries, unit test.

lint,format check passed


raghavdixit99:
not sure why ruff check fails here, after local ruff lint check as well via VScode ruff extention, the `import lancedb` block moves below , are we using an older verision of ruff in the actions workflow @AyushExel / @changhiskhan ?

raghavdixit99:
closing for now, will push a new PR

PR #1164, opened 2024-03-28
ci: fix name collision in npm artifacts for vectordb

### LINKED ISSUE #1163: bug(node): @ lancedb version 0.4.14 missing for a few platforms on NPM
### LanceDB version

v0.4.14

### What happened?

v0.4.14 depends on @ lancedb/<platform> version 0.4.14, which doesn't exist for many platforms, leading to an error on `npm install`. This will affect almost anyone who has installed `vectordb`, as "^0.4.x" will resolve to 0.4.14.

NPM search shows that only a couple of them have a version 0.4.14

Screenshot 2024-03-28 at 8 31 37 AM

Just downgrading to 0.4.12 is a perfectly fine solution for now.

And thanks for all the work that you do! When I found LanceDB it was exactly what we needed, and has played its role perfectly since then : )

_No response_

wjones127:
Fixes #1163

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Upload Darwin Artifacts
         uses: actions/upload-artifact@v4
         with:
-          name: node-native-darwin
+          name: node-native-darwin-${{ matrix.config.arch }}
           path: |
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Ah, thanks :)  this is because I changed `upload-artifact` to `v4`.  My bad

PR #1169, opened 2024-03-30
feat: add filterable countRows to remote API

QianZhu:


node/src/remote/index.ts
```diff
@@ -454,8 +454,10 @@ export class RemoteTable<T = number[]> implements Table<T> {
     }
   }
 
-  async countRows (): Promise<number> {
```
eddyxu, line 457: What is the local native API? can we make them consistent. 

Same in python / rust as well.

eddyxu: Could you add one comment that rust API has already created?

eddyxu (approved):

node/src/remote/index.ts
QianZhu: https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/remote/table.rs#L64

PR #1178, opened 2024-04-02
chore: pass str instead of String to build table names

eddyxu:


rust/lancedb/src/connection.rs
```diff
@@ -101,8 +101,8 @@ impl TableNamesBuilder {
     ///
     /// This can be combined with limit to implement pagination by setting this to
     /// the last table name from the previous page.
-    pub fn start_after(mut self, start_after: String) -> Self {
-        self.start_after = Some(start_after);
+    pub fn start_after(mut self, start_after: &str) -> Self {
+        self.start_after = Some(start_after.to_string());
```
westonpace, line 105: I'm a little confused.  What is the point if you are just going to turn around and call `to_string`?

westonpace: Maybe accept `impl Into<String>` instead?

eddyxu: Ok, changed to `impl Into<String>`

westonpace (approved):

Issue #1172, opened 2024-03-30
bug(python): More descriptive error for Unknown error: 'generator' object has no attribute 'to_batches'


tattyd:
### What happened?

Hi,

If you try to add a single class instance inheriting from LanceModel, say:

```
        tbl = self.db.open_table(table_name)
        tbl.add(
            ClassInheritsFromLanceModel(
                ...
            )
         )
```

then you get:

```OSError: LanceError(Arrow): C Data interface error: Unknown error: 'generator' object has no attribute 'to_batches'. Detail: Python exception: AttributeError, [/Users/runner/work/lance/lance/rust/lance-datafusion/src/utils.rs:38:28](https://file+.vscode-resource.vscode-cdn.net/Users/runner/work/lance/lance/rust/lance-datafusion/src/utils.rs:38:28)```

The message is a bit cryptic; the actual fix is to simply pass a list instead:

```
        tbl = self.db.open_table(table_name)
        tbl.add(
            [
                 ClassInheritsFromLanceModel(
                     ...
                 )
            ]
         )
```

It would be great to get a less cryptic error message, thanks! :)

### Are there known steps to reproduce?

See above

Issue #1124, opened 2024-03-17
FTS search index crashes as lance cannot infer vector field


natcharacter:
### What happened?

If you have an index with only FTS and no vectors, searching without passing through an vector_column_name crashes. However, it works fine if you pass in an empty string for 

### Are there known steps to reproduce?

Create any index with only an FTS index and no vector index. Then run:

`table.search(query="hi",query_type="fts").to_list()`

```
You'll get the following traceback.:
Traceback (most recent call last):
 File "/app/test.py", line 5, in <module>
 table.search(query="hi",query_type="fts").to_list()
 File "/usr/local/lib/python3.10/site-packages/lancedb/table.py", line 1329, in search
  vector_column_name = inf_vector_column_query(self.schema)
 File "/usr/local/lib/python3.10/site-packages/lancedb/util.py", line 189, in inf_vector_column_query
  raise ValueError(
ValueError: There is no vector column in the data. Please specify the vector column name for vector search
```

Passing vector_column_name='' fixes it, so easy workaround for now, but is a bit confusing.

changhiskhan:
Should be a quick fix. I'll see if I can throw up a PR tonight

PR #1195, opened 2024-04-05
python: bug fixes, ruff fix

### LINKED ISSUE #1194: bug(python): imagebind-packaged==0.1.2 contradicts with torch==1.13
### LanceDB version

_No response_

### What happened?

Error downloading the dev package, need to fix the root package so removing the imagebind dependancy from init should work.

```python
collecting imagebind-packaged==0.1.2
  Using cached imagebind-packaged-0.1.2.tar.gz (1.4 MB)
  Preparing metadata (setup.py) ... done
INFO: pip is looking at multiple versions of imagebind-packaged to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement torch==1.13.0 (from imagebind-packaged) (from versions: 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.1.2, 2.2.0, 2.2.1, 2.2.2)
ERROR: No matching distribution found for torch==1.13.0

```

_No response_

raghavdixit99:
closes #1194 #1172 #1124

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -1,0 +1,5063 @@
+{
+  "name": "vectordb",
+  "version": "0.4.16",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "vectordb",
+      "version": "0.4.16",
+      "cpu": [
+        "x64",
+        "arm64"
+      ],
+      "license": "Apache-2.0",
+      "os": [
+        "darwin",
+        "linux",
+        "win32"
+      ],
+      "dependencies": {
+        "@neon-rs/load": "^0.0.74",
+        "axios": "^1.4.0"
+      },
+      "devDependencies": {
+        "@neon-rs/cli": "^0.0.160",
+        "@types/chai": "^4.3.4",
+        "@types/chai-as-promised": "^7.1.5",
+        "@types/mocha": "^10.0.1",
+        "@types/node": "^18.16.2",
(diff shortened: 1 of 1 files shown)


python/python/lancedb/table.py
```diff
@@ -1409,7 +1411,13 @@ def search(
...
         """
         if vector_column_name is None and query is not None:
-            vector_column_name = inf_vector_column_query(self.schema)
+            if query_type == "fts":
+                try:
+                    vector_column_name = inf_vector_column_query(self.schema)
+                except ValueError:
+                    vector_column_name = ""
+            else:
+                vector_column_name = inf_vector_column_query(self.schema)
```
wjones127, line 1420: Could you add a unit test for #1124?

Also if query type is `"fts"`, why do we even need a vector column? Does it sometime imply there is also vector search at the same time?

python/python/lancedb/table.py
```diff
@@ -95,6 +95,8 @@ def _sanitize_data(
             data = _to_record_batch_generator(
                 data.data.to_batches(), schema, metadata, on_bad_vectors, fill_value
             )
+    if isinstance(data, LanceModel):
+        raise TypeError("Pass the data as a list : tbl.add([LanceModel_1(),..])")
```
wjones127, line 99: I'd also recommend adding a test that we get the expected error here. That way it doesn't get broken later when someone refactors this.

PR #1204, opened 2024-04-05
feat: expose storage options in LanceDB

### LINKED ISSUE #114: Add documentation on auth with S3/GCS
- Private S3/GCS authentication docs
- Details on design, in particular, how this works with Lance as well

### LINKED ISSUE #281: Implement common credential provider interface in JS and Python
we should be able to pass creds for AWS, GCP, Azure, etc. without breaking FFI. Currently this is not true because we use positional args for AWS creds in the JS FFI. We should fix that

### LINKED ISSUE #439: Move Custom AWS Credentials into rust core
Users can pass custom AWS credentials to connect to s3

```javascript
lancedb.connect({ uri: 's3://bucked', awsCredentials: { accessKeyId: ... }
```

Currently we store the credentials in javascript and pass them to the FFI bridge on each call. We can store them in the rust side instead, and make sure we use it for all write operations (create add, delete, createIndex,...)

### LINKED ISSUE #642: dynamically set AWS region
There was already a PR for dynamically changing AWS region when running lancedb.connect(). However, the current implementation does not work. I can only change the region if the env variable is changed.

### LINKED ISSUE #897: Feature: AWS KMS support
### SDK

Node

### Description

Support user provided KMS key per database or globally.

### LINKED ISSUE #990: Feature: add GCS credentials to `vectordb::connection::ConnectOptions`
### SDK

Rust

### Description

There are currently fields for specifying AWS or Lance Cloud credentials but there is no field for specifying GCS credentials.

### LINKED ISSUE #1082: Automatically detect region of S3 buckets
### LanceDB version

_No response_

### What happened?

I can touch a file with other systems

```python
import s3fs

s3 = s3fs.S3FileSystem()
s3.touch("s3://my-bucket/dir/myfile")  # this works fine
```

When I use LanceDB I get an issue:

```python
import lancedb
db = lancedb.connect(
    "s3://my-bucket/dir/", 
)
tbl = db.create_table("tablename", data=df)
```

```python-traceback
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In[10], line 5
      1 import lancedb
      2 db = lancedb.connect(
      3     "s3://...", 
      4 )
----> 5 tbl = db.create_table("arxiv", data=df)

File ~/mambaforge/envs/rag/lib/python3.11/site-packages/lancedb/db.py:391, in LanceDBConnection.create_table(self, name, data, schema, mode, exist_ok, on_bad_vectors, fill_value, embedding_functions)
    388 if mode.lower() not in ["create", "overwrite"]:
    389     raise ValueError("mode must be either 'create' or 'overwrite'")
--> 391 tbl = LanceTable.create(
    392     self,
    393     name,
    394     data,
    395     schema,
    396     mode=mode,
    397     exist_ok=exist_ok,
    398     on_bad_vectors=on_bad_vectors,
    399     fill_value=fill_value,
    400     embedding_functions=embedding_functions,
    401 )
    402 return tbl

File ~/mambaforge/envs/rag/lib/python3.11/site-packages/lancedb/table.py:1438, in LanceTable.create(cls, db, name, data, schema, mode, exist_ok, on_bad_vectors, fill_value, embedding_functions)
   1436 empty = pa.Table.from_pylist([], schema=schema)
   1437 try:
-> 1438     lance.write_dataset(empty, tbl._dataset_uri, schema=schema, mode=mode)
   1439 except OSError as err:
   1440     if "Dataset already exists" in str(err) and exist_ok:

File ~/mambaforge/envs/rag/lib/python3.11/site-packages/lance/dataset.py:2381, in write_dataset(data_obj, uri, schema, mode, max_rows_per_file, max_rows_per_group, max_bytes_per_file, commit_lock, progress)
   2378     params["commit_handler"] = commit_lock
   2380 uri = os.fspath(uri) if isinstance(uri, Path) else uri
-> 2381 _write_dataset(reader, uri, params)
   2382 return LanceDataset(uri)

OSError: LanceError(IO): Generic S3 error: Error performing list request: Received redirect without LOCATION, this normally indicates an incorrectly configured region, /Users/runner/work/lance/lance/rust/lance-table/src/io/commit.rs:100:26
```

This bucket already exists so it seems strange that I need to specify the region again.  AWS already has this information, and if for some reason we needed it we could query it from AWS.  But ok, I'll specify the region.  I see that there is a convenient `region=` keyword

```python
db = lancedb.connect(
    "s3://my-bucket/dir/", 
    region="us-west-2",
)
tbl = db.create_table("tablename", data=df)
```

Sadly, I get the same error.  I then read through https://github.com/lancedb/lancedb/issues/712 and see that maybe I need to specify region in an environment variable

```python
import os
os.environ["AWS_DEFAULT_REGION"] = "us-west-2"
```

And now things work.

Personally I can do this, but it makes it hard to include LanceDB in a recommended workflow to downstream users.  Is this easy to fix? or is this some downstream library's issue? I (we?) can go complain there if preferred.

_No response_

### LINKED ISSUE #1165: Feature: support passing in storage args in `connect`
### SDK

Python

### Description

For example, it is not possible to programmatically set the S3 region.  object_store can (I think) accept a `HashMap<str, str>` and we should provide that same interface to our users.

### LINKED ISSUE #1168: bug(rust): AWS Credentials are not passed to underlying object_store
### LanceDB version

0.4.14

### What happened?

Various behaviors lead me to believe that an `AwsCredential` provided to a `ConnectBuilder` does not actually get propagated to the correct places. When aws credentials are provided using environment variables, everything works as expected.

I suspect you are calling `object_store::aws::AmazonS3Builder::from_env()` somewhere you shouldn't be.

```
    let db = lancedb::connect("s3://my-bucket/lance").region("us-west-1").aws_creds(
        AwsCredential {
            key_id: KEY_ID.to_string(),
            secret_key: SECRET_KEY.to_string(),
            token: Some(TOKEN.to_string()),
        }).execute().await.unwrap();

    let table = db.open_table("all_rides").execute().await.unwrap();
```

This yields a `TableNotFound { name: "all_rides" }`. However, if I set environment variables `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN` then it works correctly.

In another case, I tried to create a new table after providing an `AwsCredential` to a `ConnectBuilder`, and it explicitly complained about not having a credential provider.

wjones127:
Exposes `storage_options` in LanceDB. This is provided for Python async, Node `lancedb`, and Node `vectordb` (and Rust of course). Python synchronous is omitted because it's not compatible with the PyArrow filesystems we use there currently. In the future, we will move the sync API to wrap the async one, and then it will get support for `storage_options`.

1. Fixes #1168
2. Closes #1165
3. Closes #1082
4. Closes #439
5. Closes #897
6. Closes #642
7. Closes #281
8. Closes #114
9. Closes #990
10. Deprecating `awsCredentials` and `awsRegion`. Users are encouraged to use `storageOptions` instead.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -107,6 +107,7 @@ jobs:
       AWS_ENDPOINT: http://localhost:4566
       # this one is for dynamodb
       DYNAMODB_ENDPOINT: http://localhost:4566
+      ALLOW_HTTP: true
     steps:
     - uses: actions/checkout@v4
       with:
--- a/docs/src/guides/storage.md
+++ b/docs/src/guides/storage.md
@@ -340,6 +340,7 @@ GCS credentials are configured by setting the `GOOGLE_SERVICE_ACCOUNT` environme
 
 === "Python"
 
+    <!-- skip-test -->
     ```python
     import lancedb
     db = await lancedb.connect_async(
@@ -386,6 +387,7 @@ Azure Blob Storage credentials can be configured by setting the `AZURE_STORAGE_A
 
 === "Python"
 
+    <!-- skip-test -->
     ```python
     import lancedb
     db = await lancedb.connect_async(
--- a/docs/src/guides/storage.md
+++ b/docs/src/guides/storage.md
@@ -55,18 +55,116 @@ LanceDB OSS supports object stores such as AWS S3 (and compatible stores), Azure
     const db = await lancedb.connect("az://bucket/path");
     ```
 
-In most cases, when running in the respective cloud and permissions are set up correctly, no additional configuration is required. When running outside of the respective cloud, authentication credentials must be provided using environment variables. In general, these environment variables are the same as those used by the respective cloud SDKs. The sections below describe the environment variables that can be used to configure each object store.
+In most cases, when running in the respective cloud and permissions are set up correctly, no additional configuration is required. When running outside of the respective cloud, authentication credentials must be provided. Credentials and other configuration options can be set in two ways: first, by setting environment variables. And second, by passing a `storage_options` object to the `connect` function. For example, to increase the request timeout to 60 seconds, you can set the `TIMEOUT` environment variable to `60s`:
 
-LanceDB OSS uses the [object-store](https://docs.rs/object_store/latest/object_store/) Rust crate for object store access. There are general environment variables that can be used to configure the object store, such as the request timeout and proxy configuration. See the [object_store ClientConfigKey](https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html) doc for available configuration options. The environment variables that can be set are the snake-cased versions of these variable names. For example, to set `ProxyUrl` use the environment variable `PROXY_URL`. (Don't let the Rust docs intimidate you! We link to them so you can see an up-to-date list of the available options.)
+```bash
+export TIMEOUT=60s
+```
+
+If you only want this to apply to one particular connection, you can pass the `storage_options` object to the `connect` function:
+
+=== "Python"
+
+    ```python
+    import lancedb
+    db = lancedb.connect("s3://bucket/path",
+                         storage_options={"timeout": "60s"})
+    ```
+
+=== "JavaScript"
+
+    ```javascript
+    const lancedb = require("lancedb");
+    const db = await lancedb.connect("s3://bucket/path",
+                                     {timeout: "60s"});
+    ```
+
+Getting even more specific, you can set the `timeout` for only a particular table:
+
+=== "Python"
+
+    ```python
+    import lancedb
+    db = lancedb.connect("s3://bucket/path")
+    table = db.open_table("table", storage_options={"connect_timeout": "60s"})
+    ```
+
+=== "JavaScript"
+
+    ```javascript
+    const lancedb = require("lancedb");
+    const db = await lancedb.connect("s3://bucket/path");
+    const table = db.openTable("table", {timeout: "60s"});
+    ```
+
+### General configuration
+
+There are several options that can be set for all object stores, mostly related to network client configuration.
+
+<!-- from here: https://docs.rs/object_store/latest/object_store/enum.ClientConfigKey.html -->
+
+| Key                        | Description                                                                                      |
+|----------------------------|--------------------------------------------------------------------------------------------------|
+| `allow_http`               | Allow non-TLS, i.e. non-HTTPS connections. Default: `False`.                                      |
+| `allow_invalid_certificates`| Skip certificate validation on HTTPS connections. Default: `False`.                               |
+| `connect_timeout`          | Timeout for only the connect phase of a Client. Default: `5s`.                                    |
+| `timeout`                  | Timeout for the entire request, from connection until the response body has finished. Default: `30s`. |
+| `user_agent`               | User agent string to use in requests.                                                             |
+| `proxy_url`                | URL of a proxy server to use for requests. Default: `None`.                                       |
+| `proxy_ca_certificate`     | PEM-formatted CA certificate for proxy connections.                                                |
+| `proxy_excludes`           | List of hosts that bypass the proxy. This is a comma-separated list of domains and IP masks. Any subdomain of the provided domain will be bypassed. For example, `example.com, 192.168.1.0/24` would bypass `https://api.example.com`, `https://www.example.com`, and any IP in the range `192.168.1.0/24`. |
 
 
 ### AWS S3
 
-To configure credentials for AWS S3, you can use the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` environment variables.
+To configure credentials for AWS S3, you can use the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` keys.
+These can be set as environment variables or passed in the `storage_options` parameter:
+
+=== "Python"
+
+    ```python
+    import lancedb
+    db = lancedb.connect(
+        "s3://bucket/path",
+        storage_options={
+            "aws_access_key_id": "my-access-key",
+            "aws_secret_access_key": "my-secret-key",
+            "aws_session_token": "my-session-token",
+        }
+    )
+    ```
+
+=== "JavaScript"
+
+    ```javascript
+    const lancedb = require("lancedb");
+    const db = await lancedb.connect(
+        "s3://bucket/path",
+        {
+            aws_access_key_id: "my-access-key",
+            aws_secret_access_key: "my-secret-key",
+            aws_session_token: "my-session-token",
+        }
+    );
+    ```
 
 Alternatively, if you are using AWS SSO, you can use the `AWS_PROFILE` and `AWS_DEFAULT_REGION` environment variables.
 
-You can see a full list of environment variables [here](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.from_env).
+The following keys can be used as both environment variables or keys in the `storage_options` parameter:
+
+| Key                                | Description                                                                                          |
+|------------------------------------|------------------------------------------------------------------------------------------------------|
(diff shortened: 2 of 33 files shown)


nodejs/lancedb/connection.ts
```diff
@@ -13,10 +13,32 @@
...
+ */
+export async function connect(
+  uri: string,
+  opts?: Partial<ConnectionOptions>,
+): Promise<Connection> {
+  opts = opts ?? {};
+  opts.storageOptions = cleanseStorageOptions(opts.storageOptions);
+  const nativeConn = await LanceDbConnection.new(uri, opts);
+  return new Connection(nativeConn);
+}
```
wjones127, line 40: Moved this function to get access to `cleanseStorageOptions`, but it also felt appropriate that it was here in the `connection`. It's still exported in the same place.

.github/workflows/node.yml
```diff
@@ -107,6 +107,7 @@ jobs:
       AWS_ENDPOINT: http://localhost:4566
       # this one is for dynamodb
       DYNAMODB_ENDPOINT: http://localhost:4566
+      ALLOW_HTTP: true
```
wjones127, line 110: We stopped hardcoding allowing unsafe connections to S3 in https://github.com/lancedb/lance/pull/2147, so have to set this explicitly now.

rust/ffi/node/src/lib.rs
wjones127, line 1: A lot of this could be deleted because we now handle this through `storage_options`.

wjones127, line 1: A lot of this could be deleted because we now handle this through `storage_options`.

docs/src/guides/storage.md
```diff
@@ -55,18 +55,139 @@ LanceDB OSS supports object stores such as AWS S3 (and compatible stores), Azure
...
+```bash
+export TIMEOUT=60s
+```
+
+!!! note "`storage_options` availability"
+
+    The `storage_options` parameter is only available in Python *async* API and JavaScript API.
+    It is not yet supported in the Python synchronous API.
+
+If you only want this to apply to one particular connection, you can pass the `storage_options` object to the `connect` function:
```
westonpace, line 69:
```suggestion
If you only want this to apply to one particular connection, you can pass the `storage_options` object when
opening a connection:
```
One method is named `connect` and the other is `connect_async` so maybe better to avoid the method name entirely (it's in the example).

docs/src/guides/storage.md
```diff
@@ -55,18 +55,139 @@ LanceDB OSS supports object stores such as AWS S3 (and compatible stores), Azure
...
+| `proxy_url`                | URL of a proxy server to use for requests. Default: `None`.                                       |
+| `proxy_ca_certificate`     | PEM-formatted CA certificate for proxy connections.                                                |
+| `proxy_excludes`           | List of hosts that bypass the proxy. This is a comma-separated list of domains and IP masks. Any subdomain of the provided domain will be bypassed. For example, `example.com, 192.168.1.0/24` would bypass `https://api.example.com`, `https://www.example.com`, and any IP in the range `192.168.1.0/24`. |
 
 
 ### AWS S3
 
-To configure credentials for AWS S3, you can use the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` environment variables.
+To configure credentials for AWS S3, you can use the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` keys.
+These can be set as environment variables or passed in the `storage_options` parameter:
```
westonpace, line 142: is `region` mandatory?  Or is that only when using an s3-compatible store?

docs/src/guides/storage.md
```diff
@@ -146,22 +267,174 @@ For **read-only access**, LanceDB will need a policy such as:
...
 
+
+The following keys can be used as both environment variables or keys in the `storage_options` parameter:
+<!-- source: https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html -->
+
+| Key                                   | Description                                  |
+|---------------------------------------|----------------------------------------------|
+| ``google_service_account``            | Path to the service account JSON file.       |
+| ``google_service_account_key``        | The serialized service account key.          |
+| ``google_application_credentials``    | Path to the application credentials.         |
```
westonpace, line 381: Should we mention here that `google_service_account` and `service_account` are the same thing?

nodejs/eslint.config.js
```diff
@@ -13,7 +13,39 @@ module.exports = tseslint.config(
   ...tseslint.configs.recommended,
   {
     rules: {
-      "@typescript-eslint/naming-convention": "error",
+      "@typescript-eslint/naming-convention": [
```
westonpace, line 16: This works though I think you could have just disabled linting on that entire file.

python/python/lancedb/table.py
```diff
@@ -1424,6 +1424,7 @@ def create(
         on_bad_vectors: str = "error",
         fill_value: float = 0.0,
         embedding_functions: List[EmbeddingFunctionConfig] = None,
+        storage_options=None,
```
westonpace, line 1427: Is this needed?  I though the sync API didn't have storage_options?

python/src/connection.rs
```diff
@@ -85,55 +85,70 @@ impl Connection {
         future_into_py(self_.py(), async move { op.execute().await.infer_error() })
     }
 
+    #[pyo3(signature = (name, mode, data, storage_options = None))]
```
westonpace, line 88: Is this needed?  Is it for the default value?  I think, in other calls, we put defaults in the `.py` and don't bother with them in the `.rs`

python/src/connection.rs
```diff
@@ -85,55 +85,70 @@ impl Connection {
...
-                .mode(mode)
-                .execute()
-                .await
-                .infer_error()?;
+            let table = builder.execute().await.infer_error()?;
             Ok(Table::new(table))
         })
     }
 
+    #[pyo3(signature = (name, mode, schema, storage_options = None))]
```
westonpace, line 113: Ditto, is this needed?

rust/ffi/node/src/lib.rs
westonpace: Could be deleted or is deleted?  Looks like you've deleted quite a bit.

rust/lancedb/Cargo.toml
```diff
@@ -46,8 +46,13 @@ tempfile = "3.5.0"
...
+aws-sdk-kms = { version = "1.0" }
+aws-config = { version = "1.0" }
 
 [features]
 default = ["remote"]
 remote = ["dep:reqwest"]
-fp16kernels = ["lance-linalg/fp16kernels"]
\ No newline at end of file
+fp16kernels = ["lance-linalg/fp16kernels"]
+integration-test = []
```
westonpace, line 58: Maybe `store-test` instead of `integration-test` since "integration test" is a pretty ambiguous term?

rust/lancedb/src/connection.rs
```diff
@@ -194,6 +237,56 @@ impl CreateTableBuilder<false, NoData> {
         }
     }
 
+    /// Apply the given write options when writing the initial data
```
westonpace, line 240: Can we expand on this to describe situations where this might be needed?  Or maybe keep it private?

westonpace (approved):
This is an incredible set of tests!

docs/src/guides/storage.md
wjones127: It used to be, but with the latest version of Lance we added automatic bucket region resolution. It's still required for s3-compatible ones, though.

nodejs/eslint.config.js
wjones127: Oh okay. That's a better idea. Thanks.

python/python/lancedb/table.py
wjones127: Ah sorry it shouldn't be in here. Thanks.

rust/ffi/node/src/lib.rs
wjones127: Uh, well could and did delete it. Just explaining why it's okay that we are deleting it all.

rust/lancedb/src/connection.rs
wjones127: I guess it's not necessary. I can remove it.

Issue #1208, opened 2024-04-07
bug(python): pydantic warning upon import lancedb


changhiskhan:
### What happened?

<img width="989" alt="image" src="https://github.com/lancedb/lancedb/assets/759245/6a5d6cd6-6a15-46e7-9881-983c55bdfe3e">

(from local install off of top of main. With pydantic 2.6.4)

raghavdixit99:
my fork of the latest pull gives me - 

`File [~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:13](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:13)
      [1](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:1) #  Copyright 2023 LanceDB Developers
      [2](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:2) #
      [3](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:3) #  Licensed under the Apache License, Version 2.0 (the "License");
   (...)
     [11](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:11) #  See the License for the specific language governing permissions and
     [12](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:12) #  limitations under the License.
---> [13](https://file+.vscode-resource.vscode-cdn.net/Users/raghavdixit/Desktop/open_source/lancedb/python/example/~/Desktop/open_source/lancedb/python/python/lancedb/utils/__init__.py:13) from .config import Config
[...]

PR #1210, opened 2024-04-08
python: Bug fixes / tests

### LINKED ISSUE #1194: bug(python): imagebind-packaged==0.1.2 contradicts with torch==1.13

raghavdixit99:
closes #1194 #1172 #1124 #1208 

@wjones127  : `if query_type != "fts":` is needed because both fts and vector search create `LanceQueryBuilder` which has `vector_column_name` as a required attribute.


python/python/tests/test_db.py
```diff
@@ -28,20 +28,33 @@ def test_basic(tmp_path):
...
         "test",
         data=[
             {"vector": [3.1, 4.1], "item": "foo", "price": 10.0},
             {"vector": [5.9, 26.5], "item": "bar", "price": 20.0},
         ],
+        schema=SimpleModel,
     )
+    try:
+        table.add(SimpleModel(item="baz", price=30.0, vector=[1.0, 2.0]))
+    except Exception as e:
```
AyushExel, line 46: Probably better to use pytest.raises(ValueError) to test for exceptions

python/python/lancedb/table.py
```diff
@@ -1336,7 +1339,7 @@ def embedding_functions(self) -> dict:
     def search(
         self,
         query: Optional[Union[VEC, str, "PIL.Image.Image", Tuple]] = None,
-        vector_column_name: Optional[str] = None,
+        vector_column_name: Optional[str] = "",
```
AyushExel, line 1342: Any specific reason to change None to "" ?

AyushExel:
Seems like you added the extra .ipynb file by mistake?

python/python/lancedb/table.py
raghavdixit99: no reason just to avoid passing it again, can change as well

raghavdixit99:
yes, I guess it got added by mistake, updating the PR with changes.

AyushExel:
please fix tests

raghavdixit99:
Locally it runs, tantivy isnt isntalled thats the ci issue

.github/workflows/run_tests/action.yml
```diff
@@ -11,7 +11,7 @@ runs:
     - name: Install lancedb
       shell: bash
       run: |
-        pip3 install $(ls target/wheels/lancedb-*.whl)[tests,dev]
+        pip3 install $(ls target/wheels/lancedb-*.whl)[tests,dev] tantivy
```
wjones127, line 14: We are selecting the `tests` and `dev` dependencies are part of the install. Could you instead add `tantivy` to `tests` dependencies?
```suggestion
        pip3 install $(ls target/wheels/lancedb-*.whl)[tests,dev]
```

It would go here:
https://github.com/lancedb/lancedb/blob/8a1227030a27dd229c2492e534682d4eae1f3cf4/python/pyproject.toml#L50-L59

python/python/lancedb/table.py
wjones127: hmm. IMO it's weird that we use empty string and not None. But that seems to be a pre-existing issue with the other APIs, so this is fine for now.

python/python/tests/test_db.py
```diff
@@ -28,20 +28,33 @@ def test_basic(tmp_path):
...
     rs = table.search([100, 100]).limit(1).to_pandas()
     assert len(rs) == 1
     assert rs["item"].iloc[0] == "bar"
 
-    rs = table.search([100, 100]).where("price < 15").limit(2).to_pandas()
+    table.create_fts_index(["item"])
+    rs = table.search("bar", query_type="fts").to_pandas()
     assert len(rs) == 1
-    assert rs["item"].iloc[0] == "foo"
+    assert rs["item"].iloc[0] == "bar"
```
wjones127, line 57: Why did you replace the vector search query? Can you instead make the FTS query an additional query in this test?

wjones127 (changes requested):

wjones127 (approved):

PR #1215, opened 2024-04-10
ci: fix failures in release scripts

wjones127:
* Python release has been running when we create a Node release. https://github.com/lancedb/lancedb/actions/runs/8635662585
* Rust is missing new enough compilers to check the kernels feature https://github.com/lancedb/lancedb/actions/runs/8635662578


westonpace (approved):
Thanks!

PR #1234, opened 2024-04-22
add rename_table fn and more data for index_stats to return

QianZhu:
1. added rename_table fn to enable dashboard to rename a table
2. added index_type and distance_type (for vector index) to index_stats so that more detailed data can be shown on the table page.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -1061,6 +1061,26 @@ impl NativeTable {
         }
     }
 
+    pub async fn get_index_type(&self, index_uuid: &str) -> Result<Option<String>> {
+        match self.load_index_stats(index_uuid).await? {
+            Some(stats) => Ok(Some(stats.index_type)),
+            None => Ok(None),
+        }
+    }
+
+    pub async fn get_distance_type(&self, index_uuid: &str) -> Result<Option<String>> {
+        match self.load_index_stats(index_uuid).await? {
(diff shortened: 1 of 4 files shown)


rust/lancedb/src/table.rs
```diff
@@ -1061,6 +1061,26 @@ impl NativeTable {
         }
     }
 
+    pub async fn get_index_type(&self, index_uuid: &str) -> Result<Option<String>> {
```
rpgreen, line 1064: Does this need to return Result if the result is always Ok?

rust/lancedb/src/table.rs
```diff
@@ -1061,6 +1061,26 @@ impl NativeTable {
...
+        match self.load_index_stats(index_uuid).await? {
+            Some(stats) => Ok(Some(stats.index_type)),
+            None => Ok(None),
+        }
+    }
+
+    pub async fn get_distance_type(&self, index_uuid: &str) -> Result<Option<String>> {
```
rpgreen, line 1071: Same comment

rpgreen (approved):

rust/lancedb/src/table.rs
QianZhu: good point. There some other functions that are having similar issues. We can fix them all together.

QianZhu:
this PR didn't cause the failed test in object_store.rs. The test will be fixed later

Issue #1099, opened 2024-03-13
Feature: Rust polars import / export


westonpace:
### SDK

Rust

### Description

In the Rust SDK we should be able to go to/from polars more easily.  For example, something like the following should work when adding data:

```
let my_df: DataFrame = ...
tbl.add(data_source::from_polars(my_df)).execute();
```

The above should also work with `LazyFrame`, although, for today at least, this will probably involve the consumption / execution of the `LazyFrame`:

```
let my_df: LazyFrame = ...
tbl.add(data_source::from_polars_lazy(my_df).execute();
```

Also note that `DataSource` doesn't exist yet.  In my brainstorming its just a module with a bunch of static methods (enabled by feature flags) for creating `Box<dyn RecordBatchReader + Send>` from various inputs.  As a result, this technique should also be usable in other methods that need data (e.g. `create_table`, `merge_insert`)

kkang2097:
I can take this

kkang2097:
okay i'm scoping some of this out so we get a nice conversion. forgive my slowness, still learning Rust

kkang2097:
Wait we should close this issue. If we need a deeper integration with Polars, I'm willing to start on that instead.

The solution here would just be:
Type conversions to LanceDB

If we made the deeper integration:
Conversions should be native to Lance

westonpace:
> Wait we should modify this issue or work on the Lance repo instead.

Now that add is based on a trait we don't even need the `data_source::from_polars` part.  We should be able to do something like:

```
let my_df: DataFrame = ...
tbl.add(my_df).execute();
```

This will require adding something like...

```
#[cfg(feature = polars)]
impl IntoArrow for polars::DataFrame {
  ...
}
```

> The solution here would just be:
Type conversions to LanceDB

I think we just need polars <-> arrow which should already be a part of polars.

> If we need a deeper integration with Polars, I'm willing to start on that instead.
> If we made the deeper integration:
Conversions should be native to Lance

What do you imagine this would look like?  What is your goal with a deeper integration?  Can you give some example use cases?

westonpace:
> Agh, have been pulled away with other projects. Sorry, will have to hand this off to someone else

No sweat, I'm a bit tied down with encodings at the moment but will try and snag this next week.

rohitrastogi:
Can I take a stab at this @westonpace?

westonpace:
> Can I take a stab at this @westonpace?

[...]

PR #1235, opened 2024-04-22
Implement convertors to and from Polars DataFrames in Rust SDK using polars-arrow high level convertors #1099

rohitrastogi:
Took a stab at #1099, but I have a few questions/concerns:
1. Because of the [blanket implementation](https://github.com/lancedb/lancedb/blob/431f94e5644936e67298e81285c76d8fd1858d29/rust/lancedb/src/arrow.rs#L117) of IntoArrow for any type that implements RecordBatchReader, it is not possible to implement IntoArrow directly for Polar's DataFrame struct. Because of this, I had to create a new type (`PolarsDataFrameRecordBatchReader`) that consumed the DataFrame and implement the trait for this type. This does not seem ideal as users will not be able to directly called `into_arrow()` directly on their Polars DataFrame. Please advise on the best way to address this. If we remove the blanket implementation, we will have to explicitly box things that are already record batch readers, which also isn't too ergonomic.

2. My implementation of `into_polars `requires calling `try_collect()` on `arrow::SendableRecordBatchStream`. `Into_polars` is fallible due to any potential errors that occur in the LanceDB query OR due to any errors that occur in the conversion process. I wasn't sure the best way to surface that union to the user (creating a new Error enum specifically for this function didn't seem right), so I decided to extend the existing LanceDB error enum with a variant for Polars Errors behind a feature flag. This seems like bad practice too as it requires me to mark the enum as non-exhaustive, which is a breaking change. This is to handle cases where crate A depends on LanceDB, but does not enable the polars feature but crate B, which depends on crate A and LanceDB, does enable the polars feature, preventing crate A from compiling. 

3. What is the purpose of SimpleRecordBatchReader and RecordBatchReader defined [here](https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/arrow.rs#L24-L52)? I don't see it referenced. When would we have a RecordBatchReader that uses the LanceDB error type and not the arrow error type?

Thanks!

### DIFF
--- a/python/src/error.rs
+++ b/python/src/error.rs
@@ -35,21 +35,15 @@ impl<T> PythonErrorExt<T> for std::result::Result<T, LanceError> {
         match &self {
             Ok(_) => Ok(self.unwrap()),
             Err(err) => match err {
-                LanceError::InvalidInput { .. } => self.value_error(),
-                LanceError::InvalidTableName { .. } => self.value_error(),
-                LanceError::TableNotFound { .. } => self.value_error(),
-                LanceError::Schema { .. } => self.value_error(),
+                LanceError::InvalidInput { .. }
+                | LanceError::InvalidTableName { .. }
+                | LanceError::TableNotFound { .. }
+                | LanceError::Schema { .. } => self.value_error(),
                 LanceError::CreateDir { .. } => self.os_error(),
-                LanceError::TableAlreadyExists { .. } => self.runtime_error(),
                 LanceError::ObjectStore { .. } => Err(PyIOError::new_err(err.to_string())),
-                LanceError::Lance { .. } => self.runtime_error(),
-                LanceError::Runtime { .. } => self.runtime_error(),
-                LanceError::Http { .. } => self.runtime_error(),
-                LanceError::Arrow { .. } => self.runtime_error(),
                 LanceError::NotSupported { .. } => {
                     Err(PyNotImplementedError::new_err(err.to_string()))
                 }
-                LanceError::Other { .. } => self.runtime_error(),
                 _ => self.runtime_error(),
             },
         }
--- a/python/src/error.rs
+++ b/python/src/error.rs
@@ -50,6 +50,7 @@ impl<T> PythonErrorExt<T> for std::result::Result<T, LanceError> {
                     Err(PyNotImplementedError::new_err(err.to_string()))
                 }
                 LanceError::Other { .. } => self.runtime_error(),
+                _ => self.runtime_error(),
             },
         }
     }
--- a/rust/lancedb/src/arrow.rs
+++ b/rust/lancedb/src/arrow.rs
@@ -201,12 +201,12 @@ impl arrow_array::RecordBatchReader for PolarsDataFrameRecordBatchReader {
 /// chunks are not guaranteed to be contiguous.
 #[cfg(feature = "polars")]
 pub trait IntoPolars {
-    fn into_polars(&mut self) -> impl std::future::Future<Output = Result<DataFrame>> + Send;
+    fn into_polars(self) -> impl std::future::Future<Output = Result<DataFrame>> + Send;
 }
 
 #[cfg(feature = "polars")]
 impl IntoPolars for SendableRecordBatchStream {
-    async fn into_polars(&mut self) -> Result<DataFrame> {
+    async fn into_polars(mut self) -> Result<DataFrame> {
         let arrow_schema = self.schema();
         let polars_schema = convert_arrow_schema_to_polars_schema(&arrow_schema);
         let mut acc_df: DataFrame = DataFrame::from(&polars_schema);
@@ -302,7 +302,7 @@ mod tests {
     async fn from_arrow_to_polars() {
         let record_batch_reader = get_record_batch_reader_from_polars();
         let schema = record_batch_reader.schema();
-        let mut stream: SendableRecordBatchStream = Box::pin(SimpleRecordBatchStream {
+        let stream: SendableRecordBatchStream = Box::pin(SimpleRecordBatchStream {
             schema: schema.clone(),
             stream: futures::stream::iter(
                 record_batch_reader
--- a/rust/lancedb/src/arrow.rs
+++ b/rust/lancedb/src/arrow.rs
@@ -17,6 +17,15 @@ use std::{pin::Pin, sync::Arc};
 pub use arrow_array;
 pub use arrow_schema;
 use futures::{Stream, StreamExt};
+use polars::frame::ArrowChunk;
+
+#[cfg(feature = "polars")]
+use {
+    futures::TryStreamExt,
+    polars::datatypes,
+    polars::prelude::{DataFrame, Field, Schema, Series},
+    polars_arrow::array,
+};
 
 use crate::error::Result;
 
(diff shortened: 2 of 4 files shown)


rust/lancedb/src/arrow.rs
```diff
@@ -119,3 +128,214 @@ impl<T: arrow_array::RecordBatchReader + Send + 'static> IntoArrow for T {
...
+impl PolarsDataFrameRecordBatchReader {
+    /// Creates a new `PolarsDataFrameRecordBatchReader` from a given Polars DataFrame.
+    pub fn new(df: DataFrame) -> Self {
+        let fields: Vec<arrow_schema::Field> = df
+            .schema()
+            .into_iter()
+            .map(|(name, dtype)| {
+                arrow_schema::Field::new(
+                    name,
+                    arrow_schema::DataType::from(dtype.to_arrow(false)),
```
westonpace, line 154: It might be a minor readability boost if we create a constant `const PL_FLAVOR: bool = false` since we use this `false` in a few places.

rust/lancedb/src/arrow.rs
```diff
@@ -119,3 +128,214 @@ impl<T: arrow_array::RecordBatchReader + Send + 'static> IntoArrow for T {
...
+    }
+}
+
+#[cfg(feature = "polars")]
+impl Iterator for PolarsDataFrameRecordBatchReader {
+    type Item = std::result::Result<arrow_array::RecordBatch, arrow_schema::ArrowError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index < self.chunks.len() {
+            let columns: Vec<arrow_array::ArrayRef> = self.chunks[self.index]
```
westonpace, line 178: Instead of storing the chunks as `Vec` and then walking through the `Vec` with an index can we store the chunks as an iterator?  E.g. either...

```
pub struct PolarsDataFrameRecordBatchReader<I: Iterator<Item = ArrowChunk>> {
    chunks: I,
    ...
```

Or just be explicit...

```
pub struct PolarsDataFrameRecordBatchReader {
    chunks: std::vec::IntoIter<ArrowChunk>
```

The advantage here is that chunks are purged from RAM as soon as they are processed.

rust/lancedb/src/arrow.rs
```diff
@@ -119,3 +128,214 @@ impl<T: arrow_array::RecordBatchReader + Send + 'static> IntoArrow for T {
...
+impl arrow_array::RecordBatchReader for PolarsDataFrameRecordBatchReader {
+    fn schema(&self) -> Arc<arrow_schema::Schema> {
+        self.arrow_schema.clone()
+    }
+
+    fn next_batch(
+        &mut self,
+    ) -> std::prelude::v1::Result<Option<arrow_array::RecordBatch>, arrow_schema::ArrowError> {
+        self.next().transpose()
+    }
```
westonpace, line 204:
```suggestion
```

You don't need this.  It's a provided method (has a [default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations)) so you only need to overload it if you need special logic.

rust/lancedb/src/arrow.rs
```diff
@@ -119,3 +128,214 @@ impl<T: arrow_array::RecordBatchReader + Send + 'static> IntoArrow for T {
...
+}
+
+#[cfg(feature = "polars")]
+impl IntoPolars for SendableRecordBatchStream {
+    async fn into_polars(self) -> Result<DataFrame> {
+        let arrow_schema = self.schema();
+        let polars_schema = convert_arrow_schema_to_polars_schema(&arrow_schema);
+        let mut acc_df: DataFrame = DataFrame::from(&polars_schema);
+        let record_batches = self.try_collect::<Vec<_>>().await?;
+        for record_batch in record_batches {
```
westonpace, line 222: No need to collect if you're just going to iterate.  You should be able to do something like...

```
while let Some(record_batch) = self.next().await? {
  ...
}
```

This still means we iterate the entire stream before returning so its a pretty minor optimization (saves us the one allocation of the vec) but that is inevitable.

rust/lancedb/src/arrow.rs
```diff
@@ -119,3 +128,214 @@ impl<T: arrow_array::RecordBatchReader + Send + 'static> IntoArrow for T {
...
+
+#[cfg(feature = "polars")]
+fn convert_record_batch_to_polars_df(
+    record_batch: &arrow::record_batch::RecordBatch,
+    polars_schema: &Schema,
+) -> Result<DataFrame> {
+    let mut columns: Vec<Series> = Vec::with_capacity(record_batch.num_columns());
+
+    for (i, column) in record_batch.columns().iter().enumerate() {
+        let polars_array = Box::<dyn array::Array>::from(&**column);
```
westonpace, line 248: `&**` feels weird but maybe it's inevitable.

rust/lancedb/src/error.rs
```diff
@@ -49,6 +50,10 @@ pub enum Error {
     Arrow { source: ArrowError },
     #[snafu(display("LanceDBError: not supported: {message}"))]
     NotSupported { message: String },
+    #[cfg(feature = "polars")]
+    #[snafu(display("Polars error: {source}"))]
+    #[cfg(feature = "polars")]
+    Polars { source: polars::error::PolarsError },
```
westonpace, line 56: Let's go ahead and just make an `External` error type similar to the one in [`arrow-rs`](https://docs.rs/arrow-schema/51.0.0/arrow_schema/enum.ArrowError.html#variant.ExternalError) or [`datafusion`](https://docs.rs/datafusion/latest/datafusion/error/enum.DataFusionError.html#variant.External).  It will accept `Box<dyn Error + Send + Sync>` and so you don't need to hide it behind the polars feature gate.

westonpace (changes requested):
This is great!  Thanks for doing this.  I've added a few suggestions inline.  Please re-request review once you've addressed them.  Let me address your top-level questions here:

> Because of the blanket implementation of IntoArrow for any type that implements RecordBatchReader...

Yeah, this is no good, but we can proceed with this PR and clean up later.  I'll investigate further.  I'd rather explicitly wrap existing record batch readers in a newtype than force all other integrations to use a newtype.  So maybe I should remove the blanket impl and make it easier to wrap record batch readers.

> My implementation of into_polars requires calling try_collect() on arrow::SendableRecordBatchStream...

Yes.  This is unavoidable if we're mapping to polars DataFrame (which I think we should).  In the future we can map to a lazy frame and perhaps avoid the eager iteration.

> I wasn't sure the best way to surface that union to the user

I've added a comment below.  We can simplify this with an `External` error type.

> What is the purpose of SimpleRecordBatchReader and RecordBatchReader

They aren't used yet so could probably remove them but I felt it better to include the complete set for users.  They are meant to be utilities for users that are using the Rust API.

`lancedb::arrow::RecordBatchReader` is exactly the same as `arrow_array::RecordBatchReader` except the item type is `Result<RecordBatch, lancedb::error::Error>` instead of `Result<RecordBatch, ArrowError>` (in other words, a different error type).  It's not used because we don't have any methods yet that return synchronous iterators (we may never so I could imagine getting rid of this one).

`SimpleRecordBatchReader` and `SimpleRecordBatchStream` are utilities that convert an iterator + schema (or a stream + schema) into the respective trait.  For example, it's easy to end up with something like...

```
let my_stream: SendableRecordBatchStream = ...
let transformed = my_stream.map_ok(|rb| /* transform record batch */).boxed();
# Unfortunately, `transformed` no longer implements SendableRecordBatchStream
# even though it is a BoxStream because it doesn't have a schema() method.
# So we need to wrap it with SimpleRecordBatchStream
```

The need to constantly wrap and rewrap streams / iterators with the appropriate error type is kind of bothersome and I haven't really wrapped my head around the best way to manage that yet so if you have any good suggestions let me know.

python/src/error.rs
```diff
@@ -50,6 +50,7 @@ impl<T> PythonErrorExt<T> for std::result::Result<T, LanceError> {
                     Err(PyNotImplementedError::new_err(err.to_string()))
                 }
                 LanceError::Other { .. } => self.runtime_error(),
+                _ => self.runtime_error(),
```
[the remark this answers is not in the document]
rohitrastogi: Done.

rust/lancedb/src/arrow.rs
```diff
@@ -119,3 +128,214 @@ impl<T: arrow_array::RecordBatchReader + Send + 'static> IntoArrow for T {
         Ok(Box::new(self))
     }
 }
+
+#[cfg(feature = "polars")]
+/// An iterator of record batches formed from a Polars DataFrame. The iterator
+/// panics if the DataFrame's chunks are not aligned. Consider calling
+/// [`polars::prelude::DataFrame::should_rechunk`] to determine whether the DataFrame
+/// should be rechunked and calling [`polars::prelude::DataFrame::align_chunks`]
+/// if the DataFrame needs to be rechunked before using the iterator.
```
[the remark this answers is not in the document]
rohitrastogi: Done. Added a note that its an expensive thing to do and may cause an allocation.

rust/lancedb/src/arrow.rs
rohitrastogi: Done.

rust/lancedb/src/arrow.rs
rohitrastogi: Done.

rust/lancedb/src/arrow.rs
rohitrastogi: Done.

rust/lancedb/src/arrow.rs
rohitrastogi: Yeah, I wasn't able to find a way around it.

rust/lancedb/src/arrow.rs
rohitrastogi: Done. Had to do option 2 - I'm not able to pull the RecordBatchIter struct which implements Iterator<Item = ArrowChunk> due to its lifetime. If we wanted to avoid the allocation into a vector, I think the dataframe would have to outlive the PolarsDataFrameRecordBatchReader struct, which we don't want.

rust/lancedb/src/arrow.rs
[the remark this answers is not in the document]
rohitrastogi: This mutates the df, which forces into_polars to take &mut self. Clippy doesn't like this and complains that "methods called into_&" usually take 'self' by value.

westonpace:
It appears we have a bit of a conundrum.  You can reproduce this by running `cargo update`.

The latest version of arrow is 51.  LanceDb is pinned to version 50 (this is somewhat unavoidable as Lance is pinned to 50 because datafusion is pinned to 50).  LanceDb is close to upgrading to 51 (Lance already has) but we should generally assume that LanceDb will lag behind the latest version of Arrow.

`polars-arrow` uses a range (`>=41`). This seems like it should work.  Version 50 would satisfy both LanceDb and `polars-arrow`.  However, that is regrettably not how the cargo resolver works (https://doc.rust-lang.org/cargo/reference/resolver.html#unexpected-dependency-duplication).  As a result we end up with two different versions of Arrow, 50 and 51.  This is why we are getting errors in the CI jobs.

We can patch the dependency, and that would allow us to build, but patches have to be at the top-level and so anyone using lancedb would then fail.

The only solution I can think of at the moment is to move away from `Arrow2Arrow` and use the FFI stuff directly (polars can export to FFI, which is unversioned, and then lancedb can import from FFI and vice versa).

rohitrastogi:
Ah I see, I was wondering why I couldn't reproduce the CI failures locally. I guess this was a good exercise in learning to run `cargo update` when taking on new dependencies.

The cargo book mentions using `cargo update` with a `--precise` flag - is that what you mean by patching locally? I guess even if you do use `cargo update --precise`, since LanceDB is a binary, the pinned locked file won't be all too useful - AFAIK even if its checked in, it is ignored during dependency resolution.

I'll take a crack at the FFI stuff next week.

rohitrastogi:
This approach won't work, so closing the PR.

PR #1239, opened 2024-04-24
chore: disable the remote feature by default

westonpace:
The rust implementation of the remote client is not yet ready.  This is understandably confusing for users since it is enabled by default.  This PR disables it by default.  We can re-enable it when we are ready (even then it is not clear this is something that should be a default feature).

### DIFF
--- a/rust/lancedb/src/lib.rs
+++ b/rust/lancedb/src/lib.rs
@@ -34,6 +34,16 @@
 //! cargo install lancedb
 //! ```
 //!
+//! ## Create Features
+//!
+//! ### Experimental Features
+//!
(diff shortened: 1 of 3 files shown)


rust/lancedb/src/lib.rs
```diff
@@ -34,6 +34,16 @@
 //! cargo install lancedb
 //! ```
 //!
+//! ## Create Features
```
wjones127, line 37:
```suggestion
//! ## Crate Features
```

wjones127 (approved):

westonpace:
It appears the macos tests were already broken.  I'm going to merge this and we can address shortly.

PR #1245, opened 2024-04-26
feat: allow the index_cache_size to be configured when opening a table

westonpace:
This was already configurable in the rust API but it wasn't actually being passed down to the underlying dataset.  I added this option to both the async python API and the new nodejs API.

I also added this option to the synchronous python API.

I did not add the option to vectordb.

### DIFF
--- a/python/python/lancedb/remote/db.py
+++ b/python/python/lancedb/remote/db.py
@@ -111,7 +111,10 @@ class RemoteDBConnection(DBConnection):
         self._client.mount_retry_adapter_for_table(name)
 
         if index_cache_size is not None:
-            raise ValueError("index_cache_size is not respected by the remote client")
+            logging.info(
+                "index_cache_size is ignored in LanceDb Cloud"
+                " (there is no local cache to configure)"
+            )
 
         # check if table exists
         if self._table_cache.get(name) is None:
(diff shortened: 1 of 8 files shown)


westonpace:
There is no great way to test this feature.  Once we address https://github.com/lancedb/lance/issues/1613 then I think we should have the following:

 * Can set index cache size in bytes
 * Can read index cache capacity in bytes
 * Can read index cache usage in bytes

This will make testing easier and should be a lot more transparent to users.

python/python/lancedb/remote/db.py
```diff
@@ -110,6 +110,9 @@ def open_table(self, name: str) -> Table:
 
         self._client.mount_retry_adapter_for_table(name)
 
+        if index_cache_size is not None:
+            raise ValueError("index_cache_size is not respected by the remote client")
```
chebbyChefNEQ, line 114: nit: Maybe just a warning?

chebbyChefNEQ (approved):

python/python/lancedb/remote/db.py
westonpace: Good point.  We need to make it as easy as possible to swap between the two and I could see an error here being very annoying.  I changed to a logging message (not even a warning, just `info` since this is technically harmless).

PR #1252, opened 2024-04-29
chore: remove "optionalDependencies"

### LINKED ISSUE #1248: bug(node): unable to build locally via yarn
### LanceDB version

`main`

### What happened?

If you try to install the dependencies via `yarn`, it asks to specify a version of `@lancedb/<triple>`. 

This appears to be due to the fact that the binaries are specified in the main `package.json`. 

the `optionalDependencies` section is traditionally added as part of the release process. 

some examples:
- [nodejs-polars](https://github.com/pola-rs/nodejs-polars/blob/main/package.json)
- [snappy](https://github.com/Brooooooklyn/snappy/blob/main/package.json)
- [blake-hash](https://github.com/Brooooooklyn/blake-hash)

```sh
git clone https://github.com/lancedb/lancedb.git
cd lancedb/nodejs
yarn 
```

universalmind303:
closes #1248 

the binding specific `optionalDependencies` are added automatically as part of the `prepublishOnly` hook, and they are not supposed to be committed to `package.json`. 

--- 

npm lifecycle scripts: 
https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts

### DIFF
--- a/nodejs/.gitignore
+++ b/nodejs/.gitignore
@@ -1,0 +1,1 @@
+yarn.lock
--- a/nodejs/package-lock.json
+++ b/nodejs/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@lancedb/lancedb",
-  "version": "0.4.16",
+  "version": "0.4.17",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@lancedb/lancedb",
-      "version": "0.4.16",
+      "version": "0.4.17",
       "cpu": [
         "x64",
         "arm64"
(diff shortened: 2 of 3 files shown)


westonpace (approved):
This seems fine to me.  As long as they are in the published package I agree that it is annoying to need to maintain them.

I don't know if it's related or not but there is also a github action `.github/workflows/update_package_lock_nodejs` which is similar.  Basically, after a release, CI will fail because the `package-lock.json` refers to version X but the package.json refers to version Y.

It's possible that workflow is no longer needed with this change (though it also probably harmless to keep running too, just technical debt to look at at some point)

universalmind303:
> I don't know if it's related or not but there is also a github action .github/workflows/update_package_lock_nodejs which is similar. Basically, after a release, CI will fail because the package-lock.json refers to version X but the package.json refers to version Y.

yes, this PR removes them from the package-lock as well, so we _should_ be able to remove that CI workflow

PR #1257, opened 2024-04-30
feat: add the optimize function to nodejs and async python

westonpace:
The optimize function is pretty crucial for getting good performance when building a large scale dataset but it was only exposed in rust (many sync python users are probably doing this via to_lance today)

This PR adds the optimize function to nodejs and to python.

I left the function marked experimental because I think there will likely be changes to optimization (e.g. if we add features like "optimize on write").  I also only exposed the `cleanup_older_than` configuration parameter since this one is very commonly used and the rest have sensible defaults and we don't really know why we would recommend different values for these defaults anyways.

### DIFF
--- a/nodejs/lancedb/table.ts
+++ b/nodejs/lancedb/table.ts
@@ -154,7 +154,7 @@ export class Table {
    */
   async update(
     updates: Map<string, string> | Record<string, string>,
-    options?: Partial<UpdateOptions>
+    options?: Partial<UpdateOptions>,
   ) {
     const onlyIf = options?.where;
     let columns: [string, string][];
--- a/nodejs/lancedb/table.ts
+++ b/nodejs/lancedb/table.ts
@@ -55,7 +55,6 @@ export interface OptimizeOptions {
   /**
    * If set then all versions older than the given date
    * be removed.  The current version will never be removed.
-   *
    * The default is 7 days
    *
    * @example
@@ -379,9 +378,9 @@ export class Table {
    *
    *  Optimization covers three operations:
    *
-   *   * Compaction: Merges small files into larger ones
-   *   * Prune: Removes old versions of the dataset
-   *   * Index: Optimizes the indices, adding new data to existing indices
+   *  - Compaction: Merges small files into larger ones
+   *  - Prune: Removes old versions of the dataset
+   *  - Index: Optimizes the indices, adding new data to existing indices
    *
    *
    *  Experimental API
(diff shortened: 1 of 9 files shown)


nodejs/lancedb/table.ts
```diff
@@ -344,6 +357,40 @@ export class Table {
...
+   *  hope that the API may be simplified (or not even need to be called) in the
+   *  future.
+   *
+   *  The frequency an application shoudl call optimize is based on the frequency of
+   *  data modifications.  If data is frequently added, deleted, or updated then
+   *  optimize should be run frequently.  A good rule of thumb is to run optimize if
+   *  you have added or modified 100,000 or more records or run more than 20 data
+   *  modification operations.
+   */
+  async optimize(options?: Partial<OptimizeOptions>): Promise<OptimizeStats> {
```
universalmind303, line 390: i think the `OptimizeOptions` are a bit confusing. 

how does the cleanup logic work? Does it clean any tables older than `now() - cleanupOlderThanMs`? or is `cleanupOlderThanMs` supposed to be a unix timestamp in milliseconds? 
What's considered valid?
```ts
// using a unix timestamp as ms
const unix = 1577836800000 // 2020-01-01T00:00:00.000Z
// milliseconds ago
const msAgo = 3600000; // 1 hour ago

await table.optimize({cleanupOlderThanMs: ms})
```

If it's the latter, I think we should accept a JS `Date` object instead of `number`

westonpace: Yes, it's the latter.  It would be nice if JS had a duration / timedelta type class.  A user's intent is not usually "I want to cleanup all versions before march 1st'  It is usually "I want to cleanup all versions older than 7 days".  For example, in python it looks like this:

```
tbl.cleanup_old_versions(timedelta(days=7))
```

In rust it looks like this:

```
tbl.cleanup_old_versions(Duration::days(7))
```

I agree though that a `number` is a little vague and a date is probably better.  I will switch.

universalmind303: > A user's intent is not usually "I want to cleanup all versions before march 1st' It is usually "I want to cleanup all versions older than 7 days". For example, in python it looks like this:

Yeah I agree that JS does make this a little more awkward than languages with a duration datatype.
Maybe we could add an example using date math

```js
const now = new Date();

// date math in JS is very ugly. 
const sevenDaysAgo = new Date((new Date()).setDate(now.getDate() - 7))

tbl.cleanupOldVersions(sevenDaysAgo)
```

wjones127 (approved):

universalmind303 (approved):

Issue #994, opened 2024-02-19
Feature: port python's embedding funtions to Rust


westonpace:
### SDK

Rust

### Description

Python has the ability to specify one or more embedding functions on table creation.  These functions will be run whenever data is added (to calculate the embedding for the new rows from the raw data) and whenever a table is queried (to calculate the appropriate embedding from the query).  Adding similar traits into Rust should be straightforward.

However, in python, these embedding functions are also persisted as part of the table and will be present when the table is loaded again in the future.  This will be slightly more complex in Rust and probably require some kind of registry of embedding functions.

changhiskhan:
the main issue is how we'll run the huggingface models. if they all have an onnx representation then this maybe possible

westonpace:
> the main issue is how we'll run the huggingface models. if they all have an onnx representation then this maybe possiblej

My thinking wasn't full compatibility in terms of "available models and libraries".  My thinking is that the rust feature would only allow custom embedding functions initially, and users would need to configure the registry themselves.  E.g. this could be an extension port for people working with things like https://github.com/rustformers/llm or https://github.com/huggingface/candle

universalmind303:
I'm wondering if these should be compatible across languages? 

as in if i create this table in python
```py

func = get_registry().get("openai").create(name="text-embedding-3-large")
# func = get_registry().get("openai").create(name="text-embedding-ada-002")
class Items(LanceModel):
    text: str = func.SourceField()
    id: int
    price: float
    vector: Vector(func.ndims()) = func.VectorField()

table = db.create_table("python", schema=Items, mode="overwrite")
table.add([
    {"id": 100, "text": "socks", "price": 10.00},
    {"id": 200, "text": "shoes", "price": 20.00},
])
```

then register the same function in rust

```rs
    let embedding = Arc::new(OpenAIEmbeddingFunction::new_with_model(
        "text",
        api_key,
        "text-embedding-3-large",
    )?);

    let db = connect(tempdir).execute().await?;
    db.embedding_registry()
        .register("openai", embedding.clone())?;
```

When the python created table is loaded in rust, should it automatically use this embedding function for the table? I'd assume we'd let the `registry` decide if it's actually the same function, but I wanted clarity on if this is functionality we want to support.

westonpace:
> I'm wondering if these should be compatible across languages?

I think cross-language use cases like this will be rare but possible.  It would be nice to support it if we could.  I think we can state that a user should only register embeddings with the same name if they are the same function.

[...]

PR #1259, opened 2024-05-01
feat: rust embedding registry

universalmind303:
Todo:

- [x] add proper documentation
- [x] add unit tests
- [x] better handling of the registry**1
- [x] allow user defined registry**2

**1 The python implementation just uses a global registry so it makes things a bit easier. I attached it to the db/connection to prevent future conflicts if running multiple connections/databases. I mostly modeled the registry & pattern off of datafusion's [FunctionRegistry](https://docs.rs/datafusion/latest/datafusion/execution/trait.FunctionRegistry.html). 

**2 Ideally, the user should be able to provide it's own registry entirely, but currently it just uses an in memory registry by default (_which isn't configurable_)

`rust/lancedb/examples/embedding_registry.rs` provides a thorough example of expected usage. 

---

Some additional notes:

This does not provide any of the out of box functionality that the python registry does.

_i.e there are no built-in embedding functions._ 

You can think of this as the ground work for adding those built in functions, So while this is part of https://github.com/lancedb/lancedb/issues/994, it does not yet offer feature parity.

### DIFF
--- a/rust/lancedb/src/embeddings.rs
+++ b/rust/lancedb/src/embeddings.rs
@@ -1,7 +1,21 @@
+// Copyright 2024 LanceDB Developers.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use lance::arrow::RecordBatchExt;
 use std::{
     borrow::Cow,
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     sync::{Arc, RwLock},
 };
 
@@ -27,14 +41,16 @@ use crate::{
 /// similar sentences to a query sentence.
 ///
 /// To use an embedding function you must first register it with the `EmbeddingsRegistry`.
-/// Then you can define it on a column in the table schema.  That embedding will then be used
+/// Then you can define it on a column in the table schema. That embedding will then be used
 /// to embed the data in that column.
 pub trait EmbeddingFunction: std::fmt::Debug + Send + Sync {
     fn name(&self) -> &str;
+    /// The type of the input data
     fn source_type(&self) -> Cow<DataType>;
+    /// The type of the output data
+    /// This should match the output of the `embed` function
     fn dest_type(&self) -> Cow<DataType>;
-
-    /// TODO: This should be async
+    /// Embed the input
     fn embed(&self, source: Arc<dyn Array>) -> Result<Arc<dyn Array>>;
 }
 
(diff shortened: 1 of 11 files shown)


rust/lancedb/src/embeddings.rs
```diff
@@ -0,0 +1,264 @@
...
+/// similar sentences to a query sentence.
+///
+/// To use an embedding function you must first register it with the `EmbeddingsRegistry`.
+/// Then you can define it on a column in the table schema.  That embedding will then be used
+/// to embed the data in that column.
+// #[async_trait]
+pub trait EmbeddingFunction: std::fmt::Debug + Send + Sync {
+    fn name(&self) -> &str;
+    fn source_type(&self) -> Cow<DataType>;
+    fn dest_type(&self) -> Cow<DataType>;
```
universalmind303, line 36: I think the `Cow` return types provide a bit more flexibility to end users as they don't need to go through extra steps to create a reference.

For reference, I opened up an [issue in datafusion](https://github.com/apache/datafusion/issues/10018) a while back highlighting the issues with using borrowed types

AyushExel:
Curious to know how/if this plans to also push pytorch model inference to rust level?

universalmind303:
> Curious to know how/if this plans to also push pytorch model inference to rust level?

I think supporting pytorch models in rust is possible, however my initial idea was to try to plug in to [candle](https://github.com/huggingface/candle) and [hf-hub](https://github.com/huggingface/hf-hub) which should allow for users to run most models available in huggingface as well as onnx based models.

universalmind303:
Here's a followup PR (on my fork) that adds the openai embedding function

https://github.com/universalmind303/lancedb/pull/1

wjones127 (commented):
> I think supporting pytorch models in rust is possible

I think the trade off worth considering here is binary size. Rust users can enable/disable features to choose what they want to include in the binaries, but Python and Node users can't. I think having one or two of the most generic frameworks integrated at the Rust level makes sense, but beyond that it seems like it would be preferable to expose Python and Node register hooks so the libraries can be dynamically imported.

rust/lancedb/src/embeddings.rs
```diff
@@ -0,0 +1,289 @@
...
+pub trait EmbeddingFunction: std::fmt::Debug + Send + Sync {
+    fn name(&self) -> &str;
+    /// The type of the input data
+    fn source_type(&self) -> Cow<DataType>;
+    /// The type of the output data
+    /// This should match the output of the `embed` function
+    fn dest_type(&self) -> Cow<DataType>;
+    /// Embed the input
+    fn embed(&self, source: Arc<dyn Array>) -> Result<Arc<dyn Array>>;
+}
```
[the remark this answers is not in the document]
universalmind303: Good call, I'll wrap all of these in `Result`

rust/lancedb/tests/embedding_registry_test.rs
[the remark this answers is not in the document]
universalmind303: No, this only supports single ones right now. I'll update PR with support for multiple embeddings.

rust/lancedb/src/embeddings.rs
```diff
@@ -0,0 +1,303 @@
...
+
+impl<R: RecordBatchReader> Iterator for WithEmbeddings<R> {
+    type Item = std::result::Result<RecordBatch, arrow_schema::ArrowError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let batch = self.inner.next()?;
+        match batch {
+            Ok(mut batch) => {
+                // todo: parallelize this
```
wjones127, line 262: Not sure which you mean, but I imagine two levels of parallelism here:
1. Within batch: multiple embedding columns at a time
2. Inter-batch: make parallel request to embed data from different batches.

wjones127 (approved):
This looks pretty good now :)

rust/lancedb/src/embeddings.rs
universalmind303: So i think the latter can be done outside of this function as this method acts on a single batch. The `todo` is mostly referring to the former. 

if multiple embeddings are specified on the table, they are applied in series. Given that these `embed` functions can potentially delegate to external services such as openai, they become primarily IO bound. Having some way to parallelize these IO bound embedding functions would greatly speed up this operation.

On the other hand, if we always parallelize multiple embeddings, it could result in contention as the user may implement the `embed` function to have it's own internal parallelism.

PR #1260, opened 2024-05-01
Implement convertors to and from Polars DataFrames in Rust SDK using convertors based on C FFI #1099

rohitrastogi:
https://github.com/lancedb/lancedb/issues/1099

Took the same general approach from: https://github.com/lancedb/lancedb/pull/1235. Instead of using high-level convertors  implemented in polars-arrow (with the arrow-rs feature flag, which adds a dependency on arrow-rs), I used convertors based on the C FFI to avoid dependency conflicts.

### DIFF
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -40,8 +40,8 @@ serde = { version = "^1" }
 serde_json = { version = "1" }
 # For remote feature
 reqwest = { version = "0.11.24", features = ["gzip", "json"], optional = true }
-polars-arrow = { version = "0.39.2", optional = true }
-polars = { version = "0.39.2", optional = true}
+polars-arrow = { version = ">=0.38", optional = true }
+polars = { version = ">=0.38", optional = true}
 
 [dev-dependencies]
 tempfile = "3.5.0"
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -40,6 +40,8 @@ serde = { version = "^1" }
 serde_json = { version = "1" }
 # For remote feature
 reqwest = { version = "0.11.24", features = ["gzip", "json"], optional = true }
+polars-arrow = { version = "0.39.2", features=["arrow_rs"], optional = true }
+polars = { version = "0.39.2", optional = true}
 
 [dev-dependencies]
 tempfile = "3.5.0"
@@ -56,3 +58,4 @@ default = ["remote"]
 remote = ["dep:reqwest"]
 fp16kernels = ["lance-linalg/fp16kernels"]
 s3-test = []
+polars = ["dep:polars-arrow", "dep:polars"]
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -40,7 +40,7 @@ serde = { version = "^1" }
 serde_json = { version = "1" }
 # For remote feature
 reqwest = { version = "0.11.24", features = ["gzip", "json"], optional = true }
-polars-arrow = { version = "0.39.2", features=["arrow_rs"], optional = true }
+polars-arrow = { version = "0.39.2", optional = true }
 polars = { version = "0.39.2", optional = true}
 
 [dev-dependencies]
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -40,8 +40,8 @@ serde = { version = "^1" }
 serde_json = { version = "1" }
 # For remote feature
 reqwest = { version = "0.11.24", features = ["gzip", "json"], optional = true }
-polars-arrow = { version = ">=0.38", optional = true }
-polars = { version = ">=0.38", optional = true}
+polars-arrow = { version = ">=0.37", optional = true }
+polars = { version = ">=0.37", optional = true}
 
 [dev-dependencies]
 tempfile = "3.5.0"
--- a/rust/lancedb/src/polars_arrow_convertors.rs
+++ b/rust/lancedb/src/polars_arrow_convertors.rs
@@ -6,8 +6,9 @@
 /// between the arrows s in polars-arrow and arrow-rs using the C FFI.
 ///
 /// The polars-arrow does implement conversions to and from arrow-rs, but
-/// requires a feature flagged dependency on arrow-rs. Unfortunately, the arrow-rs
-/// versions are not compatible, which necessitates using the C FFI.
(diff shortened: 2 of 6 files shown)


rohitrastogi:
It may be slightly more performant to:
1) create a StructArray for each arrow-rs RecordBatch/polars-arrow ArrowChunk 
2) convert the the StructArray between Rust Arrow representations via C FFI
3) downcast the StructArray and unpack the constituent arrays

instead of converting each array individually via the C FFI, but I'm not sure.

rust/lancedb/Cargo.toml
```diff
@@ -40,6 +40,8 @@ serde = { version = "^1" }
 serde_json = { version = "1" }
 # For remote feature
 reqwest = { version = "0.11.24", features = ["gzip", "json"], optional = true }
+polars-arrow = { version = "0.39.2", optional = true }
+polars = { version = "0.39.2", optional = true}
```
universalmind303, line 44: We don't want to force consumers to be locked to our version of polars. This should be wildcarded, or at least specify a wider range similar to what polars does with `arrow` dependencies

rust/lancedb/src/polars_arrow_convertors.rs
```diff
@@ -0,0 +1,123 @@
+/// Polars and LanceDB both use Arrow for their in memory-representation, but use
```
universalmind303, line 1: polars wildcards their arrow-rs dependencies, so we should be able to use their conversion traits without worrying about compatibility issues

```toml
# polars/Cargo.toml
arrow-array = { version = ">=41", default-features = false }
arrow-buffer = { version = ">=41", default-features = false }
arrow-data = { version = ">=41", default-features = false }
arrow-schema = { version = ">=41", default-features = false }
```

rohitrastogi: Please see the discussion here: https://github.com/lancedb/lancedb/pull/1235. The polars-arrow conversion traits pulled in arrow-rs 51, but LanceDB depended on arrow-rs 50. Dependency resolution did converge on version 50 and pulled in both versions. This prevented compilation.

It looks like LanceDB just upgraded to arrow-rs 51, so for now, the polars-arrow conversion traits should compile fine but @westonpace mentioned that we should assume that LanceDB will lag behind the newest version of arrow-rs.

Let me know if you think there's a way to use the polars-arrow conversion traits given this.

westonpace: > polars wildcards their arrow-rs dependencies, so we should be able to use their conversion traits without worrying about compatibility issues

That was my original thinking too but https://doc.rust-lang.org/cargo/reference/resolver.html#unexpected-dependency-duplication seems to suggest otherwise :(

rust/lancedb/Cargo.toml
rohitrastogi: Good call. Updated to >= 0.38. There were some updates to how polars handled strings/string views in 0.38.

rohitrastogi: Actually, since Polars isn’t at version 1 yet, technically anything can change between version numbers. 0.40 or 0.39.3 aren’t guaranteed to be backwards compatible. So, maybe expecting a specific version (0.39.2) is the way to go. Not sure. What do you think?

universalmind303: From reading through the code, it looks like we are only using functionality that is "stablilzed". Those api's haven't changed much _(if at all)_ since very early on. So overall, I'd say it's pretty safe to use a range like `>= 0.38` here.

westonpace:
> It may be slightly more performant to:
>
>   create a StructArray for each arrow-rs RecordBatch/polars-arrow ArrowChunk
>    convert the the StructArray between Rust Arrow representations via C FFI
>    downcast the StructArray and unpack the constituent arrays
>
> instead of converting each array individually via the C FFI, but I'm not sure.

There is actually a (very subtle) downside to the struct array approach.  The C FFI interface does not allow releasing individual children, you can only release the top-level structure.  This means, if you do the struct array approach, then any single child array will keep the other child arrays alive (you can only free/release the entire batch).  This can sometimes be a problem when writing data since different columns flush at different speeds (narrow columns like boolean rarely flush to disk and wide columns like images flush almost every batch).

I don't think it would be any faster and so I think it's better how it is.

rust/lancedb/src/polars_arrow_convertors.rs
```diff
@@ -0,0 +1,123 @@
...
+/// different Rust Arrow implementations. LanceDB uses the arrow-rs crate and
+/// Polars uses the polars-arrow crate.
+///
+/// This crate defines zero-copy conversions (of the underlying buffers)
+/// between the arrows s in polars-arrow and arrow-rs using the C FFI.
```
westonpace, line 6:
```suggestion
/// between polars-arrow and arrow-rs using the C FFI.
```

rust/lancedb/src/polars_arrow_convertors.rs
```diff
@@ -0,0 +1,123 @@
...
+/// requires a feature flagged dependency on arrow-rs. The version of arrow-rs
+/// depended on by polars-arrow and LanceDB may not be compatible,
+/// which necessitates using the C FFI.
+use crate::error::Result;
+use polars::prelude::{DataFrame, Series};
+use std::{mem, sync::Arc};
+
+/// When interpreting Polars dataframes as polars-arrow record batches,
+/// whether to use Arrow string/binary view types instead of the standard
+/// Arrow string/binary types.
```
westonpace, line 18:
```suggestion
/// one must decide whether to use Arrow string/binary view types
/// instead of the standard Arrow string/binary types.
```

rust/lancedb/src/polars_arrow_convertors.rs
```diff
@@ -0,0 +1,123 @@
...
+/// For now, we will not use string view types because conversions
+/// for string view types from polars-arrow to arrow-rs are not yet implemented.
+/// See: https://lists.apache.org/thread/w88tpz76ox8h3rxkjl4so6rg3f1rv7wt for the
+/// differences in the types.
```
westonpace, line 22: No need to adjust the comment but I think we also don't support the view types in Lance yet.

westonpace (approved):
Minor comment tweaks but otherwise this is good.  @universalmind303 did you have any further concerns?

universalmind303 (approved):

rohitrastogi:
Addressed the comment tweaks. Thanks for the reviews!

westonpace:
Thanks a lot for your help!

PR #1275, opened 2024-05-07
feat(rust): openai embedding function

universalmind303:
part of https://github.com/lancedb/lancedb/issues/994. 

Adds the ability to use the openai embedding functions.

the example can be run by the following

```sh
> EXPORT OPENAI_API_KEY="sk-..."
> cargo run --example openai --features=openai
```

which should output
```
Closest match: Winter Parka
```


rust/lancedb/examples/openai.rs
```diff
@@ -0,0 +1,83 @@
...
+        .create_table("vectors", make_data())
+        .add_embedding(EmbeddingDefinition::new(
+            "text",
+            "openai",
+            Some("embeddings"),
+        ))?
+        .execute()
+        .await?;
+
+    // there is no equivalent to '.search(<query>)' yet
```
westonpace, line 39: Is the plan to have an equivalent at some point in the future?

rust/lancedb/src/embeddings/openai.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+    pub fn api_base<S: Into<String>>(mut self, api_base: S) -> Self {
+        self.api_base = Some(api_base.into());
+        self
+    }
+
+    pub fn org_id<S: Into<String>>(mut self, org_id: S) -> Self {
+        self.org_id = Some(org_id.into());
+        self
+    }
```
westonpace, line 141: Doesn't have to be this PR but can we document these public methods at some point?  I don't know what `api_base` or `org_id` are referring to.

rust/lancedb/src/embeddings/openai.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    fn compute_source_embeddings(&self, source: ArrayRef) -> crate::Result<ArrayRef> {
+        let len = source.len();
+        let n_dims = self.model.ndims();
+        let inner = self.compute_inner(source)?;
+
+        let fsl = DataType::new_fixed_size_list(DataType::Float32, n_dims as i32, false);
+
+        // We can't use the FixedSizeListBuilder here because it always adds a null bitmap
+        // and we want to explicitly work with non-nullable arrays.
```
westonpace, line 170: I see that OpenAi only supports non-nullable but will this builder hide the nulls by dropping the nullability (so we don't get an error when we should get an error)?

rust/lancedb/src/embeddings/openai.rs
```diff
@@ -0,0 +1,262 @@
...
+                        s.expect("we already asserted that the array is non-nullable")
+                            .to_string()
+                    })
+                    .collect::<Vec<String>>();
+                EmbeddingInput::StringArray(array)
+            }
+            _ => unreachable!("This should not happen. We already checked the data type."),
+        };
+
+        let client = Client::with_config(creds);
```
westonpace, line 235: Do we want to cache / pool this client at some point in the future?  Although maybe that's more work than its worth given the amount of time it takes to calculate embeddings the connection establishment might be a small fraction anyways.

rust/lancedb/src/embeddings/openai.rs
```diff
@@ -0,0 +1,262 @@
...
+            model: self.model.to_string(),
+            input,
+            encoding_format: Some(EncodingFormat::Float),
+            user: None,
+            dimensions: None,
+        };
+
+        // TODO: request batching and retry logic
+        task::block_in_place(move || {
+            Handle::current().block_on(async {
```
westonpace, line 247: Should `compute_inner` be an async function?

westonpace (commented):
This is pretty cool!  A few initial questions.

rust/lancedb/src/embeddings/openai.rs
universalmind303: I think we'd still need to spawn an async task for this regardless unless we made `EmbeddingFunction::compute_source_embeddings` async, but we can't do that without some significant refactoring. 


We use the `RecordBatchReader` which isn't async. We'd likely have to refactor all instances of that to an async equivalent. 

However, if you prefer the wrapping to happen outside of `compute_inner`, that's totally fine & i can make those changes.

rust/lancedb/src/embeddings/openai.rs
universalmind303: no, we assert that the input array is non null, so we should never run into a scenario where we drop the nullability.

rust/lancedb/examples/openai.rs
universalmind303: IIRC, there were plans to add similar functionality to rust sdk

rust/lancedb/src/embeddings/openai.rs
universalmind303: that was my initial thought. We can definitely add this as an optimization later on if we need to.

rust/lancedb/src/embeddings/openai.rs
westonpace: We don't have to worry about it right now.  I am thinking we might eventually want `compute_source_embeddings` to be async though.  The top-level `add` and `query` methods are already async so it seems odd to go async -> sync -> async.

Also, we might want to run embeddings in parallel at some point too.  Although that might be best left to embedding function since different embeddings might have different preferences on batch size / parallel execution.

universalmind303: > We don't have to worry about it right now. I am thinking we might eventually want compute_source_embeddings to be async though. The top-level add and query methods are already async so it seems odd to go async -> sync -> async.

makes sense. Will update!

---


> Also, we might want to run embeddings in parallel at some point too. Although that might be best left to embedding function since different embeddings might have different preferences on batch size / parallel execution.

I made a similar comment on an earlier PR. https://github.com/lancedb/lancedb/pull/1259#discussion_r1591459450

rust/lancedb/src/embeddings/openai.rs
```diff
@@ -0,0 +1,266 @@
...
+impl TryFrom<&str> for EmbeddingModel {
+    type Error = Error;
+
+    fn try_from(value: &str) -> std::result::Result<Self, Self::Error> {
+        value.parse()
+    }
+}
+
+pub struct OpenAIEmbeddingFunction {
+    source_column: String,
```
westonpace, line 69: What is this field used for?  Isn't the source column already a part of the embedding definition?

rust/lancedb/src/embeddings/openai.rs
```diff
@@ -0,0 +1,266 @@
...
+                &self.api_key[self.api_key.len() - 4..]
+            )
+        } else {
+            "[INVALID]".to_string()
+        };
+
+        f.debug_struct("OpenAI")
+            .field("source_column", &self.source_column)
+            .field("model", &self.model)
+            .field("credentials", &creds_display)
```
westonpace, line 92: Minor nit: this is called `api_key` everywhere else.
```suggestion
            .field("api_key", &creds_display)
```

westonpace (approved):
Just some minor nits but looks great otherwise

rust/lancedb/src/embeddings/openai.rs
universalmind303: good catch. Removed!

PR #1279, opened 2024-05-08
feat(java) add table names java api

LuQQiu:
Add lancedb-jni and table names API


LuQQiu:
@QianZhu @eddyxu  @walterddr  PTAL the table names
I am trying to save the connection on JNI side so that the connection can be reused for create_table, open_table and other funcs 
https://github.com/LuQQiu/lancedb/tree/addJavaAPI
but encountering some JNI crash issue, still debugging.

LuQQiu:
wip testing for this PR

java/core/src/test/java/com/lancedb/lancedb/DatabaseTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Test
+  void tableNmaes() {
+    ClassLoader classLoader = getClass().getClassLoader();
+    URL lanceDbURL = classLoader.getResource("example_db");
+    List<String> tableNames = Database.tableNames(lanceDbURL.toString());
+    assertEquals("dataset_version", tableNames.get(0));
+    assertEquals("new_empty_dataset", tableNames.get(1));
+    assertEquals("test", tableNames.get(2));
+    assertEquals("write_stream", tableNames.get(3));
```
QianZhu, line 44: is the returned table names sorted in a certain way?

java/core/lancedb-jni/src/ffi.rs
```diff
@@ -0,0 +1,199 @@
...
+
+use jni::objects::{JByteBuffer, JObjectArray, JString};
+use jni::sys::jobjectArray;
+use jni::{objects::JObject, JNIEnv};
+
+use crate::error::{Error, Result};
+
+/// TODO(lu) import from lance-jni without duplicate
+/// Extend JNIEnv with helper functions.
+pub trait JNIEnvExt {
```
QianZhu, line 25: will we add unit tests?

java/core/lancedb-jni/src/error.rs
```diff
@@ -0,0 +1,215 @@
...
+        reason: String,
+        location: Location,
+    },
+    #[snafu(display("Other Lance error: {message}, {location}"))]
+    OtherLance { message: String, location: Location },
+    #[snafu(display("Other LanceDB error: {message}, {location}"))]
+    OtherLanceDB { message: String, location: Location },
+}
+
+impl Error {
```
QianZhu, line 76: prob it is easier to map errors (with an ErrorExt trait). You can see an example here: https://github.com/lancedb/lancedb/blob/e688484bd3ad7b53e0832ed3dfa30351ca16483b/python/src/error.rs#L33

java/core/lancedb-jni/src/ffi.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    /// Get Option<&[u8]> from Java Optional<ByteBuffer>.
+    fn get_bytes_opt(&mut self, obj: &JObject) -> Result<Option<&[u8]>>;
+
+    fn get_optional<T, F>(&mut self, obj: &JObject, f: F) -> Result<Option<T>>
+    where
+        F: FnOnce(&mut JNIEnv, &JObject) -> Result<T>;
+}
+
+impl JNIEnvExt for JNIEnv<'_> {
```
QianZhu, line 62: All functions are returning Result but do we need it if the result is always Ok?

java/core/lancedb-jni/src/ffi.rs
LuQQiu: Yeah, i added some examples of test cases but haven't got time to add more. 
https://github.com/lancedb/lance/blob/main/java/core/src/test/java/com/lancedb/lance/JNITest.java

Because testing JNIEnv requires the launch of an actual JVM, so i plan to add the test on the Java side.
Ideally the tests will be passing some object from Java to Rust, Rust side consumes the Java object, convert to rust object, and convert back to java object to return to the Java side. If anything is wrong, the test will fail

java/core/lancedb-jni/src/ffi.rs
LuQQiu: Using "Result" for convenient handle the returned result of the inner called methods. JNI needs special error handling differently, require calling env.throw_new(JAVA exception) to throw java exception to end user.

java/core/src/test/java/com/lancedb/lancedb/DatabaseTest.java
LuQQiu: Yeah, it's always in ascending order (alphabetical order), will add it to the method description

java/core/lancedb-jni/src/error.rs
LuQQiu: Good idea, i will test this tonight and create another PR for the error related changes

java/core/src/main/java/com/lancedb/lancedb/Database.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.lancedb.lancedb;
+
+import io.questdb.jar.jni.JarJniLoader;
+import java.util.List;
+ 
+/**
+ * Represents LanceDB database.
+*/
+public class Database {
```
eddyxu, line 23: Lets make the naming conventions be consistent with the rest of the rust/python/nodejs packages.

YOu can take a look of https://docs.rs/lancedb/latest/lancedb/

walterddr (commented):
thanks i will take a look

> I am trying to save the connection on JNI side so that the connection can be reused for create_table, open_table and other funcs

i am pretty sure it is OK to cache the connection on the Java side as well as long as the API is clear :-P

LuQQiu:
@QianZhu @eddyxu @walterddr PTAL, thanks!

java/core/src/test/java/com/lancedb/lancedb/ConnectionTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Path;
+import java.util.List;
+import java.net.URL;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class ConnectionTest {
+  @TempDir
```
QianZhu, line 27: do we need to close the connection?

java/core/src/test/java/com/lancedb/lancedb/ConnectionTest.java
```diff
@@ -0,0 +1,73 @@
...
+  void tableNmaesLimit() {
+    ClassLoader classLoader = getClass().getClassLoader();
+    URL lanceDbURL = classLoader.getResource("example_db");
+    try (Connection conn = Connection.connect(lanceDbURL.toString())) {
+      List<String> tableNames = conn.tableNames(2);
+      assertEquals("dataset_version", tableNames.get(0));
+      assertEquals("new_empty_dataset", tableNames.get(1));
+    }
+  }
+}
```
QianZhu, line 73: can we add a test where both startAfter and limit are speficied?

java/core/src/test/java/com/lancedb/lancedb/ConnectionTest.java
LuQQiu: Yeah the connections are closed in try (Connection conn = Connection.connect(databaseUri)) {

LuQQiu:
@QianZhu Addressed the comments, added test cases, PTAL

QianZhu (approved):

java/core/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+    <parent>
+        <groupId>com.lancedb</groupId>
+        <artifactId>lancedb-parent</artifactId>
+        <version>0.1-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>lancedb-core</artifactId>
+    <name>LanceDB Core</name>
+    <packaging>jar</packaging>
```
walterddr, line 17: what's the relationship between this and the lancedb/lance/pull/2115?

AFAIK these cannot be identical. (maybe this should be lancedb-core and the other one should be lance-core?)

LuQQiu: @walterddr This is the LanceDB java module with LanceDB functionalities e.g. embeddings, table names
The other one is Lance java module.

Put it to the sub module to give space if we want to put other connectors e.g. Spark connectors directly in LanceDB java side

walterddr: ok i see. so lancedb-core and lance-core are both core modules and doesn't depend on each other (and thus are independent basically?) nice i will pull in both in the trino side.

LuQQiu: Yeah yeah!! From the Java Side you need to import both
The only problem later is the packaged jar may be bigger, but we can consider that issue later

PR #1284, opened 2024-05-09
feat: support IVF_HNSW_SQ

BubbleCal:


BubbleCal:
leave docs and python to future PRs, or it's too large

rust/lancedb/src/index/vector.rs
```diff
@@ -201,3 +205,121 @@ pub(crate) fn suggested_num_sub_vectors(dim: u32) -> u32 {
...
+        self.m = m;
+        self
+    }
+
+    /// The number of candidates to evaluate during the construction of the HNSW graph.
+    /// The default value is 300.
+    pub fn ef_construction(mut self, ef_construction: usize) -> Self {
+        self.ef_construction = ef_construction;
+        self
+    }
```
westonpace, line 324: We can add this in a future PR (when we extend this to node and python perhaps) but we will want these parameters to include:

 * What happens if this is set incorrectly?
 * When should the user adjust the default?

rust/lancedb/src/index/vector.rs
```diff
@@ -201,3 +205,121 @@ pub(crate) fn suggested_num_sub_vectors(dim: u32) -> u32 {
...
+pub struct IvfHnswSqIndexBuilder {
+    // IVF
+    pub(crate) distance_type: DistanceType,
+    pub(crate) num_partitions: Option<u32>,
+    pub(crate) sample_rate: u32,
+    pub(crate) max_iterations: u32,
+
+    // HNSW
+    pub(crate) m: usize,
+    pub(crate) ef_construction: usize,
```
westonpace, line 228: Can we use `u64` instead of `usize`?  We can convert to `usize` internally.  I'd rather avoid `usize` in our public API since it is slightly more confusing to think about.

westonpace (approved):
Some minor suggestions.  I think we should decide on `usize` / `u64` before merging (as changing that would be a breaking change) but the other suggestion about the comments we can figure out later.

rust/lancedb/src/index/vector.rs
BubbleCal: sure, let me change it to `u32` cause we don't need such great value of `u64`

PR #1290, opened 2024-05-10
fix(nodejs): add better error handling when missing embedding functions

universalmind303:
note: 
running the default lint command `npm run lint -- --fix` seems to have made a lot of unrelated changes.


node/src/index.ts
```diff
@@ -830,98 +843,107 @@ export class LocalConnection implements Connection {
...
+      this._embeddings
+    );
+
     if (data instanceof ArrowTable) {
-      tbl = data
+      tbl = data;
     } else {
-      tbl = makeArrowTable(data, { schema })
+      tbl = makeArrowTable(data, { schema: schemaWithoutEmbeddings });
     }
```
universalmind303, line 945: this is the relevant code

node/src/index.ts
```diff
@@ -1106,63 +1138,105 @@ export class LocalTable<T = number[]> implements Table<T> {
...
+  async dropColumns(columnNames: string[]): Promise<void> {
+    return tableDropColumns.call(this._tbl, columnNames);
+  }
+
+  withMiddleware(middleware: HttpMiddleware): Table<T> {
+    return this;
+  }
+}
+
+function validateSchemaEmbeddings(
```
universalmind303, line 1202: this is the relevant code change. Every thing else is just from running the linter

universalmind303:
@wjones127 @westonpace, I'm unable to reproduce the CI failures locally. Could one of you take a look and see if you are able to reproduce.

westonpace:
> @wjones127 @westonpace, I'm unable to reproduce the CI failures locally. Could one of you take a look and see if you are able to reproduce.

@universalmind303 It does reproduce locally for me.  It appears that it's this line:

>         if (data[0][field.name] === undefined) {

I suspect `data[0]` is undefined.

node/src/arrow.ts
```diff
@@ -503,49 +550,92 @@ export async function fromTableToBuffer<T> (
...
+  for (const field of schema.fields) {
+    if (field.type instanceof FixedSizeList) {
+      if (!(data instanceof ArrowTable)) {
+        if (data.length !== 0 && data?.[0]?.[field.name] === undefined) {
+          missingEmbeddingFields.push(field);
+        } else {
+          fields.push(field);
+        }
+      } else {
+        if (data.getChild(field.name) !== null) {
```
westonpace, line 622: Should this be `===`?  Why is it a `missingEmbeddingField` if the column exists?

westonpace (commented):
I might be reading this wrong but does this prevent a user from providing a fixed-size-list column directly without using the embeddings function?  In other words, if a user is calculating their own embeddings (not using lancedb to do it) will it still work?

Also, any change added to `node/src/arrow.ts` is probably appropriate for `nodejs/src/arrow.ts`.

node/src/arrow.ts
universalmind303: oh, i forgot to remove this logic. Will update.

universalmind303:
> I might be reading this wrong but does this prevent a user from providing a fixed-size-list column directly without using the embeddings function? In other words, if a user is calculating their own embeddings (not using lancedb to do it) will it still work?

Users can still provide their own embeddings manually without using the `embed` functionality. It will only error out if there is a vector field in the schema with no matching implementation (manual or embed function)

Example:

```ts
     const schema = new Schema([
        new Field("id", new Int32()),
        new Field("text", new Utf8()),
        new Field(
          "vector",
          new FixedSizeList(2, new Field("item", new Float32(), true))
        ),
      ]);
```
### user provides schema and manual embeddings (OK)
```ts
      const data = [
        { id: 1, text: "foo", vector: [0.1, 0.2] },
        { id: 2, text: "bar", vector: [0.3, 0.4] },
      ];
      let table = await con.createTable({
        name: "test",
        data,
        schema,
      });
```

### user provides schema and embedding function (OK)
```ts
      const embeddingFunction = {...}
      const data = [
        { id: 1, text: "foo"},
        { id: 2, text: "bar"},
      ];
      let table = await con.createTable({
        name: "test",
        data,
        schema,
        embeddingFunction
      });
```

### User provides schema, but no embeddings or function (Not OK)
```ts
      const data = [
        { id: 1, text: "foo"},
        { id: 2, text: "bar"},
      ];
      let table = await con.createTable({
        name: "test",
        data,
        schema
      });
```

westonpace (approved):
Got it, looks good.

PR #1294, opened 2024-05-10
ci: write down breaking change policy

wjones127:
* Enforce conventional commit PR titles
* Add automatic labelling of PRs
* Write down breaking change policy.

Left for another PR:
* Validation of breaking change version bumps. (This is complicated due to separate releases for Python and other package.)


albertlockett (approved):

wjones127:
Validated scripts in: https://github.com/wjones127/lancedb/pull/1

PR #1253, opened 2024-04-29
Remove semvar strict dependancy

AmanKishore:

PR #1298, opened 2024-05-13
fix semver deprecation warnings

### LINKED ISSUE #1296: Fix semvar deprication warning
### SDK

Python

### Description

Cohere reranker uses semvar.compare which seems depricated

lancedb/rerankers/cohere.py:49: PendingDeprecationWarning: Function 'semver.compare' is deprecated. Deprecated since version 3.0.0.  Still under investigation, see #258. Use the respective 'semver.Version.compare' instead.
    and semver.compare(cohere.__version__, "5.0.0") < 0

pmeier:
Fixes #1296.

### DIFF
--- a/python/python/lancedb/pydantic.py
+++ b/python/python/lancedb/pydantic.py
@@ -37,7 +37,7 @@ import pyarrow as pa
 import pydantic
 import semver
 
-PYDANTIC_VERSION = semver.parse_version_info(pydantic.__version__)
+PYDANTIC_VERSION = semver.Version.parse(pydantic.__version__)
 try:
     from pydantic_core import CoreSchema, core_schema
 except ImportError:
--- a/python/python/lancedb/rerankers/cohere.py
+++ b/python/python/lancedb/rerankers/cohere.py
@@ -46,7 +46,7 @@ class CohereReranker(Reranker):
         # ensure version is at least 0.5.0
         if (
             hasattr(cohere, "__version__")
-            and semver.compare(cohere.__version__, "5.0.0") < 0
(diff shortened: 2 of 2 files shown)


AyushExel (approved):

AyushExel:
Thankyou!

wjones127 (changes requested):
This seems to be reverting the changes from #1253 where we were trying to expand the compatibility of `semver` dependency. Does this method support older versions of `semver`?

wjones127:
cc @AmanKishore

python/python/lancedb/pydantic.py
```diff
@@ -37,7 +37,7 @@
 import pydantic
 import semver
 
-PYDANTIC_VERSION = semver.parse_version_info(pydantic.__version__)
+PYDANTIC_VERSION = semver.Version.parse(pydantic.__version__)
```
wjones127, line 40: Yeah this function was introduced in `semver` version 3.0.0, so using this breaks compatibility with 2.x versions of `semver`.

One alternative I've suggested is using `packaging.version.parse()` from [packaging](). This is a more common dependency than `semvar`, so it's likely users already have it downloaded, and the parse function there has been around since nearly the earliest version of it and supports parsing semver versions.

pmeier: Happy to change the PR to depend on packaging instead. And I agree, that is indeed the much more common dependency.

wjones127: It might be a little more work, since the version object returned isn't quite the same. But that seems like the best solution to me.

pmeier:
Superseded by #1311.

PR #1304, opened 2024-05-14
refactor(nodejs): use biomejs instead of eslint & prettier

universalmind303:
I've been noticing a lot of friction with the current toolchain for '/nodejs'. Particularly with the usage of eslint and prettier. 

[Biome](https://biomejs.dev/) is an all in one formatter & linter that replaces the need for two different ones that can potentially clash with one another. 

I've been using it in the [nodejs-polars](https://github.com/pola-rs/nodejs-polars) repo for quite some time & have found it much more pleasant to work with.


wjones127 (commented):
I'm open to this.

We have pre-commit configuration for prettier right now. Could you switch that to biomejs?

westonpace (approved):
That works for me.  I have no particular attachment to prettier.  I just picked it because whatever we had before didn't really enforce much.  As long as biome is opinionated (don't make me think) and fast then I'm happy.

PR #1308, opened 2024-05-15
feat: js embedding registry

universalmind303:
todo: 
- [ ] add more tests
- [ ] add more examples


nodejs/lancedb/arrow.ts
```diff
@@ -375,12 +376,71 @@ function makeVector(
...
+    table.schema.fields.map((field) => [
+      field.name,
+      table.getChild(field.name)!,
+    ]),
+  );
+
+  for (const functionEntry of functions.values()) {
+    const sourceColumn = columns[functionEntry.sourceColumn];
+    const destColumn = functionEntry.vectorColumn ?? "vector";
+    if (sourceColumn === undefined) {
```
westonpace, line 396: This is just for my own knowledge but how do you decide between `sourceColumn === undefined` and `sourceColumn === undefined || sourceColumn === null` in these validation checks?  (i.e. how do you know when you only need to check for one and not both)

nodejs/lancedb/arrow.ts
```diff
@@ -375,12 +376,71 @@ function makeVector(
...
+      throw new Error(
+        "Embedding function did not return an embedding for each input element",
+      );
+    }
+    let destType: DataType;
+    const dtype = schema.fields.find((f) => f.name === destColumn)!.type;
+    if (dtype instanceof FixedSizeList) {
+      destType = dtype;
+    } else {
+      destType = newVectorType(vectors[0].length, dtype);
```
westonpace, line 425: What does it mean when `dtype` is not an instance of `FixedSizeList`?

nodejs/lancedb/arrow.ts
```diff
@@ -375,12 +376,71 @@ function makeVector(
...
+
+    const vector = makeVector(vectors, destType);
+    columns[destColumn] = vector;
+  }
+  const newTable = new ArrowTable(columns);
+  return alignTable(newTable, schema);
+}
+
 /** Helper function to apply embeddings to an input table */
 async function applyEmbeddings<T>(
```
westonpace, line 436: What's the relationship between `applyEmbeddings` and `appendVectorColumn`?

nodejs/lancedb/connection.ts
```diff
@@ -65,6 +66,8 @@ export interface CreateTableOptions {
    * The available options are described at https://lancedb.github.io/lancedb/guides/storage/
    */
   storageOptions?: Record<string, string>;
+  schema?: Schema;
+  embeddingFunction?: EmbeddingFunctionConfig;
```
westonpace, line 70: Does `open_table` load the embedding config back from the schema?

westonpace, line 70: Minor nit: maybe use the name `embeddingConfig` instead of `embeddingFunction` since there can be multiple functions?

Actually, does this support multiple embedding functions?

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -1,3 +1,7 @@
+import { DataType, Field, FixedSizeList, Float, Float32 } from "apache-arrow";
+import "reflect-metadata";
```
westonpace, line 2: What is this?

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -12,67 +16,138 @@
...
+   *   toJSON() {
+   *     return {
+   *       model: this.model,
+   *       timeout: this.timeout,
+   *     };
+   * }
+   * ```
    */
-  embeddingDimension?: number;
+  abstract toJSON(): Partial<M>;
```
westonpace, line 60: Why is the output `M` and not just `any`?

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -12,67 +16,138 @@
...
+      throw new Error("Datatype is required");
+    }
+    const metadata = new Map<string, EmbeddingFunction>();
+    metadata.set("source_column_for", this);
+
+    if (options instanceof DataType) {
+      return [options, metadata];
+    } else {
+      return [datatype, metadata];
+    }
```
westonpace, line 83: This is a little confusing.  Why does it return either a datatype or an options in the first value?

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -12,67 +16,138 @@
...
+  }
 
   /**
-   * Should the source column be excluded from the resulting table
+   * vectorField is used in combination with `LanceSchema` to provide a declarative data model
```
westonpace, line 87: I'm not sure I understand this comment

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -12,67 +16,138 @@
...
+    const metadata = new Map<string, EmbeddingFunction>();
+    metadata.set("vector_column_for", this);
+
+    return [dtype, metadata];
+  }
+
+  /** The number of dimensions of the embeddings */
+  ndims(): number | undefined {
+    return undefined;
+  }
```
westonpace, line 126: Should this just be abstract?

nodejs/lancedb/embedding/registry.ts
```diff
@@ -0,0 +1,157 @@
+import { EmbeddingFunction } from "./embedding_function";
```
westonpace, line 1: These files need license / copyright headers

nodejs/lancedb/embedding/registry.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/**
+ * This is a singleton class used to register embedding functions
+ * and fetch them by name. It also handles serializing and deserializing.
+ * You can implement your own embedding function by subclassing EmbeddingFunction
+ * or TextEmbeddingFunction and registering it with the registry
+ */
+export class EmbeddingFunctionRegistry {
+  #functions: Map<string, EmbeddingFunctionFactory> = new Map();
```
westonpace, line 24: What does the `#` mean in the name?

nodejs/lancedb/embedding/registry.ts
```diff
@@ -0,0 +1,157 @@
...
+ * Utility function to get the global instance of the registry
+ * @returns `EmbeddingFunctionRegistry` The global instance of the registry
+ * @example
+ * ```ts
+ * const registry = getRegistry();
+ * const openai = registry.get("openai").create();
+ */
+export function getRegistry(): EmbeddingFunctionRegistry {
+  return _REGISTRY;
+}
```
westonpace, line 151: Do we need both this and `getInstance` above?

westonpace (commented):
Some initial questions

nodejs/lancedb/arrow.ts
universalmind303: so I want to eventually get rid of `applyEmbeddings` in favor of `appendVectorColumn` to consolidate functionality & better align with the python implementation. It's essentially `applyEmbeddingsFromSchema` while the current `applyEmbeddings` is `applyEmbeddingsFromEmbeddingFunctions`

nodejs/lancedb/connection.ts
universalmind303: So that's actually a major issue I found while writing some additional tests. When deserializing the table using arrow's functions, it doesn't preserve the metadata like it does in other languages. So I think there's a bit more work that needs to be done on this PR before it can be merged.

universalmind303: the current implementation does not, but it should be pretty easy to add that in a follow up.

nodejs/lancedb/embedding/embedding_function.ts
universalmind303: This is the JS equivalent to python's [func.VectorField](https://github.com/universalmind303/lancedb/blob/90afcc051d35d4d7edc7cdaa832ae353670afe14/python/python/lancedb/embeddings/base.py#L114). I'm open to suggestions on improving the docs here.

nodejs/lancedb/embedding/embedding_function.ts
universalmind303: if it's abstract, it requires the user to provide an implementation, even if the function can return `undefined`. This just provides a default implementation so users don't need to implement it if they dont want.

nodejs/lancedb/embedding/registry.ts
universalmind303: the `#` means it's _actually_ private. unlike the typescript modifier of `private`, the `#` symbol makes it inaccessible from other classes

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

nodejs/lancedb/embedding/registry.ts
universalmind303: I initially ported over this from python, but I agree, we probably don't need both.

nodejs/lancedb/embedding/embedding_function.ts
universalmind303: It's a pretty commonly used library when working with metadata and decorators. 

It allows us to set some metadata on the class prototype. It's used to add the function name to the embedding function if you register it using the decorator. 

such as 

```ts
@register("my-function")
class MyEmbeddingFunction extends EmbeddingFunction<string> {
// ...
}
```
then the `register` function injects some metadata into the class prototype
https://github.com/universalmind303/lancedb/blob/90afcc051d35d4d7edc7cdaa832ae353670afe14/nodejs/lancedb/embedding/registry.ts#L51

which we can later pull back off the class when we convert the embedding functions to metadata 
https://github.com/universalmind303/lancedb/blob/90afcc051d35d4d7edc7cdaa832ae353670afe14/nodejs/lancedb/embedding/registry.ts#L115

nodejs/lancedb/arrow.ts
universalmind303: so `null` usually has to be explicitly set, while `undefined` is just the absence of a value. So as long as we never explicitly start setting values to null, we generally only need to check for `undefined`.

nodejs/lancedb/embedding/embedding_function.ts
universalmind303: I can adjust this if it's intent isn't clear. The idea was to try to allow some level of typechecking for subclasses of `EmbeddingFunction`. So if the subclass supplies the second generic such as is the case with the [OpenAIEmbeddingFunction](https://github.com/universalmind303/lancedb/blob/90afcc051d35d4d7edc7cdaa832ae353670afe14/nodejs/lancedb/embedding/openai.ts#L26), it should return the type matching the one supplied by the subclass.

nodejs/lancedb/embedding/embedding_function.ts
universalmind303: I'll rename to make the intent clearer. 

`options` is actually options or datatype: `Partial<FieldOptions> | DataType`.

nodejs/lancedb/arrow.ts
universalmind303: Good point, we should just error here instead.

wjones127 (commented):
Got some initial comments. There are a some tests I'd like to see, but this is looking well designed so far.

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -12,67 +12,141 @@
...
+  abstract embeddingDataType(): Float;
 
   /**
    * Creates a vector representation for the given values.
    */
-  embed: (data: T[]) => Promise<number[][]>;
-}
+  abstract computeSourceEmbeddings(
+    data: T[],
+  ): Promise<number[][] | Float32Array[] | Float64Array[]>;
```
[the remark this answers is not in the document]
universalmind303: These are the JS native `Float32Array` and `Float64Array`, not the arrow arrays. There is no native `Float16Array` in JS.

universalmind303:
@wjones127 
> Would a user override this? I'm wondering if it would be more appropriate as a private free function, so it's hidden from the users.

In JS, you can override anything, anywhere if you try hard enough. So we can't explicitly forbid a user from overriding this, but declaring it as not `abstract` generally means it's not meant to be overridden.

nodejs/lancedb/embedding/index.ts
```diff
@@ -1,2 +1,105 @@
...
+ *   vector: func.vectorField(),
+ *   // optional: specify the datatype and/or dimensions
+ *   vector2: func.vectorField({ datatype: new Float32(), dims: 3}),
+ * });
+ *
+ * const table = await db.createTable("my_table", data, { schema });
+ * ```
+ */
+export function LanceSchema(
+  fields: Record<string, [DataType, Map<string, EmbeddingFunction>] | DataType>,
```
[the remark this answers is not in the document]
universalmind303: String keys are ordered by insertion order, so this should always be deterministic.

nodejs/__test__/table.test.ts
```diff
@@ -420,6 +424,111 @@ describe("when dealing with versioning", () => {
   });
 });
 
+describe("embedding functions", () => {
```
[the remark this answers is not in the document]
universalmind303: similar to the python api, this'll throw when adding, not on `open`

nodejs/lancedb/embedding/embedding_function.ts
universalmind303: In JS, you can override anything, anywhere if you try hard enough. 

So we can't explicitly forbid a user from overriding this, but not declaring it as `abstract` generally means it's not meant to be overridden.

universalmind303:
Getting unrelated build errors from the lance-jni crate

```
 error[E0026]: variant `lance::Error::IO` does not have a field named `message`
   --> java/core/lancedb-jni/src/error.rs:174:32
```

nodejs/lancedb/embedding/embedding_function.ts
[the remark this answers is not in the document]
wjones127: Oh I see.

nodejs/__test__/table.test.ts
```diff
@@ -420,6 +424,161 @@ describe("when dealing with versioning", () => {
...
+    // biome-ignore lint/suspicious/noExplicitAny: test
+    const arr = (await table.query().toArray()) as any;
+    expect(arr[0].vector).toBeDefined();
+
+    // we round trip through JSON to make sure the vector properly gets converted to an array
+    // otherwise it'll be a TypedArray or Vector
+    const vector0 = JSON.parse(JSON.stringify(arr[0].vector));
+    expect(vector0).toEqual([1, 2, 3]);
+  });
+  it("should provide an error when opening a table with an unregistered embedding function", async () => {
```
wjones127, line 530: Since the behavior is explicitly not to error on opening, we should be clear about that.
```suggestion
  it("should error when appending to a table with an unregistered embedding function", async () => {
```

wjones127 (approved):
This is looking good now. Thanks for the additional tests :)

PR #1311, opened 2024-05-16
chore: replace semver dependency with packaging

### LINKED ISSUE #1296: Fix semvar deprication warning

pmeier:
Fixes #1296 per title. See https://github.com/lancedb/lancedb/pull/1298#discussion_r1603931457 Cc @wjones127


python/python/lancedb/pydantic.py
```diff
@@ -35,13 +35,13 @@
 import numpy as np
 import pyarrow as pa
 import pydantic
-import semver
+from packaging.version import Version
 
-PYDANTIC_VERSION = semver.parse_version_info(pydantic.__version__)
+PYDANTIC_VERSION = Version(pydantic.__version__)
```
pmeier, line 40: This is technically a BC break since this is a public attribute, i.e. `lancedb.pydantic.PYDANTIC_VERSION`. Thoughts?

wjones127: Eh I don't think anyone should be relying on this, and it's easy for them to change their code to not. So I don't feel the need to mark it as a breaking change.

wjones127 (approved):
Looks good. Thank you!

wjones127:
Thanks @pmeier. Apologies for the delay in merging.

PR #1319, opened 2024-05-23
fix: parse index_stats for scalar index

QianZhu:
parse the index stats for scalar index - it is different from the index stats for vector index

### DIFF
--- a/rust/lancedb/src/index/vector.rs
+++ b/rust/lancedb/src/index/vector.rs
@@ -53,9 +53,14 @@ pub struct VectorIndexMetadata {
 }
 
 #[derive(Debug, Deserialize)]
-pub struct VectorIndexStatistics {
+pub struct IndexMetadata {
     pub num_indexed_rows: usize,
     pub num_unindexed_rows: usize,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct VectorIndexStatistics {
+    pub num_indexed_and_unindexed_rows: IndexMetadata,
     pub index_type: String,
     pub indices: Vec<VectorIndexMetadata>,
 }
--- a/rust/lancedb/src/index/vector.rs
+++ b/rust/lancedb/src/index/vector.rs
@@ -24,6 +24,7 @@ use serde::Deserialize;
 use lance::table::format::{Index, Manifest};
 
 use crate::DistanceType;
+use serde_with::skip_serializing_none;
 
 pub struct VectorIndex {
     pub columns: Vec<String>,
@@ -46,22 +47,20 @@ impl VectorIndex {
     }
 }
 
+#[skip_serializing_none]
 #[derive(Debug, Deserialize)]
 pub struct VectorIndexMetadata {
-    pub metric_type: String,
-    pub index_type: String,
+    pub metric_type: Option<String>,
+    pub index_type: Option<String>,
 }
 
-#[derive(Debug, Deserialize)]
-pub struct IndexMetadata {
-    pub num_indexed_rows: usize,
-    pub num_unindexed_rows: usize,
-}
-
+#[skip_serializing_none]
 #[derive(Debug, Deserialize)]
 pub struct VectorIndexStatistics {
-    pub num_indexed_and_unindexed_rows: IndexMetadata,
-    pub index_type: String,
+    // pub row_stats: IndexRowStatistics,
+    pub num_indexed_rows: usize,
+    pub num_unindexed_rows: usize,
+    pub index_type: Option<String>,
     pub indices: Vec<VectorIndexMetadata>,
 }
 
--- a/rust/lancedb/src/index/vector.rs
+++ b/rust/lancedb/src/index/vector.rs
@@ -19,12 +19,10 @@
 //! values
 use std::cmp::max;
 
-use serde::Deserialize;
-
 use lance::table::format::{Index, Manifest};
 
 use crate::DistanceType;
-use serde_with::skip_serializing_none;
+
 
 pub struct VectorIndex {
     pub columns: Vec<String>,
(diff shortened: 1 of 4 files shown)


rust/lancedb/src/table.rs
```diff
@@ -1263,6 +1266,25 @@ impl NativeTable {
...
+            .into_iter()
+            .find(|i| i.index_uuid == index_uuid);
+        if index.is_none() {
+            return Ok(None);
+        }
+        let dataset = self.dataset.get().await?;
+        let index_stats = dataset.index_statistics(&index.unwrap().index_name).await?;
+        let index_stats: IndexMetadata = whatever!(
+            serde_json::from_str(&index_stats),
+            "error deserializing index statistics {index_stats}",
```
rpgreen, line 1282: do you need to use format! here?

rust/lancedb/src/table.rs
```diff
@@ -1263,6 +1266,25 @@ impl NativeTable {
         Ok(Some(index_stats))
     }
 
+    async fn load_index_stats(&self, index_uuid: &str) -> Result<Option<IndexMetadata>> {
+        let index = self
```
rpgreen, line 1270: Almost all of this is duplicated from above - please refactor

rust/lancedb/src/index/vector.rs
```diff
@@ -53,9 +53,14 @@ pub struct VectorIndexMetadata {
...
 #[derive(Debug, Deserialize)]
-pub struct VectorIndexStatistics {
+pub struct IndexMetadata {
     pub num_indexed_rows: usize,
     pub num_unindexed_rows: usize,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct VectorIndexStatistics {
+    pub num_indexed_and_unindexed_rows: IndexMetadata,
```
rpgreen, line 63: naming nits:

num_indexed_and_unindexed_rows is overly specific, and "Metadata" is somewhat redundant

how about:
```
row_stats: IndexRowStatistics
```

rust/lancedb/src/index/vector.rs
```diff
@@ -46,17 +47,20 @@ impl VectorIndex {
     }
 }
 
+#[skip_serializing_none]
 #[derive(Debug, Deserialize)]
 pub struct VectorIndexMetadata {
-    pub metric_type: String,
-    pub index_type: String,
+    pub metric_type: Option<String>,
```
eddyxu, line 53: why do we need to change this to optional? There are index that do not have these fields?

rust/lancedb/src/index/vector.rs
```diff
@@ -46,17 +47,20 @@ impl VectorIndex {
...
+    pub index_type: Option<String>,
 }
 
+#[skip_serializing_none]
 #[derive(Debug, Deserialize)]
 pub struct VectorIndexStatistics {
+    // pub row_stats: IndexRowStatistics,
```
eddyxu, line 60: why comment out this field?

rust/lancedb/src/index/vector.rs
QianZhu: correct, scalar index stats does not have this field.

rust/lancedb/src/index/vector.rs
QianZhu: it should be just deleted.

rust/lancedb/src/index/vector.rs
eddyxu: But this struct is called `VectorIndexMetadata`. Do we want to create a new `ScalarIndexMetadata`, or consolidate into one `IndexMetadata`?

rust/lancedb/src/index/vector.rs
```diff
@@ -46,17 +47,19 @@ impl VectorIndex {
...
 pub struct VectorIndexMetadata {
-    pub metric_type: String,
-    pub index_type: String,
+    pub metric_type: Option<String>,
+    pub index_type: Option<String>,
 }
 
+#[skip_serializing_none]
 #[derive(Debug, Deserialize)]
 pub struct VectorIndexStatistics {
```
eddyxu, line 59: Also here, are we carrying ScalarIndexStats from `VectorIndexStats`?

rust/lancedb/src/index/vector.rs
rpgreen: Probably this should not be called VectorIndexStatistics if it applies to scalar indexes too.

It would definitely be cleaner to have separate ScalarIndexMetadata and VectorIndexMetadata types instead of using optional fields. Is that feasible?

rust/lancedb/src/index/vector.rs
```diff
@@ -46,17 +47,19 @@ impl VectorIndex {
...
     pub num_indexed_rows: usize,
     pub num_unindexed_rows: usize,
-    pub index_type: String,
+    pub index_type: Option<String>,
```
rpgreen, line 62: Can this be an enum?

rust/lancedb/src/index/vector.rs
QianZhu: Yeah, it should be renamed to `IndexStatistics`
we plan to use createIndex for both vector and scalar index in the future so prob it is a good idea to be unified as `IndexMetadata`.

rust/lancedb/src/index/vector.rs
QianZhu: will change the name.

rust/lancedb/src/index/vector.rs
QianZhu: I would like to have it as enum. But it might require changes from how we write index stats for lance dataset. Currently, for a vector index, this index type is "IVF" but for a scalar index, this index type is None - ideally, it should be "BTree"

eddyxu (approved):

Issue #1288, opened 2024-05-10
Feature: explain_plan


wjones127:
### SDK

Python

### Description

For debugging purposes, it's often nice to see the underlying query plan. Lance scanner has `explain_plan`, and it wouldn't be hard to expose that in our LanceDB SDKs.

This could be in all three SDKs.

wjones127:
Hi @nuvic. Feel free to make a PR for this. mention me in it when you are ready for review or have some questions.

This PR from Weston might be a good reference: https://github.com/lancedb/lancedb/pull/1257 There he exposes a different function in the Python async and Node `lancedb` package.

PR #1328, opened 2024-05-28
feat: add the explain_plan function

nuvic:
It's useful to see the underlying query plan for debugging purposes. This exposes LanceScanner's `explain_plan` function. Addresses #1288

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -430,7 +430,7 @@ class LanceQueryBuilder(ABC):
         >>> print(plan) # doctest: +ELLIPSIS, +NORMALIZE_WHITESPACE
         Projection: fields=[vector, _distance]
           KNNFlat: k=10 metric=l2
-            LanceScan: uri=..., projection=[vector], row_id=true, ordered=false
+            LanceScan: uri=..., projection=[vector], row_id=true, row_addr=false, ordered=false
 
         Parameters
         ----------
@@ -440,7 +440,7 @@ class LanceQueryBuilder(ABC):
         Returns
         -------
         plan : str
-        """
+        """ # noqa: E501
         ds = self._table.to_lance()
         return ds.scanner(
             nearest={
@@ -1224,7 +1224,7 @@ class AsyncQueryBase(object):
         Returns
         -------
         plan : str
-        """
+        """ # noqa: E501
         return await self._inner.explain_plan(verbose)
 
 
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1214,7 +1214,7 @@ class AsyncQueryBase(object):
         >>> asyncio.run(doctest_example()) # doctest: +ELLIPSIS, +NORMALIZE_WHITESPACE
         Projection: fields=[vector, _distance]
           KNNFlat: k=10 metric=l2
-            LanceScan: uri=..., projection=[vector], row_id=true, ordered=false
+            LanceScan: uri=..., projection=[vector], row_id=true, row_addr=false, ordered=false
 
         Parameters
         ----------
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1198,6 +1198,9 @@ class AsyncQueryBase(object):
         """
         return (await self.to_arrow()).to_pandas()
 
+    async def explain_plan(self, verbose: Optional[bool] = False):
+        return await self._inner.explain_plan(verbose)
+
 
 class AsyncQuery(AsyncQueryBase):
     def __init__(self, inner: LanceQuery):
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -440,7 +440,7 @@ class LanceQueryBuilder(ABC):
         Returns
         -------
         plan : str
-        """ # noqa: E501
+        """  # noqa: E501
         ds = self._table.to_lance()
         return ds.scanner(
             nearest={
@@ -1224,7 +1224,7 @@ class AsyncQueryBase(object):
         Returns
         -------
         plan : str
-        """ # noqa: E501
+        """  # noqa: E501
         return await self._inner.explain_plan(verbose)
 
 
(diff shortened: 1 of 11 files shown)


nuvic:
Hi @wjones127, I've added the `explain_plan` function to the python package. Just wanted to check that this is the correct approach before I continue. Thanks

python/python/lancedb/query.py
```diff
@@ -417,6 +417,35 @@ def with_row_id(self, with_row_id: bool) -> LanceQueryBuilder:
         self._with_row_id = with_row_id
         return self
 
+    def explain_plan(self, verbose: Optional[bool] = False) -> str:
```
nuvic, line 420: Thanks, that makes more sense. Added to `LanceQueryBuilder`

python/python/lancedb/query.py
```diff
@@ -417,6 +417,35 @@ def with_row_id(self, with_row_id: bool) -> LanceQueryBuilder:
...
+        """Return the execution plan for this query.
+
+        Examples
+        --------
+        >>> import lancedb
+        >>> db = lancedb.connect("./.lancedb")
+        >>> table = db.open_table("my_table")
```
wjones127, line 427: For the example to work, you'll need to create the table:

```suggestion
        >>> import lancedb
        >>> db = lancedb.connect("./.lancedb")
        >>> db.create_table("my_table", [{"vector": [99, 99]}])
        >>> table = db.open_table("my_table")
```

python/python/lancedb/query.py
```diff
@@ -417,6 +417,35 @@ def with_row_id(self, with_row_id: bool) -> LanceQueryBuilder:
...
+        >>> query = [100, 100]
+        >>> plan = table.search(query).explain_plan(True)
+        >>> print(plan)
```
wjones127, line 430: To make the example more informative and get the example to pass, you should print the output. Also, you can use `...` as a wildcard so you don't have to write it all out.

```suggestion
        >>> table.search(query).explain_plan(True)
        >>> print(plan)
        Projection: fields=[i, s, vec, _distance]
          Take: columns=\"_distance, _rowid, vec, i, s\"
            SortExec: TopK(fetch=5), expr=...
              KNNIndex: name=..., k=5, deltas=1
                ScalarIndexQuery: query=i > 10
```

To run the doctests locally, you can run:

```bash
pytest --doctest-modules python/lancedb
```

wjones127 (commented):
There's a few more changes you'll need to make to get the doctests to pass, but otherwise looks good.

python/python/lancedb/query.py
nuvic: Oh I see. Okay I've updated the examples so they pass the doctest

nuvic:
I'm trying to figure out how to add this to the nodejs sdk. Looking at `nodejs/src`, there doesn't seem to be a `Scanner` exposed there, so it's not as simple as the python solution.
Since we want to put `explain_plan` in the `Query` data structure (`nodejs/src/query.rs`), we need to get access to a `dataset` ref there, to get to the scanner. What do you recommend?

wjones127:
> I'm trying to figure out how to add this to the nodejs sdk. Looking at `nodejs/src`, there doesn't seem to be a `Scanner` exposed there, so it's not as simple as the python solution.

This is true. Although you've only implemented for the legacy Python sync API, which wraps pylance. In the future, we'll also want to add to the new Python *async* API. (Eventually the legacy Python sync API will be replaced by wrapping the async API, and LanceDB will no longer depend on pylance.) The async API also doesn't have a scanner directly exposed.

> Since we want to put `explain_plan` in the `Query` data structure (`nodejs/src/query.rs`), we need to get access to a `dataset` ref there, to get to the scanner. What do you recommend?

You will need to add a `explain_plan` method to the Rust `Query` structure here:

https://github.com/lancedb/lancedb/blob/56b4fd2bd9c5c3077c1c7f6048780aecfbc577e3/nodejs/src/query.rs#L33

And that should call an `explain_plan` method you'll have to add to to the `ExecutableQuery` trait in this file:

https://github.com/lancedb/lancedb/blob/56b4fd2bd9c5c3077c1c7f6048780aecfbc577e3/rust/lancedb/src/query.rs#L430

Doing this for Python async API will be very similar to this.

nuvic:
Thanks for the explanation above. I've now implemented the method for the python async API and the nodejs API. Could you take a look?

rust/lancedb/src/table.rs
```diff
@@ -1756,6 +1757,37 @@ impl TableInternal for NativeTable {
             .await
     }
 
+    async fn explain_plan(&self, query: &VectorQuery, verbose: bool) -> Result<String> {
```
wjones127, line 1760: I feel like this replicates some (but not all) of the logic from `create_plan` above. Could we instead pull this out to some common function (`build_plan`?) and then both `create_plan` and `explain_plan` can be built off of those?

nuvic: Good point. Pulled out common functionality here: https://github.com/lancedb/lancedb/pull/1328/commits/6dc8ced985c1798edf3c5f927d80335ae561716a

nuvic:
Addressed the lint and test errors here:
- https://github.com/lancedb/lancedb/pull/1328/commits/b8702de5162ff87a0e9a3ca25ff399dfeca208bc
- https://github.com/lancedb/lancedb/pull/1328/commits/8e96a57d6508d6ea4fcae13b4fb16a91fdc523bf
- https://github.com/lancedb/lancedb/pull/1328/commits/a966c4775d821a56e95cb89e02485fb706972e23

nuvic:
Addressed the new lint errors introduced by recent changes.
https://github.com/lancedb/lancedb/pull/1328/commits/fca314185672e1b6a287eed4937fe76c33c5d435
https://github.com/lancedb/lancedb/pull/1328/commits/4d58d6f9fa06e343db6994b3c98475508d127ed6

Missed some of these because I couldn't run `cargo clippy --all --all-features -- -D warnings` locally (unable to build AVX2 f16 kernels). I'll try to figure this out.

nuvic:
Okay I got the cargo clippy all command to work locally. Fixed the remaining lint issue: https://github.com/lancedb/lancedb/pull/1328/commits/f27d4684065d73ae3e982ac1a80950dba52f3c1a

wjones127 (approved):
This is great work! Thanks for sticking with it.

PR #1333, opened 2024-05-29
fix: the bug of async connection context manager

ChenZhongPu:
- add `return` for `__enter__`

The buggy code didn't return the object, therefore it will always return None within a context manager:

```python
with await lancedb.connect_async("./.lancedb") as db:
        # db is always None
```

(BTW, why not to design an async context manager?)

- add a unit test for Async connection context manager

- update return type of `AsyncConnection.open_table` to `AsyncTable`

Although type annotation doesn't affect the functionality, it is helpful for IDEs.

### DIFF
--- a/python/python/lancedb/db.py
+++ b/python/python/lancedb/db.py
@@ -509,7 +509,7 @@ class AsyncConnection(object):
         return self._inner.__repr__()
 
     def __enter__(self):
-        self
+        return self
 
     def __exit__(self, *_):
         self.close()
@@ -779,7 +779,7 @@ class AsyncConnection(object):
         name: str,
         storage_options: Optional[Dict[str, str]] = None,
         index_cache_size: Optional[int] = None,
-    ) -> Table:
+    ) -> AsyncTable:
         """Open a Lance Table in the database.
 
         Parameters
--- a/python/python/tests/test_db.py
+++ b/python/python/tests/test_db.py
@@ -297,6 +297,13 @@ async def test_close(tmp_path):
 
 
 @pytest.mark.asyncio
+async def test_context_manager(tmp_path):
+    with await lancedb.connect_async(tmp_path) as db:
+        assert db.is_open()
+    assert not db.is_open()
+
+
+@pytest.mark.asyncio
 async def test_create_mode_async(tmp_path):
     db = await lancedb.connect_async(tmp_path)
     data = pd.DataFrame(


python/python/lancedb/db.py
```diff
@@ -509,7 +509,7 @@ def __repr__(self):
         return self._inner.__repr__()
 
     def __enter__(self):
-        self
+        return self
```
westonpace, line 512: Yikes! Thanks for catching this. I was silly and coded rust in python.

python/python/lancedb/remote/table.py
```diff
@@ -354,19 +354,19 @@ def _do_merge(
...
         params["when_not_matched_by_source_delete"] = str(
             merge._when_not_matched_by_source_delete
         ).lower()
         if merge._when_not_matched_by_source_condition is not None:
-            params[
-                "when_not_matched_by_source_delete_filt"
-            ] = merge._when_not_matched_by_source_condition
+            params["when_not_matched_by_source_delete_filt"] = (
+                merge._when_not_matched_by_source_condition
+            )
```
westonpace, line 369: Can you revert the changes to this file?  They are breaking our lint job (maybe you are using a different version of `ruff`?)

If it becomes trouble then let me know and I will fix after the fact.

westonpace (changes requested):
Thank you for the contribution.  We just need to fix formatting change and then we are good to go.

python/python/lancedb/remote/table.py
ChenZhongPu: I am using ruff 0.4.6,  and I am afraid the lint job does not use the latest ruff. I will revert those changes.

BTW, if your lint job is using a specific version of ruff, I think it would be better to specify it in README file.

wjones127 (commented):
Could we also change at least one unit test to use the context manager?

ChenZhongPu:
> Could we also change at least one unit test to use the context manager?

Good idea. But I have some issues to run the unit test, and don't know why. I also posted the question on [discussion](https://github.com/lancedb/lancedb/discussions/1327#discussioncomment-9587227).

westonpace:
> (BTW, why not to design an async context manager?)

Do you mean using the decorator: https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager ?  If so, then the answer is that I didn't know it existed.  I think it would be ok to use here.

ChenZhongPu:
> > (BTW, why not to design an async context manager?)
> 
> Do you mean using the decorator: https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager ? If so, then the answer is that I didn't know it existed. I think it would be ok to use here.

No, the decorator is just a syntax sugar for `__aenter__` and `__aexit`. 

In fact, LanceDB now adopts a context manager, *NOT* an async one. As for a context manager, we can use `with`, while as for an async context manager, we have to use `async with`. Therefore, switching to an async context manager would introduce a *break change* in API.

wjones127 (approved):

westonpace (approved):

westonpace:
> In fact, LanceDB now adopts a context manager, NOT an async one

Ok, I see.  Well, the `close` method is sync, and the `__enter__` happens in the constructor of the `AsyncConnection` which happens after the asynchronous part of `connect_async` and so I think a synchronous connection manager is ok?  I'm not an expert though.

PR #1335, opened 2024-05-29
feat: add support for new cohere models in cohere and bedrock embedding functions

### LINKED ISSUE #1329: bug(python): Cohere embedding functions use incorrect input type for query embeddings
### LanceDB version

v0.8.0

### What happened?

Cohere embeddings for search and retrieval prepend special tokens to differentiate embedding types. This is specified by `input_type` parameter in the [Cohere API](https://docs.cohere.com/reference/embed) and in the [Bedrock API for Cohere embedding models](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html).

Neither of the `generate_embeddings` methods in the [CohereEmbeddingFunction](https://github.com/lancedb/lancedb/blob/python-v0.8.0/python/python/lancedb/embeddings/cohere.py#L70C9-L85) and [BedrockText](https://github.com/lancedb/lancedb/blob/python-v0.8.0/python/python/lancedb/embeddings/bedrock.py#L106-L153) classes allow for the correct use of the `input_type` parameter, even though there are distinct `compute_query_embeddings` and `compute_source_embeddings` methods that could control the correct `input_type` value. The `BedrockText` class even [hardcodes the `"search_document"` input type for all embeddings](https://github.com/lancedb/lancedb/blob/python-v0.8.0/python/python/lancedb/embeddings/bedrock.py#L144-L148). This leads to a noticeable performance reduction when evaluating search and retrieval pipelines that use these LanceDB embedding functions when compared to generating the embedding vectors externally with the correct `input_type` parameter and passing the vectors into LanceDB.

I'd propose we expose the model keyword arguments, specifically `input_type` parameter to the `compute_query_embeddings` and `compute_source_embeddings` methods for specifying the correct `input_type` for each case.

- `compute_source_embeddings` methods should use `"search_document"` for encoding documents for embeddings that are stored in the database
- `compute_query_embeddings` methods should use `"search_query"` when querying the database for relevant documents.

_No response_

AyushExel:
Fixes #1329

Will update docs on https://github.com/lancedb/lancedb/pull/1326

### DIFF
--- a/python/python/lancedb/embeddings/bedrock.py
+++ b/python/python/lancedb/embeddings/bedrock.py
@@ -89,7 +89,11 @@ class BedRockText(TextEmbeddingFunction):
         # TODO: fix hardcoding
         if self.name == "amazon.titan-embed-text-v1":
             return 1536
-        elif self.name in ["amazon.titan-embed-text-v2:0", "cohere.embed-english-v3", "cohere.embed-multilingual-v3"]:
+        elif self.name in [
+            "amazon.titan-embed-text-v2:0",
+            "cohere.embed-english-v3",
+            "cohere.embed-multilingual-v3",
+        ]:
             # TODO: "amazon.titan-embed-text-v2:0" model supports dynamic ndims
             return 1024
         else:
@@ -105,9 +109,7 @@ class BedRockText(TextEmbeddingFunction):
     ) -> List[List[float]]:
         texts = self.sanitize_input(texts)
         # assume source input type if not passed by `compute_query_embeddings`
-        kwargs["input_type"] = ( 
-            kwargs.get("input_type") or self.source_input_type
-        )
+        kwargs["input_type"] = kwargs.get("input_type") or self.source_input_type
 
         return self.generate_embeddings(texts, **kwargs)
 
--- a/python/python/lancedb/embeddings/bedrock.py
+++ b/python/python/lancedb/embeddings/bedrock.py
@@ -73,6 +73,8 @@ class BedRockText(TextEmbeddingFunction):
     assumed_role: Union[str, None] = None
     profile_name: Union[str, None] = None
     role_session_name: str = "lancedb-embeddings"
+    source_input_type: str = "search_document"
+    query_input_type: str = "search_query"
 
     if PYDANTIC_VERSION.major < 2:  # Pydantic 1.x compat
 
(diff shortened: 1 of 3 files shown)


tanaymeh (approved):

python/python/lancedb/embeddings/cohere.py
```diff
@@ -61,14 +89,39 @@ class TextModel(LanceModel):
...
+            return 384
+        elif self.name == "embed-english-v2.0":
+            return 4096
+        elif self.name == "embed-multilingual-v2.0":
+            return 768
+        else:
+            raise ValueError(f"Model {self.name} not supported")
+
+    def compute_query_embeddings(self, query: str, *args, **kwargs) -> List[np.array]:
+        return self.compute_source_embeddings(query, input_type=self.query_input_type)
```
huffmanjohnf, line 114: [Same as above] Non-blocking: Should this use the same `kwargs.get("input_type") or self.query_input_type` logic as in `compute_source_embeddings`? 
This logic is perfectly fine for the typical search/retrieval case. I'm much less familiar with the other use cases for vector databases but I would imagine other input types (such as `"clustering"`) might be useful.

python/python/lancedb/embeddings/bedrock.py
```diff
@@ -87,21 +89,29 @@ def ndims(self):
...
             return 1024
         else:
-            raise ValueError(f"Unknown model name: {self.name}")
+            raise ValueError(f"Model {self.name} not supported")
 
     def compute_query_embeddings(
         self, query: str, *args, **kwargs
     ) -> List[List[float]]:
-        return self.compute_source_embeddings(query)
+        return self.compute_source_embeddings(query, input_type=self.query_input_type)
```
huffmanjohnf, line 105: Non-blocking: Should this use the same `kwargs.get("input_type") or self.query_input_type` logic as in `compute_source_embeddings`? 
This logic is perfectly fine for the typical search/retrieval case. I'm much less familiar with the other use cases for vector databases but I would imagine other input types (such as `"clustering"`) might be useful.

huffmanjohnf (commented):
These changes look great! Thank you for handling this so quickly. Just left a few small, non-blocking comments.

In general, are there unit tests coverage requirements for these changes?

python/python/lancedb/embeddings/cohere.py
AyushExel: Yeah these are tested in slow tests [here](https://github.com/lancedb/lancedb/blob/main/python/python/tests/test_embeddings_slow.py) .
These functions are called internally with the right params, so you can set both source and query input types of your choice. Like this:
```
import lancedb

cohere = get_registry.get("cohere").create(name="model_name", source_input_type="classification", query_input_type="clustering")
```

See docs here- https://lancedb.github.io/lancedb/embeddings/default_embedding_functions/#ollama-embeddings

I realized we missed adding cohere docs, which I'm addressing in another PR, but you can see all accepted args that you can pass here - https://github.com/lancedb/lancedb/pull/1326/files#diff-e4f8e4ca897425fa2aa42069418ab1b10a35b3b8233e30ab278d776ffb96fddcR385

python/python/lancedb/embeddings/bedrock.py
```diff
@@ -73,6 +73,8 @@ class TextModel(LanceModel):
     assumed_role: Union[str, None] = None
     profile_name: Union[str, None] = None
     role_session_name: str = "lancedb-embeddings"
+    source_input_type: str = "search_document"
+    query_input_type: str = "search_query"
```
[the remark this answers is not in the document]
AyushExel: Ahh yes makes sense.. I'll add a one-liner here and also a detailed version in the other docs PR . thanks!!

python/python/lancedb/embeddings/cohere.py
huffmanjohnf: Ahh ok, I'm following. Yea, what you have for setting the input types when creating the embedding function is even better. Thanks for the explanation!

huffmanjohnf (approved):

PR #1339, opened 2024-05-30
fix: error msg when query vector dim is wrong

QianZhu:
- changed the error msg for table.search with wrong query vector dim 
- added missing fields for listIndices and indexStats to be consistent with Python API - will make changes in node integ test

### DIFF
--- a/node/src/index.ts
+++ b/node/src/index.ts
@@ -704,6 +704,9 @@ export interface VectorIndex {
 export interface IndexStats {
   numIndexedRows: number | null
   numUnindexedRows: number | null
+  index_type: string | null
+  distance_type: string | null
+  completed_at: string | null
 }
 
 /**
--- a/node/src/remote/index.ts
+++ b/node/src/remote/index.ts
@@ -509,7 +509,8 @@ export class RemoteTable<T = number[]> implements Table<T> {
     return (await results.body()).indexes?.map((index: any) => ({
       columns: index.columns,
       name: index.index_name,
-      uuid: index.index_uuid
+      uuid: index.index_uuid,
+      status: index.status
     }))
   }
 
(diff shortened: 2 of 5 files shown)


rust/lancedb/src/utils.rs
```diff
@@ -118,14 +118,14 @@ pub(crate) fn default_vector_column(schema: &Schema, dim: Option<i32>) -> Result
         })
         .collect::<Vec<_>>();
     if candidates.is_empty() {
-        Err(Error::Schema {
-            message: "No vector column found to create index".to_string(),
+        Err(Error::InvalidInput {
+            message: "No vector column found to match with the query vector dimension".to_string(),
```
eddyxu, line 122: can we print the explicit dimension as well?

`format!("No vector column found to match with the query vector dimension: {}", dim.unwrap_or_default())`

node/src/index.ts
```diff
@@ -704,6 +704,9 @@ export interface VectorIndex {
 export interface IndexStats {
   numIndexedRows: number | null
   numUnindexedRows: number | null
+  index_type: string | null
```
eddyxu, line 707: Why are these needed in this PR?

QianZhu: not related to the vec dim. But just noticed that we did not make changes with node sdk - so wanted to add them

eddyxu (approved):

node/src/remote/index.ts
```diff
@@ -520,7 +521,10 @@ export class RemoteTable<T = number[]> implements Table<T> {
     const body = await results.body()
     return {
       numIndexedRows: body?.num_indexed_rows,
-      numUnindexedRows: body?.num_unindexed_rows
+      numUnindexedRows: body?.num_unindexed_rows,
+      index_type: body?.index_type,
```
rpgreen, line 525: why are we using a mix of camel case and snake case here?

Issue #1256, opened 2024-04-30
Feature(nodejs):  `Table.search` function for `lancedb` sdk


universalmind303:
### SDK

Node

### Description

both the python sdk, and the older js `vectordb` sdk have a `search` function on the table class. 

`tbl.search(query)`.

There does not appear to be an equivalent in the newer `lancedb` sdk

wjones127:
`Table.query()` is the equivalent of this. I just looked, and I think we already have feature parity here, so I'm closing this.

universalmind303:
@wjones127 There isn't full feature parity. In python you don't need to manually embed the query. 


```py
query = "greetings"
table.search(query).limit(1)
```

```js
const query = await func.computeQueryEmbeddings("greetings");
    
await table.query().nearestTo(query).limit(1)
```

PR #1341, opened 2024-05-30
feat(nodejs): table.search functionality

universalmind303:
closes https://github.com/lancedb/lancedb/issues/1256

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -613,3 +613,86 @@ describe("when optimizing a dataset", () => {
     expect(stats.prune.oldVersionsRemoved).toBe(3);
   });
 });
+
+describe("table.search", () => {
+  let tmpDir: tmp.DirResult;
+  beforeEach(() => {
+    tmpDir = tmp.dirSync({ unsafeCleanup: true });
+  });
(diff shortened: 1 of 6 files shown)


universalmind303:
@wjones127 
> My main question is whether we want to have a new search() method, or are we fine just using query()? Or, put another way, why not just make nearestTo take the string/embedd-able data?

good question. We could instead make `query` take in the string parameter instead. 

I think intuitively `.query("some string")` feels much more natural than starting the builder first. `.query().nearestTo("some string")`.

wjones127:
> I think intuitively .query("some string") feels much more natural than starting the builder first.

I think the idea is that `query` is more general than `search`. The `query` API can also be used to do a `scan`.

wjones127 (approved):
I think I'm good with having the separate `search()` API given the precedence in the Python sync API.

universalmind303:
@wjones127 @westonpace Do we have any way to run CI without cache? It seems like it's using cached code 

>  `Error: lancedb/table.ts(146,7): error TS2663: Cannot find name 'schema'. Did you mean the instance member 'this.schema'?`

`lancedb/table.ts(146,7)` has no reference to `schema`

Issue #1293, opened 2024-05-10
bug(node, lancedb): unable to 'add' to a table created with a `Float64` vector


universalmind303:
### LanceDB version

@lancedb/lancedb main

### What happened?

if you try to create a table with a specific schema, then try to add data to it, it seems that the arrow conversion does not account for the schema and tries to instead convert the array to a `Float32` vector

### Are there known steps to reproduce?

```js
import lancedb from "@lancedb/lancedb";


const db = await lancedb.connect("/tmp/lancedb");

const data = [
  {text: "hello", vector: Array(512).fill(1.0)},
  {text: "hello world", vector: Array(512).fill(1.0)},
]
const f64Schema = new Schema([
  new Field("text", new Utf8(), true),
  new Field("vector", new FixedSizeList(512, new Field("item", new Float64())), true)
]);

let f64Table = await db.createEmptyTable("f64", f64Schema, {mode: "overwrite"})

await f64Table.add(data)
```

```
[Error: Failed to add batches to table f64: lance error: Append with different schema: `vector` should have type fixed_size_list:double:512 but type was fixed_size_list:float:512, location: /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lance-core-0.10.18/src/datatypes/schema.rs:150:27] {
```

Macbook-Specter:
I had the same problem. I couldn't create an empty table and then add data to it first.
 "@lancedb/lancedb": "^0.5.0"
"apache-arrow": "^16.1.0"
```
const uri = "./db/sample-lancedb";
const db = await lancedb.connect(uri);
const document_fragment_schema = new Schema([
    new Field("id", new Float32(), false),
    new Field("fragment_content", new Utf8(), false),
    new Field("docment_id", new Float32(), false),
    new Field("vector", new FixedSizeList(2, new Field("vector_values", new Float32(), false)), false),
]);
const tb = db.createEmptyTable("document_fragment", document_fragment_schema);
await tb.add(
    [
        {
            id: 2,
            fragment_content: "hello world2",
            docment_id: 1,
            vector: [1, 2]
        },
        {
            id: 3,
            fragment_content: "hello world3",
            docment_id: 1,
            vector: [4, 8]
        },
    ]
);
const tb_schema = await tb.schema();
console.log(tb_schema);

const res = await tb.vectorSearch([5,6]).limit(3).toArray();
console.log(res);
```
node:internal/process/esm_loader:34
[...]

Issue #1292, opened 2024-05-10
bug(node, vectordb): error when using `Float64` embedding function


universalmind303:
### LanceDB version

main

### What happened?

I get some buffer allocation errors if i try to specify `Float64` as my embedding function type

```
Error: Invalid argument error: Need at least 8192 bytes in buffers[0] in array of type Float64, but got 4096
```

### Are there known steps to reproduce?

```js
const db = await lancedb.connect("/tmp/lancedb");

const embeddingFunction = {
  sourceColumn: "text",
  embed: async function (batch) {
    return batch.map(() => Array(512).fill(1.0))

  },
};

const data = [
  {text: "hello"},
  {text: "hello world"},
]

const schema = new Schema([
  new Field("vector", new FixedSizeList(512, new Field("item", new Float64()))),
  new Field("text", new Utf8())
]);

const name = "test";
let table = await db.createTable({name, data, schema, embeddingFunction})
```

However, if I change `Float64` to `Float32`, everything works fine.

wjones127:
That seems odd. We might hard code f32 in some places, but in Lance we do support f16 and f64, so we should fix any places that assume f32.

PR #1343, opened 2024-05-30
fix(nodejs): better support for f16 and f64

universalmind303:
closes https://github.com/lancedb/lancedb/issues/1292
closes https://github.com/lancedb/lancedb/issues/1293


wjones127 (approved):
Great work here :)

PR #1344, opened 2024-05-30
chore: add precommit hooks for rust

universalmind303:
I've noticed a few failures slip through the cracks, such as https://github.com/lancedb/lancedb/pull/1330. Since we are already using precommit hooks, it makes sense to validate rust like we do with ts and py.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,12 +22,14 @@ repos:
 - repo: local
   hooks:
   - id: fmt
-    name: fmt
+    name: cargo fmt
     description: Format files with cargo fmt.
     entry: cargo fmt
     language: system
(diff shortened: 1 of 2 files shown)


wjones127:
Why pre-commit hooks and not CI jobs that cover this? It's usually the CI jobs that block merges more than the commit hooks.

wjones127:
I tried to look for the smoking gun where our CI failed to find the change, but I couldn't quite see where it happened. A bunch of stuff started to fail after merging https://github.com/lancedb/lancedb/pull/1279, but the `lint` job passed in the PR CI https://github.com/lancedb/lancedb/actions/runs/9073480677/job/24930635163

Any idea what actually went wrong here?

PR #1351, opened 2024-06-03
feat(nodejs): merge insert

universalmind303:
closes https://github.com/lancedb/lancedb/issues/1349

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -148,7 +148,7 @@ describe("merge insert", () => {
     ];
     await table
       .mergeInsert("a")
-      .whenMatchedUpdateAll("target.b = 'b'")
+      .whenMatchedUpdateAll({ where: "target.b = 'b'" })
       .execute(newData);
 
     const expected = [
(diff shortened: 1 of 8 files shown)


nodejs/__test__/table.test.ts
```diff
@@ -102,6 +102,140 @@ describe.each([arrow, arrowOld])("Given a table", (arrow: any) => {
...
+  });
+  test("conditional update", async () => {
+    const newData = [
+      { a: 2, b: "x" },
+      { a: 3, b: "y" },
+      { a: 4, b: "z" },
+    ];
+    await table
+      .mergeInsert("a")
+      .whenMatchedUpdateAll("target.b = 'b'")
```
wjones127, line 151: I feel like I'd perfer the kwargs style for some of these. Here, it's slightly ambiguous whether `target.b = 'b'` means `WHERE target.b = 'b'` or `SET target.b = 'b'`.

wjones127 (commented):
This looks pretty good to me. Just two tweaks are necessary IMO.

nodejs/__test__/table.test.ts
```diff
@@ -102,6 +102,140 @@ describe.each([arrow, arrowOld])("Given a table", (arrow: any) => {
...
+    const newData = [
+      { a: 2, b: "x" },
+      { a: 4, b: "z" },
+    ];
+    await table
+      .mergeInsert("a")
+      .whenMatchedUpdateAll()
+      .whenNotMatchedInsertAll()
+      .whenNotMatchedBySourceDelete()
+      .execute(newData);
```
wjones127, line 223: I wonder if this particular combination of parameters, should just perform overwrite under the hood. Don't have to implement now, but something for us to think about.

PR #1352, opened 2024-06-04
feat: make it possible to opt in to using the v2 format

westonpace:
This also exposed the max_batch_length configuration option in python/node (it was needed to verify if we are actually in v2 mode or not)


westonpace:
This will fail until 0.12.0 is released

wjones127 (approved):

PR #1354, opened 2024-06-04
feat: add fast path to dataset reload

chebbyChefNEQ:
most of the time we don't need to reload. Locking the write lock and performing IO is not an ideal pattern.

This PR tries to make the critical section of `.write()` happen less frequently.

This isn't the most ideal solution. The most ideal solution should not lock until the new dataset has been loaded. But that would require too much refactoring.


wjones127 (approved):
This seems ago. I do intend to add async reload later though.

PR #1356, opened 2024-06-05
feat: enable stemming

### LINKED ISSUE #1315: Enable stemming and choosing tokenizer, when doing full text search in tantivy
### SDK

Python

### Description

Enabling stemming and using a language specific tokenizer tend to improve recall quite a bit, when doing full text search.

Tantivy has support for this through the [tokenizer_name argument in add_text_field](https://tantivy-py.readthedocs.io/en/latest/tutorials/#building-an-index-and-populating-it).

As far as I can tell the change needed is to add tokenizer_name argument to the following [line](https://github.com/lancedb/lancedb/blob/5349e8b1db57dda7ec7c61d6ee55b81625bf3acc/python/python/lancedb/fts.py#L59)

And then add the tokenizer_name argument to the [create_fts_index](https://github.com/lancedb/lancedb/blob/5349e8b1db57dda7ec7c61d6ee55b81625bf3acc/python/python/lancedb/table.py#L1171) method.

I would personally really prefer if the argument could be exposed instead of just enabling the usage of the english stemmer. [Tantivy supports a few different language tokenizers](https://docs.rs/tantivy/latest/tantivy/tokenizer/enum.Language.html), which I think a lot of people would like to use instead of english

I can create a pull request with the suggested changes if you think it is a good idea :-).

josca42:
Added the ability to specify tokenizer_name, when creating a full text search index using tantivy. This enables the use of language specific stemming. 

Also updated the [guide on full text search](https://lancedb.github.io/lancedb/fts/) with a short section on choosing tokenizer.

Fixes #1315

### DIFF
--- a/docs/src/fts.md
+++ b/docs/src/fts.md
@@ -56,7 +56,7 @@ This returns the result as a list of dictionaries as follows.
     LanceDB automatically searches on the existing FTS index if the input to the search is of type `str`. If you provide a vector as input, LanceDB will search the ANN index instead.
 
 ## Tokenization
-By default the text is tokenized by splitting on punctuation and whitespaces and then removing tokens that are longer than 40 chars. For more language specific tokenization then provide the argument tokenizer_name with the 2 letter language code followed by "en_stem". So for english it would be "en_stem".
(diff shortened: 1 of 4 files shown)


wjones127 (commented):
Could you add a unit test where we specify the tokenizer?

wjones127 (approved):

PR #1361, opened 2024-06-06
feat(nodejs): table.indexStats

universalmind303:
closes https://github.com/lancedb/lancedb/issues/1359


nodejs/src/table.rs
```diff
@@ -328,16 +328,36 @@ impl Table {
             .map(IndexConfig::from)
             .collect::<Vec<_>>())
     }
+
+    #[napi]
+    pub async fn index_stats(
+        &self,
+        index_uuid: String,
+    ) -> napi::Result<Either<IndexStatistics, ()>> {
+        // We use Either<T, ()> instead of `Option` to avoid returning null. We'd rather return 'undefined'
```
wjones127, line 337: why is this?

nodejs/lancedb/table.ts
```diff
@@ -478,4 +478,12 @@ export class Table {
...
     return await this.query().toArrow();
   }
+  /** List all the stats of a specified index
+   * @param {string} indexUUID The UUID of the index to get stats for
+   * @returns {IndexStatistics | undefined} The stats of the index
+   * If the index does not exist, `undefined` is returned
+   */
+  async indexStats(indexUUID: string): Promise<IndexStatistics | undefined> {
+    return await this.inner.indexStats(indexUUID);
+  }
```
wjones127, line 488: So the `indexStats()` method in the existing SDKs was actually incorrectly designed. The UUIDs here are meant to be internal ids, and users should instead use the `name` of the index.

This is because there can be multiple "parts" of an index, and each have their own internal UUID. But the name of the index is what they all have in common.

If you look at the underlying code to get the stats, you'll notice that it actually just uses the UUID to get the name of the index, and uses that to retrieve the stats. (`dataset.index_statistics(&index.unwrap().index_name)`.)

https://github.com/lancedb/lancedb/blob/27e4ad3f11117965fd4b57ce4c84808dfd5b8335/rust/lancedb/src/table.rs#L1252-L1262

I think it's best we just use name here from the start. We can fix the other SDKs in a follow-up PR.

nodejs/src/table.rs
```diff
@@ -430,3 +451,40 @@ pub struct AddColumnsSql {
...
+    /// The type of the index
+    pub index_type: Option<String>,
+    /// The metadata for each index
+    pub indices: Vec<IndexMetadata>,
+}
+impl From<lancedb::index::IndexStatistics> for IndexStatistics {
+    fn from(value: lancedb::index::IndexStatistics) -> Self {
+        Self {
+            num_indexed_rows: value.num_indexed_rows as f64,
+            num_unindexed_rows: value.num_unindexed_rows as f64,
```
wjones127, line 470: 🫠

wjones127 (changes requested):
Sorry I didn't point this out earlier, but the existing APIs are flawed and need to be fixed soon. See my comment on using `name` over `uuid`.

nodejs/src/table.rs
universalmind303: So this may be personal preference, and I'm fine with using `null` here. But in general, I think it's much easier for consumers to only have to deal with one `None` type instead of having to handle both `undefined` and `null`.

nodejs/src/table.rs
universalmind303: `f64` is the native equivalent to javascript's `number` type.

nodejs/lancedb/table.ts
universalmind303: @wjones127 does this mean that all of these public methods that accept `index_uuid` should instead take in `index_name` as well? 

- [count_indexed_rows](https://github.com/universalmind303/lancedb/blob/2b040f7f00b59554d0154dfba7854b3edc60225f/rust/lancedb/src/table.rs#L1209)
- [count_unindexed_rows](https://github.com/universalmind303/lancedb/blob/2b040f7f00b59554d0154dfba7854b3edc60225f/rust/lancedb/src/table.rs#L1216)
- [get_index_type](https://github.com/universalmind303/lancedb/blob/2b040f7f00b59554d0154dfba7854b3edc60225f/rust/lancedb/src/table.rs#L1230)
- [get_distance_type](https://github.com/universalmind303/lancedb/blob/2b040f7f00b59554d0154dfba7854b3edc60225f/rust/lancedb/src/table.rs#L1230)

wjones127: Yes, though should have. Also they should not have been separate methods, because they are so often called together and therefore make duplicate requests. 🤦

nodejs/lancedb/table.ts
```diff
@@ -479,6 +482,18 @@ export class Table {
   async toArrow(): Promise<ArrowTable> {
     return await this.query().toArrow();
   }
+  /** List all the stats of a specified index
+   *
+   * @param {string} name The name of the index. The index name should always be `${column}_idx`
```
wjones127, line 487: >  The index name should always be `${column}_idx`

I don't know if that's something we need to enforce at all. In Lance, you can name it whatever you want. But I suppose LanceDB probably names it for you, based on the column name.

I'd maybe say instead where you can check what indices are currently on the table and what names they have. Can say the convention is `${column}_idx`.

wjones127 (commented):
Just a comment on the docs.

nodejs/lancedb/table.ts
universalmind303: I did add a comment on `createIndex` that states that we don't yet support custom named indices.
```ts
   * @note We currently don't support custom named indexes,
   * The index name will always be `${column}_idx`
```

So i'll just remove that comment from `listIndices()`

wjones127 (approved):

universalmind303:
@wjones127 the failure is unrelated. I opened up an issue for it [here](https://github.com/lancedb/lancedb/issues/1394)

PR #1372, opened 2024-06-09
fix(java): the JVM crash when using jdk 8

beinan:
The Optional::isEmpty does not exist in java 8, so we should use isPresent instead


LuQQiu:
Looks like i didn't enable java 8 test in this repo...

PR #1378, opened 2024-06-11
feat(nodejs): feature parity [1/N] - remote table

universalmind303:
closes https://github.com/lancedb/lancedb/issues/1362


nodejs/lancedb/remote/connection.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  async createTable(
+    tableName: string,
+    data: Data,
+    options?: Partial<CreateTableOptions> | undefined,
+  ): Promise<Table> {
+    if (options?.mode) {
+      console.warn(`mode is not supported in remote connection, ignoring.`);
+    }
```
wjones127, line 115: Could you describe what mode behaviour it is following here? It it going to error if the table already exists? Or overwrite the table?

nodejs/lancedb/remote/table.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  display(): string {
+    return `RemoteTable(${this.#dbName}; ${this.#name})`;
+  }
+
+  async schema(): Promise<import("apache-arrow").Schema> {
+    const resp = await this.#client.post(`${this.#tablePrefix}/describe/`);
+    // TODO: parse this into a valid arrow schema
+    return resp.schema;
```
wjones127, line 63: Is this handled in one of the follow-up PRs?

wjones127 (commented):
Good work! I have a couple minor comments.

nodejs/lancedb/remote/table.ts
universalmind303: I don't have this handled yet.

The old `vectorDB` sdk was returning this AS IS and silently deviating from the type hints. Currently this mimics the same functionality, but I wanted to call out that this is not the correct behavior.

nodejs/lancedb/remote/connection.ts
universalmind303: We can definitely make this a bit more informative. _FWIW , I was copying the python message_

I'd assume cloud behavior is to error on exists?

nodejs/lancedb/remote/table.ts
wjones127: Okay. This is fine for now.

wjones127 (approved):

PR #1380, opened 2024-06-13
feat(nodejs): feature parity [2/N] - add `table.name` and `lancedb.connect({args})`

universalmind303:
depends on https://github.com/lancedb/lancedb/pull/1378

see proper diff here https://github.com/universalmind303/lancedb/compare/remote-table-node...universalmind303:lancedb:table-name


nodejs/lancedb/index.ts
```diff
@@ -76,15 +71,61 @@ export * as embedding from "./embedding";
...
+ * ```ts
+ * const conn = await connect({
+ *   uri: "/path/to/database",
+ *   storageOptions: {timeout: "60s"}
+ * });
+ * ```
+ */
+export async function connect(
+  opts: Partial<RemoteConnectionOptions | ConnectionOptions> & { uri: string },
+): Promise<Connection>;
```
wjones127, line 110: Should we mark the old style variant with `@deprecated`?

nodejs/src/connection.rs
```diff
@@ -56,12 +56,6 @@ impl Connection {
     #[napi(factory)]
     pub async fn new(uri: String, options: ConnectionOptions) -> napi::Result<Self> {
         let mut builder = ConnectBuilder::new(&uri);
-        if let Some(api_key) = options.api_key {
-            builder = builder.api_key(&api_key);
-        }
-        if let Some(host_override) = options.host_override {
-            builder = builder.host_override(&host_override);
-        }
```
wjones127, line 64: This was never used, right?

universalmind303: yeah. these are now unused as the js side handles the remote connection.

nodejs/lancedb/index.ts
universalmind303: I'm indifferent. I don't think it's much of a hassle supporting both signatures going forward.

wjones127: Okay that seems like a fine position for now.

wjones127 (approved):

PR #1381, opened 2024-06-13
feat: upgrade lance to 0.12.2-beta.2

westonpace:


.github/workflows/python.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           workspaces: python
       - name: Install
         run: |
-          pip install -e .[tests,dev,embeddings]
+          pip install --extra-index-url https://pypi.fury.io/lancedb/ -e .[tests,dev,embeddings]
```
wjones127, line 68: I think you might also need the `--pre` tag if you want beta releases.

westonpace: I confirmed this works.  The `--pre` argument will consider beta releases when resolving version ranges.  However, if you are specifying the beta release directly (e.g. `pip install lance==0.12.2-beta.2`) then it is not needed.

I didn't want to include it because I don't want to pick up beta releases from other components.

wjones127 (approved):

PR #1392, opened 2024-06-20
feat(nodejs): feature parity [6/N] - make public interface work with multiple arrow versions

universalmind303:
previously we didnt have great compatibility with other versions of apache arrow. This should bridge that gap a bit. 

depends on https://github.com/lancedb/lancedb/pull/1391
see actual diff here https://github.com/universalmind303/lancedb/compare/query-filter...universalmind303:arrow-compatibility


wjones127 (approved):
Thanks, it's awesome we are able to do this.

Issue #1394, opened 2024-06-21
bug(node): build command failing 


universalmind303:
### What happened?

`npm run build` fails for node. 
```
> vectordb@0.5.2 build
> npm run tsc && cargo-cp-artifact --artifact cdylib lancedb-node index.node -- cargo build --message-format=json


> vectordb@0.5.2 tsc
> tsc -b

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.30s
Did not copy "cdylib:lancedb-node"
```

### Are there known steps to reproduce?

```sh
> cd node
> npm i 
> npm run build
```

universalmind303:
This is due to updates in cargo in 1.79. File names with dashes are normalized to `_`

https://github.com/rust-lang/cargo/pull/12783/

PR #1396, opened 2024-06-21
fix(node): node build

universalmind303:
i have no idea why this fixes the build.


universalmind303:
closes https://github.com/lancedb/lancedb/issues/1394

wjones127 (approved):

PR #1407, opened 2024-06-26
feat: support to create more vector index types

BubbleCal:


python/python/lancedb/remote/table.py
```diff
@@ -111,6 +111,7 @@ def create_index(
         num_sub_vectors: Optional[int] = None,
         replace: Optional[bool] = None,
         accelerator: Optional[str] = None,
+        index_type="IVF_PQ",
```
rpgreen, line 114: Can this default to "vector" for backwards-compatiblity with older server versions?

https://github.com/lancedb/sophon/pull/2189/files#diff-3f4ef59e62aab203d6a9c873f3d07a0fdee2d007427635de47c5aaabe7adf2e0R72

rpgreen (approved):

PR #1408, opened 2024-06-26
feat(nodejs): add better typehints for registry

universalmind303:
previously the `registry` would return `undefined | EmbeddingFunction` even for built in functions such as "openai"

now it'll return the correct type for `getRegistry().get("openai")

as well as pass in the correct options type to `create`

### before
```ts
const options: {model: 'not-a-real-model'}
// this'd compile just fine, but result in runtime error
const openai: EmbeddingFunction | undefined = getRegistry().get("openai").create(options)
// this'd also compile fine
const openai: EmbeddingFunction | undefined = getRegistry().get("openai").create({MODEL: ''})
```
### after
```ts
const options: {model: 'not-a-real-model'}

const openai: OpenAIEmbeddingFunction = getRegistry().get("openai").create(options)
// Type '"not-a-real-model"' is not assignable to type '"text-embedding-ada-002" | "text-embedding-3-large" | "text-embedding-3-small" | undefined'


```

### DIFF
--- a/nodejs/lancedb/embedding/openai.ts
+++ b/nodejs/lancedb/embedding/openai.ts
@@ -43,8 +43,11 @@ export class OpenAIEmbeddingFunction extends EmbeddingFunction<
     }
     const modelName = options?.model ?? "text-embedding-ada-002";
 
+    /**
+     * @type {import("openai").default}
+     */
     // eslint-disable-next-line @typescript-eslint/naming-convention
-    let Openai: import("openai").default;
+    let Openai;
     try {
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       Openai = require("openai");
--- a/nodejs/lancedb/embedding/openai.ts
+++ b/nodejs/lancedb/embedding/openai.ts
@@ -13,16 +13,14 @@
 // limitations under the License.
 
 import type OpenAI from "openai";
+import { type EmbeddingCreateParams } from "openai/resources";
 import { Float, Float32 } from "../arrow";
 import { EmbeddingFunction } from "./embedding_function";
 import { register } from "./registry";
 
 export type OpenAIOptions = {
   apiKey: string;
-  model:
-    | "text-embedding-ada-002"
-    | "text-embedding-3-large"
-    | "text-embedding-3-small";
+  model: EmbeddingCreateParams["model"];
 };
 
 @register("openai")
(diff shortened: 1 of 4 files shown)


nodejs/lancedb/embedding/openai.ts
```diff
@@ -18,19 +18,26 @@ import { EmbeddingFunction } from "./embedding_function";
 import { register } from "./registry";
 
 export type OpenAIOptions = {
-  apiKey?: string;
-  model?: string;
+  apiKey: string;
+  model:
+    | "text-embedding-ada-002"
+    | "text-embedding-3-large"
+    | "text-embedding-3-small";
```
wjones127, line 25: IIUC, each time OpenAI releases a new model, we'll have to update our package. Users will have to wait until a newer release of LanceDB to use those models. Is that correct?

nodejs/lancedb/embedding/registry.ts
```diff
@@ -69,18 +65,25 @@ export class EmbeddingFunctionRegistry {
...
    * @param name The name of the function
    */
-  get<T extends EmbeddingFunction<unknown> = EmbeddingFunction>(
-    name: string,
-  ): EmbeddingFunctionCreate<T> | undefined {
+  get<T extends EmbeddingFunction<unknown>, Name extends string = "">(
+    name: Name extends "openai" ? "openai" : string,
+  ): Name extends "openai"
+    ? EmbeddingFunctionCreate<OpenAIEmbeddingFunction>
+    : EmbeddingFunctionCreate<T> | undefined {
```
wjones127, line 72: 😵 Could you leave a comment describing what this means?

universalmind303: TS type system is pretty ugly...

This makes it so that you can use string constants as "types", or use an explicitly supplied type
so 

`registry.get("openai") -> EmbeddingFunctionCreate<OpenAIEmbeddingFunction>`


`registry.get<MyCustomEmbeddingFunction>("my_func") -> EmbeddingFunctionCreate<MyCustomEmbeddingFunction> | undefined`

the reason this is important is that we always know our built in functions are defined so the user isnt forced to do a non null/undefined assertion

```ts
// before
const openai: OpenAIEmbeddingFunction | undefined = registry.get("openai")?.create();

// after
const openai: OpenAIEmbeddingFunction = registry.get("openai").create()

```

I'll also add this to the source code.

nodejs/lancedb/embedding/openai.ts
universalmind303: It looks like we currently have a hard dependency on openai, so users will be locked to our version regardless.

wjones127 (approved):

PR #1414, opened 2024-06-28
feat(nodejs): catch unwinds in node bindings

universalmind303:
this bumps napi version to 2.16 which contains a few bug fixes.
Additionally, it adds `catch_unwind` to any method that may unintentionally panic. 

`catch_unwind` will unwind the panics and return a regular JS error instead of panicking.

### DIFF
--- a/nodejs/src/table.rs
+++ b/nodejs/src/table.rs
@@ -70,7 +70,7 @@ impl Table {
     }
 
     /// Return Schema as empty Arrow IPC file.
-    #[napi]
+    #[napi(catch_unwind)]
     pub async fn schema(&self) -> napi::Result<Buffer> {
         let schema =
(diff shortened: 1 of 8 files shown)


nodejs/src/table.rs
```diff
@@ -108,7 +108,7 @@ impl Table {
         })
     }
 
-    #[napi]
+    #[napi(catch_unwind)]
```
wjones127, line 111: A bit unfortunate we have to repeat this.

wjones127 (approved):
Nice work!

nodejs/src/table.rs
universalmind303: yeah, it'd be nice if there was a way to enable this by default.

PR #1420, opened 2024-07-01
docs: add merge insert, create index and create scalar index to public rest api doc

eddyxu:
Added 3 APIs doc publicly:
- `merge_insert`
- `create_index`
- `create_scalar_index`

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -252,7 +252,7 @@ paths:
           in: query
           description: |
             The column to use as the primary key for the merge operation.
-          required: false
+          required: true
           schema:
             type: string
         - name: when_matched_update_all
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -232,13 +232,32 @@ paths:
       responses:
         "200":
           description: Insert successful
-
         "400":
           $ref: "#/components/responses/invalid_request"
         "401":
           $ref: "#/components/responses/unauthorized"
         "404":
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -219,20 +219,87 @@ paths:
...
+        - Data
+      summary: Merge Insert
+      operationId: mergeInsert
+      parameters:
+        - $ref: "#/components/parameters/table_name"
+        - name: on
+          in: query
+          description: |
+            The column to use as the primary key for the merge operation.
+          required: false
```
westonpace, line 255: I believe this is required (or should be)

docs/openapi.yml
```diff
@@ -219,20 +219,87 @@ paths:
...
+            Rows that exist in both the source table (new data) and
+            the target table (old data) will be updated, replacing
+            the old row with the corresponding matching row.
+          required: false
+          schema:
+            type: boolean
+        - name: when_matched_update_all_filt
+          in: query
+          description: |
+            The filter expression that specifies the rows to update.
```
westonpace, line 270:
```suggestion
            If present then only rows that satisfy the filter expression will
            be updated
```

westonpace (approved):
I didn't check the actual phalanx endpoints but these descriptions seem correct.

docs/openapi.yml
eddyxu: Fixed.

PR #1438, opened 2024-07-09
docs: add fine tuning section in retriever guide and minor fixes

AyushExel:


tanaymeh:
@AyushExel LGTM except that in both notebooks, there is a lot of redundant output like cell execution, warnings, etc.

Can you suppress the warnings using
```python
import warnings
warnings.simplefilter('ignore')
```

And also clear the imports and installation cell outputs?

tanaymeh (approved):

AyushExel:
Okk will do in a follow up PR!

PR #1440, opened 2024-07-10
feat: support update over binary field

eddyxu:


eddyxu:
Pending lance release with https://github.com/lancedb/lance/pull/2579

python/python/lancedb/util.py
```diff
@@ -231,6 +222,15 @@ def _(value: str):
     return f"'{value}'"
 
 
+@value_to_sql.register(bytes)
+def _(value: bytes):
```
chebbyChefNEQ, line 226: do we want to support more types? like str?

chebbyChefNEQ (approved):

python/python/lancedb/util.py
wjones127: ? `str` is the function above.

eddyxu: string is already supports (see 2 lines above)

wjones127 (approved):

python/python/lancedb/util.py
chebbyChefNEQ: I meant coercing str to bytes automatically. nvm doesn't seem possible from this interface

PR #1442, opened 2024-07-12
chore: bump lance to 0.14.1

BubbleCal:


rust/lancedb/src/table.rs
```diff
@@ -1503,7 +1503,7 @@ impl NativeTable {
         }
 
         let mut dataset = self.dataset.get_mut().await?;
-        let lance_idx_params = lance::index::scalar::ScalarIndexParams {};
+        let lance_idx_params = lance::index::scalar::ScalarIndexParams::default();
```
albertlockett, line 1506: Since this is in the `create_btree_index` method, should this be:

```
         let lance_idx_params = lance::index::scalar::ScalarIndexParams {
             force_index_type: Some(ScalarIndexType::BTree),
         };
```

I guess currently B-tree is the default:
https://github.com/lancedb/lance/blob/4cdcb22bc46432ff819d3a37baccf492bf6512f8/rust/lance/src/index/scalar.rs#L120-L128


But this comment makes it seem like in the future we might try to automatically detect what is the kind of scalar index to build?
https://github.com/lancedb/lance/blob/4cdcb22bc46432ff819d3a37baccf492bf6512f8/rust/lance/src/index/scalar.rs#L46-L47

albertlockett (approved):

PR #1447, opened 2024-07-16
feat(rust): huggingface sentence-transformers

universalmind303:


rust/lancedb/src/embeddings/sentence_transformers.rs
```diff
@@ -0,0 +1,447 @@
...
+            tokenizer,
+            device,
+            n_dims: self.n_dims,
+        }
+    }
+}
+
+macro_rules! from_cpu_storage {
+    ($cpu_storage:ident, $data:ident, $builder:ident, $layout:ident, $embeddings:ident, $d1:ident, $d2:ident) => {{
+        let mut builder = $builder::new();
```
wjones127, line 185: I feel like you could pre-size the builder here.
```suggestion
        let mut builder = $builder::with_capacity(dim1 * dim2);
```

rust/lancedb/src/embeddings/sentence_transformers.rs
```diff
@@ -0,0 +1,447 @@
...
+            model,
+            tokenizer,
+            device,
+            n_dims: self.n_dims,
+        }
+    }
+}
+
+macro_rules! from_cpu_storage {
+    ($cpu_storage:ident, $data:ident, $builder:ident, $layout:ident, $embeddings:ident, $d1:ident, $d2:ident) => {{
```
wjones127, line 184: I feel like you could probably write this as a generic function and it would be a bit easier to maintain.
```rust
fn from_cpu_storage<T: ArrowPrimitiveType>(
    buffer: &[T::Native],
    layout: &Layout,
    embeddings: ???,
    dim1: usize,
    dim2: usize,
) -> PrimitiveArray<T> { ... }
```

wjones127 (changes requested):
This looks pretty well done. I have a few suggestions around how the data transfer code could work. I'd be especially interested in getting non-CPU data working, which seems like an easy fix.

rust/lancedb/src/embeddings/sentence_transformers.rs
universalmind303: i originally was trying to write this as a generic function. I can't recall the issue I ran into. I'll take another look

wjones127 (approved):

PR #1451, opened 2024-07-17
docs: minor updates for js migration guides

universalmind303:


docs/src/basic_legacy.ts
```diff
@@ -43,32 +74,33 @@ const example = async () => {
...
-        vector: Array.from(Array(dim), Math.random)
-      })),
-      { f16_schema }
-    )
-  const table = await db.createTable('f16_tbl', data)
+    Array.from(Array(total), (_, i) => ({
+      id: i,
+      vector: Array.from(Array(dim), Math.random),
+    })),
+    { f16_schema },
```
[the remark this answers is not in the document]
universalmind303: lol, i didn't notice this from the old code. This is definitely not valid

docs/src/js/classes/Index.md
```diff
@@ -35,7 +35,7 @@
 
 #### Defined in
 
-[indices.ts:118](https://github.com/lancedb/lancedb/blob/9d178c7/nodejs/lancedb/indices.ts#L118)
+[indices.ts:118](https://github.com/universalmind303/lancedb/blob/833b375/nodejs/lancedb/indices.ts#L118)
```
wjones127, line 38: Having your fork in the URL doesn't seem desirable long term. Should this be fixed somehow?

universalmind303: just removed the "Defined In" sections from the api reference docs

docs/src/migration.md
```diff
@@ -80,17 +80,114 @@ we plan to support them soon.
 
 For JS/TS users, we offer a brand new SDK [@lancedb/lancedb](https://www.npmjs.com/package/@lancedb/lancedb)
 
+### Changes to DB APIs
+
+We tried to keep the API as similar as possible to the previous version, but there are a few small changes.
```
wjones127, line 85: We say there are a few small changes, but don't describe any of them. Could we do that?

docs/src/migration.md
```diff
@@ -80,17 +80,114 @@ we plan to support them soon.
...
+      config: lancedb.Index.ivfPq({
+        numPartitions: 2,
+        numSubVectors: 2,
+      }),
+    });
+    ```
+
+### Embedding Functions
+
+The embedding api has been completely reworked, and it now more closely resembles the python API, including the new [embedding registry](./js/classes/embedding.EmbeddingFunctionRegistry.md)
```
wjones127, line 138:
```suggestion
The embedding API has been completely reworked, and it now more closely resembles the Python API, including the new [embedding registry](./js/classes/embedding.EmbeddingFunctionRegistry.md)
```

docs/src/migration.md
universalmind303: don't the following sub sections explain the differences?

- creating tables
- creating indices
- embedding functions

wjones127: Oh. I was confused because the next subheader is also an h3, so I didn't read that as a subsection. I guess `Creating tables` should be an h4 then?

wjones127: The hierarchy you have right now is

* TypeScript/JavaScript
  * Changes to DB APIs
  * Creating Tables
  * Changes to Table APIs
    * Creating Indices
  * Embedding Functions

Is that what you meant?

Is "Changes to DB APIs" meant to refer just to connection objects? Or to all the APIs? If the latter, I wonder if that header is even necessary.

PR #1462, opened 2024-07-20
feat(nodejs): huggingface compatible transformers

universalmind303:


nodejs/lancedb/embedding/transformers.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    this.#ndims = options.ndims;
+    this.#modelName = modelName;
+  }
+  toJSON() {
+    // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+    const obj: Record<string, any> = {
+      model: this.#modelName,
+      // tokenizerOptions: this.#tokenizerOptions,
+      // ndims: this.#ndims,
```
wjones127, line 78: Should we remove this?

wjones127 (commented):
I noticed there aren't any unit tests? Is that intentional? What should be our testing strategy to make sure this continues to work in the future?

wjones127 (approved):
Shame we weren't able to get the unit tests working. Hopefully their API remains stable.

Issue #1358, opened 2024-06-05
Remove source `.ts` code from published npm package for `@lancedb/lancedb


universalmind303:
### SDK

Node

### Description

The [published package](https://www.npmjs.com/package/@lancedb/lancedb?activeTab=code) contains a bunch of unnecessary code. Usually the npm package contains the compiled `/dist`, but not the source typescript code.

SiluPanda:
Hi maintainers — I’d like to work on this if it’s still available. I’ll keep the first PR focused and add/update tests or docs where appropriate. Please let me know if there’s any preferred implementation direction before I start.

PR #1467, opened 2024-07-22
fix: remove source .ts code from published npm package

gagan-bhullar-tech:
This PR is for issue - https://github.com/lancedb/lancedb/issues/1358

### DIFF
--- a/nodejs/.npmignore
+++ b/nodejs/.npmignore
@@ -15,3 +15,11 @@ renovate.json
 src
 lancedb
 examples
+nodejs-artifacts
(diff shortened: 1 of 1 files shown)


universalmind303 (commented):
most published packages I see have **all** dev related code stripped out. 

I think the published package should only contain the following.

.
|-- dist/
|-- LICENSE
|-- package.json
|-- README.md

gagan-bhullar-tech:
@universalmind303 updated PR as per your comments

PR #1479, opened 2024-07-29
docs: custom embedding function for ts

universalmind303:


nodejs/examples/custom_embedding_function.ts
```diff
@@ -0,0 +1,63 @@
...
+  vector: sentenceTransformer.vectorField(),
+  text: sentenceTransformer.sourceField(),
+});
+
+const db = await lancedb.connect("/tmp/db");
+const table = await db.createEmptyTable("table", schema, { mode: "overwrite" });
+
+await table.add([{ text: "hello" }, { text: "world" }]);
+
+const results = await table.search("hello").limit(1).toArray();
```
wjones127, line 62: minor nit: if we are doing semantic search, seems odd to pass same value. More illustrative to pass something semantically similar.
```suggestion
const results = await table.search("greeting").limit(1).toArray();
```

docs/src/embeddings/custom_embedding_function.md
```diff
@@ -15,198 +15,225 @@ There is another optional layer of abstraction available: `TextEmbeddingFunction
...
+        @cached(cache={})
+        def _embedding_model(self):
+            return sentence_transformers.SentenceTransformer(name)
+    ```
+
+=== "TypeScript"
+
+    ```ts
+    --8<--- "nodejs/examples/custom_embedding_function.ts:imports"
+    --8<--- "nodejs/examples/custom_embedding_function.ts:embedding_impl"
```
wjones127, line 51: It might be nice to have some whitespace here:
```suggestion
    --8<--- "nodejs/examples/custom_embedding_function.ts:imports"
    
    --8<--- "nodejs/examples/custom_embedding_function.ts:embedding_impl"
```

docs/src/embeddings/custom_embedding_function.md
```diff
@@ -15,198 +15,225 @@ There is another optional layer of abstraction available: `TextEmbeddingFunction
...
-tbl.add(pd.DataFrame({"text": ["halo", "world"]}))
-result = tbl.search("world").limit(5)
-```
+=== "TypeScript"
+
+    ```ts
+    --8<--- "nodejs/examples/custom_embedding_function.ts:call_custom_function"
+    ```
 
 NOTE:
```
wjones127, line 83: Not related to this PR, but would you mind fixing this so it uses a proper `!!! note` admonition?

docs/src/embeddings/custom_embedding_function.md
```diff
@@ -15,198 +15,225 @@ There is another optional layer of abstraction available: `TextEmbeddingFunction
...
-            image_features /= image_features.norm(dim=-1, keepdim=True)
-        return image_features.cpu().numpy().squeeze()
-```
+                image_features /= image_features.norm(dim=-1, keepdim=True)
+            return image_features.cpu().numpy().squeeze()
+    ```
+
+=== "TypeScript"
+
+    Coming Soon!
```
wjones127, line 239: Do we have an issue number we can link here? If not, can we make one? Would be nice if users could see where to track progress? Or ask if it is already done and needs to be added to docs.

wjones127 (commented):
This looks good overall. I have a few minor suggestions :)

wjones127 (approved):

PR #1481, opened 2024-07-30
feat(python): multi-vector reranking support

AyushExel:
Currently targeting the following usage:
```
from lancedb.rerankers import CrossEncoderReranker

reranker = CrossEncoderReranker()

query = "hello"

res1 = table.search(query, vector_column_name="vector").limit(3)
res2 = table.search(query, vector_column_name="text_vector").limit(3)
res3 = table.search(query, vector_column_name="meta_vector").limit(3)

reranked = reranker.rerank_multivector(
               [res1, res2, res3],  
              deduplicate=True,
              query=query # some reranker models need query
)
```
- This implements rerank_multivector function in the base reranker so that all rerankers that implement rerank_vector will automatically have multivector reranking support
- Special case for RRF reranker that just uses its existing rerank_hybrid fcn to multi-vector reranking.

------
 questions:
- Some features might need access to `_rowid` that I've exposed via a flag. Not sure if there's a reason it wasn't exposed yet.


python/python/lancedb/rerankers/base.py
```diff
@@ -130,12 +134,91 @@ def merge_results(self, vector_results: pa.Table, fts_results: pa.Table):
...
         # deduplicate
-        mask = np.full((combined.shape[0]), False)
+        combined = self._deduplicate(combined)
+
+        return combined
+
+    def rerank_multivector(
+        self,
+        vector_results: Union[List[pa.Table], List["LanceVectorQueryBuilder"]],
+        query: Union[str, None] = None,  # Some rerankers might not need the query
```
AyushExel, line 146: Here, I was not sure about order of the args. `query` comes 2nd compared to other rerank_ methods because:
- These are meant to be directly called by the user 
- And `query` might be optional for some rerankers

python/python/lancedb/table.py
```diff
@@ -1464,14 +1466,19 @@ def search(
...
-        return LanceQueryBuilder.create(
+        query_builder = LanceQueryBuilder.create(
             self,
             query,
             query_type,
             vector_column_name=vector_column_name,
             ordering_field_name=ordering_field_name,
         )
 
+        if with_row_id:
```
AyushExel, line 1477: Is exposing `with_row_id` safe?

python/python/lancedb/rerankers/base.py
```diff
@@ -130,12 +134,91 @@ def merge_results(self, vector_results: pa.Table, fts_results: pa.Table):
...
+        deduplicate: bool = False,
+        # deduplicator: callable = None,
```
westonpace, line 148:
```suggestion
```

python/python/lancedb/rerankers/base.py
```diff
@@ -130,12 +134,91 @@ def merge_results(self, vector_results: pa.Table, fts_results: pa.Table):
...
+    ):
+        """
+        This is a rerank function that receives the results from multiple
+        vector searches. Could be using different vector columns
+
```
westonpace, line 153:
```suggestion
        This is a rerank function that receives the results from multiple
        vector searches. For example, this can be used to combine the
        results of two vector searches with different embeddings.

```

python/python/lancedb/rerankers/base.py
westonpace: I think the order is fine.  However, maybe `query` should not have a default value but clearly document that it could be None?  This way you force the user to explicitly state "I know my reranker does not need the query"?

python/python/lancedb/table.py
westonpace: I don't think there is a problem exposing `with_row_id`.  However, I believe the user can already do:

```
my_tbl.search(...).with_row_id()
```

Do we need to add `with_row_id` to `search`?

westonpace (approved):
Some suggestions but no major concerns

python/python/lancedb/table.py
AyushExel: Ahh I see.. yeah this is redundant then. will update

python/python/lancedb/rerankers/base.py
AyushExel: yeah sounds good.. Updated

PR #1483, opened 2024-07-31
feat!: migrate FTS from tantivy to lance-index

BubbleCal:
Lance now supports FTS, so add it into lancedb Python, TypeScript and Rust SDKs.

For Python, we still use tantivy based FTS by default because the lance FTS index now misses some features of tantivy.

For Python:
- Support to create lance based FTS index
- Support to specify columns for full text search (only available for lance based FTS index)

For TypeScript:
- Change the search method so that it can accept both string and vector
- Support full text search

For Rust
- Support full text search

The others:
- Update the FTS doc

BREAKING CHANGE: 
- for Python, this renames the attached score column of FTS from "score" to "_score", this could be a breaking change for users that rely the scores


python/python/lancedb/table.py
```diff
@@ -1188,11 +1189,25 @@ def create_index(
...
         )
 
-    def create_scalar_index(self, column: str, *, replace: bool = True):
+    def create_scalar_index(
+        self,
+        column: str,
+        index_type: Literal["BTREE", "BITMAP", "LABEL_LIST", "INVERTED"] = "BTREE",
+        *,
+        replace: bool = True,
+    ):
```
westonpace, line 1198: This is a breaking change (no change needed, just noting)

python/python/lancedb/table.py
```diff
@@ -1188,11 +1189,25 @@ def create_index(
             index_cache_size=index_cache_size,
         )
 
-    def create_scalar_index(self, column: str, *, replace: bool = True):
+    def create_scalar_index(
```
westonpace, line 1192: Does the API need to change for the remote table too?

rust/lancedb/src/query.rs
```diff
@@ -351,6 +352,17 @@ pub trait QueryBase {
...
+    /// Perform a full text search on the table.
+    ///
+    /// The results will be returned in order of BM25 scores.
+    ///
+    /// This method is only valid on tables that have a full text search index.
+    ///
+    /// ```ignore
+    /// query.full_text_search(FullTextSearchQuery::new("hello world"))
+    /// ```
+    fn full_text_search(self, query: FullTextSearchQuery) -> Self;
```
westonpace, line 364: Do we expect that we will eventually allow FTS to be combined with vector search in a single query?  Or will reranking always involve two queries?

westonpace (approved):
It doesn't have to be this PR but we should probably add the query method and FTS index type to the async python API as well.

Otherwise, some suggestions, but nothing major.

Thanks for doing this!

westonpace:
I changed **breaking change** to `BREAKING CHANGE:`.  The [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) spec requires that exact phrasing (and tools look for it).  I also added `!` in the PR title.  This should allow our automated tooling to detect the breaking changes.

wjones127 (commented):
I'm not sure we should deprecate the old way until we have further tested the new one and gotten closer to feature parity.

docs/src/fts.md
```diff
@@ -35,42 +35,83 @@ table = db.create_table(
...
+
+    ```python
+    table.create_scalar_index("text", "INVERTED")
+    ```
+
+=== "<= 0.11.0"
+
+    ```python
+    table.create_fts_index("text")
+    ```
```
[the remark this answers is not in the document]
westonpace: We could also combine the two create_index methods (scalar and vector).  The inverted index *can* be used in the same way as scalar indices (speeding up filters) but it can also be used for full text search (which is a search and not a filter)

nodejs/src/query.rs
```diff
@@ -138,6 +145,12 @@ impl VectorQuery {
         self.inner = self.inner.clone().only_if(predicate);
     }
 
+    #[napi]
+    pub fn full_text_search(&mut self, query: String, columns: Option<Vec<String>>) {
+        let query = FullTextSearchQuery::new(query).columns(columns);
+        self.inner = self.inner.clone().full_text_search(query);
```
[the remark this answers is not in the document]
BubbleCal: it can't work with vector search now

docs/src/fts.md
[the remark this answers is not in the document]
BubbleCal: I'm going to change the `create_fts_index(column: str)` to `create_fts_index(column: str, index_type="FTS", use_legacy=False)`. We haven't supported use the inverted index as filtering so it still makes sense.

the `create_index` seems has been used for vector index only, that can't be easily changed.

python/python/lancedb/table.py
BubbleCal: need to change the server side impl first, i think it doesn't respect the `index_type` param now cause we are passing `scalar` to it

rust/lancedb/src/index/scalar.rs
```diff
@@ -28,3 +28,13 @@
...
 
 impl BTreeIndexBuilder {}
+
+/// Builder for a full text search index
+///
+/// A full text search index is an index on a string column that allows for full text search
+#[derive(Debug, Clone, Default)]
+pub struct FtsIndexBuilder {}
+
+impl FtsIndexBuilder {}
```
wjones127, line 38: IIUC, this is where things like tokenizer and stemmer will eventually go?

rust/lancedb/examples/full_text_search.rs
```diff
@@ -0,0 +1,108 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This example demonstrates setting advanced parameters when building an IVF PQ index
+//!
+//! Snippets from this example are used in the documentation on ANN indices.
```
wjones127, line 17:
```suggestion
```

rust/lancedb/examples/full_text_search.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let words = random_word::all(random_word::Lang::En);
+    let batches = RecordBatchIterator::new(
+        vec![RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(Int32Array::from_iter_values(0..TOTAL as i32)),
+                Arc::new(StringArray::from_iter_values(
+                    (0..TOTAL).map(|_| words[random::<usize>() % 256]),
+                )),
```
wjones127, line 63: It's a bit of a weird example that all the documents are single tokens, IIUC. Might be worth constructing random documents from N tokens instead.

python/python/lancedb/table.py
```diff
@@ -1211,19 +1220,39 @@ def create_fts_index(
...
+            only available with use_legacy=True
         tokenizer_name: str, default "default"
             The tokenizer to use for the index. Can be "raw", "default" or the 2 letter
             language code followed by "_stem". So for english it would be "en_stem".
             For available languages see: https://docs.rs/tantivy/latest/tantivy/tokenizer/enum.Language.html
-        """
+            only available with use_legacy=True for now
+        use_legacy: bool, default False
+            If True, use the legacy full-text search implementation based on tantivy.
+            If False, use the new full-text search implementation based on lance-index.
```
wjones127, line 1240: Should this be `use_tantivy` now?

python/python/lancedb/query.py
```diff
@@ -689,6 +701,31 @@ def phrase_query(self, phrase_query: bool = True) -> LanceFtsQueryBuilder:
         return self
 
     def to_arrow(self) -> pa.Table:
+        tantivy_index_path = self._table._get_fts_index_path()
+        if Path(tantivy_index_path).exists():
+            return self.tantivy_to_arrow()
+
+        # get the scores and doc ids
```
wjones127, line 708: This comment doesn't seem to correspond with the code below at all.

```suggestion
```

python/python/lancedb/query.py
```diff
@@ -689,6 +701,31 @@ def phrase_query(self, phrase_query: bool = True) -> LanceFtsQueryBuilder:
...
+        query = self._query
+        if self._phrase_query:
+            raise NotImplementedError("Phrase query is not yet supported")
+        if self._reranker:
+            raise NotImplementedError("Reranking is not yet supported")
```
wjones127, line 713: These error messages could be more helpful:
```suggestion
        if self._phrase_query:
            raise NotImplementedError("Phrase query is not yet supported in Lance FTS. Use tantivy-based index instead for now.")
        if self._reranker:
            raise NotImplementedError("Reranking is not yet supported in Lance FTS. Use tantivy-based index instead for now.")
```

nodejs/examples/package.json
```diff
@@ -11,9 +11,22 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@lancedb/lancedb": "file:../",
-    "@xenova/transformers": "^2.17.2"
+    "@xenova/transformers": "^2.17.2",
+    "tsc": "^2.0.4"
```
wjones127, line 15: I think we should move `typescript` to `devDependencies`, but I don't think we should need `tsc`. Should be able to still run `npx tsc <file>`.
```suggestion
```

nodejs/examples/package.json
```diff
@@ -11,9 +11,22 @@
...
   },
   "peerDependencies": {
     "typescript": "^5.0.0"
+  },
+  "devDependencies": {
+    "typescript": "^5.5.4"
+  },
```
wjones127, line 22:
```suggestion
  "devDependencies": {
    "typescript": "^5.5.4"
  },
```

docs/src/fts.md
```diff
@@ -14,63 +16,118 @@ pip install tantivy==0.20.1
...
+
+    ```python
+    table.create_fts_index(["text1", "text2"])
+    ```
+
+=== "use_tantivy=False"
+
+    ```python
+    not supported yet
+    ```
```
wjones127, line 130: For these, could we link to a GH issue?
```suggestion
    **Not supported yet**
```

docs/src/fts.md
```diff
@@ -14,63 +16,118 @@ pip install tantivy==0.20.1
...
+
+    ```python
+    table.create_fts_index("text", use_tantivy=True, tokenizer_name="en_stem")
+    ```
+
+=== "use_tantivy=False"
+
+    ```python
+    not supported yet
+    ```
```
wjones127, line 112:
```suggestion
    **Not supported yet**
```

docs/src/fts.md
```diff
@@ -14,63 +16,118 @@ pip install tantivy==0.20.1
...
 
-This returns the result as a list of dictionaries as follows.
+=== "Rust"
+
+    ```rust
+    let uri = "data/sample-lancedb";
+    let db = connect(uri).execute().await?;
+    let initial_data: Box<dyn RecordBatchReader + Send> = create_some_records()?;
+    let tbl = db
+        .create_table("my_table", Box::new(initial_data))
```
wjones127, line 70: `initial_data` is already a `Box`, don't need to re-wrap it.
```suggestion
    let initial_data: Box<dyn RecordBatchReader + Send> = create_some_records()?;
    let tbl = db
        .create_table("my_table", initial_data)
```

docs/src/fts.md
```diff
@@ -14,63 +16,118 @@ pip install tantivy==0.20.1
...
+        .execute()
+        .await
+        .unwrap();
```
wjones127, line 73: We use `?` already below, so might as well use it here.
```suggestion
        .await?;
```

docs/src/fts.md
```diff
@@ -14,63 +16,118 @@ pip install tantivy==0.20.1
...
+        .select(lancedb::query::Select::Columns(vec!["text".to_owned()]))
+        .limit(10)
+        .execute()
+        .await?;
+    ```
+
+It would search on all indexed columns by default, so it's useful when there are multiple indexed columns.
+For now, this is supported in tantivy way only.
+
+Passing `fts_columns="text"` if you want to specify the columns to search, but it's not available for tantivy based full text search.
```
wjones127, line 91:
```suggestion
Passing `fts_columns="text"` if you want to specify the columns to search, but it's not available for Tantivy-based full text search.
```

docs/src/fts.md
```diff
@@ -80,19 +137,48 @@ Currently the LanceDB full text search feature supports *post-filtering*, meanin
...
+        .select(lancedb::query::Select::Columns(vec!["doc".to_owned()]))
+        .limit(10)
+        .only_if("meta='foo'")
+        .execute()
+        .await?;
+    ```
 
 ## Sorting
 
+!!! tip "Warn"
```
wjones127, line 172: It's confusing to use `tip` for a warning. Use `warning` instead:

```suggestion
!!! warning "Warn"
```

You can see all available admonitions here: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types

docs/src/fts.md
```diff
@@ -80,19 +137,48 @@ Currently the LanceDB full text search feature supports *post-filtering*, meanin
...
+    Sorting is available for only tantivy based FTS
```
wjones127, line 173:
```suggestion
    Sorting is available for only Tantivy-based FTS
```

docs/src/fts.md
```diff
@@ -116,6 +202,9 @@ table.create_fts_index(["text_field"], ordering_field_names=["sort_by_field"])
 
 ## Phrase queries vs. terms queries
 
+!!! tip "Warn"
```
wjones127, line 205:
```suggestion
!!! warning "Warn"
```

docs/src/fts.md
```diff
@@ -1,9 +1,11 @@
 # Full-text search
 
-LanceDB provides support for full-text search via [Tantivy](https://github.com/quickwit-oss/tantivy) (currently Python only), allowing you to incorporate keyword-based search (based on BM25) in your retrieval solutions. Our goal is to push the FTS integration down to the Rust level in the future, so that it's available for Rust and JavaScript users as well.  Follow along at [this Github issue](https://github.com/lancedb/lance/issues/1195)
+LanceDB provides support for full-text search via Lance (before via [Tantivy](https://github.com/quickwit-oss/tantivy) (Python only)), allowing you to incorporate keyword-based search (based on BM25) in your retrieval solutions.
 
+Currently, the Lance full text search is missing some features that are in the Tantivy full text search (custom tokenizer, phrase query, etc.), so in Python, tantivy is still the default way to do full text search, and we are leaving the instructions in place, as we finish adding the rest of this functionality.
```
wjones127, line 5: This is a very long sentence. Could we split this up?

```suggestion
Currently, the Lance full text search is missing some features that are in the Tantivy full text search. This includes phrase queries, re-ranking, and customizing the tokenizer. Thus, in Python, Tantivy is still the default way to do full text search and many of the instructions below apply just to Tantivy-based indices.
```

rust/lancedb/src/query.rs
BubbleCal: Now reranking is done in lancedb python, i think we can do FTS and vector search simultaneously in a single query, which is more efficient, but it requires moving these reranker into lance in rust.

rust/lancedb/src/index/scalar.rs
BubbleCal: Yes

docs/src/fts.md
```diff
@@ -1,9 +1,11 @@
 # Full-text search
 
-LanceDB provides support for full-text search via [Tantivy](https://github.com/quickwit-oss/tantivy) (currently Python only), allowing you to incorporate keyword-based search (based on BM25) in your retrieval solutions. Our goal is to push the FTS integration down to the Rust level in the future, so that it's available for Rust and JavaScript users as well.  Follow along at [this Github issue](https://github.com/lancedb/lance/issues/1195)
+LanceDB provides support for full-text search via Lance (before via [Tantivy](https://github.com/quickwit-oss/tantivy) (Python only)), allowing you to incorporate keyword-based search (based on BM25) in your retrieval solutions.
 
+Currently, the Lance full text search is missing some features that are in the Tantivy full text search. This includes phrase queries, re-ranking, and customizing the tokenizer. Thus, in Python, Tantivy is still the default way to do full text search and many of the instructions below apply just to Tantivy-based indices.
 
-## Installation
+
+## Installation (Only for Tantivy-based FTS)
```
eddyxu, line 9: lets add a note that using native FTS does not require install extra dependencies.

docs/src/fts.md
```diff
@@ -14,63 +16,112 @@ pip install tantivy==0.20.1
...
+    table = db.create_table(
+        "my_table",
+        data=[
+            {"vector": [3.1, 4.1], "text": "Frodo was a happy puppy"},
+            {"vector": [5.9, 26.5], "text": "There are several kittens playing"},
+        ],
+    )
 
-To search an FTS index via keywords, LanceDB's `table.search` accepts a string as input:
+    # passing `use_tantivy=False` to use lance FTS index
```
eddyxu, line 37: do we have a default value for `use_tantivy`?

Lets show that in the code as well, i.e.,

```
# Create native FTS index
table.create_fts_index("text")
```

docs/src/fts.md
```diff
@@ -14,63 +16,112 @@ pip install tantivy==0.20.1
...
+    # ...
+    ```
 
-```python
-table.search("puppy").limit(10).select(["text"]).to_list()
-```
+=== "TypeScript"
+
+    ```typescript
+    import * as lancedb from "@lancedb/lancedb";
```
eddyxu, line 47: do we need to call `createFtsIndex()` ?

nodejs/examples/full_text_search.ts
```diff
@@ -0,0 +1,38 @@
+import * as lancedb from "@lancedb/lancedb";
```
eddyxu, line 1: add license

BubbleCal: Added

docs/src/fts.md
BubbleCal: Added

docs/src/fts.md
BubbleCal: Modified

docs/src/fts.md
BubbleCal: Added

PR #1485, opened 2024-07-31
ci: public java package

eddyxu:


.github/workflows/java-publish.yml
```diff
@@ -0,0 +1,82 @@
...
+          toolchain: "1.79.0"
+          cache-workspaces: "src/rust"
+          # Disable full debug symbol generation to speed up CI build and keep memory down
+          # "1" means line tables only, which is useful for panic tracebacks.
+          rustflags: "-C debuginfo=1"
+      - name: Install dependencies
+        run: |
+          sudo apt -y -qq update
+          sudo apt install -y protobuf-compiler libssl-dev pkg-config
+      - name: Build release
```
walterddr, line 55: looks like we need to do something similar to https://github.com/lancedb/lance/pull/2613/files#r1720849909 before the Build release in order to incorporate the right GLIBC

walterddr:
thanks @LuQQiu for the update. do we have any word on the final build still containing a wrong GLIBC version? let me know if we need to work around it. thx

LuQQiu:
@walterddr If we are using ubuntu 22 for testing, should not have the wrong GLIBC version issue. If we target to use older ubuntu version, then need to set nodejs to an older version.

walterddr:
i downloaded and checked the binary it doesn't contain wrong GLIBC versions. so i think we should be good with this PR

LuQQiu (approved):

PR #1494, opened 2024-08-02
ci: fix release by upgrading to Node 18

wjones127:
Building with Node 16 produced this error:

```
npm ERR! code ENOENT
npm ERR! syscall chmod
npm ERR! path /io/nodejs/node_modules/apache-arrow-15/bin/arrow2csv.cjs
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, chmod '/io/nodejs/node_modules/apache-arrow-15/bin/arrow2csv.cjs'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```

[CI Failure](https://github.com/lancedb/lancedb/actions/runs/10117131772/job/27981475770). This looks like it is https://github.com/apache/arrow/issues/43341

Upgrading to Node 18 makes this goes away. Since Node 18 requires glibc >= 2_28, we had to upgrade the manylinux version we are using. This is fine since we already state a minimum Node version of 18.

This also upgrades the openssl version we bundle, as well as consolidates the build files.


westonpace (approved):
Good cleanup, thanks for removing the duplicate files I introduced :)

PR #1495, opened 2024-08-02
docs: add a section to describe scalar index

eddyxu:


docs/mkdocs.yml
```diff
@@ -89,9 +89,10 @@ nav:
           - Data management: concepts/data_management.md
       - 🔨 Guides:
           - Working with tables: guides/tables.md
-          - Building an ANN index: ann_indexes.md
+          - Building vector index: ann_indexes.md
```
wjones127, line 92:
```suggestion
          - Building a vector index: ann_indexes.md
```

docs/mkdocs.yml
```diff
@@ -89,9 +89,10 @@ nav:
...
           - Vector Search: search.md
           - Full-text search: fts.md
+          - Building scalar index: guides/scalar_index.md
```
wjones127, line 95:
```suggestion
          - Building a scalar index: guides/scalar_index.md
```

docs/mkdocs.yml
```diff
@@ -178,6 +179,7 @@ nav:
       - Building an ANN index: ann_indexes.md
       - Vector Search: search.md
       - Full-text search: fts.md
+      - Building scalar index: guides/scalar_index.md
```
wjones127, line 182:
```suggestion
      - Building a scalar index: guides/scalar_index.md
```

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,90 @@
+# Building Scalar Index
+
+Similar to many SQL databases, LanceDB supports several types of Scalar indices to accelerate search
+over scalar columns.
+
+- `BTREE`: The most common type is BTREE. This index is inspired by the btree data structure
+  although only the first few layers of the btree are cached in memory.
+  It will perform well on columns with a large number of unique values and few rows per value.
```
wjones127, line 8: We should probably state which operations are supported by the indices.

IIUC:

 * BTREE: `=`, `<>`, inequalities (`<=`, `>`, and so on)
 * BITMAP: `=`
 * LABEL_LIST: `array_has_any`, `array_has_all`, `array_contains`

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,90 @@
...
+- `BITMAP`: this index stores a bitmap for each unique value in the column.
+  This index is useful for columns with a small number of unique values and many rows per value.
+  For example, a column presents `categories`. In Apache Arrow terminalogy, a column can be presented
```
wjones127, line 11: Consider running a spellchecker
```suggestion
  For example, a column presents `categories`. In Apache Arrow terminology, a column can be presented
```

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,90 @@
...
+    ```python
+    import lancedb
+    db = lancedb.connect("./db")
+    table = db.open_table("my_table")
+    table.create_scalar_index("category", index_type="BITMAP")
+    ```
+
+=== "Typescript"
+
+    Only `BTree` index is supported today. `BITMAP` and `LABEL_LIST` will be added soon.
```
wjones127, line 28: Do we have a GH issue with this? We should create one if not. Might be nice to link to that issue here so users can track the progress of that or maybe see it needs doing and pick it up.

eddyxu: Added.

docs/src/guides/scalar_index.md
eddyxu: Added a table.

python/python/lancedb/remote/table.py
```diff
@@ -97,6 +97,8 @@ def index_stats(self, index_uuid: str):
     def create_scalar_index(
         self,
         column: str,
+        *,
+        index_type: Literal["BTREE", "BITMAP", "LABEL_LIST"] = "BTREE",
     ):
         """Creates a scalar index
         Parameters
```
westonpace, line 104: We need to make a decision around the python API at some point (doesn't have to be right now but this PR will make the problem worse).

If we keep adding items to the synchronous API and NOT the asynchronous API (as done here) then it makes the gap between the APIs larger and makes it harder for us to unify them.

The proper approach should be to either:
* Only add to the async API and add a PR for follow-up
* Add it to the async API and make the synchronous API call the async API (`table_names` on `LanceDBConnection` demonstrates this)

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,109 @@
...
+        const tbl = await db.openTable("images");
+
+        await tbl
+          .query()
+          .where("my_col != 7")
+          .limit(10)
+          .toArray();
+        ```
+
+Scalar indices can also speed up scans containing a vector search or full text search, and a prefilter:
```
westonpace, line 77: Why mention `full text search` here?  I'm not entirely sure what this would imply.  I don't think we ever connected the "inverted scalar index" (which isn't mentioned here) with the `LIKE` operator.

westonpace (commented):
A few concerns / thoughts

docs/src/guides/scalar_index.md
eddyxu: Adding python lancedb api here:

https://github.com/lancedb/lancedb/pull/1495/files#diff-1fd270f2427efc004849c52acb4f60a2d49396cec44dacea2a553455d820acbaR100

docs/src/guides/scalar_index.md
eddyxu: I meant to use pre-filter as part of the FTS.

eddyxu: I can just remove it

westonpace: Ah, no, that makes sense.

westonpace: Go ahead and keep it

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,108 @@
...
+- `BITMAP`: this index stores a bitmap for each unique value in the column.
+  This index is useful for columns with a finite number of unique values and many rows per value.
+  For example, columns that represent "categories", "labels", or "tags"
+- `LABEL_LIST`: a special index that is used to index list columns whose values have a finite set of possibilities.
+  For example, a column that contains lists of tags (e.g. `["tag1", "tag2", "tag3"]`) can be indexed with a `LABEL_LIST` index.
+
+| Data Type                                     | Filter                           | Index Type   |
+| --------------------------------------------- | -------------------------------- | ------------ |
+| Numeric, String, Temporal                      | `<`, `=`, `>`, `in`, `between`, `is null`                    | `BTREE`      |
+| Boolean, numbers or strings with finite range of values         | `<`, `=`, `>`, `in`, `between`, `is null`   | `BITMAP`     |
```
wjones127, line 18: I think you mean "few unique values", right? A sequence of all `u32` values has a finite range, but would be a bad bitmap index, right?
```suggestion
| Boolean, numbers or strings with few unique values        | `<`, `=`, `>`, `in`, `between`, `is null`   | `BITMAP`     |
```

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,108 @@
...
+| List of low cardinality of numbers or strings | `array_has_any`, `array_has_all` | `LABEL_LIST` |
+
+=== "Python"
+
+    ```python
+    import lancedb
+    data = [
+      {"my_col": 1, "category": "A", "vector": [1, 2]},
+      {"my_col": 2, "category": "B", "vector": [3, 4]},
```
wjones127, line 27: Don't have to fix this in this PR, but I feel like it would be easier to understand the examples if they had data that would make sense in the real world. For example, we could instead make the example data an array of metadata about books:

```suggestion
    books = [
      {"book_id": 1, "publisher": "plenty of books", "tags": ["fantasy", "adventure"]},
      {"book_id": 2, "publisher": "book town", "tags": ["non-fiction"]},
```

python/python/lancedb/remote/table.py
wjones127: Yeah, please add to the async API first. We need parity between these two because we need to transition the sync API to be a wrapper around the async API soon. Similar to Node, we don't want to have to keep maintaining two APIs.

docs/src/guides/scalar_index.md
eddyxu: https://github.com/lancedb/lancedb/pull/1495#discussion_r1710010068 
@westonpace, let's get an agreement on the wording here?

docs/src/guides/scalar_index.md
```diff
@@ -0,0 +1,109 @@
+# Building Scalar Index
+
+Similar to many SQL databases, LanceDB supports several types of Scalar indices to accelerate search
+over scalar columns.
+
+- `BTREE`: The most common type is BTREE. This index is inspired by the btree data structure
+  although only the first few layers of the btree are cached in memory.
+  It will perform well on columns with a large number of unique values and few rows per value.
+- `BITMAP`: this index stores a bitmap for each unique value in the column.
+  This index is useful for columns with a small number of unique values and many rows per value.
```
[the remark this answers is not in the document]
wjones127: heh the mathematician in me thinks finite -> could be in trillions.

How about `limited` instead of `finite`, and give a parenthetical "(<5,000)" for added clarity?

docs/src/guides/scalar_index.md
westonpace: I think in the lancedb API PR you said `where the number of unique values is small (i.e., less than a few thousands)`.  Let's use that.  So maybe:

```
Boolean, numbers or strings with fewer than 1,000 unique values
```

That's probably good for "rule of thumb".  I worry that "small" makes people think "dozens".  In some cases a bitmap index will even make sense if there are 10's of thousands of unique values but I don't know how to explain that better without confusing the user.

eddyxu: Done

eddyxu:
>Yeah, please add to the async API first. We need parity between these two because we need to transition the sync API to be a wrapper around the async API soon. Similar to Node, we don't want to have to keep maintaining two APIs.

Added to async Python API and nodejs API

westonpace (approved):

Issue #1496, opened 2024-08-02
bug(node): Unnecessary dependencies cause error in @lancedb/lancedb


wjones127:
### LanceDB version

v0.8.0

### What happened?

When trying to run a simple example, we get these TypeScript errors:

```
node_modules/@lancedb/lancedb/dist/arrow.d.ts:40:32 - error TS2307: Cannot find module 'apache-arrow-13' or its corresponding type declarations.

40 export type FloatLike = import("apache-arrow-13").Float | import("apache-arrow-14").Float | import("apache-arrow-15").Float | import("apache-arrow-16").Float | import("apache-arrow-17").Float;
```

```
node_modules/@lancedb/lancedb/dist/embedding/openai.d.ts:1:44 - error TS2307: Cannot find module 'openai/resources' or its corresponding type declarations.

1 import { type EmbeddingCreateParams } from "openai/resources";
```

It seems for typing purposes, `apache-arrow-13` and `openai` are non-optional for some reason. 

### Are there known steps to reproduce?

`index.ts`:
```ts
import * as lancedb from "@lancedb/lancedb";

(async () => {
    const db = await lancedb.connect("/tmp/testdb");
    const table = await db.createTable("test", [{ name: "id", type: "int" }]);
    console.log(await table.countRows());
})();

```

`package.json`:

```json
{
  "name": "node",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "typescript": "^5.5.4"
  },
  "dependencies": {
    "@lancedb/lancedb": "^0.8.0",
    "openai": "^4.29.2"
  }
}
```

`tsconfig.json`:
```json
{
  "compilerOptions": {
    "moduleResolution": "node16",
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "resolveJsonModule": true,
    "target": "ES2020",
    "module": "node16",
    "strict": false,
  },
}
```

PR #1503, opened 2024-08-05
fix(nodejs): address import issues in `lancedb` npm module

wjones127:
Fixes [#1496](https://github.com/lancedb/lancedb/issues/1496)

### DIFF
--- a/nodejs/lancedb/arrow.ts
+++ b/nodejs/lancedb/arrow.ts
@@ -109,30 +109,13 @@ export function isArrowTable(value: object): value is TableLike {
 }
 
 export function isDataType(value: unknown): value is DataType {
-  return (
-    value instanceof DataType ||
-    DataType.isNull(value) ||
-    DataType.isInt(value) ||
-    DataType.isFloat(value) ||
-    DataType.isBinary(value) ||
-    DataType.isLargeBinary(value) ||
-    DataType.isUtf8(value) ||
-    DataType.isLargeUtf8(value) ||
-    DataType.isBool(value) ||
-    DataType.isDecimal(value) ||
-    DataType.isDate(value) ||
-    DataType.isTime(value) ||
-    DataType.isTimestamp(value) ||
-    DataType.isInterval(value) ||
-    DataType.isDuration(value) ||
-    DataType.isList(value) ||
-    DataType.isStruct(value) ||
-    DataType.isUnion(value) ||
-    DataType.isFixedSizeBinary(value) ||
-    DataType.isFixedSizeList(value) ||
-    DataType.isMap(value) ||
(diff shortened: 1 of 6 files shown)


nodejs/lancedb/arrow.ts
```diff
@@ -103,25 +103,12 @@ export type IntoVector =
...
-  | import("apache-arrow-15").Float
-  | import("apache-arrow-16").Float
-  | import("apache-arrow-17").Float;
-export type DataTypeLike =
-  | import("apache-arrow-13").DataType
-  | import("apache-arrow-14").DataType
-  | import("apache-arrow-15").DataType
-  | import("apache-arrow-16").DataType
-  | import("apache-arrow-17").DataType;
-
```
wjones127, line 118: These cannot appear in our public API without forcing users to install all versions.

westonpace: export function isDataType(value: unknown): value is DataType {
  return sanitizeType(value) instanceof DataType;
}

westonpace (commented):
I don't think either approach works that well.

The former, as you pointed out, required all versions, and it still wouldn't work in all situations.
The latter, requires lock-step versioning, and that is extremely difficult to achieve in node.  

This is because, node will often install two copies of the same library, even if the versions are completely identical (rollup-style approaches basically inline all dependencies).

Can we use the `sanitizeType` method instead?

I added an example in a comment

nodejs/lancedb/arrow.ts
westonpace: (actually, I don't think we even need the `instanceof`, `sanitizeType` will throw if the input isn't type-like.

westonpace: Or, (sorry for stream of consciousness), we should replace all calls to `isDataType` with `sanitizeType`.

wjones127: No worries, I find this Typescript dependency resolution stuff very confusing so I appreciate the help.

westonpace (approved):
Tests pass and this matches my knowledge to the extent I understand anything.  I assume whatever integration test you are testing with is also working?

wjones127:
> Tests pass and this matches my knowledge to the extent I understand anything. I assume whatever integration test you are testing with is also working?

Indeed. Just re-ran it with these changes. I can install a local version of this.

<details><summary>Details</summary>
<p>

Create a local version

```shell
npm run build:release
npm run build-release
npm pack
```

Then you can use that package elsewhere. It runs successfully with type checking. However, I haven't figured out how to pack the native module yet, so it doesn't work at runtime 🤷 

</p>
</details>

PR #1509, opened 2024-08-06
feat(python): index config repr method

### LINKED ISSUE #1506: Feature: IndexConfig should have a __repr__ method
### SDK

Python

### Description

Trying to inspect the indices is disappointing at the moment:
```python
await table.list_indices()
```
```
[<IndexConfig at 0x149d3fb40>, <IndexConfig at 0x14d3b2f10>]
```

We should add a `__repr__` method on `IndexConfig` so we can introspect them easily here.

gagan-bhullar-tech:
PR fixes #1506

### DIFF
--- a/python/src/index.rs
+++ b/python/src/index.rs
@@ -101,7 +101,7 @@ pub struct IndexConfig {
 #[pymethods]
 impl IndexConfig {
     pub fn __repr__(&self) -> String {
-        format!("Index({}, {:?})", self.index_type, self.columns)
+        format!("Index({}, columns={:?})", self.index_type, self.columns)
     }
 }
 
--- a/python/src/index.rs
+++ b/python/src/index.rs
@@ -98,6 +98,13 @@ pub struct IndexConfig {
     pub columns: Vec<String>,
 }
 
+#[pymethods]
+impl IndexConfig {
+    pub fn __repr__(&self) -> String {
+        format!("Index({}, {:?})", self.index_type, self.columns)
+    }
+}
+
 impl From<lancedb::index::IndexConfig> for IndexConfig {
     fn from(value: lancedb::index::IndexConfig) -> Self {
         let index_type = format!("{:?}", value.index_type);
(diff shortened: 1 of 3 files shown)


wjones127:
Hi @gagan-bhullar-tech. Thanks for opening a PR. Unfortunately, this approach will not work. The `pyi` files are just type stubs and aren't meant to hold implementations of methods. Instead you'll need to implement a `repr` method on `IndexConfig` in the Rust files here: https://github.com/lancedb/lancedb/blob/d07d7a5980d1f064956bb417cd7aa1aba7d3f58a/python/src/index.rs#L91

Here is an example of a repr method being implemented in Rust:

https://github.com/lancedb/lancedb/blob/d07d7a5980d1f064956bb417cd7aa1aba7d3f58a/python/src/table.rs#L197-L202

Also, if you implement this, could you add a unit test that verifies the repr method shows data as expected?

gagan-bhullar-tech:
@wjones127 thanks for pointing it out. Will do that.

gagan-bhullar-tech:
@wjones127 can you please confirm whether test need to be written in rust or python

gagan-bhullar-tech:
> @wjones127 can you please let me know whether test needs to be written in rust or python

Nevermind. It looks like it should be in python.

gagan-bhullar-tech:
@wjones127 can you please review

python/src/index.rs
```diff
@@ -98,6 +98,13 @@ pub struct IndexConfig {
     pub columns: Vec<String>,
 }
 
+#[pymethods]
+impl IndexConfig {
+    pub fn __repr__(&self) -> String {
+        format!("Index({}, {:?})", self.index_type, self.columns)
```
wjones127, line 104: The first argument is probably clear, but the second argument is an arbitrary list of strings, so would be clearer what they mean if we added a name to them.
```suggestion
        format!("Index({}, columns={:?})", self.index_type, self.columns)
```

gagan-bhullar-tech:
@wjones127 can you please review

wjones127 (approved):
Looks great, thanks!

PR #1520, opened 2024-08-07
feat(python): add to_list to async api

### LINKED ISSUE #1517: Feature: add synchronous `to_list` method to the Python async API
### SDK

Python

### Description

Because of the difference between the two api (working with asynchronous table and synchronous table) we have to rewrite most of the code, it is not always handy and does not always get the same result.
It would be great to have the same method when working with an asynchronous table when vector_search

gagan-bhullar-tech:
PR fixes #1517

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1177,6 +1177,16 @@ class AsyncQueryBase(object):
             await batch_iter.read_all(), schema=batch_iter.schema
         )
 
+    async def to_list(self) -> List[dict]:
+        """
+        Execute the query and return the results as a list of dictionaries.
+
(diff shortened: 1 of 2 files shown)


wjones127:
Thanks for making a PR. Two general points to consider for this PR and future PRs:

1. When something is adding a new API, it is a new feature, not a bugfix. So this PR should be labelled `feat:` not `fix:`.
2. Any new API should have a corresponding unit tests. We don't merge any new features or bug fixes that don't have some form of test. Otherwise we have no guarantee it actually works and will continue working as others update the code.

gagan-bhullar-tech:
@wjones127 please review

wjones127 (approved):
Looks good. Thanks @gagan-bhullar-tech!

PR #1531, opened 2024-08-10
Remove native.d.ts from .npmignore

mattbasta:
Presently the `opts` argument to `lancedb.connect` is typed `any`, even though it shouldn't be.

image

Clicking into the type definitions for the published package, it has the correct type signature:

image

However, `ConnectionOptions` is imported from `native.js` (along with a number of other imports a bit further down):

image

This is not otherwise an issue, except that the type definitions for `native.js` are not included in the published package:

image

I haven't compiled the Rust code and run the build script, but I strongly suspect that disincluding the type definitions in `.npmignore` is ultimately the root cause here.

### DIFF
--- a/nodejs/.npmignore
+++ b/nodejs/.npmignore
@@ -20,6 +20,5 @@ Cargo.toml
 biome.json
 build.rs
 jest.config.js
-native.d.ts
 tsconfig.json
 typedoc.json


wjones127:
This makes sense to me, I'm not sure why it's excluded.

@universalmind303 in case you have any objections, since you approved the initial PR removing this https://github.com/lancedb/lancedb/pull/1467

wjones127 (approved):

universalmind303:
> This makes sense to me, I'm not sure why it's excluded.
> 
> @universalmind303 in case you have any objections, since you approved the initial PR removing this #1467

yeah this seems reasonable. I don't think it was supposed to be added to the `.npmignore`

PR #1537, opened 2024-08-13
feat: support to query/index FTS on RemoteTable/AsyncTable

BubbleCal:


python/python/lancedb/query.py
```diff
@@ -135,6 +135,8 @@ def create(
         vector_column_name: str,
         ordering_field_name: str = None,
         fts_columns: Union[str, List[str]] = None,
+        *,
+        use_tantivy=True,
```
eddyxu, line 139: This is query, we should be able to infer what fts index it has?

For example, if we have native index, whether set this `use_tantive = True | False` does not matter, IIRC?

python/python/lancedb/query.py
```diff
@@ -675,13 +684,18 @@ def __init__(
         query: str,
         ordering_field_name: str = None,
         fts_columns: Union[str, List[str]] = None,
+        *,
+        use_tantivy: bool = True,
```
eddyxu, line 688: Same here, we should be able to infer this parameter from index?

python/python/lancedb/query.py
```diff
@@ -711,23 +724,25 @@ def to_arrow(self) -> pa.Table:
...
                 "Use tantivy-based index instead for now."
             )
-        if self._reranker:
-            raise NotImplementedError(
-                "Reranking is not yet supported in Lance FTS. "
-                "Use tantivy-based index instead for now."
-            )
-        ds = self._table.to_lance()
-        return ds.to_table(
+        # if self._reranker:
```
eddyxu, line 727: lets just delete this.

python/python/lancedb/query.py
```diff
@@ -849,15 +864,20 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
     in the `rerank` method to convert the scores to ranks and then normalize them.
     """
 
-    def __init__(self, table: "Table", query: str, vector_column: str):
+    def __init__(
+        self, table: "Table", query: str, vector_column: str, *, use_tantivy=True
```
eddyxu, line 868: Dont need use_tantivy during query.

python/python/lancedb/remote/table.py
```diff
@@ -118,6 +118,22 @@ def create_scalar_index(
 
         return resp
 
+    def create_fts_index(
```
eddyxu, line 121: actually, can we make this work the same as `AsyncTable` API, which will be our future standard APIs

python/python/lancedb/remote/table.py
```diff
@@ -307,10 +324,22 @@ def search(
             - and also the "_distance" column which is the distance between the query
             vector and the returned vector.
         """
-        if vector_column_name is None:
-            vector_column_name = inf_vector_column_query(self.schema)
-        query = LanceQueryBuilder._query_to_vector(self, query, vector_column_name)
-        return LanceVectorQueryBuilder(self, query, vector_column_name)
+        if vector_column_name is None and query is not None:
+            try:
```
eddyxu, line 328: this feels wrong. Why can we pass `query_type = fts` in `vector search`?

python/python/lancedb/table.py
```diff
@@ -1488,12 +1489,19 @@ def search(
                 else:
                     raise e
 
+        use_tantivy = False
```
eddyxu, line 1492: How does this code work with s3/gcs? will it crash?

python/python/lancedb/query.py
BubbleCal: yes, this param isn't exposed to users, would set it by existence of tantivy index (true if exists)
see t`able.py`'s `search()` method

eddyxu: lets put the this logic right before you call `scan ` or `tantivy`, to reduce the context between the functions.

python/python/lancedb/remote/table.py
BubbleCal: it's all-in-one search API, can do FTS and vector search, depending on the `query_type` and the type of `query`.
this is:
- if the vector_column isn't specified, then try to infer it
- if can't find such a vector column
    - if it's FTS, then it's fine
    - if it's vector search, we need to report error

python/python/lancedb/query.py
```diff
@@ -740,24 +740,23 @@ def tantivy_to_arrow(self) -> pa.Table:
...
             raise FileNotFoundError(
                 "Fts index does not exist. "
                 "Please first call table.create_fts_index(['<field_names>']) to "
                 "create the fts index."
             )
+
+        # Check that we are on local filesystem
+        if not isinstance(fs, pa_fs.LocalFileSystem):
+            raise NotImplementedError(
+                "Full-text search is only supported on the local filesystem"
```
westonpace, line 756:
```suggestion
                "Tantivy-based full-text search is only supported on the local filesystem"
```

python/python/lancedb/remote/table.py
westonpace: Can we change the top-level if statement to `if vector_column_name is None and query is not None and query_type != "fts":`?

westonpace (commented):
Some minor comments but looks good to me otherwise.

python/python/lancedb/query.py
BubbleCal: Fixed

python/python/lancedb/table.py
BubbleCal: for S3/GCS it would throw a exception, see `query.py: 754`

python/python/lancedb/remote/table.py
BubbleCal: Yes that's good, Fixed

python/python/lancedb/remote/table.py
BubbleCal: Added

python/python/lancedb/query.py
BubbleCal: Removed

python/python/lancedb/query.py
BubbleCal: Removed

python/python/lancedb/query.py
BubbleCal: Removed

python/python/lancedb/query.py
BubbleCal: Removed

westonpace (approved):

PR #1538, opened 2024-08-13
feat: allow new scalar index types to be created in remote table

rpgreen:


eddyxu (approved):

python/python/lancedb/remote/table.py
```diff
@@ -97,15 +97,18 @@ def index_stats(self, index_uuid: str):
     def create_scalar_index(
         self,
         column: str,
+        index_type: str = "scalar",
```
westonpace, line 100: You could use a union of literals here:

```
Literal["scalar", "btree", "bitmap", "label_list"] = "scalar"
```

Also, any reason these are different than the local verisons which are `BTREE`, `BITMAP`, and `LABEL_LIST`?

westonpace (approved):

python/python/lancedb/remote/table.py
rpgreen: Updated. The server can handle both, but I've updated for consistency. "scalar" (and "vector" for vector indices) are retained for backwards-compatibility.

PR #1541, opened 2024-08-14
chore: bump the required python version to 3.9

BubbleCal:


wjones127 (approved):
Looks good. Note that our required dependency, lance, already requires python>=3.9.

https://github.com/lancedb/lance/blob/865bce7f0000c76d15bcbaa3c7d1633bc5fc7d05/python/pyproject.toml#L9

PR #1542, opened 2024-08-15
feat(python): add delete unverified parameter

### LINKED ISSUE #1527: Feature: add `delete_unverified` parameter to optimize operation
### SDK

Node, Python

### Description

Sometimes users will want to force a cleanup faster, but we don't expose this parameter to them.

https://lancedb.github.io/lance/api/python/lance.html#lance.LanceDataset.cleanup_old_versions

See example in: #1526

gagan-bhullar-tech:
PR fixes #1527

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2438,7 +2438,10 @@ class AsyncTable:
         await self._inner.restore()
 
     async def optimize(
-        self, *, cleanup_older_than: Optional[timedelta] = None
+        self,
+        *,
+        cleanup_older_than: Optional[timedelta] = None,
+        delete_unverified: bool = False,
     ) -> OptimizeStats:
         """
         Optimize the on-disk data and indices for better performance.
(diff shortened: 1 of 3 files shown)


python/python/lancedb/table.py
```diff
@@ -2438,7 +2438,10 @@ async def restore(self):
         await self._inner.restore()
 
     async def optimize(
-        self, *, cleanup_older_than: Optional[timedelta] = None
+        self,
+        *,
+        cleanup_older_than: Optional[timedelta] = None,
+        delete_unverified: bool = False,
```
wjones127, line 2444: Could you also add documentation below for the parameter? You can model it off of the docs here:

https://lancedb.github.io/lance/api/python/lance.html#lance.LanceDataset.cleanup_old_versions

wjones127 (changes requested):

python/python/lancedb/table.py
gagan-bhullar-tech: @wjones127 updated PR.

wjones127 (approved):
Excellent work!

Issue #1546, opened 2024-08-19
bug(python): ColbertReranker uses wrong architecture & other inconsistencies


AyushExel:
### LanceDB version

v5.0

### What happened?

- ColBERT predates the `trust_remote_code` option on HF, so the model architecture is actually wrong 
- MaxSim works by summing rather than averaging -- the torch.mean() at the end will mess things up (you should just sum up the score of all query tokens)
- The query length is super important, and needs a specific kind of padding with <MASK> token to take query augmentation into account, but not too many <MASK>s, as they're just augmenters.
-

Issue #1487, opened 2024-08-01
Feature: Add cuda acceleration for all on-device embedding functions & rerankers


AyushExel:
### SDK

Python

### Description

- [x] Hf
- [x] Imagebind
- [x] Colbert reranker
- [x] CrossEncoder Reranker

PR #1547, opened 2024-08-19
chore(python): update Colbert architecture and minor improvements

AyushExel:
- Update ColBertReranker architecture: The current implementation doesn't use the right arch. This PR uses the implementation in Rerankers library. Fixes https://github.com/lancedb/lancedb/issues/1546
Benchmark diff (hit rate):
Hybrid - 91 vs 87
reranked vector - 85 vs 80

- Reranking in FTS is basically disabled in main after last week's FTS updates. I think there's no blocker in supporting that? 
- Allow overriding accelerators: Most transformer based Rerankers and Embedding automatically select device. This PR allows overriding those settings by passing `device`. Fixes: https://github.com/lancedb/lancedb/issues/1487

### DIFF
--- a/python/python/lancedb/rerankers/colbert.py
+++ b/python/python/lancedb/rerankers/colbert.py
@@ -37,7 +37,7 @@ class ColbertReranker(Reranker):
         doc_ids = [i for i in range(len(docs))]
         result = self.colbert.rank(query, docs, doc_ids=doc_ids)
 
-        # get the scores of each document in the same order as the input       
+        # get the scores of each document in the same order as the input
         scores = [result.get_result_by_docid(i).score for i in doc_ids]
 
         # add the scores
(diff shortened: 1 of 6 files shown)


python/python/lancedb/rerankers/colbert.py
```diff
@@ -22,41 +20,26 @@ class ColbertReranker(Reranker):
...
-        self.torch = attempt_import_or_raise(
-            "torch"
+        rerankers = attempt_import_or_raise(
+            "rerankers"
         )  # import here for faster ops later
+        self.colbert = rerankers.Reranker(self.model_name, model_type="colbert")
 
     def _rerank(self, result_set: pa.Table, query: str):
         docs = result_set[self.column].to_pylist()
+        doc_ids = [i for i in range(len(docs))]
```
BubbleCal, line 37:
```suggestion
        doc_ids = list(range(len(docs)))
```

BubbleCal (approved):
LGTM, just left a NIT comment

PR #1552, opened 2024-08-21
docs: add evaluation example

rithikJha:
before:
Screenshot 2024-08-21 194228

After:
Screenshot 2024-08-21 195258


docs/src/examples/python_examples/evaluations.md
```diff
@@ -0,0 +1,23 @@
+**Evaluation: Assessing Text Performance with Precision 📊💡**
+====================================================================
+
+**Evaluation Fundamentals 📊**
+
+The Evaluation Application is a comprehensive tool designed to measure the performance of text-based inputs, enabling data-driven optimization and improvement 📈.
```
AyushExel, line 6: This part talks about "The evaluation application"? I think this should just be a short into to what evaluation is as there isn't any specific eval application?

PR #1556, opened 2024-08-21
feat(python, rust): expose offset in query

### LINKED ISSUE #1555: Feature: expose `offset` in query APIs
### SDK

Python, TypeScript, Rust

### Description

Support a version of pagination where users can provide `offset` in addition to `limit`. The underlying Lance APIs support this, so it shouldn't be hard to implement.

The `explain_plan` PR is a good model for how to wire up a LanceDB query API to an existing Lance feature: https://github.com/lancedb/lancedb/commit/46c6ff889d37733c277f3b40d95c317af30f1c8b

* [x] Rust
* [x] Python
* [ ] Typescript

gagan-bhullar-tech:
PR is part of #1555


python/python/tests/test_query.py
```diff
@@ -106,6 +106,12 @@ class TestModel(LanceModel):
     assert r0.float_field == 1.0
 
 
+def test_offset(table):
+    q = LanceVectorQueryBuilder(table, [0, 0], "vector").offset(1)
+    results = q.to_pandas()
+    assert len(results) == 2
```
wjones127, line 112: I'm confused. The table only has two rows. If you pass `offset(1)`, shouldn't there only be one result?

It's not clear to me this test is doing anything meaningful. You might need to update the `_execute_query(` method above.

wjones127 (changes requested):
It looks like you only attempted to add a unit test for the sync Python API. Please make sure to test the Python async API and the Rust API.

gagan-bhullar-tech:
@wjones127 thanks. Can you please review.

python/python/tests/test_query.py
```diff
@@ -106,6 +107,12 @@ class TestModel(LanceModel):
     assert r0.float_field == 1.0
 
 
+def test_offset(table):
+    q = LanceVectorQueryBuilder(table, [0, 0], "vector").offset(1)
+    results = q.to_pandas()
+    assert len(results) == 1
```
wjones127, line 113: Just want to keep this test robust in case someone changes how `table` is defined.
```suggestion
    assert table.count_rows() == 2
    q = LanceVectorQueryBuilder(table, [0, 0], "vector").offset(1)
    results = q.to_pandas()
    assert len(results) == 1
```

rust/lancedb/src/query.rs
```diff
@@ -338,6 +338,12 @@ pub trait QueryBase {
     /// it will default to 10.
     fn limit(self, limit: usize) -> Self;
 
+    // Set the offset of the query.
+
```
wjones127, line 342:
```suggestion
    /// Set the offset of the query.
    ///
```

python/python/tests/test_query.py
```diff
@@ -270,6 +277,7 @@ async def test_query_async(table_async: AsyncTable):
         expected_columns=["foo", "bar"],
     )
     await check_query(table_async.query().limit(1), expected_num_rows=1)
+
```
wjones127, line 280: Could you add a test for async table for offset as well?

rust/lancedb/src/query.rs
```diff
@@ -870,6 +886,7 @@ mod tests {
             new_vector
         );
         assert_eq!(query.base.limit.unwrap(), 100);
+        assert_eq!(query.base.offset.unwrap(), 1);
```
wjones127, line 889: Please also test offset in `test_execute` below. This just tests the builder.

wjones127 (changes requested):

gagan-bhullar-tech:
@wjones127 please review

gagan-bhullar-tech:
@wjones127 please review

wjones127 (approved):
Looks great. I'll leave a todo in the original issue for Typescript support, but we can check off Python and Rust :)

PR #1558, opened 2024-08-22
fix: add appropriate QueryBuilder overloads to LanceTable.search

### LINKED ISSUE #1550: bug(python): Attribute "rerank" is unknown
### LanceDB version

v0.12.0

### What happened?

Attempting to use `.rerank` method on `table.search` results generates `reportAttributeAccessIssue` in `pyright` and `pylance`.
```python
assert isinstance(table, LanceTable)
result3 = (
    table.search(query, query_type="hybrid")
    .where(criterion, prefilter=True)
    .limit(5)
    .rerank(reranker=reranker)  # <- Attribute "rerank" is unknown
)
```
image

### Expected:
Should not issue type checker error.

Create a LanceTable instance, and try to perform reranking on a search result.

```python
import lancedb  # type: ignore[import-untyped]
from lancedb.embeddings import get_registry  # type: ignore[import-untyped]
from lancedb.pydantic import LanceModel, Vector  # type: ignore[import-untyped]
from lancedb.rerankers import LinearCombinationReranker  # type: ignore[import-untyped]

model_registry = get_registry().get("sentence-transformers")
model = model_registry.create(name="BAAI/bge-small-en-v1.5")

class Document(LanceModel):
    text: str = model.SourceField()
    vector: Vector(model.ndims()) = model.VectorField()

db = lancedb.connect(".test_db")
assert isinstance(db, lancedb.LanceDBConnection)
table = db.create_table("test_table", schema=Document, exist_ok=True)
table.create_fts_index("text")

reranker = LinearCombinationReranker(weight=0.3, return_score="all")

result = (
    table.search("foo", query_type="hybrid")
    .where("category = 'bar'", prefilter=True)
    .limit(5)
    .rerank(reranker=reranker)
)
```

---

This happens because `LanceTable.search` returns a `LanceQueryBuilder`. It should dispatch different types of query builders based on `query_type` parameter. Adding a few overloads on `LanceTable.search` based on `Literal` query types should solve the issue.

sayandipdutta:
- Add overloads to Table.search, to preserve the return information
of different types of QueryBuilder objects for LanceTable
- Fix fts_column type annotation by including making it `Optional`

resolves #1550


AyushExel:
> resolves https://github.com/lancedb/lancedb/pull/1554
Thanks for the PR. I think this is independent of #1554, no?

sayandipdutta:
Well, it solves more problems, but certainly solves #1554 as well. The `rerank` attribute was not found because even after secifying `query_type="hybrid"`, type checker didn't have the information that `search` was returning a `VectorQueryBuilder` (which has the `rerank` method).

sayandipdutta:
Ah, I see what you mean. I pasted the wrong link. Fixing it

wjones127 (approved):
Thank you!

PR #1563, opened 2024-08-25
feat: support fuzzy matching in full text search

MagnusS0:
This pull request introduces support for fuzzy matching in full text search. The changes include modifications to the `search_index` function and related classes to accept and handle fuzzy matching parameters. It uses **tantivy-py's**  `FuzzyTermQuery`.

**Changes**

- Modified `search_index` function in `fts.py` to accept `fuzzy_fields `parameter.
- Updated create method in `query.py` to pass `fuzzy_fields` to the `search_index` function.
- Added `fuzzy_field`s parameter to `LanceQueryBuilder` class in `query.py`.
- Updated search method in `table.py` to accept and pass `fuzzy_fields` to the `LanceQueryBuilder`.

**Example usage**
```python
fuzzy_fields = {
    'keywords': (True, 2, True),
    'id': (False, 0, True),
    'title': (True, 2, True),
    'combined_text': (False, 2, True),
}

input = ""
fts = tbl.search(input, query_type="fts", fuzzy_fields=fuzzy_fields).limit(10).to_pandas()
```

### DIFF
--- a/python/python/lancedb/fts.py
+++ b/python/python/lancedb/fts.py
@@ -14,7 +14,7 @@
 """Full text search index using tantivy-py"""
 
 import os
-from typing import List, Tuple
+from typing import List, Tuple, Dict
 
 import pyarrow as pa
 
@@ -171,7 +171,11 @@ def resolve_path(schema, field_name: str) -> pa.Field:
 
 
 def search_index(
-    index: tantivy.Index, query: str, limit: int = 10, ordering_field=None
+    index: tantivy.Index,
+    query: str,
+    limit: int = 10,
+    ordering_field: str = None,
+    fuzzy_fields: Dict[str, Tuple[bool, int, bool]] = None,
 ) -> Tuple[Tuple[int], Tuple[float]]:
     """
     Search an index for a query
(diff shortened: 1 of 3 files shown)


wjones127:
Hi @MagnusS0. Thanks for your PR.

I'm not sure if you've been following this work, but we've been moving away from using Tantivy for FTS in favor of our own FTS implementation. The current tantivy-py implementation is limited to Python, and not available to Rust or Node users. It also doesn't support object storage, like S3. We didn't have a viable way to address these limitations while sticking with Tantivy, so that's why we have made our own. This work is ongoing and tracked in https://github.com/lancedb/lance/issues/1195

Given this, I'm reluctant to expose additional features that are based on `tantivy`. Instead, I'd prefer we made PRs or feature requests to add those features into our native FTS indices.

MagnusS0:
Hi @wjones127 

Thank you for the feedback and for the update on the direction of the project regarding FTS.

I should have checked the ongoing work before submitting the PR. I’ll look into how fuzzy search might be implemented within the native FTS indices or if it’s already being addressed in lancedb/lance#1195.

PR #1567, opened 2024-08-27
fix(node): read consistency level fix

### LINKED ISSUE #1565: bug(node): `readConsistencyInterval` not passed down in `vectordb`
### LanceDB version

0.4.20

### What happened?

`readConsistencyInterval` doesn't seem to be properly passed down. This is likely never 
receiving the correct argument here:

https://github.com/lancedb/lancedb/blob/9c259981108768e6ab01cb5f7948389266f2b197/rust/ffi/node/src/lib.rs#L51-L52

We should test this like we do in the newer SDK:

https://github.com/lancedb/lancedb/blob/9c259981108768e6ab01cb5f7948389266f2b197/nodejs/__test__/table.test.ts#L540

_No response_

gagan-bhullar-tech:
PR fixes #1565

### DIFF
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -105,6 +105,7 @@ describe("LanceDB client", function () {
       })
       const table2 = await db2.openTable("vectors");
 
+      assert.equal(await table2.countRows(), 2);
       await table1.add([
         {
           id: 3,
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -93,6 +93,29 @@ describe("LanceDB client", function () {
       const con = await lancedb.connect(uri);
       assert.deepEqual(await con.tableNames(), ["vectors"]);
     });
+
+    it("read consistency level", async function () {
+      const uri = await createTestDB();
+      const db1 = await lancedb.connect({ uri });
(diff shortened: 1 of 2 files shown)


gagan-bhullar-tech:
@wjones127 please review.

node/src/test/test.ts
```diff
@@ -93,6 +93,29 @@ describe("LanceDB client", function () {
...
+      const db1 = await lancedb.connect({ uri });
+      const table1 = await db1.openTable("vectors");
+
+      const db2 = await lancedb.connect({
+        uri,
+        readConsistencyInterval: 0
+      })
+      const table2 = await db2.openTable("vectors");
+
+      await table1.add([
```
wjones127, line 108: We should make this test more robust by asserting the state before the change as well:
```suggestion
      assert.equal(await table2.countRows(), 2);
      await table1.add([
```

wjones127 (changes requested):

gagan-bhullar-tech:
@wjones127 Updated PR. Please review.

wjones127 (approved):

wjones127:
Thanks @gagan-bhullar-tech !

PR #1570, opened 2024-08-28
docs: add hnsw docs

AyushExel:


jacketsj (changes requested):
I think there are few changes needed for the HNSW description to be correct. I left a suggested version for each section that I think needs changing.  
Also, I strongly encourage not discussing NSW, since it's not actually as similar to HNSW as plain old k-NN graphs (and it looks like it got you confused you too with the high/low degree vertices, which aren't used in HNSW).

jacketsj (approved):
lgtm
figures would be nice in the future

PR #1572, opened 2024-08-28
fix: specify column to search for FTS

BubbleCal:
Before this we ignored the `fts_columns` parameter, and for now we support to search on only one column, it could lead to an error if we have multiple indexed columns for FTS

### DIFF
--- a/python/python/tests/test_fts.py
+++ b/python/python/tests/test_fts.py
@@ -142,10 +142,75 @@ def test_search_fts(table, use_tantivy):
     assert len(results) == 5
 
 
+def test_search_fts_specify_column(table):
+    table.create_fts_index("text", use_tantivy=False)
+    table.create_fts_index("text2", use_tantivy=False)
+
+    results = table.search("puppy", fts_columns="text").limit(5).to_list()
+    assert len(results) == 5
+
+    results = table.search("puppy", fts_columns="text2").limit(5).to_list()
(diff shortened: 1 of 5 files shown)


python/python/tests/test_fts.py
```diff
@@ -142,12 +142,77 @@ def test_search_fts(table, use_tantivy):
...
 
+def test_search_fts_specify_column(table):
+    table.create_fts_index("text", use_tantivy=False)
+    table.create_fts_index("text2", use_tantivy=False)
+
+    results = table.search("puppy", fts_columns="text").limit(5).to_list()
+    assert len(results) == 5
+
+    results = table.search("puppy", fts_columns="text2").limit(5).to_list()
+    assert len(results) == 5
```
wjones127, line 153: Could we change this so it uses two different text columns? Right now `text` and `text2` are identical. Because of this, this test wouldn't be able to detect whether this parameter is ignored and just the first text column available is used. The results of this query should be different, and we should be able to assert they are different.

wjones127 (changes requested):
One suggestion on how to make one of the tests more robust.

python/python/tests/test_fts.py
BubbleCal: Fixed

wjones127 (approved):

Issue #937, opened 2024-02-06
Feature: Better API for handling hybrid queries


AyushExel:
### SDK

Python

### Description

Currently, there is only 1 documented pathway for performing hybrid search.
- `table.search("this is a query", query_type="hybrid")
This'll only work if the Embedding API is used.
- `table.search((vector, string))` -- Undocumented, Temporary solution to allow search without Embedding API usage.

~Need to provide a better way to take user input of vector and fts query input that's more explicit. Something like
`table.search(vector=, text=) ` would be better as parameters as explicit~

### Based on [this](https://github.com/lancedb/lancedb/pull/1013#pullrequestreview-1913154943) comment:

overall i'm wary of having `query`, `vector` and `text` parameters.

how about this. For hybrid search, if you have an embedding function setup, then you can do it like this:

```
table.search("user query", query_type="hybrid")
    .rerank(...)
    .limit(10)
    .to_arrow()
```

if you don't have an embedding function setup, you can do this:

```
table.search(query_type="hybrid")
    .vector(<vec>)
    .text(<str>)
    .rerank(...)
    .limit(10)
    .to_arrow()
```

if you have embedding function setup but you want to parse your own terms from the user query:

```
table.search(query="user question", query_type="hybrid")
    .text(<str>)
    .rerank(...)
    .limit(10)
    .to_arrow()
```

PR #1575, opened 2024-08-29
feat!: better api for manual hybrid queries

AyushExel:
Currently, the only documented way of performing hybrid search is by using embedding API and passing string queries that get automatically embedded. There are use cases where users might like to pass vectors and text manually instead. 
This ticket contains more information and historical context - https://github.com/lancedb/lancedb/issues/937

This breaks a undocumented pathway that allowed passing (vector, text) tuple queries which was intended to be temporary, so this is marked as a breaking change. For all practical purposes, this should not really impact most users

### usage
```
results = table.search(query_type="hybrid")
                .vector(vector_query)
                .text(text_query)
                .limit(5)
                .to_pandas()
```


python/python/lancedb/query.py
```diff
@@ -206,8 +207,6 @@ def _resolve_query(cls, table, query, query_type, vector_column_name):
         elif query_type == "auto":
             if isinstance(query, (list, np.ndarray)):
                 return query, "vector"
-            if isinstance(query, tuple):
-                return query, "hybrid"
```
BubbleCal, line 210: can we remain this? or it's a breaking change

python/python/lancedb/query.py
```diff
@@ -895,40 +926,69 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
...
-                "The query must be either a string or a tuple of (vector, string)."
+                "You can either provide query or its vector representation"
+                "but not both."
             )
 
+        vector_query = query or vector
+        if not isinstance(vector_query, (str, list, np.ndarray)):
+            raise ValueError("Vector query must be either a string or a vector")
+
+        text_query = query or text
```
BubbleCal, line 953: I think here we need to exchange the order of parameters, consider these cases:
`table.search(some_string).text(fts_query)`
then it's supposed to `vector_query=some_string` and `text_query=fts_query`
and for `table.search(some_string).vector(vec_query)`
then it's supposed to `vector_query=vec_query` and `text_query=some_string`

python/python/lancedb/query.py
```diff
@@ -1150,7 +1129,15 @@ def refine_factor(self, refine_factor: int) -> LanceHybridQueryBuilder:
         LanceHybridQueryBuilder
             The LanceHybridQueryBuilder object.
         """
-        self._vector_query.refine_factor(refine_factor)
+        self._refine_factor = refine_factor
+        return self
+
+    def vector(self, vector: Union[np.ndarray, list]) -> LanceHybridQueryBuilder:
```
BubbleCal, line 1135: NIT: it can be a string right?

python/python/lancedb/query.py
```diff
@@ -895,40 +926,69 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
...
-                    "The query must be a tuple of (vector_query, fts_query)."
-                )
-            if not isinstance(query[0], (list, np.ndarray, pa.Array, pa.ChunkedArray)):
-                raise ValueError(f"The vector query must be one of {VEC}.")
-            if not isinstance(query[1], str):
-                raise ValueError("The fts query must be a string.")
-            return query[0], query[1]
-        else:
+    def _validate_query(self, query, vector=None, text=None):
+        if query is not None and vector is not None:
```
BubbleCal, line 943: if we don't allow the users to pass both query & vector, we should ban text as well?

python/python/lancedb/query.py
AyushExel: this will be a breaking change. The tuple pathway was temporary for internal benchmarking while we didn't have the API finalised

python/python/lancedb/query.py
AyushExel: Right..makes sense

python/python/lancedb/query.py
AyushExel: Actually I wasn't sure about this based on [this comment](https://github.com/lancedb/lancedb/pull/1013#pullrequestreview-1913154943) (see the last usage). But your suggestion sounds better as it'll be less confusing for the user. Updated it!

python/python/lancedb/query.py
AyushExel: No I think this one can't be string. This is meant for cases where user wants to explicitly pass the vector without relying on our embedding API

AyushExel:
@BubbleCal addressed reviews. Can you take a look again?

python/python/lancedb/query.py
BubbleCal: So if the users want to use our embedding API, it's
```
table.search(vec_string).text(fts_query)
```
right?

BubbleCal (approved):

python/python/lancedb/query.py
AyushExel: With embedding API the string will automatically be embedded and it will also be used for fts. 
'Search(text_query, query_type="hybrid")'
This is the current usage. 
With the new API, we'll allow passing vector and text queries separately.

Symbolk:
Does Hybrid Search only support Python API now? I failed to find the corresponding JS/TS API.

PR #1576, opened 2024-08-29
fix: can't do full text search when queryType='auto'

### LINKED ISSUE #1557: bug(node): TypeError: lancedb.Index.fts is not a function
### LanceDB version

0.9.0

### What happened?

I am trying to implement full-text search using JavaScript, but I encountered an error when testing the official sample code.

`import * as lancedb from "@lancedb/lancedb";
const uri = "data/sample-lancedb"
const db = await lancedb.connect(uri);

const data = [
{ vector: [3.1, 4.1], text: "Frodo was a happy puppy" },
{ vector: [5.9, 26.5], text: "There are several kittens playing" },
];
const tbl = await db.createTable("my_table", data, { mode: "overwrite" });
await tbl.createIndex("text", {
    config: lancedb.Index.fts(),
});

await tbl
    .search("puppy")
    .select(["text"])
    .limit(10)
    .toArray();`

BubbleCal:
When queryType='auto' and the query is a string,
we would check whether there are any embedding functions registered, if so, then it's a vector search with the embedding, otherwise it's full text search.

The current TypeScript implementation always register 2 embedding functions, then it breaks the check. But it won't enable the embedding functions if the schema doesn't contain the metadata "embedding_functions".

fix #1557

### DIFF
--- a/nodejs/lancedb/table.ts
+++ b/nodejs/lancedb/table.ts
@@ -602,18 +602,12 @@ export class LocalTable extends Table {
       });
     }
 
-    // The query type is auto or vector
-    // fall back to full text search if no embedding functions are defined and the query is a string
-    if (queryType === "auto" && getRegistry().length() === 0) {
-      return this.query().fullTextSearch(query, {
-        columns: ftsColumns,
-      });
-    }
-
     // TODO: Support multiple embedding functions
     const embeddingFunc: EmbeddingFunctionConfig | undefined =
       this.getEmbeddingFunctions().values().next().value;
     if (!embeddingFunc) {
+      // The query type is auto and string
+      // fall back to full text search if no embedding functions are defined and the query is a string
       if (queryType === "auto" && typeof query === "string") {
         return this.query().fullTextSearch(query, {
           columns: ftsColumns,
--- a/nodejs/lancedb/table.ts
+++ b/nodejs/lancedb/table.ts
@@ -613,7 +613,11 @@ export class LocalTable extends Table {
           columns: ftsColumns,
         });
       }
-      throw new Error("No embedding functions are defined in the table");
+      return this.query().nearestTo(
+        Promise.reject(
+          new Error("No embedding functions are defined in the table"),
+        ),
+      );
     }
     const queryPromise = embeddingFunc.function.computeQueryEmbeddings(query);
     return this.query().nearestTo(queryPromise);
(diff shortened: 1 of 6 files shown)


nodejs/lancedb/table.ts
```diff
@@ -456,10 +456,12 @@ export abstract class Table {
...
 export class LocalTable extends Table {
   private readonly inner: _NativeTable;
+  private readonly schema_: Schema;
 
-  constructor(inner: _NativeTable) {
+  constructor(inner: _NativeTable, schema: Schema) {
     super();
     this.inner = inner;
+    this.schema_ = schema;
   }
```
westonpace, line 465: Are you caching this to avoid the async call?  I'm not sure it will work.  The reason that `schema()` is asynchronous is because it is possible for the table to have been changed (by other processes) and we want to respect the `read_consistency_interval` and detect this change automatically.

nodejs/lancedb/table.ts
```diff
@@ -603,29 +602,24 @@ export class LocalTable extends Table {
...
+      if (queryType === "auto" && typeof query === "string") {
+        return this.query().fullTextSearch(query, {
+          columns: ftsColumns,
+        });
+      }
+      return this.query().nearestTo(
+        Promise.reject(
+          new Error("No embedding functions are defined in the table"),
+        ),
+      );
```
westonpace, line 620: It looks like you are passing a failed promise to the `nearestTo` function.  Is this intentional?

westonpace (commented):
Thanks for working on this.  I have a question about caching the schema.

nodejs/lancedb/table.ts
BubbleCal: Fixed, cache only the registered embedding functions

nodejs/lancedb/table.ts
BubbleCal: I think I did this for passing some tests, just fixed it

BubbleCal:
It seems there's very weird behavior when opening the dataset, now I'm a little suspicious that whether we should allow doing FTS when `queryType=auto` because I noticed that the python SDK can't do this as well.

westonpace:
Should we close this or do you plan on coming back to it?

BubbleCal:
Close it
We need consistent API for all SDKs

PR #1580, opened 2024-08-30
feat: update default reranker to RRF

AyushExel:
- Both LinearCombination (the current default) and RRF are pretty fast compared to model based rerankers. RRF is slightly faster.
- In our tests RRF has also been slightly more accurate.

This PR:
- Makes RRF the default reranker
- Removed duplicate docs for rerankers


westonpace (approved):
Seems like a good change to me.  Are the docs duplicated because we have individual pages now for each of the rerankers (e.g. cross encoder is now in `cross_encoder.md` and doesn't need to be in `hybrid_search.md`?)

AyushExel:
Yup. Initially rerankers were limited to hybrid search but now you can reranker any query so they have their own section. 
https://lancedb.github.io/lancedb/reranking/

PR #1593, opened 2024-09-03
feat: support to create table from record batch iterator

### LINKED ISSUE #1562: bug(python): `create_table()` can't take a generator as data parameter when connect to remote database
### LanceDB version

v0.12.0

### What happened?
This is from [discord report](https://discord.com/channels/1030247538198061086/1105226910528716912/1277230905584259185)

Get error when the try to create table with `data=some_data_generator` in remote database:
```
File ~/miniconda3/envs/hl-reposearch/lib/python3.11/site-packages/lancedb/remote/arrow.py:20, in to_ipc_binary(table)
     18 """Serialize a PyArrow Table to IPC binary."""
...
---> 20 with pa.ipc.new_stream(sink, table.schema) as writer:
     21     writer.write_table(table)
     22 return sink.getvalue().to_pybytes()

AttributeError: 'generator' object has no attribute 'schema'
```
This happens only if connected to a remote database.

```
table = db.create_table(
    lance_db_table_name,
    get_batches_generator(),
    schema=Item.to_arrow_schema(),
)

def get_batches_generator():
    result = []
    batches = [nodes[i : i + batch_size] for i in range(0, len(nodes), batch_size)]
    for i, batch in enumerate(batches):
        yield pd.DataFrame.from_records(batch)

class Item(LanceModel):  
    content: str = func.SourceField() 
    file_path: str
    file_name: str
```

BubbleCal:
resolve #1562
- fix that RemoteTable can't create table from iterator
- infer the schema from the iterator
- refactor the checks for creating table

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -149,20 +149,11 @@ def sanitize_create_table(
             on_bad_vectors=on_bad_vectors,
             fill_value=fill_value,
         )
-    else:
-        if schema is None:
+    if schema is None:
+        if data is None:
             raise ValueError("Either data or schema must be provided")
         elif hasattr(data, "schema"):
             schema = data.schema
-        elif isinstance(data, Iterable):
-            if metadata:
-                raise TypeError(
-                    (
-                        "Persistent embedding functions not yet "
-                        "supported for generator data input"
-                    )
-                )
-        data = pa.Table.from_pylist([], schema=schema)
 
     if metadata:
         schema = schema.with_metadata(metadata)
@@ -1647,7 +1638,7 @@ class LanceTable(Table):
             data, schema, metadata, on_bad_vectors, fill_value
         )
 
-        empty = pa.Table.from_pylist([], schema=schema)
+        empty = pa.Table.from_batches([], schema=schema)
         try:
             lance.write_dataset(empty, tbl._dataset_uri, schema=schema, mode=mode)
         except OSError as err:
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -117,6 +117,8 @@ def _sanitize_data(
         data = _sanitize_schema(
             data, schema=schema, on_bad_vectors=on_bad_vectors, fill_value=fill_value
         )
+        if schema is None:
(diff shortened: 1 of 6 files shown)


python/python/lancedb/table.py
```diff
@@ -117,15 +117,59 @@ def _sanitize_data(
...
+            schema = data.schema
+        elif isinstance(data, Iterable):
+            if metadata:
+                raise TypeError(
+                    (
+                        "Persistent embedding functions not yet "
+                        "supported for generator data input"
+                    )
+                )
+        data = pa.Table.from_pylist([], schema=schema)
```
wjones127, line 165: small nit: it's faster to not go through the Python -> Arrow conversion methods if you just want to create an empty table.

```suggestion
        data = pa.Table.from_batches([], schema=schema)
```

```python
In [5]: %timeit pa.Table.from_pylist([], schema)
6.1 µs ± 519 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [6]: %timeit pa.Table.from_batches([], schema)
287 ns ± 4.87 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

python/python/lancedb/table.py
```diff
@@ -117,15 +117,59 @@ def _sanitize_data(
...
+        elif hasattr(data, "schema"):
+            schema = data.schema
+        elif isinstance(data, Iterable):
+            if metadata:
+                raise TypeError(
+                    (
+                        "Persistent embedding functions not yet "
+                        "supported for generator data input"
+                    )
+                )
```
wjones127, line 164: I'm confused. In this branch, `data is None`, right? So neither of these will ever apply?

wjones127 (commented):
Some initial questions

python/python/lancedb/table.py
BubbleCal: Fixed

python/python/lancedb/table.py
BubbleCal: Fixed

wjones127 (approved):

PR #1602, opened 2024-09-05
ci: enable java auto release

LuQQiu:
Enable bump java pom.xml versions
Enable auto java release when detect stable github release

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -33,7 +33,7 @@ pre_commit_hooks = [
     echo "Constructed new version: $NEW_VERSION"
     cd java && mvn versions:set -DnewVersion=$NEW_VERSION && mvn versions:commit
     cd ..
-    git add java/core/pom.xml java/pom.xml
+    git add java/
     """,
 ]
 
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -33,8 +33,16 @@ pre_commit_hooks = [
     fi
     echo "Constructed new version: $NEW_VERSION"
     cd java && mvn versions:set -DnewVersion=$NEW_VERSION && mvn versions:commit
-    cd ..
-    git add java/
+
+    # Check for any modified but unstaged pom.xml files
+    MODIFIED_POMS=$(git ls-files -m | grep pom.xml)
+    if [ ! -z "$MODIFIED_POMS" ]; then
+        echo "The following pom.xml files were modified but not staged. Adding them now:"
+        echo "$MODIFIED_POMS" | while read -r file; do
+            git add "$file"
+            echo "Added: $file"
+        done
+    fi
     """,
 ]
 
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -24,6 +24,19 @@ tag = true
 tag_message = "Bump version: {current_version} → {new_version}"
 tag_name = "v{new_version}"
 
+pre_commit_hooks = [
+  """
+    NEW_VERSION="${BVHOOK_NEW_MAJOR}.${BVHOOK_NEW_MINOR}.${BVHOOK_NEW_PATCH}"
+    if [ ! -z "$BVHOOK_NEW_PRE_L" ] && [ ! -z "$BVHOOK_NEW_PRE_N" ]; then
+        NEW_VERSION="${NEW_VERSION}-${BVHOOK_NEW_PRE_L}.${BVHOOK_NEW_PRE_N}"
+    fi
+    echo "Constructed new version: $NEW_VERSION"
+    cd java && mvn versions:set -DnewVersion=$NEW_VERSION && mvn versions:commit
+    cd ..
+    git add java/core/pom.xml java/pom.xml
+    """,
+]
+
 [tool.bumpversion.parts.pre_l]
 optional_value = "final"
 values = ["beta", "final"]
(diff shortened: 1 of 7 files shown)


.bumpversion.toml
```diff
@@ -1,57 +1,86 @@
...
+replace = """
+        <groupId>com.lancedb</groupId>
+        <artifactId>lancedb-parent</artifactId>
+        <version>{new_version}</version>
+"""
+search = """
+        <groupId>com.lancedb</groupId>
+        <artifactId>lancedb-parent</artifactId>
+        <version>{current_version}</version>
+"""
```
LuQQiu, line 86: This approach may be broken easily since it does not take care of the complicated java dependencies

https://github.com/LuQQiu/lancedb/blob/f08af7e747ade81af988b594057402336083902c/ci/bump_version.sh#L13

Another approach is to leverage "mvn versions" command, but it does not integrate seamlessly with python bump-my-version.

.bumpversion.toml
```diff
@@ -1,57 +1,70 @@
...
-tag_name = "v{new_version}"
+tag = true
 tag_message = "Bump version: {current_version} → {new_version}"
-allow_dirty = true
-commit = true
-message = "Bump version: {current_version} → {new_version}"
-commit_args = ""
+tag_name = "v{new_version}"
+
+pre_commit_hooks = [
```
LuQQiu, line 27: Only this in the file is the actual change, all others are just auto formatter

LuQQiu:
@wjones127 @westonpace PTAL, thanks!

java/core/pom.xml
```diff
@@ -8,7 +8,7 @@
     <parent>
         <groupId>com.lancedb</groupId>
         <artifactId>lancedb-parent</artifactId>
-        <version>0.0.3</version>
+        <version>0.10.0-beta.1</version>
```
walterddr, line 11: naming convention here is alpha/beta/release-candidate? if not i think -SNAPSHOT should be suffice?

LuQQiu: yep, this is the lancedb naming convention

.bumpversion.toml
```diff
@@ -24,34 +24,48 @@ commit = true
...
+  """
+    NEW_VERSION="${BVHOOK_NEW_MAJOR}.${BVHOOK_NEW_MINOR}.${BVHOOK_NEW_PATCH}"
+    if [ ! -z "$BVHOOK_NEW_PRE_L" ] && [ ! -z "$BVHOOK_NEW_PRE_N" ]; then
+        NEW_VERSION="${NEW_VERSION}-${BVHOOK_NEW_PRE_L}.${BVHOOK_NEW_PRE_N}"
+    fi
+    echo "Constructed new version: $NEW_VERSION"
+    cd java && mvn versions:set -DnewVersion=$NEW_VERSION && mvn versions:commit
+    cd ..
+    git add java/
```
wjones127, line 37: Hmm that seems like it could be a bit dangerous. Someone might have intentionally unstaged changes in that directory. I feel like most commit hooks would just detect if anything changed and then error if it did, right? That would give the developer the chance to review and add the changes themself.

release_process.md
```diff
@@ -1,11 +1,11 @@
 # Release process
 
-There are five total packages we release. Three are the `lancedb` packages
-for Python, Rust, and Node.js. The other two are the legacy `vectordb`
+There are six total packages we release. Four are the `lancedb` packages
+for Python, Rust, Java, and Node.js. The other two are the legacy `vectordb`
 packages for Rust and node.js.
```
wjones127, line 5: While we are updating this, we can remove the `vectordb` Rust package, which hasn't been part of our release process for a little while.
```suggestion
There are five total packages we release. Four are the `lancedb` packages
for Python, Rust, Java, and Node.js. The other one is the legacy `vectordb`
package node.js.
```

release_process.md
```diff
@@ -1,11 +1,11 @@
...
 
-The Python package is versioned and released separately from the Rust and Node.js
-ones. For Rust and Node.js, the release process is shared between `lancedb` and
+The Python package is versioned and released separately from the Rust, Java, and Node.js
+ones. For Rust, and Node.js, the release process is shared between `lancedb` and
```
wjones127, line 8:
```suggestion
ones. For Node.js the release process is shared between `lancedb` and
```

release_process.md
```diff
@@ -24,20 +24,23 @@ indexes.
...
-                                └─────────────┘                          
+  ┌─────────────────────┐
+  │Create Release Commit│
+  └─┬───────────────────┘
+    │                           ┌────────────┐ ┌──►Python GH Release
+    ├──►(tag) python-vX.Y.Z ───►│PyPI Publish├─┤
+    │                           └────────────┘ └──►Python Wheels
+    │
+    │                           ┌───────────┐
+    └──►(tag) vX.Y.Z ───┬──────►│NPM Publish├──┬──►Rust/Node GH Release
```
wjones127, line 35: Just a note for us in the future: we should probably add `java` to these release notes. And also consider taking seriously the language labels in the release notes. Think about what we need to do to make them easy to read for both Node and Java developers.

wjones127 (changes requested):
I'm not sure about the commit hook `git add`. Could we error if there is a version update not checked in?

.bumpversion.toml
LuQQiu: Good catch! Agree that it could be error-prone. modified the script to detect changes to pom.xml files and add those files. And ensure we can get information of what happened

LuQQiu:
@wjones127 PTAL, thanks!

wjones127 (approved):
Thanks. Excellent work!

PR #1603, opened 2024-09-05
fix(python): remove unmaintained ratelimiter dependency

antoniomdk:
The `ratelimiter` package hasn't been updated in ages and is no longer maintained. This PR removes the dependency on `ratelimiter` and replaces it with a custom rate limiter implementation.


wjones127 (approved):
Seems fine. Nice simplification.

PR #1612, opened 2024-09-06
feat: add flag to enable faster manifest paths

### LINKED ISSUE #1583: Feature: support enabling V2 manifest paths
### SDK

Python

### Description

Exposing this feature from Lance: https://github.com/lancedb/lance/issues/2790

wjones127:
The new V2 manifest path scheme makes discovering the latest version of a table constant time on object stores, regardless of the number of versions in the table. See benchmarks in the PR here: https://github.com/lancedb/lance/pull/2798

Closes #1583

### DIFF
--- a/python/src/table.rs
+++ b/python/src/table.rs
@@ -304,6 +304,18 @@ impl Table {
         })
     }
 
+    pub fn uses_v2_manifest_paths(self_: PyRef<'_, Self>) -> PyResult<Bound<'_, PyAny>> {
+        let inner = self_.inner_ref()?.clone();
+        future_into_py(self_.py(), async move {
+            inner
+                .as_native()
+                .ok_or_else(|| PyValueError::new_err("This cannot be run on a remote table"))?
+                .uses_v2_manifest_paths()
+                .await
+                .infer_error()
+        })
+    }
+
     pub fn migrate_manifest_paths_v2(self_: PyRef<'_, Self>) -> PyResult<Bound<'_, PyAny>> {
         let inner = self_.inner_ref()?.clone();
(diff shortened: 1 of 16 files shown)


python/src/table.rs
```diff
@@ -303,4 +303,17 @@ impl Table {
...
+    pub fn migrate_manifest_paths_v2(self_: PyRef<'_, Self>) -> PyResult<Bound<'_, PyAny>> {
+        let inner = self_.inner_ref()?.clone();
+        future_into_py(self_.py(), async move {
+            inner
+                .as_native()
+                .ok_or_else(|| PyValueError::new_err("This cannot be run on a remote table"))?
+                .migrate_manifest_paths_v2()
+                .await
+                .infer_error()
+                .map(|_| ())
```
westonpace, line 316: What's getting mapped away here?  Doesn't the function already return `Result<()>`?

westonpace (approved):
It might be nice if there were some way to inspect a table to see if it is using the v2 manifest naming scheme or not.  I know the operation is idempotent but it still requires setting up a quiet period to use so it might be convenient for a user to check if it needs to be performed before they go and do it.

Also, I wonder if we should add a link to https://lancedb.github.io/lance/format.html#manifest-naming-schemes somewhere.

wjones127:
> It might be nice if there were some way to inspect a table to see if it is using the v2 manifest naming scheme or not. I know the operation is idempotent but it still requires setting up a quiet period to use so it might be convenient for a user to check if it needs to be performed before they go and do it.

Thanks, this is a good point. I've exposed a method to let users check this.

PR #1623, opened 2024-09-09
feat: `fast_search` in Python and Node

wjones127:
Sometimes it is acceptable to users to only search indexed data and skip and new un-indexed data. For example, if un-indexed data will be shortly indexed and they don't mind the delay. In these cases, we can save a lot of CPU time in search, and provide better latency. Users can activate this on queries using `fast_search()`.

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -402,6 +402,29 @@ describe("When creating an index", () => {
     expect(rst.numRows).toBe(1);
   });
 
+  it("should be able to query unindexed data", async () => {
+    await tbl.createIndex("vec");
+    await tbl.add([
+      {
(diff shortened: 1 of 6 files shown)


BubbleCal (approved):
Is this for only vector index?
I'm thinking what about the FTS/scalar index because flat FTS can be slow as well

wjones127:
> Is this for only vector index? I'm thinking what about the FTS/scalar index because flat FTS can be slow as well

I made it so this option can be passed down for FTS and other searches. However, I'm not sure that we have implemented any changes to the planner to take advantage of this flag for FTS yet.

bfcrampton:
+1 on optimizations for FTS/scalar indices! 😄

PR #1639, opened 2024-09-12
feat: implement Remote connection for LanceDB Rust

wjones127:
* Adding a simple test facility, which allows you to mock a single endpoint at a time with a closure.
* Implementing all the database-level endpoints

Table-level APIs will be done in a follow up PR.

### DIFF
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -174,7 +174,7 @@ pub mod test_utils {
         async fn send(&self, request: reqwest::RequestBuilder) -> Result<reqwest::Response> {
             let request = request.build().unwrap();
             let response = (self.f)(request);
-            Ok(reqwest::Response::from(response))
+            Ok(response)
         }
     }
 
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -35,6 +35,7 @@ pub trait HttpSend: Clone + Send + Sync + std::fmt::Debug + 'static {
     fn send(&self, req: RequestBuilder) -> impl Future<Output = Result<Response>> + Send;
 }
 
+// Default implementation of HttpSend which sends the request normally with reqwest
 #[derive(Clone, Debug)]
 pub struct Sender;
 impl HttpSend for Sender {
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
+use std::{future::Future, time::Duration};
 
 use reqwest::{
     header::{HeaderMap, HeaderValue},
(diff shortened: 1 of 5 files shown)


rust/lancedb/src/remote/client.rs
```diff
@@ -21,13 +21,65 @@ use reqwest::{
...
+    sender: S,
 }
 
-impl RestfulLanceDbClient {
+pub trait HttpSend: Clone + Send + Sync + std::fmt::Debug + 'static {
+    fn send(&self, req: RequestBuilder) -> impl Future<Output = Result<Response>> + Send;
+}
+
+#[derive(Clone, Debug)]
+pub struct Sender;
```
westonpace, line 39:
```suggestion

// Default implementation of HttpSend which sends the request normally with reqwest
#[derive(Clone, Debug)]
pub struct Sender;
```

rust/lancedb/src/remote/client.rs
```diff
@@ -21,13 +21,65 @@ use reqwest::{
 
 use crate::error::{Error, Result};
 
+// We use the `HttpSend` trait to abstract over the `reqwest::Client` so that
+// we can mock responses in tests. Based on the patterns from this blog post:
+// https://write.as/balrogboogie/testing-reqwest-based-clients
 #[derive(Clone, Debug)]
-pub struct RestfulLanceDbClient {
+pub struct RestfulLanceDbClient<S: HttpSend = Sender> {
```
westonpace, line 28: Overly opinionated personal comment (not even rising to the standards of minor nit): I feel like parameterizing this (instead of using `Box<dyn HttpSend>`) is probably overkill.

rust/lancedb/src/remote/db.rs
```diff
@@ -88,15 +107,24 @@ impl ConnectionInternal for RemoteDatabase {
...
+        let req = self
             .client
             .post(&format!("/v1/table/{}/create/", options.name))
             .body(data_buffer)
             .header(CONTENT_TYPE, ARROW_STREAM_CONTENT_TYPE)
             // This is currently expected by LanceDb cloud but will be removed soon.
-            .header("x-request-id", "na")
-            .send()
-            .await?;
+            .header("x-request-id", "na");
```
westonpace, line 116: Not related but I wonder if we should check if this is still needed?

rust/lancedb/src/remote/db.rs
```diff
@@ -88,15 +107,24 @@ impl ConnectionInternal for RemoteDatabase {
...
+        let rsp = self.client.send(req).await?;
+
+        if rsp.status() == StatusCode::BAD_REQUEST {
+            let body = rsp.text().await?;
+            if body.contains("already exists") {
+                return Err(crate::Error::TableAlreadyExists { name: options.name });
```
westonpace, line 122: Pedantic nitpick unrelated to the PR: we could use HTTP 409 (the request is syntactically valid but can't be processed in the given server state) for "already exists" type errors instead of 400 (the request would never be valid)

westonpace (approved):
This is very cool.  Please excuse my unrelated and grumpy comments :)

rust/lancedb/src/remote/client.rs
```diff
@@ -21,13 +21,65 @@ use reqwest::{
...
+        let parsed_url = url::Url::parse(db_url)?;
+        debug_assert_eq!(parsed_url.scheme(), "db");
+        if !parsed_url.has_host() {
+            return Err(Error::Http {
+                message: format!("Invalid database URL (missing host) '{}'", db_url),
+            });
+        }
+        let db_name = parsed_url.host_str().unwrap();
+        let client = reqwest::Client::builder()
+            .timeout(Duration::from_secs(30))
```
rpgreen, line 62: should we parameterize this?

wjones127: This is a good point. I imagine you've fixed a few issues in the existing remote SDKs. I've created a follow up issue for configuring the client. If you have other things you think we should do, feel free to add to that issue: https://github.com/lancedb/lancedb/issues/1642

rust/lancedb/src/remote/db.rs
wjones127: Yes I agree. I have quite a few nits for the API (why is everything a POST request?). It would be a lot easier to detect this error if we could detect from a status code.

rust/lancedb/src/remote/db.rs
wjones127: I will check. Can address in a follow up if needed.

PR #1648, opened 2024-09-14
chore: make remote client a context manager

eddyxu:
Allow `RemoteLanceDBClient` to be used as context manager


westonpace (approved):
Looks good to me.

Do we expect users to use the client class directly?  The remote table should already be a context manager.

Or is this just to make life easier for us as developers?

eddyxu:
yea this is mostly make my life easier :)

PR #1663, opened 2024-09-18
feat(rust): remote client query endpoint

wjones127:


rust/lancedb/src/remote/table.rs
```diff
@@ -196,38 +291,79 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
...
-        Err(Error::NotSupported {
-            message: "explain_plan is not supported on LanceDB cloud.".into(),
-        })
+        let request = self.client.post(&format!("/table/{}/query/", self.name));
+
+        let mut body = serde_json::Value::Object(Default::default());
+        Self::apply_query_params(&mut body, &query.base)?;
+
+        body["prefilter"] = query.prefilter.into();
+        body["k"] = query.base.limit.unwrap_or(10).into();
```
westonpace, line 306: Doesn't `apply_query_params` also apply `k`?

rust/lancedb/src/remote/table.rs
```diff
@@ -266,11 +402,79 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
...
+            _ => {
+                return Err(Error::NotSupported {
+                    message: "Index type not supported".into(),
+                })
+            }
+        };
+        body["index_type"] = serde_json::Value::String(index_type.into());
+        if let Some(distance_type) = distance_type {
+            body["distance_type"] = serde_json::Value::String(distance_type.to_string());
+        }
```
westonpace, line 467: Didn't you calculate `distance_type` above in the `match index.index` block?

rust/lancedb/src/remote/table.rs
```diff
@@ -728,4 +953,205 @@ mod tests {
...
+            .full_text_search(
+                FullTextSearchQuery::new("hello world".into())
+                    .columns(Some(vec!["a".into(), "b".into()])),
+            )
+            .limit(10)
+            .execute()
+            .await
+            .unwrap();
+
+        // WAND factor and limit are not supported.
```
westonpace, line 1107: True but why mention that here?

rust/lancedb/src/remote/table.rs
```diff
@@ -728,4 +953,205 @@ mod tests {
...
+                });
+                if let Some(distance_type) = distance_type {
+                    expected_body["distance_type"] = distance_type.into();
+                }
+                assert_eq!(body, expected_body);
+
+                http::Response::builder().status(200).body("{}").unwrap()
+            });
+
+            // How should Auto be handled?
```
westonpace, line 1153: Looks like you are handling auto above right?

rust/lancedb/src/table.rs
```diff
@@ -391,7 +390,12 @@ pub(crate) trait TableInternal: std::fmt::Display + std::fmt::Debug + Send + Syn
         query: &Query,
         options: QueryExecutionOptions,
     ) -> Result<DatasetRecordBatchStream>;
-    async fn explain_plan(&self, query: &VectorQuery, verbose: bool) -> Result<String>;
+    async fn explain_plan(&self, query: &VectorQuery, verbose: bool) -> Result<String> {
+        let plan = self.create_plan(query, Default::default()).await?;
+        let display = DisplayableExecutionPlan::new(plan.as_ref());
+
+        Ok(format!("{}", display.indent(verbose)))
+    }
```
westonpace, line 398: I suppose this works for remote table but it will always just show `OneShotExec` right?

westonpace (approved):
Thanks for taking on the burden of finishing these up

rust/lancedb/src/table.rs
wjones127: It does. I was thinking maybe one day we will have a `RemoteQueryExec` that can handle federation or something. Like maybe do a client side join could be chained on after.

rust/lancedb/src/remote/table.rs
wjones127: True

rust/lancedb/src/remote/table.rs
wjones127: I did, but here I am adding it to the body.

rust/lancedb/src/remote/table.rs
wjones127: Left over note from testing.

PR #1667, opened 2024-09-19
fix: hnsw default partitions

### LINKED ISSUE #1662: Change HNSW default num_partitions
For small tables, we generally recommend 1 partition.

We could probably do something like: `num_rows*dim/(256*5million)` 

Also check that the IVF_PQ default in Python sync API is correct.

gagan-bhullar-tech:
PR fixes #1662

### DIFF
--- a/rust/lancedb/src/index/vector.rs
+++ b/rust/lancedb/src/index/vector.rs
@@ -17,7 +17,7 @@
 //!
 //! Vector indices are only supported on fixed-size-list (tensor) columns of floating point
 //! values
-use std::{cmp::max, ops::Div};
+use std::cmp::max;
 
 use lance::table::format::{Index, Manifest};
 
--- a/rust/lancedb/src/index/vector.rs
+++ b/rust/lancedb/src/index/vector.rs
@@ -215,7 +215,7 @@ pub(crate) fn suggested_num_partitions(rows: usize) -> u32 {
 }
 
 pub(crate) fn suggested_num_partitions_for_hnsw(rows: usize, dim: u32) -> u32 {
-    let num_partitions = ((rows as f64) * (dim as f64)).div((256 * 5000000) as f64) as u32;
+    let num_partitions = (((rows as u64) * (dim as u64)) / (256 * 5_000_000)) as u32;
     max(1, num_partitions)
 }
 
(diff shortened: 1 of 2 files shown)


gagan-bhullar-tech:
@wjones127 please review

rust/lancedb/src/index/vector.rs
```diff
@@ -214,6 +214,11 @@ pub(crate) fn suggested_num_partitions(rows: usize) -> u32 {
     max(1, num_partitions)
 }
 
+pub(crate) fn suggested_num_partitions_for_hnsw(rows: usize, dim: u32) -> u32 {
+    let num_partitions = ((rows as f64) * (dim as f64)).div((256 * 5000000) as f64) as u32;
```
wjones127, line 218: I would keep as an integer. u64 should be enough.
```suggestion
    let num_partitions = (((rows as u64) * (dim as u64)) / (256 * 5_000_000)) as u32;
```

rust/lancedb/src/table.rs
```diff
@@ -1457,6 +1453,12 @@ impl NativeTable {
             }?
         };
 
+        let num_partitions = if let Some(n) = index.num_partitions {
+            n
+        } else {
+            suggested_num_partitions_for_hnsw(self.count_rows(None).await?, num_sub_vectors)
+        };
```
wjones127, line 1460: Could you do the same in `create_ivf_hnsw_sq_index`?

rust/lancedb/src/table.rs
```diff
@@ -1457,6 +1453,12 @@ impl NativeTable {
             }?
         };
 
+        let num_partitions = if let Some(n) = index.num_partitions {
+            n
+        } else {
+            suggested_num_partitions_for_hnsw(self.count_rows(None).await?, num_sub_vectors)
```
wjones127, line 1459: I think `dims` was meant to be original dimensions, not the compressed subvectors. Could you get that? It's the `n` inside of the `arrow_schema::DataType::FixedSizeList`

wjones127 (changes requested):
Thanks for working on this. Have a few suggestions.

gagan-bhullar-tech:
@wjones127 updated PR. Can you please review.

wjones127 (approved):

PR #1687, opened 2024-09-23
feat!: enforce all rerankers always return relevance score & deprecate linear combination fixes

AyushExel:
- Enforce all rerankers always return _relevance_score. This was already loosely done in tests before but based on user feedback its better to always have _relevance_score present in all reranked results
- Deprecate LinearCombinationReranker in docs. And also fix a case where it would not return _relevance_score if one result set was missing

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1121,12 +1121,12 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
 
         Parameters
         ----------
+        reranker: Reranker, default RRFReranker()
+            The reranker to use. Must be an instance of Reranker class.
         normalize: str, default "score"
             The method to normalize the scores. Can be "rank" or "score". If "rank",
             the scores are converted to ranks and then normalized. If "score", the
             scores are normalized directly.
-        reranker: Reranker, default RRFReranker()
-            The reranker to use. Must be an instance of Reranker class.
         Returns
         -------
         LanceHybridQueryBuilder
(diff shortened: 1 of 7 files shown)


python/python/lancedb/query.py
```diff
@@ -1112,21 +1112,21 @@ def _normalize_scores(self, results: pa.Table, column: str, invert=False):
 
     def rerank(
         self,
-        normalize="score",
         reranker: Reranker = RRFReranker(),
+        normalize: str = "score",
```
BubbleCal, line 1116: is this a breaking change? the order changed

python/python/lancedb/rerankers/linear_combination.py
```diff
@@ -58,14 +59,42 @@ def rerank_hybrid(
...
-        if len(vector_results) == 0 and len(fts_results) == 0:
-            return vector_results
-        # If one is empty then return the other
+        # If one is empty then return the other and add _relevance_score
+        # column equal the existing vector or fts score
         if len(vector_results) == 0:
-            return fts_results
+            results = fts_results.append_column(
+                "_relevance_score",
+                pa.array(fts_results["_score"].to_pylist(), type=pa.float32()),
```
BubbleCal, line 67: is `fts_results["_score"]` already pyarrow array?
maybe we don't need to reconstruct it again

python/python/lancedb/query.py
AyushExel: yes marked

BubbleCal (approved):

Issue #1681, opened 2024-09-21
bug(python): ollama embedding getting client not serializable error


rjrobben:
### LanceDB version

v0.14.0b0

### What happened?

```
from lancedb.embeddings import get_registry
from lancedb.pydantic import LanceModel, Vector
from pydantic import BaseModel
from typing import List, Optional

class Cases(LanceModel):
    content_plain_text: str = model.SourceField()
    vector: Vector(model.ndims()) = model.VectorField()

model = get_registry().get("ollama").create(name='Losspost/stella_en_1.5b_v5')

table = await async_db.create_table("cases", exist_ok=True, schema=Cases)

```
Getting the error below:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 table = await async_db.create_table("cases", exist_ok=True, schema=Cases, mode="overwrite")

File ~/lancedb/python/python/lancedb/db.py:748, in AsyncConnection.create_table(self, name, data, schema, mode, exist_ok, on_bad_vectors, fill_value, storage_options, data_storage_version, use_legacy_format, enable_v2_manifest_paths)
    745 if fill_value is None:
    746     fill_value = 0.0
--> 748 data, schema = sanitize_create_table(
    749     data, schema, metadata, on_bad_vectors, fill_value
    750 )
    751 validate_schema(schema)
    753 if exist_ok is None:

File ~/lancedb/python/python/lancedb/table.py:168, in sanitize_create_table(data, schema, metadata, on_bad_vectors, fill_value)
    161 def sanitize_create_table(
    162     data, schema, metadata=None, on_bad_vectors="error", fill_value=0.0
    163 ):
    164     if inspect.isclass(schema) and issubclass(schema, LanceModel):
    165         # convert LanceModel to pyarrow schema
    166         # note that it's possible this contains
    167         # embedding function metadata already
--> 168         schema = schema.to_arrow_schema()
    170     if data is not None:
    171         data, schema = _sanitize_data(
    172             data,
    173             schema,
   (...)
    176             fill_value=fill_value,
    177         )

File ~/lancedb/python/python/lancedb/pydantic.py:327, in LanceModel.to_arrow_schema(cls)
    323 if len(functions) > 0:
    324     # Prevent circular import
    325     from .embeddings import EmbeddingFunctionRegistry
--> 327     metadata = EmbeddingFunctionRegistry.get_instance().get_table_metadata(
    328         functions
    329     )
[...]

PR #1688, opened 2024-09-24
fix(EmbeddingFunction): modify safe_model_dump to explicitly exclude class fields with underscore

rjrobben:
Resolve issue #1681


wjones127 (changes requested):
Does this fix the reported issue? Could you add a unit test that verifies this?

rjrobben:
Can confirm in my environment that the reported issue #1681 only happens under pydantic version < 2. Now shall be resolved by filtering out the underscored property explicitly and I added the tests.

Thanks for the feedback!

PR #1693, opened 2024-09-24
feat: add fast_search to vectordb

albertlockett:


rust/ffi/node/src/query.rs
```diff
@@ -46,6 +46,10 @@ impl JsQuery {
             .get::<JsBoolean, _, _>(&mut cx, "_prefilter")?
             .value(&mut cx);
 
+        let fast_search = query_obj
```
westonpace, line 49: This is more a gripe against node-ffi than anything but you have two boolean properties here (`prefilter` and `fast_search`) and are fetching them in two different ways (`.get..value(&mut cx)` and `.get_opt..map(..)`).  Might be nice if they were consistent.

westonpace (approved):
Just the most minor of nits

Issue #1638, opened 2024-09-12
feat(python): implement AsyncRemoteTable


eddyxu:
### LanceDB version

0.13.0

### What happened?

When use async Python SDK with lancedb cloud, it reports error

```python
/usr/local/lib/python3.10/dist-packages/lancedb/__init__.py in connect_async(uri, api_key, region, host_override, read_consistency_interval, request_thread_pool, storage_options)
    174 
    175     return AsyncConnection(
--> 176         await lancedb_connect(
    177             sanitize_uri(uri),
    178             api_key,

RuntimeError: Runtime error: cannot connect to LanceDb Cloud unless the 'remote' feature is enabled
```

### Are there known steps to reproduce?

```python
import lancedb

db = await lancedb.connect_async(
    "db://...",
    api_key="sk_...",
)
list(await db.table_names())
```

PR #1700, opened 2024-09-25
feat(python): bind python async remote client to rust client

wjones127:
Closes [#1638](https://github.com/lancedb/lancedb/issues/1638)

This just binds the Python Async client to the Rust remote client.

### DIFF
--- a/python/python/lancedb/remote/__init__.py
+++ b/python/python/lancedb/remote/__init__.py
@@ -144,6 +144,12 @@ class RetryConfig:
 
         You can also set this via the environment variable
         `LANCE_CLIENT_RETRY_BACKOFF_JITTER`.
+    statuses: Optional[List[int]
+        The HTTP status codes for which to retry the request. Default is
+        [429, 500, 502, 503].
+
+        You can also set this via the environment variable
+        `LANCE_CLIENT_RETRY_STATUSES`. Use a comma-separated list of integers.
     """
 
     retries: Optional[int] = None
--- a/python/python/lancedb/remote/__init__.py
+++ b/python/python/lancedb/remote/__init__.py
@@ -12,9 +12,12 @@
 #  limitations under the License.
 
 import abc
+from dataclasses import dataclass
+from datetime import timedelta
 from typing import List, Optional
 
 import attrs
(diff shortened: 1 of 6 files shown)


python/python/lancedb/remote/__init__.py
```diff
@@ -62,3 +65,103 @@ class LanceDBClient(abc.ABC):
...
+    backoff_factor: Optional[float]
+        The backoff factor to apply between retries. Default is 0.25. Between each retry
+        the client will wait for the amount of seconds:
+        `{backoff factor} * (2 ** ({number of previous retries}))`. So for the default
+        of 0.25, the first retry will wait 0.25 seconds, the second retry will wait 0.5
+        seconds, the third retry will wait 1 second, etc.
+
+        You can also set this via the environment variable
+        `LANCE_CLIENT_RETRY_BACKOFF_FACTOR`.
+    backoff_jitter: Optional[float]
```
westonpace, line 138: What happens if the random value is larger than the `backoff_factor`.  Or, to phrase it another way, do I need to worry about keeping this lower than `backoff_factor`?

python/python/lancedb/remote/__init__.py
```diff
@@ -62,3 +65,103 @@ class LanceDBClient(abc.ABC):
...
+        You can also set this via the environment variable
+        `LANCE_CLIENT_RETRY_BACKOFF_JITTER`.
+    """
+
+    retries: Optional[int] = None
+    connect_retries: Optional[int] = None
+    read_retries: Optional[int] = None
+    backoff_factor: Optional[float] = None
+    backoff_jitter: Optional[float] = None
+    statuses: Optional[List[int]] = None
```
westonpace, line 154: `statuses` is not documented

westonpace (approved):
This seems well thought out, thanks

python/python/lancedb/remote/__init__.py
wjones127: The random value comes from `(0, backoff_jitter]` seconds, so nothing bad will happen. I suppose it shouldn't be negative and I don't yet enforce that but could.

PR #1702, opened 2024-09-26
feat: remote index stats

wjones127:
BREAKING CHANGE: the return value of `index_stats` method has changed and all `index_stats` APIs now take index name instead of UUID. Also several deprecated index statistics methods were removed.

* Removes deprecated methods for individual index statistics
* Aligns public `IndexStatistics` struct with API response from LanceDB Cloud.
* Implements `index_stats` for remote Rust SDK and Python async API.

### DIFF
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -888,7 +888,7 @@ describe("LanceDB client", function () {
       expect(indices[0].columns).to.have.lengthOf(1);
       expect(indices[0].columns[0]).to.equal("vector");
 
-      const stats = await table.indexStats(indices[0].uuid);
+      const stats = await table.indexStats(indices[0].name);
       expect(stats.numIndexedRows).to.equal(300);
       expect(stats.numUnindexedRows).to.equal(0);
     }).timeout(50_000);
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -891,6 +891,8 @@ describe("LanceDB client", function () {
       const stats = await table.indexStats(indices[0].name);
       expect(stats.numIndexedRows).to.equal(300);
       expect(stats.numUnindexedRows).to.equal(0);
+      expect(stats.indexType).to.equal("IVF_PQ");
+      expect(stats.distanceType).to.equal("l2");
     }).timeout(50_000);
   });
 
--- a/node/src/test/test.ts
+++ b/node/src/test/test.ts
@@ -893,6 +893,7 @@ describe("LanceDB client", function () {
       expect(stats.numUnindexedRows).to.equal(0);
       expect(stats.indexType).to.equal("IVF_PQ");
       expect(stats.distanceType).to.equal("l2");
+      expect(stats.numIndices).to.equal(1);
     }).timeout(50_000);
   });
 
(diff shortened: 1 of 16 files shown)


BubbleCal:
@wjones127 btw do you want to add `num_indices` in the resp?
lance's `index_stats` return value contains it, it could be useful for some compact/reindex policy

nodejs/src/table.rs
```diff
@@ -480,32 +480,17 @@ pub struct IndexStatistics {
     /// The number of rows not indexed
     pub num_unindexed_rows: f64,
     /// The type of the index
-    pub index_type: Option<String>,
-    /// The metadata for each index
-    pub indices: Vec<IndexMetadata>,
+    pub index_type: String,
+    /// The type of the distance function used by the index
```
westonpace, line 484:
```suggestion
    /// The distance type used by the index
```

When is this None?

rust/ffi/node/src/table.rs
```diff
@@ -470,49 +470,31 @@ impl JsTable {
...
-                        let null = cx.null();
-                        output.set(&mut cx, "numUnindexedRows", null)?;
-                    }
-                };
+                if let Some(stats) = stats {
+                    let output = JsObject::new(&mut cx);
+                    let num_indexed_rows = cx.number(stats.num_indexed_rows as f64);
+                    output.set(&mut cx, "numIndexedRows", num_indexed_rows)?;
+                    let num_unindexed_rows = cx.number(stats.num_unindexed_rows as f64);
+                    output.set(&mut cx, "numUnindexedRows", num_unindexed_rows)?;
```
westonpace, line 492: distance type?

rust/lancedb/src/index.rs
```diff
@@ -129,16 +152,32 @@ pub struct IndexConfig {
...
     pub indices: Vec<IndexMetadata>,
+    // Sometimes, the index type is provided at this level.
+    pub index_type: Option<IndexType>,
+}
+
+#[skip_serializing_none]
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct IndexStatistics {
+    pub num_indexed_rows: usize,
+    pub num_unindexed_rows: usize,
```
westonpace, line 177: Document these fields since public facing?

node/src/test/test.ts
```diff
@@ -888,7 +888,7 @@ describe("LanceDB client", function () {
       expect(indices[0].columns).to.have.lengthOf(1);
       expect(indices[0].columns[0]).to.equal("vector");
 
-      const stats = await table.indexStats(indices[0].uuid);
+      const stats = await table.indexStats(indices[0].name);
```
westonpace, line 891: Maybe also check the new fields here?

westonpace (approved):

rust/ffi/node/src/table.rs
wjones127: Good catch. There was a type advertising more fields that we actually return.

wjones127:
> btw do you want to add `num_indices` in the resp?

Good idea @BubbleCal. I've added that field as well. It will be missing from remote tables for now until we plumb that through in Phalanx.

wjones127:
> A general question, how would a user know the index name? Do we have some message to indicate that it is`col name_idx`?

Right now they have to call `list_indices()` and find the index they created. You can see us doing that in some of the unit tests.

In Lance, users can pass down the name of the index they want. We could eventually do the same in LanceDB.

PR #1704, opened 2024-09-27
feat: expose underlying dataset uri of the table

chebbyChefNEQ:
We frequently need to recover this uri so exposing it to make it easy

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -325,6 +325,9 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
             message: "table_definition is not supported on LanceDB cloud.".into(),
         })
     }
+    fn dataset_uri(&self) -> &str {
+        "NOT_SUPPORTED"
+    }
 }
 
 #[derive(Serialize)]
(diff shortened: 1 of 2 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -325,6 +325,9 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
             message: "table_definition is not supported on LanceDB cloud.".into(),
         })
     }
+    fn dataset_uri(&self) -> &str {
+        "NOT_SUPPORTED"
```
BubbleCal, line 329: throwing an exception could be better?

BubbleCal (approved):

rust/lancedb/src/remote/table.rs
westonpace: Why can't we support it?  We can just latch the `uri` when the table is constructed.

westonpace (approved):
This is fine but I think we can support it for remote tables too.

Also, we will eventually want to expose this to python & node I assume

chebbyChefNEQ:
Yeah we def want to start supporting all of these. Just unblocking for the time being will create a ticket

PR #1708, opened 2024-09-27
ci: change to use github runner

eddyxu:
Use github runner

### DIFF
--- a/.github/workflows/docs_test.yml
+++ b/.github/workflows/docs_test.yml
@@ -24,7 +24,7 @@ env:
 jobs:
   test-python:
     name: Test doc python code
-    runs-on: "warp-ubuntu-latest-x64-4x"
+    runs-on: ubuntu-24.04
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/docs_test.yml
```diff
@@ -24,7 +24,7 @@ env:
 jobs:
   test-python:
     name: Test doc python code
-    runs-on: "warp-ubuntu-latest-x64-4x"
+    runs-on: ubuntu-24.04
```
chebbyChefNEQ, line 27: maybe use larger worker?

chebbyChefNEQ (approved):

.github/workflows/docs_test.yml
eddyxu: Tried 8x, still the same OOM.  Is 16x too overkill? :)

eddyxu: Ubuntu-24.04 is 4 core for OSS projects. Do we need 8x/16x machine?

Issue #1711, opened 2024-09-29
bug(python): create LabelList.    but list_indices it calls IVFPQ


oswardlx:
### LanceDB version

0.13.0

### What happened?

i create index like:
    await tbl.create_index("tags", config=LabelList())

but when i run :
    config = await tbl.list_indices();
 
 config print :
[Index(**IvfPq,** columns=["tags"])]

Is this a display issue or is the underlying data ivfQp

wjones127:
Thanks for reporting @oswardlx .

This appears to be related to https://github.com/lancedb/lance/issues/2039

I'll look into fixing this.

PR #1715, opened 2024-09-30
fix: `list_indices` returns correct index type

wjones127:
Fixes https://github.com/lancedb/lancedb/issues/1711

Doesn't address this https://github.com/lancedb/lance/issues/2039

Instead we load the index statistics, which seems to contain the index type. However, this involves more IO than previously. I'm not sure whether we care that much. If we do, we can fix that upstream Lance issue.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -2804,7 +2804,7 @@ mod tests {
         let index_configs = table.list_indices().await.unwrap();
         assert_eq!(index_configs.len(), 1);
         let index = index_configs.into_iter().next().unwrap();
-        assert_eq!(index.index_type, crate::index::IndexType::IvfPq);
+        assert_eq!(index.index_type, crate::index::IndexType::IvfHnswSq);
         assert_eq!(index.columns, vec!["embeddings".to_string()]);
         assert_eq!(table.count_rows(None).await.unwrap(), 512);
         assert_eq!(table.name(), "test");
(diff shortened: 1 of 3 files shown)


rust/lancedb/src/table.rs
```diff
@@ -2023,28 +2024,28 @@ impl TableInternal for NativeTable {
     async fn list_indices(&self) -> Result<Vec<IndexConfig>> {
         let dataset = self.dataset.get().await?;
         let indices = dataset.load_indices().await?;
-        indices.iter().map(|idx| {
-            let mut is_vector = false;
+        futures::stream::iter(indices.as_slice()).then(|idx| async {
```
LuQQiu, line 2027: Is it needed? is it common to have a lot of indexes?

LuQQiu (approved):
Left a minor comment, LGTM

rust/lancedb/src/table.rs
wjones127: > Is it needed to use "futures::stream"?

What's the alternative you have in mind? A `while let ...` loop?

> is it common to have a lot of indexes?

For hybrid search you need both an FTS index and a ANN index. And it's common to have a scalar index or two on top of that.

LuQQiu: I am more thinking of load_indexes would directly get all the index statistics
but may be index_statistics is a call heavier than i expected

Issue #1624, opened 2024-09-10
bug(node): openai still seems to be required as dependency


wjones127:
### LanceDB version

v0.10.0-beta.1

### What happened?

Trying to compile LanceDB gives this error:

```
node_modules/@lancedb/lancedb/dist/embedding/openai.d.ts:1:44 - error TS2307: Cannot find module 'openai/resources/index' or its corresponding type declarations.

1 import type { EmbeddingCreateParams } from "openai/resources/index";
```

### Are there known steps to reproduce?

Just try to build an application without installing `openai`.

If we add this to dev dependencies:

```
    "openai": "^4.29.2"
```

We eliminate that error. Maybe it's required to match the right version? We should make this import more robust somehow.

This was originally an issue in https://github.com/lancedb/lancedb/issues/1496

wjones127:
I think I can use this approach to solve this: https://github.com/Dawar/typescript-library-tutorial

PR #1716, opened 2024-09-30
fix: try to fix openapi dependency issue

wjones127:
Fixes https://github.com/lancedb/lancedb/issues/1624

### DIFF
--- a/nodejs/lancedb/embedding/openai.ts
+++ b/nodejs/lancedb/embedding/openai.ts
@@ -13,10 +13,13 @@
 // limitations under the License.
 
 import type OpenAI from "openai";
-import type { EmbeddingCreateParams } from "openai/resources/index";
 import { Float, Float32 } from "../arrow";
 import { EmbeddingFunction } from "./embedding_function";
 import { register } from "./registry";
+
+type EmbeddingCreateParams =
+  | import("openai/resources/embeddings").EmbeddingCreateParams
+  | Partial<{ model: string }>;
 
 export type OpenAIOptions = {
   apiKey: string;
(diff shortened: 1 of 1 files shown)


wjones127:
This is really hard to test that it works for sure. It works in local tests and in a separate test project if I reference it with a `file://` path. But so does `main`. The only way I can get it to fail is if I reference the built artifacts on npm. I don't know any easy way to create those same artifacts from a branch right now. But I think in theory these changes should be what we want (make the type optional in public API).

Here are the changes to the compiled typescript declarations:

```diff
-import type { EmbeddingCreateParams } from "openai/resources/index";
 import { Float } from "../arrow";
 import { EmbeddingFunction } from "./embedding_function";
+type EmbeddingCreateParams = import("openai/resources/embeddings").EmbeddingCreateParams | +Partial<{
+    model: string;
+}>;
 export type OpenAIOptions = {
     apiKey: string;
     model: EmbeddingCreateParams["model"];
 };
 export declare class OpenAIEmbeddingFunction extends EmbeddingFunction<string, 
 Partial<OpenAIOptions>> {
     #private;
     constructor(options?: Partial<OpenAIOptions>);
     toJSON(): {
-         model: (string & {}) | "text-embedding-ada-002" | "text-embedding-3-small" | "text-embedding-3-large";
+         model: string | (string & {}) | undefined;
     };
     ndims(): number;
     embeddingDataType(): Float;
     computeSourceEmbeddings(data: string[]): Promise<number[][]>;
     computeQueryEmbeddings(data: string): Promise<number[]>;
 }
```

wjones127:
On second thought, I don't think this approach will work. I think what I really need to do is make openai a separate import path somehow.

wjones127:
Closing in favor of #1809

PR #1722, opened 2024-10-03
feat: better errors for remote SDK

wjones127:
* Adds nicer errors to remote SDK, that expose useful properties like `request_id` and `status_code`.
* Makes sure the Python tracebacks print nicely by mapping the `source` field from a Rust error to the `__cause__` field.

### DIFF
--- a/python/python/lancedb/remote/errors.py
+++ b/python/python/lancedb/remote/errors.py
@@ -23,7 +23,7 @@ class LanceDBClientError(RuntimeError):
     message: str
         The error message.
     request_id: str
-        The request id of the failed. This can be provided in error reports
+        The id of the request that failed. This can be provided in error reports
         to help diagnose the issue.
     status_code: int
         The HTTP status code of the response. May be None if the request
@@ -46,7 +46,7 @@ class HttpError(LanceDBClientError):
     message: str
         The error message.
     request_id: str
-        The request id of the failed. This can be provided in error reports
+        The id of the request that failed. This can be provided in error reports
         to help diagnose the issue.
     status_code: int
         The HTTP status code of the response. May be None if the request
@@ -71,9 +71,9 @@ class RetryError(LanceDBClientError):
     Attributes
     ----------
     message: str
-        The error message.
+        The retry error message, which will describe which retry limit was hit.
     request_id: str
-        The request id of the failed. This can be provided in error reports
+        The id of the request that failed. This can be provided in error reports
         to help diagnose the issue.
     request_failures: int
         The number of request failures.
--- a/python/python/lancedb/remote/errors.py
+++ b/python/python/lancedb/remote/errors.py
@@ -12,5 +12,102 @@
 #  limitations under the License.
 
 
+from typing import Optional
+
+
 class LanceDBClientError(RuntimeError):
+    """An error that occurred in the LanceDB client.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
+    request_id: str
+        The request id of the failed. This can be provided in error reports
+        to help diagnose the issue.
+    status_code: int
+        The HTTP status code of the response. May be None if the request
+        failed before the response was received.
+    """
+
+    def __init__(
+        self, message: str, request_id: str, status_code: Optional[int] = None
+    ):
+        super().__init__(message)
+        self.request_id = request_id
+        self.status_code = status_code
+
+
+class HttpError(LanceDBClientError):
+    """An error that occurred during an HTTP request.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
+    request_id: str
+        The request id of the failed. This can be provided in error reports
+        to help diagnose the issue.
+    status_code: int
+        The HTTP status code of the response. May be None if the request
+        failed before the response was received.
+    """
+
     pass
+
+
+class RetryError(LanceDBClientError):
+    """An error that occurs when the client has exceeded the maximum number of retries.
+
+    The retry strategy can be adjusted by setting the
+    [retry_config](lancedb.remote.ClientConfig.retry_config) in the client
+    configuration. This is passed in the `client_config` argument of
+    [connect](lancedb.connect) and [connect_async](lancedb.connect_async).
+
+    The __cause__ attribute of this exception will be the last exception that
+    caused the retry to fail. It will be an
+    [HttpError][lancedb.remote.errors.HttpError] instance.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
+    request_id: str
+        The request id of the failed. This can be provided in error reports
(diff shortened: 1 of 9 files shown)


wjones127:
Here's a comparison of the error messages before and after. A few points I think are improved:

1. We don't have the nested prefixes like `RuntimeError: Runtime error: ...` that come from wrapping errors by formatting them to a string for the new message.
2. All the client errors provide a request id in the output message
3. The actual status code is always provided.
4. The retry error has the underlying error provided above. 

# Error - No retry

Response is 507 status with "Internal Server Error" body.

Before:
```
RuntimeError: Runtime error: Internal Server Error
```

After:
```
lancedb.remote.errors.HttpError: Http error: (request_id=214db021-cc8e-4f57-8bd7-176999c13064) 507 Insufficient Storage: Internal Server Error
```

# 429 Error with retry

Before:
```
RuntimeError: Runtime error: Request failed after 2 retries with status code 429 Too Many Requests
```

After:

```
lancedb.remote.errors.HttpError: 429 Too Many Requests: Try again later

The above exception was the direct cause of the following exception:

lancedb.remote.errors.HttpError: Http error: (request_id=f8f9a923-4a36-459b-8ed7-15522c2ee4d5) 429 Too Many Requests: Try again later

The above exception was the direct cause of the following exception:

lancedb.remote.errors.RetryError: Hit retry limit for request_id=f8f9a923-4a36-459b-8ed7-15522c2ee4d5 (request_failures=2/2, connect_failures=0/3, read_failures=0/3)
```

python/python/lancedb/remote/errors.py
```diff
@@ -12,5 +12,102 @@
...
+
 class LanceDBClientError(RuntimeError):
+    """An error that occurred in the LanceDB client.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
+    request_id: str
+        The request id of the failed. This can be provided in error reports
```
westonpace, line 26:
```suggestion
        The id of the request that failed. This can be provided in error reports
```

python/python/lancedb/remote/errors.py
```diff
@@ -12,5 +12,102 @@
...
+        to help diagnose the issue.
+    status_code: int
+        The HTTP status code of the response. May be None if the request
+        failed before the response was received.
+    """
+
+    def __init__(
+        self, message: str, request_id: str, status_code: Optional[int] = None
+    ):
```
westonpace, line 35: Minor nit: would it make sense to move `_check_status` into this constructor?

python/python/lancedb/remote/errors.py
```diff
@@ -12,5 +12,102 @@
...
+
+class HttpError(LanceDBClientError):
+    """An error that occurred during an HTTP request.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
+    request_id: str
+        The request id of the failed. This can be provided in error reports
```
westonpace, line 49:
```suggestion
        The id of the request that failed. This can be provided in error reports
```

python/python/lancedb/remote/errors.py
```diff
@@ -12,5 +12,102 @@
...
+    The __cause__ attribute of this exception will be the last exception that
+    caused the retry to fail. It will be an
+    [HttpError][lancedb.remote.errors.HttpError] instance.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
+    request_id: str
+        The request id of the failed. This can be provided in error reports
```
westonpace, line 76:
```suggestion
        The id of the request that failed. This can be provided in error reports
```

python/python/lancedb/remote/errors.py
```diff
@@ -12,5 +12,102 @@
...
+    [connect](lancedb.connect) and [connect_async](lancedb.connect_async).
+
+    The __cause__ attribute of this exception will be the last exception that
+    caused the retry to fail. It will be an
+    [HttpError][lancedb.remote.errors.HttpError] instance.
+
+    Attributes
+    ----------
+    message: str
+        The error message.
```
westonpace, line 74: Is this the error message of the last response?  Or is that embedded inside this string?

rust/lancedb/src/error.rs
```diff
@@ -46,8 +46,37 @@ pub enum Error {
...
+        /// successful but there was an error decoding the response.
+        status_code: Option<reqwest::StatusCode>,
+    },
+    #[cfg(feature = "remote")]
+    #[snafu(display(
+        "Hit retry limit for request_id={request_id} (\
+        request_failures={request_failures}/{max_request_failures}, \
+        connect_failures={connect_failures}/{max_connect_failures}, \
+        read_failures={read_failures}/{max_read_failures})"
+    ))]
```
westonpace, line 67: `source` is not included in the `display`.  Does that mean the last failure is NOT included in the error returned to python?

rust/lancedb/src/error.rs
```diff
@@ -46,8 +46,37 @@ pub enum Error {
     ObjectStore { source: object_store::Error },
     #[snafu(display("lance error: {source}"))]
     Lance { source: lance::Error },
-    #[snafu(display("Http error: {message}"))]
-    Http { message: String },
+    #[cfg(feature = "remote")]
+    #[snafu(display("Http error: (request_id={request_id}) {source}"))]
```
westonpace, line 50: How does snafu actually stringify `source`?  I'm wondering if there is value in forcing it to include the debug representation (reqwest error `to_string`'s are sometimes incomplete IIRC)

westonpace (approved):

python/python/lancedb/remote/errors.py
wjones127: I'm about to delete it, so I'm not inclined to modify that function right now.

python/python/lancedb/remote/errors.py
wjones127: It's an error message that describe which retry limit was hit.

rust/lancedb/src/error.rs
wjones127: It isn't included in the message of the top-level exception. But in python we travel along the chain of `source` errors to build up a chain of Python exceptions. Python itself prints these error messages nicely.

A big point of this refactor is to move away from sticking the source's error into a string message and instead keep it as a `source` field, so that we can present a nicer final error message to the user.

rust/lancedb/src/error.rs
wjones127: I think ideally we are printing the full chain of errors. (Going through the source fields.) Though snafu doesn't seem to be the best for this. There's a good discussion about this here that I want to look through more deeply: https://stackoverflow.com/questions/62869360/should-an-error-with-a-source-include-that-source-in-the-display-output

PR #1725, opened 2024-10-04
feat(python): drop support for dictionary in Table.add

### LINKED ISSUE #1706: Feature: drop support for dictionary as DATA in `Table.add()`
### SDK

Python

### Description

It's confusing and broken, as illustrated in this issue: https://github.com/lancedb/lancedb/issues/1705#issuecomment-2379540005

We could change it to accept a dictionary as a single row. But we want to nudge users away from writing one row at a time anyways. So better to just not accept a dictionary type.

gagan-bhullar-tech:
PR closes #1706

### DIFF
--- a/python/python/lancedb/common.py
+++ b/python/python/lancedb/common.py
@@ -20,7 +20,7 @@ from .util import safe_import_pandas
 
 pd = safe_import_pandas()
 
-DATA = Union[List[dict], dict, "pd.DataFrame", pa.Table, Iterable[pa.RecordBatch]]
+DATA = Union[List[dict], "pd.DataFrame", pa.Table, Iterable[pa.RecordBatch]]
 VEC = Union[list, np.ndarray, pa.Array, pa.ChunkedArray]
 URI = Union[str, Path]
 VECTOR_COLUMN_NAME = "vector"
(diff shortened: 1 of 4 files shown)


gagan-bhullar-tech:
@wjones127 please review

wjones127 (changes requested):
Thank you for working on this. Here's a few other changes you will need to make:

* Change the type definition of `DATA` to not include `dict` anymore
   * https://github.com/lancedb/lancedb/blob/e61ba7f4e21ed801912e9ea75f07d8c40b29adc2/python/python/lancedb/common.py#L23
* Change docstrings for every method that also uses the `DATA` type as a paramter. Such as `create_table`.
* Please add a test asserting that a user gets a clear error message if they pass a dictionary. It should say something like: "Cannot add a single dictionary to a table. Use a list."
  * Can be similar to this error message: https://github.com/lancedb/lancedb/blob/e61ba7f4e21ed801912e9ea75f07d8c40b29adc2/python/python/lancedb/table.py#L87-L88

gagan-bhullar-tech:
@wjones127 updated PR. Please review.

wjones127 (approved):
Excellent work. Thank you!

PR #1742, opened 2024-10-09
docs(node): update `indexStats` signature and regenerate docs

wjones127:
`indexStats` still referenced UUID even though in https://github.com/lancedb/lancedb/pull/1702 we changed it to take name instead.


westonpace (approved):
I assume all the markdown changes just came from rerunning the "generate md from code" script?

PR #1744, opened 2024-10-11
fix(node): make embedding dependencies optional

### LINKED ISSUE #1624: bug(node): openai still seems to be required as dependency

wjones127:
Fixes #1624

* [ ] bring back sentence-transformers.js 
* [ ] Test with mock openai
* [ ] Test with and without optional dependencies
* [ ] Check docs still look good.


wjones127:
Closing in favor of #1809

PR #1745, opened 2024-10-11
docs: clarify infrastructure requirements for S3 Express One Zone

o-alexandrov:
Clarifies infrastructure requirements for S3 Express One Zone to avoid a possible misunderstanding.
When I read the docs, I was thinking: S3 Express One Zone would work only with EC2.


changhiskhan (approved):
Thanks!

Issue #1480, opened 2024-07-30
Feature: replace dependency on retry with a more actively maintained dependency that does not raise CVE-2022-42969


finnhacks42:
### SDK

Python

### Description

lancedb depends on retry, which depends on py which has a "high severity" CVE raised against it (https://nvd.nist.gov/vuln/detail/CVE-2022-42969). Although this vulnerability is rubbish (disputed, applies only if dealing with subversion projects), its pretense in the dependency tree blocks installation of lancedb in locked down environments.

dayesouza:
There's also an issue on retry package https://github.com/invl/retry/issues/58 where people pointed that a while ago but no action.
There's also some talk about alternatives, since people from other repositories are also facing the same issue

ojturner:
Would also love for this to get looked at for the same reason as the original post

AlonsoGuevara:
Is there any plan to remove this dependency?
We use lancedb on our project https://github.com/microsoft/graphrag and we have a security risk that cannot be addressed

PR #1749, opened 2024-10-15
fix(python): remove dependency on retry package

jameswu1991:
## user story

fixes https://github.com/lancedb/lancedb/issues/1480

https://github.com/invl/retry has not had an update in 8 years, one if its sub-dependencies via requirements.txt (https://github.com/pytest-dev/py) is no longer maintained and has a high severity vulnerability (CVE-2022-42969).

retry is only used for a single function in the python codebase for a deprecated helper function `with_embeddings`, which was created for an older tutorial (https://github.com/lancedb/lancedb/pull/12) [but is now deprecated](https://lancedb.github.io/lancedb/embeddings/legacy/).

## changes

i backported a limited range of functionality of the `@retry()` decorator directly into lancedb so that we no longer have a dependency to the `retry` package.

## tests

```
/Users/james/src/lancedb/python $ ruff check .
All checks passed!
/Users/james/src/lancedb/python $ pytest python/tests/test_embeddings.py
python/tests/test_embeddings.py .......s....                                                                                                                        [100%]
================================================================ 11 passed, 1 skipped, 2 warnings in 7.08s ================================================================
```

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -5,7 +5,6 @@ dependencies = [
     "deprecation",
     "pylance==0.18.2",
     "requests>=2.31.0",
-    "retry>=0.9.2",
     "tqdm>=4.27.0",
     "pydantic>=1.10",
     "attrs>=21.3.0",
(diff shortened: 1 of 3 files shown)


wjones127 (commented):
Thanks for this. Please run `ruff format python` and then this should be good to go.

jameswu1991:
Whoops, I always forget to re-format before submitting the PR. It's updated, thanks for taking a look!

Issue #933, opened 2024-02-06
bug(python): urllib3 requires >=2


chebbyChefNEQ:
### What happened?

there should be a constraint in `pyproject.toml` for `urllib3>=2`

LuQQiu:
This ticket is not valid anymore
as will has removed this dependency fully

PR #831, opened 2024-01-19
fix: remote python client closes idle connections

albertlockett:
Adds an HTTPAdapter to the remote python client that will close idle connections.

This should resolve the issue where our load balancer front-end would decide to close the TCP connection, but the client (using requests module) might send a request before it receives the FIN/ACK packet (resulting in `requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))`). 

We'll handle this by closing idle connections early on the client side. The default idle time is 300 seconds, but can be overridden by passing env var `LANCE_CLIENT_CONNECTION_TIMEOUT` (expects an int value)

PR #1761, opened 2024-10-22
fix(python): add urllib3>=2.0.0 python dependency constraint

LuQQiu:
fixes https://github.com/lancedb/lancedb/issues/933

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -12,6 +12,7 @@ dependencies = [
     "packaging",
     "cachetools",
     "overrides>=0.7",
+    "urllib3>=2.0.0"
 ]
 description = "lancedb"
 authors = [{ name = "LanceDB Devs", email = "dev@lancedb.com" }]


wjones127:
Remind me, what do we use urllib for? I ask because I am removing the required dependency on `requests` in #1701

westonpace (approved):
@wjones127 we use `urllib3` in the python remote client (`client.py` and `connection_timeout.py`).  It appears it was added in https://github.com/lancedb/lancedb/pull/917 and https://github.com/lancedb/lancedb/pull/831

westonpace:
> I ask because I am removing the required dependency on requests in https://github.com/lancedb/lancedb/pull/1701

Ah, I see.  Your PR does remove both files that were depending on `urllib3`.

LuQQiu:
seems this is not needed anymore lol

LuQQiu:
Close because this dependency is not needed anymore

PR #1762, opened 2024-10-22
feat!: upgrade lance to 0.19.1

BubbleCal:
BREAKING CHANGE: default tokenizer no longer does stemming or stop-word removal. Users should explicitly turn that option on in the future.

- upgrade lance to 0.19.1
- update the FTS docs
- update the FTS API

Upstream change notes: https://github.com/lancedb/lance/releases/tag/v0.19.1


rpgreen (approved):

Cargo.toml
```diff
@@ -20,15 +20,15 @@ keywords = ["lancedb", "lance", "database", "vector", "search"]
...
-lance-testing = { "version" = "=0.18.3", git = "https://github.com/lancedb/lance.git", tag = "v0.18.3-beta.2" }
-lance-datafusion = { "version" = "=0.18.3", git = "https://github.com/lancedb/lance.git", tag = "v0.18.3-beta.2" }
-lance-encoding = { "version" = "=0.18.3", git = "https://github.com/lancedb/lance.git", tag = "v0.18.3-beta.2" }
+], git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
+lance-index = { "version" = "=0.19.1", git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
+lance-linalg = { "version" = "=0.19.1", git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
+lance-table = { "version" = "=0.19.1", git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
+lance-testing = { "version" = "=0.19.1", git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
+lance-datafusion = { "version" = "=0.19.1", git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
+lance-encoding = { "version" = "=0.19.1", git = "https://github.com/lancedb/lance.git", tag = "v0.19.1" }
```
wjones127, line 31: When we upgrade to stable versions (non-beta), we should reference the cargo released artifacts, not the git repo.

So:

```toml
lance = "=0.19.1"
...
```

docs/src/fts_tantivy.md
```diff
@@ -0,0 +1,163 @@
+# Full-text search (Tantivy-based FTS)
+
+LanceDB also provides support for full-text search via [Tantivy](https://github.com/quickwit-oss/tantivy), allowing you to incorporate keyword-based search (based on BM25) in your retrieval solutions.
+
+The tantivy-based FTS is available in only Python, for the other languages you can use the [native FTS](fts.md).
```
wjones127, line 5: Maybe also mention here that Tantivy doesn't support object stores or incremental indexing either.

wjones127 (changes requested):

wjones127:
It looks like in the Python sync API we kept the default tokenizer as english. But I don't think you did the same for Nodejs or async Python. I mention this because upstream we made a breaking change to alter the default tokenizer to no language, as described here: https://github.com/lancedb/lance/pull/2992

If we fix it all to default to english, then there is no breaking change. But if we pass through a breaking change, we should label this PR as a breaking change too.

BubbleCal:
> It looks like in the Python sync API we kept the default tokenizer as english. But I don't think you did the same for Nodejs or async Python. I mention this because upstream we made a breaking change to alter the default tokenizer to no language, as described here: [lancedb/lance#2992](https://github.com/lancedb/lance/pull/2992)
> 
> If we fix it all to default to english, then there is no breaking change. But if we pass through a breaking change, we should label this PR as a breaking change too.

Yes we should mark the breaking change
the language matters only if stemming or remove_stop_words enabled. So it's all the same for all the APIs

Cargo.toml
BubbleCal: will change it to upgrade to v0.19.2-beta1 for `FTS search on multiple columns` feature

Cargo.toml
wjones127, line 1: Could we coordinate and get a stable release of LanceDB out with 0.19.1, before we upgrade to a beta version? We are due for a stable release, but can't do that if we are depending on a beta version of Lance.

BubbleCal: Sure

wjones127 (approved):

PR #1769, opened 2024-10-25
feat: support optimize indices in sync API

BubbleCal:


python/python/lancedb/table.py
```diff
@@ -1889,6 +1908,23 @@ def compact_files(self, *args, **kwargs):
...
+            self.checkout_latest()
+            return result
+
+    async def _async_optimize_indices(self, **kwargs):
+        conn = await lancedb_connect(
+            sanitize_uri(self._conn.uri),
+        )
+        table = AsyncTable(await conn.open_table(self.name))
+        return await table.optimize()
+
```
westonpace, line 1927: This will be misleading.  `AsyncTable::optimize` does more than just `optimize_indices`.  It also runs compaction and cleans up older versions.  Can you change the sync method name to `optimize` and change the method description to be the same as `AsyncTable::optimize`:

```
    def optimize(
        self,
        *,
        cleanup_older_than: Optional[timedelta] = None,
        delete_unverified: bool = False,
    ) -> OptimizeStats:
        """
        Optimize the on-disk data and indices for better performance.

        Modeled after ``VACUUM`` in PostgreSQL.

        Optimization covers three operations:

         * Compaction: Merges small files into larger ones
         * Prune: Removes old versions of the dataset
         * Index: Optimizes the indices, adding new data to existing indices

        Parameters
        ----------
        cleanup_older_than: timedelta, optional default 7 days
            All files belonging to versions older than this will be removed.  Set
            to 0 days to remove all versions except the latest.  The latest version
            is never removed.
        delete_unverified: bool, default False
            Files leftover from a failed transaction may appear to be part of an
            in-progress operation (e.g. appending new data) and these files will not
            be deleted unless they are at least 7 days old. If delete_unverified is True
            then these files will be deleted regardless of their age.

        Experimental API
        ----------------

        The optimization process is undergoing active development and may change.
        Our goal with these changes is to improve the performance of optimization and
        reduce the complexity.

        That being said, it is essential today to run optimize if you want the best
        performance.  It should be stable and safe to use in production, but it our
        hope that the API may be simplified (or not even need to be called) in the
        future.

        The frequency an application shoudl call optimize is based on the frequency of
        data modifications.  If data is frequently added, deleted, or updated then
        optimize should be run frequently.  A good rule of thumb is to run optimize if
        you have added or modified 100,000 or more records or run more than 20 data
        modification operations.
        """
    ...
```

westonpace (changes requested):

python/python/lancedb/table.py
BubbleCal: ok, do we still need to make exsiting `compact_files` and `cleanup_old_versions` based on async impl?

westonpace: > ok, do we still need to make exsiting compact_files and cleanup_old_versions based on async impl?

No, we can leave those as sync-only and eventually deprecate.

python/python/lancedb/remote/table.py
```diff
@@ -564,6 +565,21 @@ def compact_files(self, *_):
...
+
+    def optimize(
+        self,
+        *,
+        cleanup_older_than: Optional[timedelta] = None,
+        delete_unverified: bool = False,
+    ):
+        """optimize() is not supported on the LanceDB cloud"""
+        raise NotImplementedError("optimize() is not supported on the LanceDB cloud")
+
```
westonpace, line 582: Can we make a slight change from `"optimize() is not supported on the LanceDB cloud"` to `"optimize() is not supported on the LanceDB cloud.  Indices are optimized automatically."`

python/python/lancedb/table.py
```diff
@@ -1969,6 +1993,81 @@ def compact_files(self, *args, **kwargs):
...
+        data modifications.  If data is frequently added, deleted, or updated then
+        optimize should be run frequently.  A good rule of thumb is to run optimize if
+        you have added or modified 100,000 or more records or run more than 20 data
+        modification operations.
+        """
+        try:
+            asyncio.get_running_loop()
+            self.compact_files()
+            self.cleanup_old_versions(older_than=cleanup_older_than)
+            self.to_lance().optimize.optimize_indices()
```
westonpace, line 2047: We should just raise an error if we reach this point:

```
raise AssertionError("Synchronous method called in asynchronous context.  If you are writing an asynchronous application then please use the asynchronous APIs")
```

Otherwise we will end up with two implementations of everything.

The only reason we had the sync-fallback path in `table_names` was for backwards compatibility.

python/python/lancedb/table.py
```diff
@@ -1969,6 +1993,81 @@ def compact_files(self, *args, **kwargs):
...
+
+    async def _async_optimize(
+        self,
+        cleanup_older_than: Optional[timedelta] = None,
+        delete_unverified: bool = False,
+    ):
+        conn = await lancedb_connect(
+            sanitize_uri(self._conn.uri),
+        )
+        table = AsyncTable(await conn.open_table(self.name))
```
westonpace, line 2066: This is fine here (optimize is expensive so calling `open_table` again is no problem) but I kind of forgot we had to do this and we will probably want a cheaper way to go from sync table to async table in the future.

CC @wjones127

python/python/lancedb/table.py
```diff
@@ -893,6 +896,27 @@ def compact_files(self, *args, **kwargs):
...
+
+        This incrementally updates the indices of the table,
+        to make queries faster.
+
+        Note: This function is not available in LanceDb Cloud (since LanceDb
+        Cloud manages index optimization for you automatically)
+
+        Arguments are passed onto
+        :meth:`lance.dataset.DatasetOptimizer.optimize_indices`.
+        For most cases, the default should be fine.
```
westonpace, line 917: Is this true?  I don't see any `kwargs` being passed on.

westonpace (changes requested):
Thanks for taking this on.  A few suggestions.

python/python/lancedb/table.py
BubbleCal: Oh I think it's from `optimize_indices`, fixed

python/python/lancedb/remote/table.py
BubbleCal: fixed

python/python/lancedb/table.py
BubbleCal: Oh I think it's from optimize_indices, fixed

python/python/lancedb/remote/table.py
```diff
@@ -478,6 +479,27 @@ def compact_files(self, *_):
             "compact_files() is not supported on the LanceDB cloud"
         )
 
+    def optimize_indices(self, **kwargs):
+        """optimize_indices() is not supported on the LanceDB cloud.
+        Indices are optimized automatically."""
+        raise NotImplementedError(
+            "optimize_indices() is not supported on the LanceDB cloud. "
+            "Indices are optimized automatically."
+        )
```
wjones127, line 488: Why add an `optimize_indices` method here? It's not implemented and we don't have one for the local table?

python/python/lancedb/table.py
```diff
@@ -1971,6 +2023,83 @@ def compact_files(self, *args, **kwargs):
...
+        you have added or modified 100,000 or more records or run more than 20 data
+        modification operations.
+        """
+        try:
+            asyncio.get_running_loop()
+            raise AssertionError(
+                "Synchronous method called in asynchronous context. "
+                "If you are writing an asynchronous application "
+                "then please use the asynchronous APIs"
+            )
```
wjones127, line 2079: I added nest_asyncio as a dependency. I think it's simpler if we just use that so this is allowed.

westonpace: I worry that leads to subtle issues down the road but will admit it does enable things like a synchronous middleman.

BubbleCal: i think here we throw an exception intentionally to let the users switch to async API cc @westonpace

python/python/lancedb/remote/table.py
BubbleCal: removed

wjones127 (approved):

westonpace (approved):

PR #1770, opened 2024-10-25
ci: arm64 windows build support

umuthopeyildirim:
Adds support for 'aarch64-pc-windows-msvc'.


.github/workflows/rust.yml
```diff
@@ -139,3 +139,24 @@ jobs:
...
+          Set-Location C:\protoc
+          Invoke-WebRequest https://github.com/protocolbuffers/protobuf/releases/download/v21.12/protoc-21.12-win64.zip -OutFile C:\protoc\protoc.zip
+          7z x protoc.zip
+          Add-Content $env:GITHUB_PATH "C:\protoc\bin"
+        shell: powershell
+      - name: Run tests
+        run: |
+          $env:VCPKG_ROOT = $env:VCPKG_INSTALLATION_ROOT
+          cargo build --target aarch64-pc-windows-msvc
+          cargo test --target aarch64-pc-windows-msvc
```
wjones127, line 162: I don't think this will work since the `windows-latest` runner is x86_64.

node/package.json
```diff
@@ -84,14 +84,16 @@
...
     }
   },
   "optionalDependencies": {
     "@lancedb/vectordb-darwin-arm64": "0.11.1-beta.1",
     "@lancedb/vectordb-darwin-x64": "0.11.1-beta.1",
     "@lancedb/vectordb-linux-arm64-gnu": "0.11.1-beta.1",
     "@lancedb/vectordb-linux-x64-gnu": "0.11.1-beta.1",
-    "@lancedb/vectordb-win32-x64-msvc": "0.11.1-beta.1"
+    "@lancedb/vectordb-win32-x64-msvc": "0.11.1-beta.1",
+    "@lancedb/vectordb-win32-arm64-msvc": "0.11.1-beta.1"
```
wjones127, line 97: If you modify this, there are corresponding places you need to `.bumpversion.toml`:

https://github.com/lancedb/lancedb/blob/04e1f1ee4c0c56efc86f7b36e9f08f06983f1de7/.bumpversion.toml#L69-L93

.github/workflows/rust.yml
wjones127: It doesn't look like there is a OSS/free option for arm Windows runner. That does make me worry about maintaining this https://github.com/actions/runner-images/issues/10820

wjones127 (commented):
Thanks for working on this! I haven't kept up with the state of adoption of Windows ARM, but it would be interesting to support. Left a few comments on small things that need to change. But Would like to also test that we can successfully produce the artifacts and run them on Windows. I can look later for an appropriate VM to try this out on.

It looks like the lockfile also needs to be updated with `npm install`.

node/package.json
umuthopeyildirim: I believe it's already present in `.bumpversion.toml`. Is that wrong?https://github.com/lancedb/lancedb/pull/1770/files#diff-c157bf6a4fcea5e19ccb39979beb75b999c94a5e950492d7e423760fd4320208

.github/workflows/rust.yml
umuthopeyildirim: I know :( I was quite confused with GH Runner docs... I think `lancedb` org needs to create a windows arm runner for it to work.

umuthopeyildirim:
Hi @wjones127 thank you for the feedback updated lock files!

node/package.json
wjones127: Ah I missed that. Thanks for pointing it out

.github/workflows/rust.yml
umuthopeyildirim: Hey @wjones127 What should we do? What are your thoughts?

.github/workflows/rust.yml
```diff
@@ -139,3 +139,24 @@ jobs:
           $env:VCPKG_ROOT = $env:VCPKG_INSTALLATION_ROOT
           cargo build
           cargo test
+  windows-arm64:
+    runs-on: windows-latest
```
wjones127, line 143: I've setup a runner for this:
```suggestion
    runs-on: windows-arm
```

.github/workflows/rust.yml
wjones127: I've setup a runner we can use to test arm64 windows.

wjones127:
Could you also update and test the release jobs? https://github.com/lancedb/lancedb/blob/main/.github/workflows/npm-publish.yml

One way to test them is comment out the final publish step and add a pull-request trigger.

umuthopeyildirim:
Merged from main

umuthopeyildirim:
It looks like `7z` is not preinstalled on `windows-4x-arm`. I can add an install script, but is this a GitHub-hosted runner or a VM you created in Azure?

wjones127:
> It looks like `7z` is not preinstalled on `windows-4x-arm`. I can add an install script, but is this a GitHub-hosted runner or a VM you created in Azure?

It's a github runner.

umuthopeyildirim:
Added `7z` setup 🤞🏻

umuthopeyildirim:
This GH runner has nothing pre-installed LOL added Rust

umuthopeyildirim:
Added build tools

timothycarambat:
@umuthopeyildirim & @wjones127 I wanted to thank you both for getting this on its way. We use Lancedb in [AnythingLLM](https://github.com/Mintplex-Labs/anything-llm) and unlocking Windows `aarch64` would essentially unlock a whole new class of consumer devices that leverage lancedb without x86 emulation for fast speeds.

I know Windows ARM runner stuff is a nightmare to configure. I just wanted to say I appreciate y'alls work on this. Looking forward to its merge in a future release.

umuthopeyildirim:
Hey @wjones127 i think its ready now

wjones127 (approved):
Thanks for working through this. To finish this, you'll also want to adjust the release workflow. But I'll merge this and let you do that in a follow up PR.

PR #1776, opened 2024-10-29
fix: explain plan options

### LINKED ISSUE #1768: bug(python): explain_plan does not include prefiltering/scalar indices
### LanceDB version

_No response_

### What happened?

https://github.com/lancedb/lancedb/blob/04e1f1ee4c0c56efc86f7b36e9f08f06983f1de7/python/python/lancedb/query.py#L472-L504

From what I can see from the implementation here, we're not carrying forward `k`, `nprobes`, or `refine_factor` inside of the `nearest` argument and are not passing down any scalar filters/preflter designations.

_No response_

gagan-bhullar-tech:
PR fixes #1768

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -500,7 +500,11 @@ class LanceQueryBuilder(ABC):
             nearest={
                 "column": self._vector_column,
                 "q": self._query,
+                "k": self._limit,
+                "nprobes": self._nprobes,
             },
+            prefilter=self._prefilter,
+            filter=self._str_query,
         ).explain_plan(verbose)
 
     def vector(self, vector: Union[np.ndarray, list]) -> LanceQueryBuilder:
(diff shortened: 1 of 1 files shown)


wjones127 (commented):
Did you see my comment https://github.com/lancedb/lancedb/issues/1768#issuecomment-2436458354 ? This is missing a lot more parameters still.

gagan-bhullar-tech:
@wjones127 added parameters

wjones127 (changes requested):
Thanks. That looks a bit better. Would still be nice to share the implementation so we don't duplicate the passing of these parameters. But for now, I think we can consider this to be fine.

Could you address the doctest error?

gagan-bhullar-tech:
@wjones127 updated PR

wjones127 (approved):

PR #1782, opened 2024-10-31
fix(rust): fix delete, update, query in remote SDK

wjones127:
Fixes several minor issues with Rust remote SDK:

* Delete uses `predicate` not `filter` as parameter
* Update does not return the row value in remote SDK
* Update takes tuples
* Content type returned by query node is wrong, so we shouldn't validate it. https://github.com/lancedb/sophon/issues/2742
* Data returned by query endpoint is actually an Arrow IPC file, not IPC stream.

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -1,3 +1,4 @@
+use std::io::Cursor;
 use std::sync::{Arc, Mutex};
 
 use crate::index::Index;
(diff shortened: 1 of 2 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -115,39 +115,14 @@ impl<S: HttpSend> RemoteTable<S> {
...
-            return Err(Error::Http {
-                source: format!(
-                    "Expected content type {}, got {}",
-                    ARROW_STREAM_CONTENT_TYPE, content_type
-                )
-                .into(),
-                request_id: request_id.to_string(),
-                status_code: None,
-            });
-        }
```
westonpace, line 145: Do we not verify the content-type any longer?

westonpace (approved):
The change seems ok but I'm not sure I understand why we removed the content-type verification.

rust/lancedb/src/remote/table.rs
wjones127: See this issue: https://github.com/lancedb/sophon/issues/2742

The server returned the incorrect content type. None of the existing clients check this right now, so I think it's best for now we remove the check.

PR #1783, opened 2024-10-31
feat(python): support post filter on FTS

wjones127:


BubbleCal (approved):
Thanks for working on this!
iirc we do post-filtering in sync API by default, we changed that in async API?

PR #1784, opened 2024-11-01
feat(python,node): support with_row_id in Python and remote

wjones127:
Needed to support hybrid search in Remote SDK.


westonpace (approved):
Do you plan to also add this to the nodejs API?

PR #1799, opened 2024-11-06
feat: voyageai support

fzowl:
Adding VoyageAI embedding and rerank support


python/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "lancedb-python"
-version = "0.16.0-beta.0"
+version = "0.16.0-beta.1"
```
wjones127, line 3: You do not need to bump the version. That happens automatically in release process.

fzowl: No, i don't need - i did not find the rules, so i thought the minor versions are raised every time. I'm reverting back this change.

fzowl: @wjones127 Can you please check the PR again?

AyushExel:
Hey @fzowl looks good, can you please address the following:
- Fix linting 
- we'll also need to add it to the docs pages for better discoverability - [Rerankers](https://lancedb.github.io/lancedb/reranking/) , [Embedding functions](https://lancedb.github.io/lancedb/embeddings/default_embedding_functions/). You can also do it in a follow up PR.

wjones127 (approved):

PR #1805, opened 2024-11-07
feat: added lancedb and vectordb release ci for win32-arm64-msvc npmjs only

umuthopeyildirim:


wjones127 (approved):
Thanks for working on this @umuthopeyildirim

PR #1818, opened 2024-11-11
fix: use pandas with pydantic embedding column

eddyxu:
* Make Pandas `DataFrame` works with embedding function + Subset of columns
* Make `lancedb.create_table()` work with embedding function

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import asyncio
 import inspect
+import json
 import time
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -73,6 +74,22 @@ pl = safe_import_polars()
 QueryType = Literal["vector", "fts", "hybrid", "auto"]
 
 
+def _schema_without_embedding_funcs(schema: Optional[pa.Schema]) -> Optional[pa.Schema]:
+    """Return a schema without any embedding function columns"""
+    if schema is None:
+        return None
+    embedding_metadata = (
+        json.loads(schema.metadata.get(b"embedding_functions", b"[]").decode("utf-8"))
+        if schema.metadata
+        else {}
+    )
+    if not embedding_metadata:
+        return schema
+
+    vector_columns = set([meta.get("vector_column") for meta in embedding_metadata])
+    return pa.schema([field for field in schema if field.name not in vector_columns])
+
+
 def _coerce_to_table(data, schema: Optional[pa.Schema] = None) -> pa.Table:
     if _check_for_hugging_face(data):
         # Huggingface datasets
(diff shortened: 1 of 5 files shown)


python/python/lancedb/table.py
```diff
@@ -73,6 +74,22 @@
 QueryType = Literal["vector", "fts", "hybrid", "auto"]
 
 
+def _schema_without_embedding_funcs(schema: Optional[pa.Schema]) -> Optional[pa.Schema]:
+    """Return a schema without any embedding function columns"""
+    if schema is None:
+        return None
+    embedding_metadata = (
```
changhiskhan, line 81: there's a `.embedding_functions` property on the table i believe. So you don't need to duplicate json decoding

python/python/tests/test_embeddings.py
```diff
@@ -129,6 +121,53 @@ class Schema(LanceModel):
...
+        }
+    )
+    db = lancedb.connect(tmp_path)
+    tbl = db.create_table("test", schema=TestSchema, mode="overwrite", data=df)
+    schema = tbl.schema
+    assert schema.field("text").type == pa.string()
+    assert schema.field("val").type == pa.int64()
+    assert schema.field("vector").type == pa.list_(pa.float32(), 128)
+
+    df = pd.DataFrame(
```
changhiskhan, line 157: maybe add unit test for:
1. if the dataframe contains the vector column does it still work?
2. what happens when you have multiple vector columns? does it work with 0, partial, and all supplied? If partial fails that's probably ok for now

python/python/lancedb/table.py
eddyxu: Changed to use `EmbeddingFunctionRegistry.parse_functions()`, because the Table might not exist yet.

python/python/lancedb/table.py
```diff
@@ -5,6 +5,7 @@
 
 import asyncio
 import inspect
+import json
```
changhiskhan, line 8:
```suggestion
```

python/python/tests/test_embeddings.py
eddyxu: `if the dataframe contains the vector column does it still work?`

I'm not sure what the existing behavior is here. Or what should we expect.

python/python/tests/test_embeddings.py
```diff
@@ -129,6 +121,141 @@ class Schema(LanceModel):
...
+
+    class Schema(LanceModel):
+        text: str = model.SourceField()
+        vector: Vector(model.ndims()) = model.VectorField()
+
+    db = lancedb.connect(tmp_path)
+    tbl = db.create_table("test", schema=Schema, mode="overwrite")
+    tbl.add([{"text": "hello world", "vector": np.zeros(128).tolist()}])
+
+    print(tbl.to_arrow())
```
changhiskhan, line 146: `print`

changhiskhan (approved):
comment on one of the unit tests otherwise lgtm pending CI

PR #1824, opened 2024-11-13
docs: update hybrid search example to latest langchain

AyushExel:


QianZhu:
let's update the imports as well:
`from langchain_community.document_loaders import PyPDFLoader`
`from langchain_openai import OpenAIEmbeddings`
and remove 
`from lancedb.pydantic import Vector, LanceModel`

also include `langchain_openai` from `pip install`

QianZhu (approved):

PR #1827, opened 2024-11-13
feat: flexible null handling and insert subschemas in Python

wjones127:
* Test that we can insert subschemas (omit nullable columns) in Python.
    * More work is needed to support this in Node. See: https://github.com/lancedb/lancedb/issues/1832
* Test that we can insert data with nullable schema but no nulls in non-nullable schema.
* Add `"null"` option for `on_bad_vectors` where we fill with null if the vector is bad.
* Make null values not considered bad if the field itself is nullable.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2217,7 +2217,10 @@ def _sanitize_vector_column(
     """
     # ChunkedArray is annoying to work with, so we combine chunks here
     vec_arr = data[vector_column_name].combine_chunks()
-    field = table_schema.field(vector_column_name)
+    if table_schema is not None:
+        field = table_schema.field(vector_column_name)
+    else:
+        field = None
     typ = data[vector_column_name].type
     if pa.types.is_list(typ) or pa.types.is_large_list(typ):
         # if it's a variable size list array,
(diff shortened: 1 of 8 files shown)


python/python/lancedb/table.py
```diff
@@ -1851,7 +1851,7 @@ def create(
             data but will validate against any schema that's specified.
         on_bad_vectors: str, default "error"
             What to do if any of the vectors are not the same size or contains NaNs.
-            One of "error", "drop", "fill".
+            One of "error", "drop", "fill", "null".
```
LuQQiu, line 1854: i thought fill is full with Nans?

wjones127: Fill is fill vectors with values. For example, the input (two 2-dimensional vectors):

```
[[2.0, NaN], [1.0, 3.0]]
```
Becomes (with default fill value 0.0):

```
[[2.0, 0.0], [1.0, 3.0]]
```

`null` replaces the entire vector with null, so that same input would become:

```
[null, [1.0, 3.0]]
```

LuQQiu: Got it, could you help add that info in the doc?
kind of confusing on what the fill fills

LuQQiu (approved):
Looks good

python/python/lancedb/table.py
wjones127: Added a section in the user guide.

PR #1828, opened 2024-11-13
feat: support remote empty query

chebbyChefNEQ:
Support sending empty query types to remote lancedb. also include offset and limit, where were previously omitted.


python/python/lancedb/query.py
wjones127, line 1: Want to add a unit test?

You can model it off of this one

https://github.com/lancedb/lancedb/blob/0fd8a50bd78d7cfd5df6ed84f7fdcb2c67338747/python/python/tests/test_remote_db.py#L181-L197

wjones127 (approved):

PR #1844, opened 2024-11-18
feat: support to sepcify ef search param

BubbleCal:


nodejs/__test__/table.test.ts
```diff
@@ -477,6 +477,54 @@ describe("When creating an index", () => {
...
+      .limit(2)
+      .nearestTo(queryVec)
+      .column("vec")
+      .toArrow();
+    expect(rst2.numRows).toBe(2);
+    expect(rst.toString()).toEqual(rst2.toString());
+
+    // test offset
+    rst = await tbl.query().limit(2).offset(1).nearestTo(queryVec).toArrow();
+    expect(rst.numRows).toBe(1);
```
westonpace, line 521: These seem unrelated to your test?  Or, if we want this to be a test of all parameters on an hnswEf index then maybe rename the test description to `should create and search IVF_HNSW indices`?

python/python/lancedb/query.py
```diff
@@ -638,6 +641,30 @@ def nprobes(self, nprobes: int) -> LanceVectorQueryBuilder:
...
         return self
 
+    def ef(self, ef: int) -> LanceVectorQueryBuilder:
+        """Set the number of candidates to consider during search.
+
+        Higher values will yield better recall (more likely to find vectors if
+        they exist) at the expense of latency.
+
+        This only applies to the HNSW-related index.
+        The default value is 1.5 * k.
```
westonpace, line 651:
```suggestion
        The default value is 1.5 * limit.
```

python/python/lancedb/query.py
```diff
@@ -1197,6 +1227,29 @@ def nprobes(self, nprobes: int) -> LanceHybridQueryBuilder:
...
 
+    def ef(self, ef: int) -> LanceHybridQueryBuilder:
+        """
+        Set the number of candidates to consider during search.
+
+        Higher values will yield better recall (more likely to find vectors if
+        they exist) at the expense of latency.
+
+        This only applies to the HNSW-related index.
+        The default value is 1.5 * k.
```
westonpace, line 1238:
```suggestion
        The default value is 1.5 * limit.
```

python/python/lancedb/query.py
```diff
@@ -1640,6 +1693,23 @@ def nprobes(self, nprobes: int) -> AsyncVectorQuery:
...
 
+    def ef(self, ef: int) -> AsyncVectorQuery:
+        """
+        Set the number of candidates to consider during search
+
+        This argument is only used when the vector column has an HNSW index.
+        If there is no index then this value is ignored.
+
+        Increasing this value will increase the recall of your query but will also
+        increase the latency of your query.  The default value is 1.5 * k.  This
```
westonpace, line 1704:
```suggestion
        increase the latency of your query.  The default value is 1.5 * limit.  This
```

westonpace (approved):
A few minor suggestions.

Can we add a test case verifying a nice error is returned if the user sets this value to be less than `limit`?

BubbleCal:
> A few minor suggestions.
> 
> Can we add a test case verifying a nice error is returned if the user sets this value to be less than `limit`?

Thanks for reminding, just checked it doesn't raise an error, I'm thinking whether we should allow this, and suggestions? cc @eddyxu @chebbyChefNEQ

PR #1847, opened 2024-11-18
feat(python): check vector query is not None

eddyxu:
Fix the type hints of `nearest_to` method, and raise `ValueError` when the input is None

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1495,7 +1495,8 @@ class AsyncQuery(AsyncQueryBase):
         return pa.array(vec)
 
     def nearest_to(
-        self, query_vector: Optional[Union[VEC, Tuple, List[VEC]]] = None
+        self,
+        query_vector: Union[VEC, Tuple, List[VEC]],
     ) -> AsyncVectorQuery:
         """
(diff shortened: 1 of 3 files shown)


eddyxu:
It was reporting the following error message, which is difficult to understand.

```
    self._inner.nearest_to(AsyncQuery._query_vec_to_array(query_vector))
python/lancedb/query.py:1495: in _query_vec_to_array
    return pa.array(vec)
pyarrow/array.pxi:370: in pyarrow.lib.array
    ???
pyarrow/array.pxi:42: in pyarrow.lib._sequence_to_array
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: 'NoneType' object is not iterable
```

wjones127 (approved):

PR #1848, opened 2024-11-18
fix: iter in merge_insert

philipp-horstenkamp:
`iter()` isn't a function but calls the dunder function `__iter__()`.

Otherwise a list and other iterables can't be used for this.

Just try `list([42].iter())` vs. `list(iter([42]))` == `list([43].__iter__())`

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -655,7 +655,7 @@ class Table(ABC):
         2  3  y
         3  4  z
         """
-        on = [on] if isinstance(on, str) else list(on.iter())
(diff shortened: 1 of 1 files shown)


wjones127:
Thanks @philipp-horstenkamp. This might be a bit moot as I think the implementation doesn't support more than 1 key column yet. See https://github.com/lancedb/lance/issues/3124

philipp-horstenkamp:
That you might want to note in your documentation / change the typehint.
This is currently only iritating to read / use.

wjones127:
I ended up making this same change in https://github.com/lancedb/lancedb/pull/2083

PR #1850, opened 2024-11-19
feat: add list_versions to typescript, rust and remote python sdks

albertlockett:
Will require update to lance dependency to bring in this change which makes the version serializable https://github.com/lancedb/lance/pull/3143

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -19,7 +19,7 @@ use http::header::CONTENT_TYPE;
 use http::StatusCode;
 use lance::arrow::json::JsonSchema;
 use lance::dataset::scanner::DatasetRecordBatchStream;
-use lance::dataset::{ColumnAlteration, NewColumnTransform};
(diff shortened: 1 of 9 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -1489,6 +1518,51 @@ mod tests {
...
+        assert_eq!(
+            versions[0].timestamp,
+            "2024-01-01T00:00:00Z".parse::<DateTime<Utc>>().unwrap()
+        );
+        assert_eq!(versions[1].version, 2);
+        assert_eq!(
+            versions[1].timestamp,
+            "2024-02-01T00:00:00Z".parse::<DateTime<Utc>>().unwrap()
+        );
+        // assert_eq!(versions, expected);
```
wjones127, line 1563: Do you need this anymore?
```suggestion
```

wjones127 (approved):
I'm okay with it, but I do feel a little cautious about this version metadata field. Users have no way to write to it now, so not sure we should expose reading it.

PR #1851, opened 2024-11-20
feat: schema evolution APIs in all SDKs

wjones127:
* Support `add_columns`, `alter_columns`, `drop_columns` in Remote SDK and async Python
* Add `data_type` parameter to node
* Docs updates

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -849,8 +849,8 @@ rewriting the column, which can be a heavy operation.
 
     ```python
     import pyarrow as pa
-    table.alter_column({"path": "double_price", "rename": "int_price",
-                        "data_type": pa.int32(), "nullable": False})
+    table.alter_column({"path": "double_price", "rename": "dbl_price",
+                        "data_type": pa.float32(), "nullable": False})
     ```
     **API Reference:** [lancedb.table.Table.alter_columns][]
 
@@ -873,7 +873,7 @@ will remove the column from the schema.
 === "Python"
 
     ```python
-    table.drop_columns(["int_price"])
+    table.drop_columns(["dbl_price"])
     ```
     **API Reference:** [lancedb.table.Table.drop_columns][]
 
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -792,17 +792,19 @@ Use the `drop_table()` method on the database to remove a table.
 
 ## Changing schemas
 
-LanceDB tables require a schema to be specified when creating a table, but the
-schema can change over time. Use `add_columns`, `alter_columns`, and `drop_columns`
-to alter the schema.
+While tables must have a schema specified when they are created, you can
+change the schema over time. There's three methods to alter the schema of
+a table:
+
+* `add_columns`: Add new columns to the table
+* `alter_columns`: Alter the name, nullability, or data type of a column
+* `drop_columns`: Drop columns from the table
 
 ### Adding new columns
 
 You can add new columns to the table with the `add_columns` method. New columns
 are filled with values based on a SQL expression. For example, you can add a new
-column `y` to the table and fill it with the value of `x + 1`. If you want to
-fill it with null, you can use `cast(NULL as <data_type>)` as the SQL expression
-to fill the column with nulls, while controlling the data type of the column.
+column `y` to the table and fill it with the value of `x + 1`.
 
 === "Python"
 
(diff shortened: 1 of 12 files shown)


QianZhu:
https://lancedb.github.io/lancedb/python/python/#lancedb.table.Table.add_columns
let's remove `This is not yet available in LanceDB Cloud.` for the 3 APIs.

docs/src/guides/tables.md
```diff
@@ -790,6 +790,93 @@ Use the `drop_table()` method on the database to remove a table.
...
+    **API Reference:** [lancedb.table.Table.add_columns][]
+
+=== "Typescript"
+
+    ```typescript
+    --8<-- "nodejs/examples/basic.test.ts:add_columns"
+    ```
+    **API Reference:** [lancedb.Table.addColumns](../js/classes/Table.md/#addcolumns)
+
+If you want to fill it with null, you can use `cast(NULL as <data_type>)` as
```
westonpace, line 823: I think my first question as a user is "what am I suppose to use for `<data_type>`?"  Is this an SQL data type?  An arrow data type?  Can we give an example?

docs/src/guides/tables.md
```diff
@@ -790,6 +790,93 @@ Use the `drop_table()` method on the database to remove a table.
...
+
+Changing the name or nullability of a column just updates the metadata. Because
+of this, it's a fast operation. Changing the data type of a column requires
+rewriting the column, which can be a heavy operation.
+
+=== "Python"
+
+    ```python
+    import pyarrow as pa
+    table.alter_column({"path": "double_price", "rename": "my_name",
```
westonpace, line 844: Maybe `rename` can be `int_price` to match the fact that we are casting to an int?  Though I think that means fixing all the examples below this so not a big deal.

nodejs/src/table.rs
```diff
@@ -178,16 +178,20 @@ impl Table {
...
     pub async fn alter_columns(&self, alterations: Vec<ColumnAlteration>) -> napi::Result<()> {
         for alteration in &alterations {
-            if alteration.rename.is_none() && alteration.nullable.is_none() {
+            if alteration.rename.is_none()
+                && alteration.nullable.is_none()
+                && alteration.data_type.is_none()
+            {
                 return Err(napi::Error::from_reason(
-                    "Alteration must have a 'rename' or 'nullable' field.",
+                    "Alteration must have a 'rename', 'dataType', or 'nullable' field.",
```
westonpace, line 186: `dataType` or `data_type`?  I suppose (from the user's perspective) it could be either.  What do we use here?

nodejs/src/table.rs
```diff
@@ -433,24 +437,36 @@ pub struct ColumnAlteration {
     /// The new name of the column. If not provided then the name will not be changed.
     /// This must be distinct from the names of all other columns in the table.
     pub rename: Option<String>,
+    /// A new data type for the column. If not provided then the data type will not be changed.
+    /// This must be a valid data type for the column.
```
westonpace, line 441: What is "a valid data type for the column"?

rust/lancedb/src/remote/table.rs
```diff
@@ -643,25 +643,85 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
     }
     async fn add_columns(
         &self,
-        _transforms: NewColumnTransform,
+        transforms: NewColumnTransform,
```
westonpace, line 646: Should we be using `lancedb::table::NewColumnTransform` instead of `lance::dataset::NewColumnTransform` (same question for some of the other types here)?

rust/lancedb/src/utils.rs
```diff
@@ -175,6 +176,25 @@ pub fn supported_vector_data_type(dtype: &DataType) -> bool {
     }
 }
 
+/// Note: this is temporary until we get a proper datatype conversion in Lance.
+pub fn string_to_datatype(s: &str) -> Option<DataType> {
```
westonpace, line 180: Ah, so we are using arrow types :).  I do think we will want this behavior documented more clearly at some point.  Doesn't have to be this PR.

westonpace (approved):

nodejs/src/table.rs
wjones127: I think in Node you are expected to pass the camelCase variable. So that felt like it makes more sense.

PR #1861, opened 2024-11-21
docs: full-text search doc update

QianZhu:


docs/src/fts.md
```diff
@@ -114,12 +114,45 @@ table.create_fts_index("text",
...
+
+    ```python
+    table.search("puppy").limit(10).where("meta='foo'", prefilte=True).to_list()
+    ```
+
+=== "TypeScript"
+
+    ```typescript
+    await tbl
+    .search("apple")
```
BubbleCal, line 130: do we want to use the same query string to make all examples consistent?
```suggestion
    .search("puppy")
```

BubbleCal (approved):

PR #1863, opened 2024-11-21
feat: support for checkout and checkout_latest in remote sdks

albertlockett:


python/python/lancedb/table.py
```diff
@@ -1012,6 +1012,24 @@ def drop_columns(self, columns: Iterable[str]):
             The names of the columns to drop.
         """
 
+    @abstractmethod
+    def checkout(self):
+        """Checkout a version of the table. This is an in-place operation.
```
westonpace, line 1017: What does "This is an in-place operation" mean?

Why not reuse the comments from the async `checkout` / `checkout_latest` methods?

python/python/lancedb/table.py
```diff
@@ -1012,6 +1012,24 @@ def drop_columns(self, columns: Iterable[str]):
...
+
+        This allows viewing previous versions of the table.
+
+        Calling this method will set the table into time-travel mode. If you
+        wish to return to standard mode, call `checkout_latest`.
```
westonpace, line 1022: I might be nice to link to the spot where time-travel mode is defined but not mandatory

westonpace (approved):
Looks good, just a few very minor thoughts

PR #1873, opened 2024-11-23
fix(python)!: handle bad openai embeddings gracefully

### LINKED ISSUE #1577: Empty String Causes Embedding Error with OpenAI Endpoint
### LanceDB version

v0.13.0b1

### What happened?

Can't embed empty string. This might seem not all that useful, but there are indeed cases, especially in bulk loading tools like dlt where we currently have to [patch it](https://github.com/dlt-hub/dlt/blob/e9c9ecfa8a644fdb516dd74aabca3bf75bafb154/dlt/destinations/impl/lancedb/models.py#L30-L32).

```py
import os
import shutil

import lancedb

os.environ["OPENAI_API_KEY"] = "sk-***"


shutil.rmtree("/tmp/db", ignore_errors=True)
db = lancedb.connect("/tmp/db")

func = get_registry().get("openai").create(name="text-embedding-ada-002")


class Words(LanceModel):


table = db.create_table("words", schema=Words, mode="overwrite")
table.add([{"text": "hello world"}, {"text": ""}])

query = "greetings"
actual = table.search(query).limit(1).to_pydantic(Words)[0]
print(actual.text)
```

```console
WARNING:root:Error occurred: Error code: 400 - {'error': {'message': "'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.", 'type': 'invalid_request_error', 'param': None, 'code': None}}
 Retrying in 2.928719034145293 seconds (retry 1 of 7) 
```

### LINKED ISSUE #1677: Feature: openai embedding function returns None if the result is not valid
### SDK

Python

### Description

Be able to handle error easier, we can

1. Set the return embedding as None
2. logging the line where it is error. 
3. Allow user either drop the row or store the row with None embedding, so users can later to rerun the ingestion with filter `vector_column == null`

eddyxu:
BREAKING-CHANGE: change Pydantic Vector field to be nullable by default.
Closes #1577

### DIFF
--- a/python/python/tests/test_embeddings_slow.py
+++ b/python/python/tests/test_embeddings_slow.py
@@ -1,15 +1,6 @@
-#  Copyright (c) 2023. LanceDB Developers
-#
-#  Licensed under the Apache License, Version 2.0 (the "License");
-#  you may not use this file except in compliance with the License.
-#  You may obtain a copy of the License at
-#      http://www.apache.org/licenses/LICENSE-2.0
-#
-#  Unless required by applicable law or agreed to in writing, software
-#  distributed under the License is distributed on an "AS IS" BASIS,
-#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  See the License for the specific language governing permissions and
-#  limitations under the License.
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
 import importlib
 import io
 import os
@@ -17,6 +8,7 @@ import os
 import lancedb
 import numpy as np
 import pandas as pd
+import pyarrow as pa
 import pytest
 from lancedb.embeddings import get_registry
 from lancedb.pydantic import LanceModel, Vector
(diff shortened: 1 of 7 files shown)


python/python/tests/test_embeddings_slow.py
```diff
@@ -444,6 +436,31 @@ class TextModel(LanceModel):
...
+        text: str = model.SourceField()
+        vector: Vector(model.ndims()) = model.VectorField()
+
+    df = pd.DataFrame({"text": ["hello world", ""]})
+    db = lancedb.connect(tmp_path)
+    tbl = db.create_table("test", schema=TextModel, mode="overwrite")
+
+    tbl.add(df, on_bad_vectors="skip")
+    tb = tbl.to_arrow()
+    print(tb)
```
QianZhu, line 456: remove this?

QianZhu (approved):

python/python/tests/test_embeddings_slow.py
eddyxu: Done

PR #1880, opened 2024-11-25
feat: connection pool for sync client

chebbyChefNEQ:
add a connection pool so sync client can be used in threaded context


chebbyChefNEQ:
close in favor of https://github.com/lancedb/lancedb/pull/1856

PR #1883, opened 2024-11-26
feat: support overwrite and exist_ok mode for remote create_table

albertlockett:
Support passing modes "overwrite" and "exist_ok" when creating a remote table.

### DIFF
--- a/rust/lancedb/src/connection.rs
+++ b/rust/lancedb/src/connection.rs
@@ -133,7 +133,7 @@ impl IntoArrow for NoData {
 
 /// A builder for configuring a [`Connection::create_table`] operation
 pub struct CreateTableBuilder<const HAS_DATA: bool, T: IntoArrow> {
-    parent: Arc<dyn ConnectionInternal>,
+    pub(crate) parent: Arc<dyn ConnectionInternal>,
     pub(crate) name: String,
     pub(crate) data: Option<T>,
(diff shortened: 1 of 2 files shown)


rust/lancedb/src/remote/db.rs
```diff
@@ -130,17 +131,49 @@ impl<S: HttpSend> ConnectionInternal for RemoteDatabase<S> {
             .await
             .unwrap()?;
 
+        let mode = match options.mode {
```
rpgreen, line 134: nit: you could just implement a Into<String> on the enum

rust/lancedb/src/connection.rs
```diff
@@ -133,7 +133,7 @@ impl IntoArrow for NoData {
 
 /// A builder for configuring a [`Connection::create_table`] operation
 pub struct CreateTableBuilder<const HAS_DATA: bool, T: IntoArrow> {
-    parent: Arc<dyn ConnectionInternal>,
+    pub(crate) parent: Arc<dyn ConnectionInternal>,
```
rpgreen, line 136: needed? seems like a leaky abstraction?

albertlockett: I feel like we do need to expose this, so we can use it to construct OpenTableBuilder which gets passed to the callback of `CreateTableMode::ExistOK(<callback>)`.

PR #1892, opened 2024-11-27
docs: better formatting for Node API docs

wjones127:
* Sets `"useCodeBlocks": true`
* Adds a post-processing script `nodejs/typedoc_post_process.js` that puts the parameter description on the same line as the parameter name, like it is in our Python docs. This makes the text hierarchy clearer in those sections and also makes the sections shorter.


wjones127:
## Before

Screenshot 2024-11-26 at 4 38 48 PM


## After

Screenshot 2024-11-26 at 4 38 32 PM

QianZhu (approved):

PR #1895, opened 2024-11-29
feat: support remote options for remote lancedb connection

rpgreen:
* Support subset of storage options as remote options
* Send Azure storage account name via HTTP header

### DIFF
--- a/rust/lancedb/src/remote/db.rs
+++ b/rust/lancedb/src/remote/db.rs
@@ -255,7 +255,7 @@ impl<S: HttpSend> ConnectionInternal for RemoteDatabase<S> {
     }
 }
 
-/// A subset of StorageOptions that are compatible with Remote LanceDB client
+/// RemoteOptions contains a subset of StorageOptions that are compatible with Remote LanceDB connections
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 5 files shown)


albertlockett (approved):

rust/lancedb/src/remote/db.rs
```diff
@@ -243,13 +252,37 @@ impl<S: HttpSend> ConnectionInternal for RemoteDatabase<S> {
...
+#[derive(Clone, Debug, Default)]
+pub struct RemoteOptions(pub HashMap<String, String>);
+
+impl RemoteOptions {
+    pub fn new(options: HashMap<String, String>) -> Self {
+        Self(options)
+    }
+}
+
+impl From<StorageOptions> for RemoteOptions {
```
albertlockett, line 265: Should we consider using TryFrom here in case there's a passed storage option that can't be converted to a remote option, we could return an error rather than ignoring it? WDYT?

rpgreen: I don't think so, because we explicitly allow specific storage options as remote options, all others would be ignored. Unless we want to add additional validation for values that are remote-specific. I don't think we would need that.

Issue #1852, opened 2024-11-20
bug(python): Query limit([-1 | 0 | None]) not being respected


bfcrampton:
### LanceDB version

v0.16.0

### What happened?

The latest version of lancedb is not respecting the Query limit when set to a "no limit" value. See https://lancedb.github.io/lancedb/python/python/#lancedb.query.LanceQueryBuilder.limit:
> Call this method and pass 0, a negative value, or None to remove the limit.

### Are there known steps to reproduce?
```
results_df = (
    db.open_table(lance_schemas.TABLE_NAME)
    .search()
    .select(select_columns)
    .limit(-1)
    .to_pandas()
)
```

In lancedb v0.15.0 this returns all the results of my table. In lancedb v0.16.0, it only returns the default 10 results.

QianZhu:
this is the change introduced in lancedb=v0.16.0. If the query is empty, the `limit` will be set to its default value which would be 10. 
We will update the documentation. Thank you for pointing it out.

bfcrampton:
Hey @QianZhu, thanks so much for the response!

I'm curious – is there any change in the efficiency using this method (i.e. counting rows before executing the query)?

wjones127:
> I'm curious – is there any change in the efficiency using this method (i.e. counting rows before executing the query)?

There isn't any meaningful change. Counting the rows simply uses metadata in the table and doesn't require any additional IO.

bfcrampton:
Terrific, thanks both!

wjones127:
> It looks like the workaround no longer works in the latest LanceDB version, I still only get the top 10 results _even_ when setting `.limit(table.count_rows())`.

Sorry it took a bit to reply. I'm not able to reproduce this though. I'm seeing you can get back more than the default results when requested.

```python
import lancedb
import numpy as np
import pyarrow as pa

vectors = np.random.rand(1000, 32)
data = pa.table({
    "id": range(1000),
    "vector": pa.FixedShapeTensorArray.from_numpy_ndarray(vectors).storage,
})

db = lancedb.connect("memory://")
table = db.create_table("example", data)

(
    table.search(np.random.rand(32))
    .limit(table.count_rows())
    .to_pandas()
)
# Shows 1k rows in results
```

alikefia:
I would vote for defaulting to 10 only for vector search
- Avoids adding limit statement when not needed (and set it when makes sense)
- Avoids hidden bugs : we had already few on consolidation jobs (behavior changed without being noticed)

[...]

PR #1902, opened 2024-12-03
feat: add a pyarrow dataset adapater for LanceDB tables

westonpace:
This currently only works for local tables (remote tables cannot be queried)
This is also exclusive to the sync interface.  However, since the pyarrow dataset interface is synchronous I am not sure if there is much value in making an async-wrapping variant.

### DIFF
--- a/python/python/lancedb/integrations/pyarrow.py
+++ b/python/python/lancedb/integrations/pyarrow.py
@@ -1,0 +1,247 @@
+import logging
+from typing import Any, List, Optional, Tuple, Union, Literal
+
+import pyarrow as pa
+
+from ..table import Table
+
+Filter = Union[str, pa.compute.Expression]
+Keys = Union[str, List[str]]
+JoinType = Literal[
+    "left semi",
+    "right semi",
+    "left anti",
+    "right anti",
+    "inner",
+    "left outer",
+    "right outer",
+    "full outer",
+]
+
+
+class PyarrowScannerAdapter(pa.dataset.Scanner):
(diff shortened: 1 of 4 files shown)


wjones127:
> I didn't want to change the default behavior (None) for the sync API but I did change it so that limit=0 in the sync API now means "return all results".

@QianZhu wanted this changed earlier, though personally I'm not opposed to changing it back. See: https://github.com/lancedb/lancedb/issues/1852#issuecomment-2489837804

python/python/lancedb/integrations/pyarrow.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def from_fragment(self, **kwargs):
+        raise NotImplementedError
+
+    def head(self, num_rows: int):
+        return self.to_reader(limit=num_rows).read_all()
+
+    @property
+    def projected_schema(self):
+        return self.table.schema
```
wjones127, line 67: Not sure if it matters, but shouldn't this account for the `columns` parameter in the constructor?

wjones127 (approved):

python/python/lancedb/integrations/pyarrow.py
westonpace: Good point.  Changed this method to return `self.head(1).schema`

westonpace:
@QianZhu @wjones127 I've removed the breaking change.  Let's just keep it simple for this PR.

wjones127 (approved):

PR #1903, opened 2024-12-03
fix: create and use AsyncLanceMergeInsertBuilder for async batch merge

borisarzentar:
Sync merge couldn't be executed on the async table, so I created an async version of `LanceMergeInsertBuilder` for this purpose.

### DIFF
--- a/python/python/lancedb/merge.py
+++ b/python/python/lancedb/merge.py
@@ -105,3 +105,13 @@ class LanceMergeInsertBuilder(object):
             The value to use when filling vectors. Only used if on_bad_vectors="fill".
         """
         return self._table._do_merge(self, new_data, on_bad_vectors, fill_value)
+
+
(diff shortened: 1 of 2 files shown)


wjones127 (changes requested):
Hi @borisarzentar, thanks for looking at this. Could you add a test that shows what would fail previously, and is fixed by this pull request?

async merge insert is currently tested here: https://github.com/lancedb/lancedb/blob/d8c758513c49af47d2cebb4a9967bdae7ab5dc9a/python/python/tests/test_table.py#L796-L797

To be honest, it is not clear to me whether this actually fixes a bug. But of course you wouldn't have bothered to make a pull request if there was no bug :)

wjones127:
@borisarzentar I'm going to close this for now. Feel free to re-open if you are able to create a failing test.

PR #1905, opened 2024-12-04
feat(python)!: async-sync feature parity on Connections

### LINKED ISSUE #1764: Feature: Same way of connecting to DB synchronously as asynchronously
### SDK

Python

### Description

Currently, the `connect_async` function expects a `storage_options` parameter. However, the `connect` function does the connection differently, and I cannot use the `storage_options`.
Would it be possible to connect in the same way?

### LINKED ISSUE #1791: Transition `LanceDBConnection` to wrap async API
## Requirements

 * Make `connect()` and `connect_async()` accept the same parameters.
     * Can close https://github.com/lancedb/lancedb/issues/1764
 * Make `LanceDBConnection` a wrapper around `AsyncConnection`
     * Use `RemoteTable` for inspiration on how to use the event loop. https://github.com/lancedb/lancedb/blob/c998a47e178fe9d840e53688e8e88d8cc14f3867/python/python/lancedb/remote/db.py#L123
 * **DO NOT** handle `LanceTable`. That is left for [a future issue](https://github.com/lancedb/lancedb/issues/1792).
     * Keep the necessary data at the connection. Pass `_uri`, `read_consistency_interval`, `_entered`, and `storage_options` down to `LanceTable`.
 * Update docs to eliminate differences in SDKs:
    * [Migration guide](https://lancedb.github.io/lancedb/migration/)
    * [Storage options guide](https://lancedb.github.io/lancedb/guides/storage/)

### `connect` vs `connect_async`

Just need to add `storage_options` to `connect()`. These can be propagated and passed down to `LanceTable`.

https://github.com/lancedb/lancedb/blob/c998a47e178fe9d840e53688e8e88d8cc14f3867/python/python/lancedb/__init__.py#L30-L40

https://github.com/lancedb/lancedb/blob/c998a47e178fe9d840e53688e8e88d8cc14f3867/python/python/lancedb/__init__.py#L117-L126

### LINKED ISSUE #1897: [docs update request] install the `adlfs>=2024.2.0` when use the azure blob
> Tips: Remember to install the `adlfs>=2024.2.0` when use the azure blob

@QianZhu @wjones127 Can you please include this information in the docs: https://lancedb.github.io/lancedb/guides/storage/#azure-blob-storage

_Originally posted by @fcakyon in https://github.com/lancedb/lancedb/issues/1102#issuecomment-2509029096_

albertlockett:
Closes #1791
Closes #1764
Closes #1897 (Makes this unnecessary)

BREAKING CHANGE: when using azure connection string `az://...` the call to connect will fail if the azure storage credentials are not set. this is breaking from the previous behaviour where the call would fail after connect, when user invokes methods on the connection.


python/python/lancedb/db.py
```diff
@@ -13,41 +13,38 @@
...
 if TYPE_CHECKING:
+    import pyarrow as pa
     from .pydantic import LanceModel
     from datetime import timedelta
 
     from ._lancedb import Connection as LanceDbConnection
     from .common import DATA, URI
     from .embeddings import EmbeddingFunctionConfig
 
+LOOP = BackgroundEventLoop()
```
wjones127, line 46: Perhaps we should have the `RemoteConnection` share this loop? I don't think there's strong reason to have separate ones.

python/python/lancedb/db.py
```diff
@@ -338,9 +339,27 @@ def __init__(
...
+                sanitize_uri(uri),
+                None,
+                None,
+                None,
+                read_consistency_interval_secs,
+                None,
+                storage_options,
+            )
+
+        self._conn = AsyncConnection(LOOP.run(do_connect()))
```
wjones127, line 362: I don't think the wrapper does much. Can just directly pass the Future/Coroutine to `LOOP.run()`.
```suggestion
        do_connect = lancedb_connect(
            sanitize_uri(uri),
            None,
            None,
            None,
            read_consistency_interval_secs,
            None,
            storage_options,
        )

        self._conn = AsyncConnection(LOOP.run(do_connect))
```

wjones127 (commented):
Looks great! Some minor comments. Also, could you remove the warning here:

https://github.com/lancedb/lancedb/blob/bd82e1f66d184bc13869030f03783cb4cb71da18/docs/src/guides/storage.md?plain=1#L90-L93

from https://lancedb.github.io/lancedb/guides/storage/

python/python/lancedb/db.py
albertlockett: It seems like if we do that, we get an error when we call lancedb_connect:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/albertlockett/Development/lancedb/python/main.py", line 3, in <module>
    db = connect(
        # "~/Desktop/lance_datasets/db05",
    ...<5 lines>...
        }
    )
  File "/Users/albertlockett/Development/lancedb/python/python/lancedb/__init__.py", line 118, in connect
    return LanceDBConnection(
        uri,
        read_consistency_interval=read_consistency_interval,
        storage_options=storage_options,
    )
  File "/Users/albertlockett/Development/lancedb/python/python/lancedb/db.py", line 361, in __init__
    do_connect = lancedb_connect(
        sanitize_uri(uri),
    ...<5 lines>...
        storage_options,
    )
RuntimeError: no running event loop
```
Wrapping the call in this function seems to relieve the error, which is why it's here. If you know an cleaner way to avoid this error, happy to try fix this up.

wjones127: Ah okay. Nevermind. There's probably some async runtime magic going on I don't understand 🤷

python/python/lancedb/db.py
albertlockett: Yeah makes sense. I'll make this change

wjones127 (approved):
Excellent work!

PR #1914, opened 2024-12-06
feat: replace sync implementation with async

### LINKED ISSUE #1792: Transition `LanceTable` to wrap async API
To support `to_lance()`, we should keep the URI and storage options as an attribute on the class. But otherwise, we should dispatch to the LanceDB table.

* Consider making `metric` and `distance_type` aliases of each other https://github.com/lancedb/lancedb/issues/1814
* Another parity: https://github.com/lancedb/lancedb/issues/1817
* Can also fix: https://github.com/lancedb/lancedb/issues/1857

wjones127:
Closes #1792

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3,9 +3,7 @@
 
 from __future__ import annotations
 
-import asyncio
 import inspect
-import time
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime, timedelta
@@ -25,6 +23,7 @@ from typing import (
 from urllib.parse import urlparse
 
 import lance
+from lancedb.db import LOOP
 from .dependencies import _check_for_pandas
 import numpy as np
 import pyarrow as pa
@@ -33,8 +32,9 @@ import pyarrow.fs as pa_fs
 from lance import LanceDataset
 from lance.dependencies import _check_for_hugging_face
 
-from .common import DATA, VEC, VECTOR_COLUMN_NAME, sanitize_uri
+from .common import DATA, VEC, VECTOR_COLUMN_NAME
 from .embeddings import EmbeddingFunctionConfig, EmbeddingFunctionRegistry
+from .index import BTree, IvfPq, Bitmap, LabelList, HnswPq, HnswSq
 from .merge import LanceMergeInsertBuilder
 from .pydantic import LanceModel, model_to_dict
 from .query import (
@@ -58,14 +58,13 @@ from .util import (
 )
 from .index import lang_mapping
 
-from ._lancedb import connect as lancedb_connect
 
 if TYPE_CHECKING:
     import PIL
     from lance.dataset import CleanupStats, ReaderLike
     from ._lancedb import Table as LanceDBTable, OptimizeStats
     from .db import LanceDBConnection
-    from .index import BTree, IndexConfig, IvfPq, Bitmap, LabelList, FTS, HnswPq, HnswSq
+    from .index import IndexConfig, FTS
 
 pd = safe_import_pandas()
 pl = safe_import_polars()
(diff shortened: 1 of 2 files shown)


docs/src/migration.md
```diff
@@ -1,81 +1,14 @@
...
 
-In an effort to ensure all clients have the same set of capabilities we have begun migrating the
-python and node clients onto a common Rust base library. In python, this new client is part of
-the same lancedb package, exposed as an asynchronous client. Once the asynchronous client has
-reached full functionality we will begin migrating the synchronous library to be a thin wrapper
-around the asynchronous client.
+In an effort to ensure all clients have the same set of capabilities we have
+migrated the Python and Node clients onto a common Rust base library. In Python,
+both the synchronous and asynchronous clients are based on this implementation.
+In Node, the new client is available as `@lancedb/lancedb`, which will replace
```
westonpace, line 6:
```suggestion
In Node, the new client is available as `@lancedb/lancedb`, which replaces
```

python/python/lancedb/index.py
```diff
@@ -87,41 +86,57 @@ class LabelList:
...
 
 
+@dataclass
 class FTS:
     """Describe a FTS index configuration.
 
     `FTS` is a full-text search index that can be used on `String` columns
 
     For example, it works with `title`, `description`, `content`, etc.
+
```
westonpace, line 99: Ah, I see your comments on making these data classes.  I think this is a good change, thanks

python/python/lancedb/index.py
```diff
@@ -232,30 +247,17 @@ class HnswPq:
...
-            distance_type=distance_type,
-            num_partitions=num_partitions,
-            num_sub_vectors=num_sub_vectors,
-            num_bits=num_bits,
-            max_iterations=max_iterations,
-            sample_rate=sample_rate,
-            m=m,
-            ef_construction=ef_construction,
-        )
+    distance_type: Literal["l2", "cosine", "dot"] = "l2"
```
westonpace, line 250: Is it intentional to use `l2` instead of `None` as the default here?

python/python/lancedb/index.py
```diff
@@ -232,30 +247,17 @@ class HnswPq:
...
+    num_partitions: Optional[int] = None
+    num_sub_vectors: Optional[int] = None
+    num_bits: int = 8
+    max_iterations: int = 50
+    sample_rate: int = 256
+    m: int = 20
+    ef_construction: int = 300
```
westonpace, line 257: Same thing.  Doesn't this mean we have to change the default in two places instead of one?

python/python/lancedb/table.py
```diff
@@ -457,28 +480,41 @@ def create_scalar_index(
     ):
         """Create a scalar index on a column.
 
+        **Experimental API**
```
westonpace, line 483: Is it still experimental?

python/python/lancedb/table.py
```diff
@@ -967,6 +1003,28 @@ def optimize(
         modification operations.
         """
 
+    @abstractmethod
+    def list_indices(self) -> Iterable[IndexConfig]:
+        """
+        List all indices that have been created with Self::create_index
```
westonpace, line 1009: `Self::create_index` is not very pythonic

python/python/lancedb/table.py
```diff
@@ -1210,17 +1208,14 @@ def _dataset_path(self) -> str:
...
-        return self._ref.dataset_mut
-
-    def to_lance(self) -> LanceDataset:
+    def to_lance(self, **kwargs) -> LanceDataset:
         """Return the LanceDataset backing this table."""
-        return self._dataset
+        return lance.dataset(
+            self._dataset_path,
+            version=self.version,
+            storage_options=self._conn.storage_options,
```
westonpace, line 1216: Hurray for adding storage options, I don't think the async version had that.

python/src/index.rs
```diff
@@ -12,224 +12,147 @@
...
-            .take()
-            .ok_or_else(|| PyRuntimeError::new_err("cannot use an Index more than once"))
+pub fn extract_index_params(source: &Option<Bound<'_, PyAny>>) -> PyResult<LanceDbIndex> {
+    if let Some(source) = source {
+        let class_name: String = source
+            .getattr(intern!(source.py(), "__class__"))?
+            .getattr(intern!(source.py(), "__name__"))?
+            .extract()?;
+
+        match class_name.rsplit_once('.').map(|v| v.1).unwrap_or(&class_name) {
```
westonpace, line 36: Hmm, interesting pattern

python/src/index.rs
```diff
@@ -12,224 +12,147 @@
...
+            "BTree" => Ok(LanceDbIndex::BTree(BTreeIndexBuilder::default())),
+            "Bitmap" => Ok(LanceDbIndex::Bitmap(Default::default())),
+            "LabelList" => Ok(LanceDbIndex::LabelList(Default::default())),
+            "FTS" => {
+                let params = source.extract::<FtsParams>()?;
+                let inner_opts = TokenizerConfig::default()
+                    .base_tokenizer(params.base_tokenizer)
+                    .language(&params.language)
+                    .map_err(|_| PyValueError::new_err(format!("LanceDB does not support the requested language: \"{}\"", params.language)))?
```
westonpace, line 45: If I'm being pedantic about errors it doesn't seem like this one is adding valuable context.  It seems there should be some way to just pass through the original error.  Was the original error not helpful or clear in some way?

westonpace (approved):
This is a heroic effort, great work!  A few nits but nothing significant.

python/src/index.rs
wjones127: The original error would have been a JSON deserialization error, which I think would have been confusing to the user. I can add a list of supported languages to the error, and that would make it more useful.

PR #1926, opened 2024-12-08
feat: support azure account name storage options in sync db.connect

LuQQiu:
db.connect with azure storage account name is supported in async connect but not sync connect. 
Add this functionality

### DIFF
--- a/python/python/lancedb/remote/db.py
+++ b/python/python/lancedb/remote/db.py
@@ -44,8 +44,17 @@ class RemoteDBConnection(DBConnection):
         client_config: Union[ClientConfig, Dict[str, Any], None] = None,
         connection_timeout: Optional[float] = None,
         read_timeout: Optional[float] = None,
+        storage_options: Optional[Dict[str, str]] = None,
     ):
         """Connect to a remote LanceDB database."""
+        if storage_options:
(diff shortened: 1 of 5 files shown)


python/python/lancedb/remote/db.py
```diff
@@ -44,8 +44,17 @@ def __init__(
         client_config: Union[ClientConfig, Dict[str, Any], None] = None,
         connection_timeout: Optional[float] = None,
         read_timeout: Optional[float] = None,
+        storage_options: Optional[Dict[str, str]] = None,
     ):
         """Connect to a remote LanceDB database."""
+        if storage_options:
```
eddyxu, line 50: Can we check this on the server side?

LuQQiu:
@eddyxu @wjones127 Addressed the comment, PTAL

QianZhu:
@LuQQiu Do we plan to add `storage_options` to our nodejs SDK as well? Don't have to address it in this PR

LuQQiu:
@QianZhu will double check with them about their SDK, but yeah, should be supported in both
i will also check the Nodejs SDK part, likely not supported yet

eddyxu (approved):

PR #1931, opened 2024-12-10
fix: support list of numpy f16 floats as query vector

eddyxu:
User reported on Discord, when using `table.vector_search([np.float16(1.0), np.float16(2.0), ...])`, it yields `TypeError: 'numpy.float16' object is not iterable`

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1634,9 +1634,7 @@ class AsyncQuery(AsyncQueryBase):
         if (
             isinstance(query_vector, list)
             and len(query_vector) > 0
-            and not isinstance(
-                query_vector[0], (float, int, np.float16, np.float32, np.float64)
-            )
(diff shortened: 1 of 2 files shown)


python/python/lancedb/query.py
```diff
@@ -1644,7 +1634,7 @@ def nearest_to(
         if (
             isinstance(query_vector, list)
             and len(query_vector) > 0
-            and not isinstance(query_vector[0], (float, int))
+            and not np.isrealobj(query_vector[0])
```
westonpace, line 1637: `isrealobj` seems like an odd choice.  We want to test if query_vector[0] is "array like" in some way right?

eddyxu: You are right, changed to check it is a list / np.ndarray / pa.array

westonpace (approved):

PR #1941, opened 2024-12-15
docs: add async examples to doc

QianZhu:
- added sync and async tabs for python examples
- moved python code to tests/docs


docs/src/python/polars_arrow.md
```diff
@@ -2,6 +2,9 @@
 
 LanceDB supports [Polars](https://github.com/pola-rs/polars), a blazingly fast DataFrame library for Python written in Rust. Just like in Pandas, the Polars integration is enabled by PyArrow under the hood. A deeper integration between Lance Tables and Polars DataFrames is in progress, but at the moment, you can read a Polars DataFrame into LanceDB and output the search results from a query to a Polars DataFrame.
 
+!!! note
+    Asynchronous `to_polars` is still work-in-progress. Therefore, only the synchonous APIs are used in this example
+
```
QianZhu, line 7: @wjones127, let me know whether you are ok with this

docs/src/ann_indexes.md
```diff
@@ -24,18 +24,27 @@ Lance supports `IVF_PQ` index type by default.
...
-
     # Create and train the index - you need to have enough data in the table for an effective training step
     tbl.create_index(num_partitions=256, num_sub_vectors=96)
+
+    # Asynchronous client
+    async_db = await lancedb.connect_async(uri)
+    # Add the vectors to a table
+    async_tbl = await async_db.create_table("my_vectors", data=data)
+    # Create and train the index - you need to have enough data in the table for an effective training step
+    await async_tbl.create_index("vector", config=IvfPq(num_partitions=256, num_sub_vectors=96))
```
wjones127, line 47: Could we do tabs for async vs sync? I worry having both and using comments will make the docs pages long and cluttered. Easier if someone can choose up front which they want and then all the following examples will keep with that language.

docs/src/fts.md
```diff
@@ -94,7 +113,10 @@ Stemming is useful for improving search results by reducing words to their root
 
 For example, to enable stemming for English:
 ```python
-table.create_fts_index("text", use_tantivy=True, tokenizer_name="en_stem")
+# Synchronous client
+table.create_fts_index("text", tokenizer_name="en_stem")
+# Asynchronous client
+await async_table.create_index("text", config=FTS(tokenizer_name="en_stem"))
```
wjones127, line 119: There is no `tokenizer_name` parameter in FTS. I think you instead want:

```suggestion
await async_table.create_index("text", config=FTS(language="English", stem=True, remove_stop_words=True))
```

Also, maybe we should move more of these into `python/tests/docs` so that the examples are tested? That would catch mistakes like this.

docs/src/python/duckdb.md
```diff
@@ -4,6 +4,9 @@ In Python, LanceDB tables can also be queried with [DuckDB](https://duckdb.org/)
 
 This integration is done via [Apache Arrow](https://duckdb.org/docs/guides/python/sql_on_arrow), which provides zero-copy data sharing between LanceDB and DuckDB. DuckDB is capable of passing down column selections and basic filters to LanceDB, reducing the amount of data that needs to be scanned to perform your query. Finally, the integration allows streaming data from LanceDB tables, allowing you to aggregate tables that won't fit into memory. All of this uses the same mechanism described in DuckDB's blog post *[DuckDB quacks Arrow](https://duckdb.org/2021/12/03/duck-arrow.html)*.
 
+!!! note
+    Asynchronous `to_lance` support is coming soon. We'll provide examples once it's ready.
```
wjones127, line 8: `to_lance` returns a synchronous API. I think users who want to use `to_lance` should just use sync `connect()`. We don't need to promise this for the async API.

docs/src/python/polars_arrow.md
```diff
@@ -2,6 +2,9 @@
 
 LanceDB supports [Polars](https://github.com/pola-rs/polars), a blazingly fast DataFrame library for Python written in Rust. Just like in Pandas, the Polars integration is enabled by PyArrow under the hood. A deeper integration between Lance Tables and Polars DataFrames is in progress, but at the moment, you can read a Polars DataFrame into LanceDB and output the search results from a query to a Polars DataFrame.
 
+!!! note
+    This example uses synchronous APIs as asynchronous `to_polars` support is under development.
+
```
wjones127, line 7: Same thing here. We don't plan on implementing async to_polars (Polars itself is entirely sync), so let's not promise this in our docs.

wjones127 (changes requested):
I think it would be a lot more readable if we had tabs for `sync` / `async`, like we did for the old/new Node SDK.

I also think we should eventually move all examples into the test files, but that should probably be a future PR.

wjones127 (approved):
Nice work!

PR #1947, opened 2024-12-17
feat(python): add `bypass_vector_index` to sync api

### LINKED ISSUE #535: Add `bypass_vector_index` to Python sync API

H3zi:
Hi lancedb team,

This PR adds the `bypass_vector_index` logic to the sync API, as described in [Issue #535](https://github.com/lancedb/lancedb/issues/535). (Closes #535).

Iv'e implemented it only for the regular vector search. If you think it should also be supported for FTS, Hybrid, or Empty queries and for the cloud solution, please let me know, and I’ll be happy to extend it.

Since there’s no `CONTRIBUTING.md` or contribution guidelines, I opted for the simplest implementation to get this started.

Looking forward to your feedback!

Thanks!

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -515,6 +515,7 @@ class LanceQueryBuilder(ABC):
                 "metric": self._metric,
                 "nprobes": self._nprobes,
                 "refine_factor": self._refine_factor,
+                "use_index": self._use_index,
             },
             prefilter=self._prefilter,
             filter=self._str_query,
@@ -807,7 +808,7 @@ class LanceVectorQueryBuilder(LanceQueryBuilder):
         self._str_query = query_string if query_string is not None else self._str_query
         return self
 
-    def bypass_vector_index(self) -> LanceQueryBuilder:
+    def bypass_vector_index(self) -> LanceVectorQueryBuilder:
         """
         If this is called then any vector index is skipped
 
@@ -820,7 +821,7 @@ class LanceVectorQueryBuilder(LanceQueryBuilder):
         Returns
         -------
         LanceVectorQueryBuilder
-            The LanceQueryBuilder object.
+            The LanceVectorQueryBuilder object.
         """
         self._use_index = False
         return self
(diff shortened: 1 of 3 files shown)


wjones127 (changes requested):
Thanks @H3zi. Thanks for working on this. The implementation changes make sense.

You are right we don't have a `CONTRIBUTING.md` yet. I will add one. But my first piece of guidance is that we don't accept PRs for new features (or bug fixes) that don't have any unit tests. Please add some :)

If you're wondering how you can test `bypass_vector_index`, one tip is you can print out the `.explain_plan()` output and test for the absence of the index steps in the plan.

wjones127:
> If you think it should also be supported for FTS, Hybrid, or Empty queries and for the cloud solution

Yes, it should be implemented for FTS, Hybrid, and empty queries.

For the Cloud API, you shouldn't need any additional changes. So I wouldn't worry about that.

H3zi:
> > If you think it should also be supported for FTS, Hybrid, or Empty queries and for the cloud solution
> 
> Yes, it should be implemented for FTS, Hybrid, and empty queries.
> 
> For the Cloud API, you shouldn't need any additional changes. So I wouldn't worry about that.

Diving deeper, I’m uncertain about how `bypass_vector_index` is relevant to FTS, considering there’s no “vector” search involved. WDYT?

wjones127:
> Diving deeper, I’m uncertain about how `bypass_vector_index` is relevant to FTS, considering there’s no “vector” search involved. WDYT?

Ah nevermind. It was originally just called `use_index`, so could also apply to FTS and scalar indices. But since it's now just for vector indices, it should only make sense on vector and hybrid searches.

H3zi:
@wjones127  I’ve added the logic to `LanceHybridQueryBuilder` and included a test for the regular vector search case. However, I couldn’t find an existing synchronous hybrid query test to build upon for testing the bypass logic...

wjones127 (approved):
This is looking good. Thanks!

PR #1955, opened 2024-12-18
feat: support IVF_FLAT, binary vectors and hamming distance

BubbleCal:
binary vectors and hamming distance can work on only IVF_FLAT, so introduce them all in this PR.

### DIFF
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -113,28 +113,17 @@ indexes work in LanceDB.
 
 ## Binary vector
 
-LanceDB supports binary vectors as a data type, and has the ability to search binary vectors with hamming distance. The binary vectors are stored as uint8 arrays:
+LanceDB supports binary vectors as a data type, and has the ability to search binary vectors with hamming distance. The binary vectors are stored as uint8 arrays (every 8 bits are stored as a byte).:
+
+!!! note
+    The dim of the binary vector must be a multiple of 8. A vector of dim 128 will be stored as a uint8 array of size 16.
 
 === "Python"
 
     ```python
-    import lancedb
-    import numpy as np
+    --8<-- "python/python/tests/docs/test_binary_vector.py:sync_binary_vector"
 
-    db = lancedb.connect("data/sample-lancedb")
-
-    data = [
-        {
-            "id": i,
-            "vector": np.random.randint(0, 256, size=128),
-        }
-        for i in range(1024)
-    ]
-
-    tbl = db.create_table("my_binary_vectors", data=data)
-
-    query = np.random.randint(0, 256, size=128)
-    results = tbl.search(query).to_arrow()
+    --8<-- "python/python/tests/docs/test_binary_vector.py:async_binary_vector"
     ```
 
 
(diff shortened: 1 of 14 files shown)


docs/src/search.md
```diff
@@ -107,6 +111,33 @@ an ANN search means that using an index often involves a trade-off between recal
 See the [IVF_PQ index](./concepts/index_ivfpq.md) for a deeper description of how `IVF_PQ`
 indexes work in LanceDB.
 
+## Binary vector
+
+LanceDB supports binary vectors as a data type, and has the ability to search binary vectors with hamming distance. The binary vectors are stored as uint8 arrays:
```
wjones127, line 116: Is each bit in the uint8 array a value? or each byte? Could you describe that format a little more?

docs/src/search.md
```diff
@@ -107,6 +111,33 @@ an ANN search means that using an index often involves a trade-off between recal
...
+            "vector": np.random.randint(0, 256, size=128),
+        }
+        for i in range(1024)
+    ]
+
+    tbl = db.create_table("my_binary_vectors", data=data)
+
+    query = np.random.randint(0, 256, size=128)
+    results = tbl.search(query).to_arrow()
+    ```
```
wjones127, line 138: Going forward, I think our preference is to put these tests in `python/tests/docs` and reference them with the snip operator `--8<--`. For example, these Python examples:

https://lancedb.github.io/lancedb/basic/#connect-to-a-database

Are here: 

https://github.com/lancedb/lancedb/blob/59b57055e78c723b84793688eeb108105026df37/docs/src/basic.md?plain=1#L137-L142

And the actual code is here:

https://github.com/lancedb/lancedb/blob/59b57055e78c723b84793688eeb108105026df37/python/python/tests/docs/test_basic.py#L15-L19

If we use these, all the examples are well tested.

rust/lancedb/src/table.rs
```diff
@@ -1870,12 +1919,22 @@ impl TableInternal for NativeTable {
...
             }
 
-            let query_vector = query_vector.as_primitive::<Float32Type>();
-            scanner.nearest(
-                &column,
-                query_vector,
-                query.base.limit.unwrap_or(DEFAULT_TOP_K),
-            )?;
+            if is_binary {
+                let query_vector = arrow::compute::cast(&query_vector, &DataType::UInt8)?;
```
wjones127, line 1923: is query_vector here already forced to be a float32? I worry about the accuracy of it here...

wjones127 (changes requested):
Could you make sure to add a test that queries an IVFFlat binary index? Right now, you just create an index with it. I'm particularly worried that the casting of the query vector will mess with the accuracy.

docs/src/search.md
BubbleCal: added

docs/src/search.md
BubbleCal: added

python/python/lancedb/index.py
```diff
@@ -355,6 +355,97 @@ class HnswSq:
     ef_construction: int = 300
 
 
+@dataclass
+class IvfFlat:
```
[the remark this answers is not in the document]
BubbleCal: added

rust/lancedb/src/table.rs
BubbleCal: yeah it's float32 here.
the values must be integers in range 0~255 (even stored as float32s) so it won't lose any accuracy?
I just added a test to verify this

wjones127 (approved):

PR #1956, opened 2024-12-18
Python v0.9.0 patch

rpgreen:


python/python/tests/test_table.py
```diff
@@ -735,7 +735,7 @@ def test_create_scalar_index(db):
     indices = table.to_lance().list_indices()
     assert len(indices) == 1
     scalar_index = indices[0]
-    assert scalar_index["type"] == "Scalar"
```
eddyxu, line 738: Curious why this is changed. Does it break things? Or is this something we did not catch before.

rpgreen: I was wondering too - these were picked up from @LuQQiu's original branch

PR #1957, opened 2024-12-18
docs: improve optimization docs

wjones127:
* Add `See Also` section to `cleanup_old_files` and `compact_files` so they know it's linked to `optimize`.
* Fixes link to `compact_files` arguments
* Improves formatting of note.

### DIFF
--- a/python/python/lancedb/remote/table.py
+++ b/python/python/lancedb/remote/table.py
@@ -15,6 +15,7 @@ from datetime import timedelta
 import logging
 from functools import cached_property
 from typing import Dict, Iterable, List, Optional, Union, Literal
+import warnings
 
 from lancedb._lancedb import IndexConfig
 from lancedb.embeddings.base import EmbeddingFunctionConfig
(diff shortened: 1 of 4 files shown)


QianZhu:
can we clean up these two places as well:
https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/remote/table.py#L483
https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/remote/table.py#L489

wjones127:
> can we clean up these two places as well:

I've updated the docstrings and also hidden those methods from the API docs. I figure simpler they are hidden than showing them as methods only to just say "Not supported". It's explained elsewhere that they aren't necessary on Cloud. And the `NotSupported` error message also explains this, just in case they try to call it anyways.

QianZhu (approved):

python/python/lancedb/remote/table.py
```diff
@@ -481,15 +481,25 @@ def update(
...
     def cleanup_old_versions(self, *_):
-        """cleanup_old_versions() is not supported on the LanceDB cloud"""
+        """
+        cleanup_old_versions() is not supported nor necessary on LanceDB Cloud.
+
+        Tables are automatically cleaned up and optimized.
+        """
         raise NotImplementedError(
-            "cleanup_old_versions() is not supported on the LanceDB cloud"
+            "cleanup_old_versions() is not supported on the LanceDB Cloud. "
```
changhiskhan, line 490: not a blocker since it's existing behavior, but should we raise an exception or just log a warning?
From a code migration perspective from local to cloud/enterprise, this means the user won't have to change their code. But if we don't raise an exception for Remote, would that present any footguns from a user perspective?

python/python/lancedb/table.py
```diff
@@ -936,21 +933,38 @@ def cleanup_old_versions(
...
         For most cases, the default should be fine.
+
+        See Also
+        --------
+        [Table.optimize][lancedb.table.Table.optimize]: A more comprehensive
+            optimization operation that includes cleanup as well as other operations.
+
+        Notes
+        -----
+        This function is not available in LanceDB Cloud (since LanceDB
```
changhiskhan, line 966: If we make the remote methods no-ops instead of raising an error then we just remove this?

wjones127: I think we can at least say it's unnecessary in the docs.

PR #1965, opened 2024-12-19
feat(python)!: support inserting and upserting subschemas

wjones127:
BREAKING CHANGE: For a field "vector", list of integers will now be converted to binary (uint8) vectors instead of f32 vectors. Use float values instead for f32 vectors.

* Adds proper support for inserting and upserting subsets of the full schema. I thought I had previously implemented this in #1827, but it turns out I had not tested carefully enough.
* Refactors `_santize_data` and other utility functions to be simpler and not require `numpy` or `combine_chunks()`.
* Added a new suite of unit tests to validate sanitization utilities.

## Examples

```python
import pandas as pd
import lancedb

db = lancedb.connect("memory://demo")
intial_data = pd.DataFrame({
    "a": [1, 2, 3],
    "b": [4, 5, 6],
    "c": [7, 8, 9]
})
table = db.create_table("demo", intial_data)

# Insert a subschema
new_data = pd.DataFrame({"a": [10, 11]})
table.add(new_data)
table.to_pandas()
```
```
    a    b    c
0   1  4.0  7.0
1   2  5.0  8.0
2   3  6.0  9.0
3  10  NaN  NaN
4  11  NaN  NaN
```

```python
# Upsert a subschema
upsert_data = pd.DataFrame({
    "a": [3, 10, 15],
    "b": [6, 7, 8],
})
table.merge_insert(on="a").when_matched_update_all().when_not_matched_insert_all().execute(upsert_data)
table.to_pandas()
```
```
    a    b    c
0   1  4.0  7.0
1   2  5.0  8.0
2   3  6.0  9.0
3  10  7.0  NaN
4  11  NaN  NaN
5  15  8.0  NaN
```

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2459,6 +2459,20 @@ def _infer_target_schema(table: pa.Table) -> pa.Schema:
             )
 
             schema = schema.set(i, new_field)
+        elif (
+            field.name == VECTOR_COLUMN_NAME
+            and (pa.types.is_list(field.type) or pa.types.is_large_list(field.type))
+            and pa.types.is_integer(field.type.value_type)
+        ):
+            # Use the most common length of the list as the dimensions
+            dim = _modal_list_size(table.column(i))
+            new_field = pa.field(
+                VECTOR_COLUMN_NAME,
+                pa.list_(pa.uint8(), dim),
+                nullable=field.nullable,
+            )
+
+            schema = schema.set(i, new_field)
 
     return schema
 
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -25,7 +25,6 @@ from urllib.parse import urlparse
 import lance
 from lancedb.background_loop import LOOP
 from .dependencies import _check_for_pandas
-import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc
 import pyarrow.fs as pa_fs
@@ -165,6 +164,8 @@ def _sanitize_data(
     metadata: Optional[dict] = None,  # embedding metadata
     on_bad_vectors: str = "error",
     fill_value: float = 0.0,
+    *,
+    allow_subschema: bool = False,
 ) -> Tuple[pa.Table, pa.Schema]:
     data = _coerce_to_table(data, schema)
 
@@ -173,8 +174,9 @@ def _sanitize_data(
         metadata.update(data.schema.metadata or {})
         data = data.replace_schema_metadata(metadata)
 
-    # TODO improve the logics in _sanitize_schema
-    data = _sanitize_schema(data, schema, on_bad_vectors, fill_value)
+    data = _sanitize_schema(
+        data, schema, on_bad_vectors, fill_value, allow_subschema=allow_subschema
+    )
     if schema is None:
         schema = data.schema
 
@@ -2175,8 +2177,10 @@ def _sanitize_schema(
     schema: pa.Schema = None,
     on_bad_vectors: str = "error",
     fill_value: float = 0.0,
+    *,
+    allow_subschema: bool = False,
 ) -> pa.Table:
-    """Ensure that the table has the expected schema.
+    """Ensure vector columns have the expected data type.
 
     Parameters
     ----------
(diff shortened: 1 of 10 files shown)


python/python/tests/test_embeddings.py
```diff
@@ -198,7 +198,6 @@ class TestSchema(LanceModel):
         {
             "text": ["hello world", "goodbye world"],
             "val": [1, 2],
-            "not-used": ["s1", "s3"],
```
wjones127, line 201: @eddyxu You added this test asserting we could insert junk columns and they would be ignored. (#1818) Is that important behavior to keep? This PR breaks it.

python/python/lancedb/table.py
```diff
@@ -2246,141 +2368,118 @@ def _sanitize_vector_column(
...
+            field.name == VECTOR_COLUMN_NAME
+            and (pa.types.is_list(field.type) or pa.types.is_large_list(field.type))
+            and pa.types.is_floating(field.type.value_type)
+        ):
+            # Use the most common length of the list as the dimensions
+            dim = _modal_list_size(table.column(i))
+
+            new_field = pa.field(
+                VECTOR_COLUMN_NAME,
+                pa.list_(pa.float32(), dim),
```
LuQQiu, line 2457: Could be stupid question, would it overflow if the list element length is bigger than this dim

wjones127: It will error at the casting step.

python/python/lancedb/table.py
```diff
@@ -154,27 +139,172 @@ def _coerce_to_table(data, schema: Optional[pa.Schema] = None) -> pa.Table:
...
+    ----------
+    target_schema : Optional[pa.Schema], default None
+        The schema to cast the table to. This is typically the schema of the table
+        if it already exists. Otherwise it might be a user-requested schema.
+    allow_subschema : bool, default False
+        If True, the input table is allowed to omit columns from the target schema.
+        The target schema will be filtered to only include columns that are present
+        in the input table before casting.
+    metadata : Optional[dict], default None
+        The embedding metadata to add to the schema.
```
westonpace, line 198: What is the reasoning for this?  Do we know?

python/python/lancedb/table.py
```diff
@@ -154,27 +139,172 @@ def _coerce_to_table(data, schema: Optional[pa.Schema] = None) -> pa.Table:
...
+        if target_field is None:
+            raise ValueError(f"Field {field.name} not found in target schema")
+        fields.append(target_field)
+    reordered_schema = pa.schema(fields, metadata=target_schema.metadata)
+    if not allow_subschema and len(reordered_schema) != len(target_schema):
+        raise ValueError(
+            "Input table has different number of columns than target schema"
+        )
+
+    if allow_subschema:
```
westonpace, line 260:
```suggestion
    if allow_subschema and len(reordered_schema) != len(target_schema):
```
Save some work if we have all the fields?

python/python/lancedb/table.py
```diff
@@ -2165,74 +2308,53 @@ def migrate_v2_manifest_paths(self):
...
     on_bad_vectors: str = "error",
     fill_value: float = 0.0,
 ) -> pa.Table:
     """
-    Ensure that the vector column exists and has type fixed_size_list(float32)
+    Ensure that the vector column exists and has type fixed_size_list(float)
+
+    If the table has a schema, the vector column is created with the data
+    type from the schema. If the schema is not provided, the vector column
+    is cast to fixed_size_list(float32) if necessary.
```
westonpace, line 2357: How is the schema not provided?  Or how is it provided?  It looks like you removed the `table_schema` argument?

westonpace (approved):
Looks like a good cleanup.  Just some nits.

python/python/lancedb/table.py
```diff
@@ -154,27 +139,172 @@ def _coerce_to_table(data, schema: Optional[pa.Schema] = None) -> pa.Table:
...
 
+def _process_iterator(data: Iterable) -> pa.Table:
+    batches = []
+    schema = None  # Will get schema from first batch
+    for batch in data:
+        batch_table = _into_pyarrow_table(batch)
+        if schema is not None:
+            if batch_table.schema != schema:
+                try:
+                    batch_table = batch_table.cast(schema)
```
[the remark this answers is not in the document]
wjones127: It's because we have this wonderful unit test that passes in an iterable where each batch is a different kind of object.

https://github.com/lancedb/lancedb/blob/0e496ed3b5749f54c4c2eba243169a6abe66c063/python/python/tests/test_db.py#L115-L156

python/python/lancedb/table.py
wjones127: A schema might not be passed down, but metadata still might be passed down in those cases. For that case, the schema is inferred from the data and the metadata is added to it.

python/python/lancedb/table.py
wjones127: I can remove that docstring. The function isn't concerned with casting anymore.

PR #1967, opened 2024-12-20
feat(python): `flatten` in `AsyncQuery`

### LINKED ISSUE #1949: feat(python): Add `flatten` option to async `to_pandas()` method
In `LanceQueryBuilder.to_pandas()` there is an option `flatten`:

https://github.com/lancedb/lancedb/blob/37c359ed40e860d8f2eec3b72e53ab91a6ad82c8/python/python/lancedb/query.py#L274

This is just implemented in Python:

https://github.com/lancedb/lancedb/blob/37c359ed40e860d8f2eec3b72e53ab91a6ad82c8/python/python/lancedb/query.py#L290-L306

We should implement the same function (likely a good idea to pull out as a utility) in the Async version:

https://github.com/lancedb/lancedb/blob/37c359ed40e860d8f2eec3b72e53ab91a6ad82c8/python/python/lancedb/query.py#L1498

gagan-bhullar-tech:
PR fixes #1949


python/python/lancedb/query.py
```diff
@@ -1516,11 +1499,19 @@ async def to_pandas(self) -> "pd.DataFrame":
...
         """Return the execution plan for this query.
 
+        Parameters
+        ----------
+        flatten: Optional[Union[int, bool]]
+            If flatten is True, flatten all nested columns.
+            If flatten is an integer, flatten the nested columns up to the
+            specified depth.
+            If unspecified, do not flatten the nested columns.
+
```
wjones127, line 1514: Why is this on `explain_plan`? Shouldn't it be in the `to_pandas()` section?

python/python/lancedb/util.py
```diff
@@ -172,6 +172,37 @@ def safe_import_polars():
...
+def flatten_columns(tbl: pa.Table, flatten: Optional[Union[int, bool]] = None):
+    """
+        Flatten all struct columns in a table.
+
+        Parameters
+        ----------
+        flatten: Optional[Union[int, bool]]
+            If flatten is True, flatten all nested columns.
+            If flatten is an integer, flatten the nested columns up to the
+            specified depth.
```
wjones127, line 185:
```suggestion
    Flatten all struct columns in a table.

    Parameters
    ----------
    flatten: Optional[Union[int, bool]]
        If flatten is True, flatten all nested columns.
        If flatten is an integer, flatten the nested columns up to the
        specified depth.
```

wjones127 (changes requested):
Thanks for working on this, @gagan-bhullar-tech 

Could you also add a unit test for `AsyncQuery.to_pandas()`?

gagan-bhullar-tech:
@wjones127 please review

Issue #1950, opened 2024-12-17
feat(python): support `.rerank()` on non-hybrid queries in Async API


wjones127:
In the Sync API, `.rerank()` is available in [FTS](https://github.com/lancedb/lancedb/blob/37c359ed40e860d8f2eec3b72e53ab91a6ad82c8/python/python/lancedb/query.py#L959) and [Vector queries](https://github.com/lancedb/lancedb/blob/37c359ed40e860d8f2eec3b72e53ab91a6ad82c8/python/python/lancedb/query.py#L769), in addition to [Hybrid queries](https://github.com/lancedb/lancedb/blob/37c359ed40e860d8f2eec3b72e53ab91a6ad82c8/python/python/lancedb/query.py#L1209C9-L1209C15).

In Async API, this is only available on hybrid queries. We should add support for vector and FTS queries to also be re-ranked.

renato2099:
Hi, I'd like to work on this please

renato2099:
Hi @wjones127 , is this what you had in mind? https://github.com/lancedb/lancedb/pull/1972

PR #1972, opened 2024-12-22
feat(python): support .rerank() on non-hybrid queries in Async API (WIP)

renato2099:
Fixes https://github.com/lancedb/lancedb/issues/1950


renato2099:
Hi @wjones127 , I am missing where/how to hook up the rerank method with the FTSQuery and the provided builders, or is the unit test wrong and I should focus on the queryBuilder rather than in chaining the operations?
Any help is much appreciated!

python/python/lancedb/query.py
```diff
@@ -1727,10 +1744,22 @@ class AsyncFTSQuery(AsyncQueryBase):
     def __init__(self, inner: LanceFTSQuery):
         super().__init__(inner)
         self._inner = inner
+        self._reranker = RRFReranker()
```
albertlockett, line 1747: We can probably allow this to be None by default, similar to how it's done for the synchronous API:
https://github.com/lancedb/lancedb/blob/50c30c5d34dab481ca00897ed9fb63aa8b033567/python/python/lancedb/query.py#L844

python/python/lancedb/query.py
```diff
@@ -1727,10 +1744,22 @@ class AsyncFTSQuery(AsyncQueryBase):
...
+    def rerank(
+        self,
+        reranker: Reranker = RRFReranker(),
+    ) -> AsyncFTSQuery:
+        if reranker and not isinstance(reranker, Reranker):
+            raise ValueError("reranker must be an instance of Reranker class.")
+
+        self._reranker = reranker
+
+        return AsyncFTSQuery(self._inner.rerank(reranker))
```
albertlockett, line 1761: calling `self._inner.rerank` calls into the underlying rust implementation of `FTSQuery`:
https://github.com/lancedb/lancedb/blob/50c30c5d34dab481ca00897ed9fb63aa8b033567/python/src/query.rs#L146

But since there's no `rerank` method on this struct, I don't think this call will succeed. 

Probably instead we'd want to do something like:

```suggestion
        self._reranker = reranker
        return self
```

python/python/lancedb/query.py
```diff
@@ -1815,6 +1844,7 @@ def __init__(self, inner: LanceVectorQuery):
         """
         super().__init__(inner)
         self._inner = inner
+        self._reranker = RRFReranker()
```
albertlockett, line 1847: Similar to my comment on AsyncFTSQuery, by default the reranker can probably be `None`

This is similar to the synchronous python SDK:
https://github.com/lancedb/lancedb/blob/50c30c5d34dab481ca00897ed9fb63aa8b033567/python/python/lancedb/query.py#L610

python/python/lancedb/query.py
```diff
@@ -1378,6 +1379,10 @@ def __init__(self, inner: Union[LanceQuery | LanceVectorQuery]):
         """
         self._inner = inner
 
+    def rerank(self, reranker=Reranker) -> AsyncQuery:
+        self._reranker = reranker
+        return self._inner.rerank(reranker)
+
```
albertlockett, line 1385: I'm not sure this method is needed. Since reranking only makes sense in the context of certain query types (vector, FTS or Hybrid search), we could probably refrain from adding reranker to the base async query.

python/python/lancedb/query.py
```diff
@@ -1605,6 +1610,7 @@ def __init__(self, inner: LanceQuery):
         """
         super().__init__(inner)
         self._inner = inner
+        self._reranker = RRFReranker()
```
albertlockett, line 1613: I'm not sure this is needed. The reranker can be a member of the query builder for the concrete search type (vector, FTS or Hybrid)

python/python/lancedb/query.py
```diff
@@ -1691,6 +1697,17 @@ def nearest_to(
...
+    def rerank(
+        self,
+        reranker: Reranker = RRFReranker(),
+    ) -> AsyncFTSQuery:
+        if reranker and not isinstance(reranker, Reranker):
+            raise ValueError("reranker must be an instance of Reranker class.")
+
+        self._reranker = reranker
+
+        return AsyncFTSQuery(self._inner.rerank(reranker))
```
albertlockett, line 1709: I'm not sure this method is needed. If the user wanted to do FTS search with a reranker, they could probably do something like:
```py
table.query()
  .nearest_to_text("foo") # creates an AsyncFTSQuery
  .reranker(RRFReranker())
```

python/python/lancedb/query.py
```diff
@@ -1026,7 +1026,8 @@ def rerank(self, reranker: Reranker) -> LanceEmptyQueryBuilder:
         LanceEmptyQueryBuilder
             The LanceQueryBuilder object.
         """
-        raise NotImplementedError("Reranking is not yet supported.")
+        self._reranker = reranker
+        return self
```
albertlockett, line 1030: I'm not sure this change is needed

python/python/lancedb/_lancedb.pyi
```diff
@@ -66,6 +68,7 @@ class Query:
     def limit(self, limit: int): ...
     def offset(self, offset: int): ...
     def nearest_to(self, query_vec: pa.Array) -> VectorQuery: ...
+    def rerank(self, Reranker) -> FTSQuery: ...
```
albertlockett, line 71: I don't believe this change is needed. These definitions are effectively to give us intellisense for the underlying rust types in this file:
https://github.com/lancedb/lancedb/blob/main/python/src/query.rs#L146

But since we haven't modified this Rust code (and I don't think we'll need to for this PR, since our rerankers are already all written in python), we won't need to make changes in this file either.

python/python/lancedb/query.py
albertlockett: This is also why the unit test is failing

albertlockett (commented):
hi @renato2099 -- thanks again for working on this. I think it will be a good contribution!

I left some feedback about how I think we can improve how the builders are wired up, and hopefully this will also fix your test.

FYI, at some point I think we'll also need to call the rerank method on the results, similar to how we do it in the synchronous SDK here:
https://github.com/lancedb/lancedb/blob/50c30c5d34dab481ca00897ed9fb63aa8b033567/python/python/lancedb/query.py#L892-L894

python/python/lancedb/query.py
renato2099: yeah it is not needed, it was me trying desperately, I have removed in the latest version of the PR

python/python/lancedb/query.py
renato2099: done in commit 3134eade69c0dcd4d72903ce080f9fb95d95b4b8

python/python/lancedb/query.py
renato2099: thanks! now it all makes sense :)

python/python/lancedb/query.py
renato2099: done in commit 3134eade69c0dcd4d72903ce080f9fb95d95b4b8

python/python/lancedb/query.py
renato2099: done in the latest version of the PR

python/python/lancedb/query.py
renato2099: you are right, it is not needed. While I was tinkering with this I thought that the `rerank` method was not being recognized because of some class hierarchy issue, so I ended up trying to add it in a few places.

Removed this method in latest version of the PR

python/python/lancedb/query.py
renato2099: it is not, sorry for the noise !

python/python/lancedb/_lancedb.pyi
renato2099: totally makes sense , thanks!
I thought it could be needed because of the type checking done over in query.py 
```
if TYPE_CHECKING:
    import PIL
    import polars as pl

    from ._lancedb import Query as LanceQuery
    from ._lancedb import FTSQuery as LanceFTSQuery
    from ._lancedb import HybridQuery as LanceHybridQuery
    from ._lancedb import VectorQuery as LanceVectorQuery
    from .common import VEC
    from .pydantic import LanceModel
    from .table import Table
```
such that all methods would have a defined return type as well, but I see now that it is not required.

Dropped this change as well.

renato2099:
> I left some feedback about how I think we can improve how the builders are wired up, and hopefully this will also fix your test.

Thanks for the pointers @albertlockett ! things are clearer now :)
I applied your feedback, let me know if there are other things we could do to improve this PR though

> FYI, at some point I think we'll also need to call the rerank method on the results, similar to how we do it in the synchronous SDK here:

I see, I get the high-level motivation, but I am struggling a bit to know where exactly in the code that would be, are you thinking about somewhere like `LanceTable`/`AsyncTable`?

albertlockett:
> > I left some feedback about how I think we can improve how the builders are wired up, and hopefully this will also fix your test.
> 
> Thanks for the pointers @albertlockett ! things are clearer now :) I applied your feedback, let me know if there are other things we could do to improve this PR though
> 
> > FYI, at some point I think we'll also need to call the rerank method on the results, similar to how we do it in the synchronous SDK here:
> 
> I see, I get the high-level motivation, but I am struggling a bit to know where exactly in the code that would be, I mean the async APIs -> sync APIs, and it seems like all sync APIs are already doing that, maybe in the TableAsyncAPI? but I am probably just missing more context :) happy to work on it anyway

Hey @renato2099!

As far as I know, our Async query builders aren't actually calling the underlying synchronous query. E.g. `AsyncFTSQuery` isn't calling `LanceFtsQueryBuilder`when it is returning the results. So even if the synchronous APIs are already invoking the rerankers, we still need to have the async queries invoke them as well. 

I think you'd want to invoke the reranker in the `to_arrow`method of the Async Queries.

You can see how we do this for the synchronous queries here:
https://github.com/lancedb/lancedb/blob/445a312667dcf975a01c10f6b8e390a79ec844b0/python/python/lancedb/query.py#L892-L894

For example, one possibility could be in AsyncFtsQuery we'd want to do something like:
```
async def to_arrow():
  results = await super().to_arrow()
  if self._reranker:
     results = self._reranker.rerank_fts(results)
   return results
```

renato2099:
oh I see what you mean @albertlockett ! I thought that there was some additional magic that would be hooking up the syncQueryBuilders and the AsyncQueries (pushed a commit with your suggestion, thanks!)

Another alternative would be to use `LanceFtsQueryBuilder` and `LanceVectorQueryBuilder` directly in the corresponding Async classes, but would that be less desirable because AsyncQueryBase already abstracts this away through the Rust bindings? I just find it a bit confusing to class hierarchy + rust bindings vs using the lance-sync-query-builders

albertlockett (approved):

albertlockett:
Hey @renato2099 Thanks again for this contribution! 
>  I thought that there was some additional magic that would be hooking up the syncQueryBuilders and the AsyncQueries (pushed a commit with your suggestion, thanks!)

> Another alternative would be to use LanceFtsQueryBuilder and LanceVectorQueryBuilder directly in the corresponding Async classes, but would that be less desirable because AsyncQueryBase already abstracts this away through the Rust bindings? I just find it a bit confusing to class hierarchy + rust bindings vs using the lance-sync-query-builders

There's no additional magic :). For a bit of historical context (in case your interested), the synchronous python API was written first, and then earlier last year we added the async python API which was intended to be a thin wrapper around the rust bindings. The idea behind the async API being structured that way was to avoid a lot of duplicate logic across duplicate SDKs (the nodejs SDK is also a thin wrapper around the rust SDK).

PR #1974, opened 2024-12-24
ci(node): remove hardcoded toolchain from typescript release build

wjones127:
We upgraded the toolchain in #1960, but didn't realize we hardcoded it in `npm-publish.yml`. I found if I just removed the hard-coded toolchain, it selects the correct one.

This didn't fully fix Windows Arm, so I created a follow-up issue here: https://github.com/lancedb/lancedb/issues/1975

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,9 +1,9 @@
 name: NPM Publish
 
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


wjones127:
Fixed runs: https://github.com/lancedb/lancedb/actions/runs/12485469407?pr=1974

Issue #1983, opened 2024-12-26
bug(python): plain search limited to 10 by default (regression)


alikefia:
### LanceDB version

v0.17.2

### What happened?

New behavior : all query are limited to 10. Typically table.to_arrow() is limited now. According to the BaseQuery doc, plan queries are not limited.

### Are there known steps to reproduce?

```
db = await lancedb.connect_async(uri)
t = await db.open_table(table)
print(await t.to_arrow())
```

fcakyon:
@QianZhu @wjones127 The linked answer does not resolve this issue since it's not related to `search` or `query` functions.

This snippet is used to return all rows in LanceDB open source:

```python
db = await lancedb.connect(uri)
table = db.open_table(table)
all_results = table.to_pandas()
```

Now, it returns only the first 10 rows, and the limit cannot be increased since to_pandas doesn't accept any input.

Please revert this change or make to_pandas accept `limit` or `num_items` parameter.

adobelis:
I wrote this in the other thread, but since I guess there are two threads for this. 

Here is a workaround that worked for me:

df = lance_table.to_lance().to_table().to_pandas()

PR #1984, opened 2024-12-26
fix(python): do not enforce query limit to 10 (expected behavior for plain queries)

alikefia:


alikefia:
Fix #1983

changhiskhan (changes requested):
Not having a default limit doesn't really make sense for vector search. This would also be a breaking change that would catch a lot of users by surprise. Is this just for consistency between vector vs non-vector query? Or do you need a total ordering of the table based on distance from a query vector?

alikefia:
Nope, the default is enforced here: https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/table.rs#L1928

alikefia:
Double checked :)

alikefia:
Added tests

alikefia:
@changhiskhan there is more to check on that one? It is blocking for us for consolidation jobs (copying full tables). Thx for support.

alikefia:
Fixed remote table defaults

alikefia:
Guys, can you please run the ci pipeline on that one ? Thx a lot !

bfcrampton:
Fixes https://github.com/lancedb/lancedb/issues/1852

wjones127 (commented):
This looks reasonable. But I think you should update the docs as well.

alikefia:
The proposed implem is aligned with the doc so far 
- https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/query.rs#L346
- https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/query.py#L1481
- https://github.com/lancedb/lancedb/blob/main/nodejs/lancedb/query.ts#L232

This is why for me, it is more a bug fix than a change on the API

wjones127 (approved):
I see. Those are documented correctly then.

wjones127:
@changhiskhan you have any further objection?

judahrand:
> The proposed implem is aligned with the doc so far
> 
> * https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/query.rs#L346
> * https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/query.py#L1481
> * https://github.com/lancedb/lancedb/blob/main/nodejs/lancedb/query.ts#L232
> 
> This is why for me, it is more a bug fix than a change to the API

I was referring to the docs on lancedb.query.LanceQueryBuilder found here https://lancedb.github.io/lancedb/python/python/#lancedb.query.LanceQueryBuilder.limit

alikefia:
@judahrand : indeed. I aligned to others for more consistency.

alikefia:
@changhiskhan any update on that ?

rust/lancedb/src/remote/table.rs
```diff
@@ -206,6 +206,11 @@ impl<S: HttpSend> RemoteTable<S> {
     ) -> Result<Vec<serde_json::Value>> {
         Self::apply_query_params(&mut body, &query.base)?;
 
+        // Fill default top_k for vector queries
+        if query.base.limit.is_none() {
+            body["k"] = serde_json::Value::Number(serde_json::Number::from(DEFAULT_TOP_K));
+        }
+
```
wjones127, line 213: This is only applied to remote queries, not to local lanceDB ones.

Instead, I think the appropriate place to apply the default limit is when either `nearest_to()` or `full_text_search()` is called. That should capture both local and remote queries.

PR #1992, opened 2025-01-03
docs: fix full-text search (Native FTS) TypeScript doc error

ahaapple:
Fix

```
Cannot find name 'queryType'.ts(2304)
any
```


albertlockett (approved):

albertlockett:
thanks for the fix @ahaapple!

PR #1993, opened 2025-01-03
feat: support vector search with distance thresholds

BubbleCal:


python/python/tests/test_query.py
```diff
@@ -94,6 +94,49 @@ def test_with_row_id(table: lancedb.table.Table):
...
+    min_dist = dists[0]
+    max_dist = dists[-1]
+
+    res = table.search(q).distance_range(lower_bound=max_dist).to_arrow()
+    assert len(res) == 1
+    assert res["_distance"].to_pylist() == [max_dist]
+
+    res = table.search(q).distance_range(upper_bound=max_dist).to_arrow()
+    assert len(res) == 1
+    assert res["_distance"].to_pylist() == [min_dist]
```
westonpace, line 110: If `upper_bound=max_dist` wouldn't `len(res)` be greater than 1?  Did you mean `min_dist` here?

python/python/tests/test_query.py
```diff
@@ -94,6 +94,49 @@ def test_with_row_id(table: lancedb.table.Table):
     assert rs["_rowid"].to_pylist() == [0, 1]
 
 
+def test_distance_range(table: lancedb.table.Table):
```
westonpace, line 97: Can we add a test that returns more than 1 result?  Maybe also one with 0 results?

What happens if the number of results is larger than the limit?  Do we still return the highest K results in that range?

rust/lancedb/src/query.rs
```diff
@@ -831,6 +837,14 @@ impl VectorQuery {
         self
     }
 
+    /// Set the distance range for vector search,
+    /// only rows with distances in the range [lower_bound, upper_bound) will be returned
+    pub fn distance_range(mut self, lower_bound: Option<f32>, upper_bound: Option<f32>) -> Self {
+        self.lower_bound = lower_bound;
+        self.upper_bound = upper_bound;
```
westonpace, line 844: Should we return an error if both `lower_bound` and `upper_bound` are `None`?

westonpace (approved):
This looks great.  Glad to finally be getting this feature, it's been asked for a while :)

westonpace:
Oh, also, are we leaving typescript for a follow-up?

python/python/tests/test_query.py
BubbleCal: The results in range [lower, upper) will be returned.
So here we have only 2 rows and upper_bound=max_dist, so the farther one will be filtered out

rust/lancedb/src/query.rs
BubbleCal: I'd prefer don't return an error for this case
It's equal to be without any distance threshold, or "revert the distance range parameter"

python/python/tests/test_query.py
BubbleCal: Yes it still returns at most k results

Issue #1395, opened 2024-06-21
Feature request: Use `typing.Annotated[]` for fixed size vector hinting in Lance -> Pydantic conversion


ntjess:
### SDK

Python

### Description

https://lancedb.github.io/lancedb/basic/#using-the-embedding-api suggests the following class to convert a table query into a Pydantic list:

```python
class Words(LanceModel):
    text: str = func.SourceField()
    vector: Vector(func.ndims()) = func.VectorField()
```

However, type checkers like Pyright correctly complain that this is an "unfair" use of type hints:
![image](https://github.com/lancedb/lancedb/assets/23620506/39aa177f-c48e-4f45-aed5-e05486c22185)

One solution is to use `typing.Annotated[FixedSizeList, Vector(func.ndims())]` instead. This allows the static checker to correctly infer the type, and allows the conversion tools to access the "annotated" vector information for conversion.

This change would 1) prevent "antipatterns" of runtime-dependent static hints, and 2) give static intellisense knowledge of the object's type (which currently defaults to `Unknown`)

One downside is greater technical complexity. However, Pydantic already makes extensive use of `Annotated` attributes for validation, metadata insertion, etc. -- so any user interested in converting to Pydantic traits can expect to see these patterns elsewhere in models.

changhiskhan:
@ntjess i think this makes sense. We currently don't have bandwidth to tackle this so if it's more urgent, we'd be very grateful for a PR. And probably at least in the beginning, we should keep the existing API for compat purposes for a some time before deprecating it. Thanks!

sayandipdutta:
I can work on it. What I have seen of the implementation, an intermediate base class would solve the problem. But there is an opportunity for bikeshedding regarding the name of that class, because it will be the base expression in the Annotated field. Would be glad if you can suggest a name.

sayandipdutta:
Current workaround to this problem is:
```python
from typing import NewType

FixedSizeList = NewType("FixedSizeList", Vector(func.ndims()))

class Words(LanceModel):
    text: str = func.SourceField()
    vector: FixedSizeList = func.VectorField()
```

PR #1997, opened 2025-01-05
feat: provide pydantic model Vector column constraints as Annotated metadata

sayandipdutta:
# Current Behavior:
```python
from lancedb.pydantic import Vector

class MyModel(LanceModel):
    test: str = model.SoureField()
    vector: Vector(model.ndims()) = model.VectorField()
#           ^^^^^^^^^^^^^^^^^^^^^ typechecker error here, function call not supported in type hint
```

# After this PR:
```python
from typing import Annotated
from lancedb.pydantic import FixedSizeList, Vector

class MyModel(LanceModel):
    test: str = model.SoureField()
    vector: Annotated[FixedSizeList, Vector(model.ndims())] = model.VectorField()  # ok
```

It is fully backward compatible (You can still provide `vector: Vector(ndim)`).

closes: #1395

### DIFF
--- a/python/python/lancedb/pydantic.py
+++ b/python/python/lancedb/pydantic.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 import inspect
 import sys
 import types
@@ -12,6 +13,7 @@ from abc import ABC, abstractmethod
 from datetime import date, datetime
 from typing import (
     TYPE_CHECKING,
+    Annotated,
     Any,
     Callable,
     Dict,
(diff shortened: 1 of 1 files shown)


sayandipdutta:
I have not updated the docs, or all the places this appears in doctests and tests. Will do so after feedback/review.

msu-reevo:
@sayandipdutta are you waiting on review? I was going to do this but then found your PR. Do we just need to get attention from the lancedb team?

sayandipdutta:
@msu-reevo Actually there is one little problem with this PR. You can see the failing test case. The new style annotation that I implemented, doesn't work with pydantic v1. I didn't manage to make it work. The older style still works though. Currently I can't find time to try and fix it. My guess is, it can't be made to work with pydantic v1. So, either you can take over, and see if your implementation works with pydantic v1, or we want someone from lance db team, to take a decision here.

PR #1999, opened 2025-01-07
feat: support distance range in queries

BubbleCal:
this also updates the docs

### DIFF
--- a/docs/src/search.md
+++ b/docs/src/search.md
@@ -136,6 +136,29 @@ LanceDB supports binary vectors as a data type, and has the ability to search bi
         --8<-- "python/python/tests/docs/test_binary_vector.py:async_binary_vector"
         ```
 
+## Search with distance range
+
+You can also search for vectors within a specific distance range from the query vector. This is useful when you want to find vectors that are not just the nearest neighbors, but also those that are within a certain distance. This can be done by using the `distance_range` method.
+
+=== "Python"
+
+    === "sync API"
+
+        ```python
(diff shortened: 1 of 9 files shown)


docs/src/search.md
```diff
@@ -136,6 +136,39 @@ LanceDB supports binary vectors as a data type, and has the ability to search bi
         --8<-- "python/python/tests/docs/test_binary_vector.py:async_binary_vector"
         ```
 
+## Search with distance range
+
+You can also search for vectors within a specific distance range from the query vector. This is useful when you want to find vectors that are not just the nearest neighbors, but also those that are within a certain distance. This can be done by using the `distance_range` method.
```
QianZhu, line 141: Do we have recommendation on how to choose a distance threshold, per distance metric used?

wjones127: That probably depends more on the users data and use case than the distance metric. Plus, it's a query-time setting, so it's easy to experiment with quickly. I think users won't have a hard time tuning that to their use case.

wjones127 (approved):
Nice work

nodejs/package.json
```diff
@@ -88,6 +88,7 @@
     "version": "napi version"
   },
   "dependencies": {
+    "@lancedb/lancedb": "^0.14.1",
```
wjones127, line 91: This seems like a bug
```suggestion
```

wjones127: That's the source of the test failure.

wjones127 (changes requested):

wjones127 (approved):

Issue #1872, opened 2024-11-22
Feature: change default from post-filtering to pre-filtering


wjones127:
### SDK

Python

### Description

Python sync API uses post-filting by default, but other SDKS (async Python, Node, Rust) use prefiltering. Pre-filtering is what users want most of the time, and having different default across SDKs can be confusing. We can either do this before, or after we start transitioning APIs to the Python async-sync consolidation.

westonpace:
I agree that prefiltering is a better default.

PR #2000, opened 2025-01-07
feat!: change default from postfiltering to prefiltering for sync python

albertlockett:
BREAKING CHANGE: prefiltering is now the default in the synchronous python SDK

resolves: #1872

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -762,7 +762,7 @@ class LanceVectorQueryBuilder(LanceQueryBuilder):
 
         return result_set
 
-    def where(self, where: str, prefilter: bool = False) -> LanceVectorQueryBuilder:
+    def where(self, where: str, prefilter: bool = True) -> LanceVectorQueryBuilder:
         """Set the where clause.
 
         Parameters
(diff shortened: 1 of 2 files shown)


westonpace (commented):
It's a little worrisome this didn't break any tests.  Can you add a very basic prefilter / postfiltering test that would break without this change?

westonpace:
(though maybe it's just that it's hard to detect prefiltering vs. postfiltering without a vector index and we don't have a lot of vector index filtered search tests)

albertlockett:
> It's a little worrisome this didn't break any tests. Can you add a very basic prefilter / postfiltering test that would break without this change?
> (though maybe it's just that it's hard to detect prefiltering vs. postfiltering without a vector index and we don't have a lot of vector index filtered search tests)

Absolutely I can

albertlockett:
@westonpace would you mind having another look at this? Turns out I had made the change in the wrong query builder, and after making the change in the LancedbVectorQueryBuilder, we see the impact on the tests we had expected.

westonpace (approved):
Awesome!  Thanks for tackling this.  Let's just make sure we announce the breaking change clearly in release notes.

I suppose we could do a deprecation cycle (one release where we say "you aren't specifying a value, the default will change, if you want the old behavior change it to this") but I think it would probably be overkill.

PR #2003, opened 2025-01-08
chore: add deprecation warning to vectordb

wjones127:


westonpace (approved):
Maybe we're not there yet but it is possible to deprecate the package on NPM as well: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

wjones127:
> Maybe we're not there yet but it is possible to deprecate the package on NPM as well: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

Yup, that's my next step

PR #2004, opened 2025-01-09
feat: expose dataset config

albertlockett:
Expose methods on NativeTable for updating schema metadata and dataset config & getting the dataset config via the manifest.


wjones127 (approved):
Looks good!

Issue #1975, opened 2024-12-24
LanceDB Typescript Windows Arm release fails with linking error


wjones127:
Example run:

https://github.com/lancedb/lancedb/actions/runs/12485469407/job/34844290690?pr=1974

We will disable this for now.

ahaapple:
lancedb/lancedb  0.14.1 Installation failed:

```
    "@lancedb/lancedb": "0.14.1",
```
```
[0.22ms] ".env"
bun install v1.1.4 (fbe2fe0c)
error: @lancedb/lancedb-win32-arm64-msvc@0.14.1 failed to resolve
```


Change the @lancedb/lancedb version to 0.14.0. bun install successfully.

PR #2006, opened 2025-01-09
ci: try to fix windows arm build

wjones127:
I've worked to reproduce the same error we had in CI in a local docker file.

I've fixed the issue with the missing dbghelp.lib, but not it's asking for `msvcrt.lib`:

```
lld: error: could not open 'msvcrt.lib': No such file or directory
```

Trying to fix [#1975](https://github.com/lancedb/lancedb/issues/1975)


FuPeiJiang:
@wjones127
`msvcrt.lib` is found in `\VC\Tools\MSVC\14.41.34117.0\lib\x64\msvcrt.lib`, but I don't see an `arm64` version on my installation

I'll try to add that to the ci if that's the correct fix, I'm not sure if the current build is using MSVCRT or UCRT, I'll check which commit added this dependency

FuPeiJiang:
`Contents/VC/Tools/MSVC/14.16.27023/lib/arm64/msvcrt.lib`
and
`Contents/VC/Tools/MSVC/14.16.27023/lib/arm64/store/msvcrt.lib`
are located in:
"Microsoft.VisualC.14.16.CRT.ARM64.Store.vsix"

"https://download.visualstudio.microsoft.com/download/pr/bac0afd7-cc9e-4182-8a83-9898fa20e092/2038def3bd4551ece3789ddc0d11f38347126a4219c7c5681a9f9a4328d51d60/Microsoft.VisualC.14.16.CRT.ARM64.Store.vsix"

but I don't think `msvcrt.lib` is needed:
I've successfully built f835ff310f020968b914b6b90122aec4627ba6e5 by only adding `dbghelp.lib`
I'll check for `main` now

FuPeiJiang:
@wjones127
> I've worked to reproduce the same error we had in CI in a local docker file.

why not use https://github.com/nektos/act ?
it does the docker config automatically for you (it uses docker to run the workflows)
this is what I've been using: `act -W '.github/workflows/npm-publish.yml'`

the best part is that if it fails, you can `docker ps` then `docker exec -it <CONTAINER ID> sh` into it, effectively getting an `ssh` with the paused state

PR #2007, opened 2025-01-09
test: fix test failure from merge

wjones127:


albertlockett (approved):

python/python/tests/test_query.py
```diff
@@ -449,10 +449,10 @@ async def test_query_to_pandas_flatten_async(table_struct_async: AsyncTable):
 @pytest.mark.asyncio
 async def test_query_to_polars_async(table_async: AsyncTable):
     df = await table_async.query().to_polars()
-    assert df.shape == (2, 4)
+    assert df.shape == (2, 5)
 
     df = await table_async.query().where("id < 0").to_polars()
-    assert df.shape == (0, 4)
+    assert df.shape == (0, 5)
```
westonpace, line 455: Why is this change needed?  Were we just not running `asyncio` tests before?  Or did something change in behavior?

wjones127: Someone added a column to the test data in a PR at the same time this test was added. So I think it just slipped through because they were both merged around the same time, and we aren't strict about rebasing and rerunning CI after each merge.

PR #2013, opened 2025-01-10
fix: broken voyageai embedding API

PrashantDixit0:
This PR fixes the broken Embedding API for Voyageai.

### DIFF
--- a/python/python/lancedb/embeddings/voyageai.py
+++ b/python/python/lancedb/embeddings/voyageai.py
@@ -59,7 +59,7 @@ class VoyageAIEmbeddingFunction(EmbeddingFunction):
         .create(name="voyage-3")
 
     class TextModel(LanceModel):
-        data: str = voyageai.SourceField()
+        text: str = voyageai.SourceField()
         vector: Vector(voyageai.ndims()) =  voyageai.VectorField()
 
     data = [ { "text": "hello world" },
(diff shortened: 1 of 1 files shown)


python/python/lancedb/embeddings/voyageai.py
```diff
@@ -115,14 +115,21 @@ def generate_text_embeddings(self, text: str, **kwargs) -> np.ndarray:
 
         truncation: Optional[bool]
         """
-        if self.name in ["voyage-multimodal-3"]:
-            rs = VoyageAIEmbeddingFunction._get_client().multimodal_embed(
-                inputs=[[text]], model=self.name, **kwargs
-            )
-        else:
+
+        if self.name in [
```
QianZhu, line 119: can you put model names as a constant

python/python/lancedb/embeddings/voyageai.py
```diff
@@ -115,14 +115,21 @@ def generate_text_embeddings(self, text: str, **kwargs) -> np.ndarray:
...
+            "voyage-3",
+            "voyage-3-lite",
+            "voyage-finance-2",
+            "voyage-law-2",
+            "voyage-code-2",
+        ]:
             rs = VoyageAIEmbeddingFunction._get_client().embed(
                 texts=[text], model=self.name, **kwargs
             )
+        elif self.name in ["voyage-multimodal-3"]:
```
QianZhu, line 129: same here, use a constant

python/python/lancedb/embeddings/voyageai.py
```diff
@@ -115,14 +115,21 @@ def generate_text_embeddings(self, text: str, **kwargs) -> np.ndarray:
...
+            rs = VoyageAIEmbeddingFunction._get_client().multimodal_embed(
+                inputs=[[text]], model=self.name, **kwargs
+            )
```
QianZhu, line 133: what happens if the model is unsupported? raise an exception?

python/python/lancedb/embeddings/voyageai.py
```diff
@@ -115,14 +115,21 @@ def generate_text_embeddings(self, text: str, **kwargs) -> np.ndarray:
...
-        else:
+
+        if self.name in [
+            "voyage-3",
+            "voyage-3-lite",
+            "voyage-finance-2",
+            "voyage-law-2",
+            "voyage-code-2",
+        ]:
             rs = VoyageAIEmbeddingFunction._get_client().embed(
```
QianZhu, line 126: can we do `client = VoyageAIEmbeddingFunction._get_client()`, then this and line 130 would be `client.embed(...)`

PrashantDixit0: Done

python/python/lancedb/embeddings/voyageai.py
PrashantDixit0: Added multimodal supported models as class constants

python/python/lancedb/embeddings/voyageai.py
PrashantDixit0: Done

QianZhu (approved):

PR #2021, opened 2025-01-14
feat: add get dataset method on NativeTable

Akagi201:
I want to public the dataset method from native table, then I can use more lance method like order_by which is not exposed in the lancedb crate.


wjones127 (approved):

Akagi201:
@wjones127 I have rebased my code, you can check again.

Issue #1666, opened 2024-09-18
Design a more coherent index API


wjones127:
There's some design work we need to do on indices. We don't represent them in our APIs very coherently. There's sort of three levels we can imagine:

1. A configuration for an index ("This is an IVF_PQ index")
2. An index on a column ("These are the index files for an IVF_PQ index")
3. A piece / segment of an index ("This is one index file for an IVF_PQ index covering fragments ...")

In most databases, users kind of expect they are creating 1. That's why they can do (1) create table, (2) create index, then (3) insert data. They are confused when the need to insert data before creating an index.

What we have right now is 3. This relates to the whole UUID confusion we had earlier, where not only does the UUID change as you incrementally update the index. But also you can have multiple UUIDs for the same index name! That's because in Lance an "index" really refers to a piece of an index.

LuQQiu:
So list_indices with same name are showing the **Index files, unique by UUID** instead  of top-level indices?

wjones127:
> So list_indices with same name are showing the **Index files, unique by UUID** instead of top-level indices?

yes, list indices shows files. There isn't a proper top-level index concept in Lance yet. So the "duplicates" you see are just different pieces of an index.

rpgreen:
We may also want to consider exposing index status and async jobs as part of this model.

PR #2023, opened 2025-01-14
feat: add index version to list_indices

LuQQiu:
Solve the case that list_indices show three exactly same index and confuse users

### DIFF
--- a/rust/lancedb/src/index.rs
+++ b/rust/lancedb/src/index.rs
@@ -177,6 +177,8 @@ pub struct IndexConfig {
     /// Currently this is always a Vec of size 1.  In the future there may
     /// be more columns to represent composite indices.
     pub columns: Vec<String>,
+    /// The version of the index
+    pub version: u64,
 }
 
 #[skip_serializing_none]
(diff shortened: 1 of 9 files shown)


rust/lancedb/src/index.rs
```diff
@@ -177,6 +177,8 @@ pub struct IndexConfig {
     /// Currently this is always a Vec of size 1.  In the future there may
     /// be more columns to represent composite indices.
     pub columns: Vec<String>,
+    /// The version of the index
+    pub version: u64,
```
wjones127, line 181: What is an index's "version"? Where does that come from?

Is this derived from `dataset_version` or something else?

https://github.com/lancedb/lance/blob/b572905d22b1f168fccb2768c211642ec894bff3/rust/lance-table/src/format/index.rs#L26-L27

wjones127 (changes requested):

wjones127:
I think the confusion you are describing stems from this topic: https://github.com/lancedb/lancedb/issues/1666#issue-2534538525

plus the fact that the index statistics are lies right now: https://github.com/lancedb/lance/issues/3374

rust/lancedb/src/index.rs
LuQQiu: yes, it's dataset_version

LuQQiu:
Thanks @wjones127, looks like require a bigger change as you mention https://github.com/lancedb/lancedb/issues/1666#issue-2534538525 to really show top level index information,

PR #2025, opened 2025-01-15
docs: add missing example code

BubbleCal:


python/python/tests/docs/test_multivector.py
```diff
@@ -0,0 +1,70 @@
...
+shutil.rmtree("data/multivector_demo", ignore_errors=True)
+
+
+def test_multivector():
+    # --8<-- [start:sync_multivector]
+    db = lancedb.connect("data/multivector_demo")
+    schema = pa.schema(
+        [
+            pa.field("id", pa.int64()),
+            pa.field("vector", pa.list_(pa.list_(pa.float32(), 256))),
```
eddyxu, line 19: let's add commment here what data type we accept for multi vector

eddyxu (approved):
Minor comment

python/python/tests/docs/test_multivector.py
QianZhu: add it to our docs? @BubbleCal

BubbleCal: will file a new PR to do this after upgrading lance, need that because now it supports only float32

PR #2026, opened 2025-01-15
fix: can't infer the vector column for multivector

BubbleCal:


rust/lancedb/src/utils.rs
```diff
@@ -188,6 +188,20 @@ pub fn supported_vector_data_type(dtype: &DataType) -> bool {
     }
 }
 
+// TODO: remove this after we expose the same function in Lance.
+pub fn infer_vector_dim(data_type: &DataType) -> Result<usize> {
+    match data_type {
+        DataType::FixedSizeList(_, dim) => Ok(*dim as usize),
+        DataType::List(inner) => infer_vector_dim(inner.data_type()),
```
westonpace, line 195: Technically I think this would succeed on `List<List<List<FixedSizeList<Float32, 128>>>>` but I don't know how likely that is.  Maybe pass a bool into `infer_vector_dim` that marks if this is the first time we've called it?

westonpace (approved):

rust/lancedb/src/utils.rs
BubbleCal: fixed

PR #2034, opened 2025-01-16
feat: add custom object_store in ConnectBuilder

zensh:
Check out this example of using LanceDB's custom object_store feature to connect with the Object Store canister on the ICP blockchain:  
https://github.com/ldclabs/anda/blob/main/anda_lancedb/src/knowledge.rs#L239

### DIFF
--- a/rust/lancedb/src/connection.rs
+++ b/rust/lancedb/src/connection.rs
@@ -23,7 +23,7 @@ use arrow_array::{RecordBatchIterator, RecordBatchReader};
 use arrow_schema::SchemaRef;
 use lance::dataset::{ReadParams, WriteMode};
 use lance::io::{ObjectStore, ObjectStoreParams, ObjectStoreRegistry, WrappingObjectStore};
-use object_store::{aws::AwsCredential, local::LocalFileSystem};
+use object_store::{aws::AwsCredential, local::LocalFileSystem, DynObjectStore};
 use snafu::prelude::*;
 
 use crate::arrow::IntoArrow;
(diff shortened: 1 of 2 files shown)


rust/lancedb/src/connection.rs
```diff
@@ -873,9 +881,19 @@ impl Database {
...
                 let registry = Arc::new(ObjectStoreRegistry::default());
                 let storage_options = options.storage_options.clone();
-                let os_params = ObjectStoreParams {
-                    storage_options: Some(storage_options.clone()),
-                    ..Default::default()
+                let os_params = if let Some(os) = &options.object_store {
+                    ObjectStoreParams {
+                        storage_options: Some(storage_options.clone()),
+                        object_store: Some((os.0.clone(), url)),
+                        use_constant_size_upload_parts: os.1.is_some(),
```
LuQQiu, line 888: is it a specific configuration for some objectstore? why need explicitly take out here?

rust/lancedb/src/connection.rs
```diff
@@ -873,9 +881,19 @@ impl Database {
...
+                        block_size: os.1,
```
wjones127, line 889: Sorry for the confusion. `block_size` is for reads. It determines when range requests are coalesced.

`use_constant_size_upload_parts` is for writes. It determines whether multipart uploads only use fixed-size parts, or variable size parts (higher max object size).

These two shouldn't be controlled by the same parameter.

wjones127 (commented):
Thanks for making this. TBH I think this might belong in the `lance` library, not LanceDB. And perhaps we should make some sort of builder for object store that makes it easier to set all the additional parameters.

zensh:
@LuQQiu @wjones127 

In our use case, to achieve data encryption, data must be saved using a fixed block_size, as seen here:
https://github.com/ldclabs/ic-cose/blob/main/src/ic_object_store/src/client.rs#L118
For reading data, matching the block_size will be the most efficient.

This PR is a minimal implementation to meet our needs. The lancedb codebase is quite large and complex, and this is all I can manage.

Issue #2014, opened 2025-01-11
bug(python): Hybrid search `inverted_fts_score` passed as `score1` in `self._combine_score`


JazJaz426:
### LanceDB version

v0.17.1b4

### What happened?

Hi,

In `linear_combination.py`-`merge_result` function, I noticed when combining scores, the vector search's weight doesn't always get assigned to the corresponding score. 
```
elif vector_list[i]["_rowid"] < fts_list[j]["_rowid"]:
124  ->	                vi["_relevance_score"] = self._combine_score(vi["_distance"], fill)
125  	                combined_list.append(vi)
126  	                i += 1
127  	            else:
128  	                fj["_relevance_score"] = self._combine_score(inverted_fts_score, fill)
129  	                combined_list.append(fj)
130  	                j += 1
```
```
def _combine_score(self, score1, score2):
152  	        # these scores represent distance
153  	        return 1 - (self.weight * score1 + (1 - self.weight) * score2)
```

Here's the real example I've stepped through in pdb. For example, I specified vector search weight = 0.7 as I wanted to give more priority to the semantic search result. When stepped through the pair below

```
vi = {'inner_id': 206, '_distance': 0.41053229570388794, '_rowid': 204} 
fj = {'inner_id': 138, '_score': 0.1949227750301361, '_rowid': 136}
```
Since rowid 136<204, we calculate relevance score for fj first. Since fj or inner id 138 is only returned by FTS but not vector search so the distance for vector search side is default to fill (i.e., 1) while the distance for FTS score is inverted_fts_score (i.e., 1-0.19=0.81). However, in the following combine_score step, it doesn't make sense to me why inverted_fts_score is passed as score 1 where 0.7 the vector search weight is applied to. Shouldn't score 1 always be the score from vector search side (either the actual distance or refill = 1)?

Internally we found our linear reranker weight doesn't work as expected so I wonder if this could be the cause. Would be great if someone can clarify this logic a bit!

LuQQiu:
@BubbleCal  FYI

BubbleCal:
@JazJaz426 
this should be fixed #2035

PR #2035, opened 2025-01-16
fix: linear reranker applies wrong score to combine

BubbleCal:
related to #2014 
this fixes:
- linear reranker may lost some results if the merging consumes all vector results earlier than fts results
- linear reranker inverts the fts score but only vector distance can be inverted

### DIFF
--- a/python/python/lancedb/rerankers/linear_combination.py
+++ b/python/python/lancedb/rerankers/linear_combination.py
@@ -109,14 +109,9 @@ class LinearCombinationReranker(Reranker):
 
         combined_list = []
         for row_id, result in results.items():
-            if "_distance" not in result:
-                result["_distance"] = fill
-            if "_score" not in result:
-                result["_score"] = fill
-            vector_score = self._invert_score(result["_distance"])
-            result["_relevance_score"] = self._combine_score(
-                vector_score, result["_score"]
-            )
+            vector_score = self._invert_score(result.get("_distance", fill))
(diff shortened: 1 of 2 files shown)


python/python/lancedb/rerankers/linear_combination.py
```diff
@@ -95,43 +95,24 @@ def merge_results(
...
+        for vector_result in vector_results.to_pylist():
+            row_id = vector_result["_rowid"]
+            if row_id not in results:
+                results[row_id] = dict()
+            results[vector_result["_rowid"]]["_distance"] = vector_result["_distance"]
+        for fts_result in fts_results.to_pylist():
+            row_id = fts_result["_rowid"]
+            if row_id not in results:
+                results[row_id] = dict()
+            results[fts_result["_rowid"]]["_score"] = fts_result["_score"]
```
wjones127, line 108: I'd recommend [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) here:
```suggestion
        results = defaultdict(dict)
        for vector_result in vector_results.to_pylist():
            row_id = vector_result["_rowid"]
            results[row_id]["_distance"] = vector_result["_distance"]
        for fts_result in fts_results.to_pylist():
            row_id = fts_result["_rowid"]
            results[row_id]["_score"] = fts_result["_score"]
```

wjones127 (approved):

PR #2038, opened 2025-01-16
chore: upgrade lance to v0.22.1-beta.3

albertlockett:


chebbyChefNEQ (approved):
we need to bump the python dep too https://github.com/lancedb/lancedb/blob/main/python/pyproject.toml#L7

PR #2039, opened 2025-01-16
feat: add `drop_index()` method

### LINKED ISSUE #1665: Add delete_index API
Right now you can only add or update indices. You can't delete them. But in some cases we might no longer want them.

wjones127:
Closes #1665

### DIFF
--- a/docs/src/js/classes/VectorQuery.md
+++ b/docs/src/js/classes/VectorQuery.md
@@ -128,6 +128,24 @@ whose data type is a fixed-size-list of floats.
 
 ***
 
+### distanceRange()
+
+```ts
+distanceRange(lowerBound?, upperBound?): VectorQuery
+```
+
+#### Parameters
+
+* **lowerBound?**: `number`
+
+* **upperBound?**: `number`
+
+#### Returns
+
+[`VectorQuery`](VectorQuery.md)
+
+***
+
 ### distanceType()
 
 ```ts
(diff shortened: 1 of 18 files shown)


docs/src/js/classes/VectorQuery.md
```diff
@@ -128,6 +128,24 @@ whose data type is a fixed-size-list of floats.
...
+* **lowerBound?**: `number`
+
+* **upperBound?**: `number`
+
+#### Returns
+
+[`VectorQuery`](VectorQuery.md)
+
+***
+
```
westonpace, line 148: Not related to the PR but this method is under-documented.

docs/src/js/classes/VectorQuery.md
```diff
@@ -528,6 +546,22 @@ distance between the query vector and the actual uncompressed vector.
...
+#### Parameters
+
+* **reranker**: [`Reranker`](../namespaces/rerankers/interfaces/Reranker.md)
+
+#### Returns
+
+[`VectorQuery`](VectorQuery.md)
+
+***
+
```
westonpace, line 564: Also underdocumented

nodejs/lancedb/table.ts
```diff
@@ -226,6 +226,17 @@ export abstract class Table {
...
   ): Promise<void>;
+
+  /**
+   * Drop an index from the table.
+   *
+   * @param name The name of the index.
+   *
+   * @note This does not delete the index from disk, it just removes it from the table.
+   * To delete the index, run {@link Table#optimize} after dropping the index.
+   */
```
westonpace, line 237: In `pylance` I added a note about how to find the `name` of the index.  Do you think we want that here?  I agree the note on deletion is a good addition.

westonpace (approved):
Thanks!  Only nit is really unrelated to this PR.  All the reranking stuff seems to be missing docstrings in typescript.  Feel free to just create an issue instead of fixing here.

nodejs/lancedb/table.ts
wjones127: Ah that would probably be good to include. I will add it.

PR #2044, opened 2025-01-20
docs: pandas example

vincent0426:
Fix example for section ## From pandas DataFrame


westonpace (approved):
Thanks, I ran through the changes manually and this works.

Kind of embarrassing we weren't actually using pandas here :laughing:

PR #2045, opened 2025-01-20
feat: support multivector on remote table

BubbleCal:


rust/lancedb/src/remote/table.rs
```diff
@@ -457,28 +460,12 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
...
-        if streams.len() == 1 {
-            let stream = streams.into_iter().next().unwrap();
-            Ok(Arc::new(OneShotExec::new(stream)))
-        } else {
-            let stream_execs = streams
-                .into_iter()
-                .map(|stream| Arc::new(OneShotExec::new(stream)) as Arc<dyn ExecutionPlan>)
-                .collect();
-            Table::multi_vector_plan(stream_execs)
-        }
```
westonpace, line 481: What was this old code?  Are we changing the way we support multi-vector?  Or was this just a guess and we never actually supported multi-vector?

rust/lancedb/src/remote/table.rs
```diff
@@ -457,28 +460,12 @@ impl<S: HttpSend> TableInternal for RemoteTable<S> {
...
-        let streams = futures::future::try_join_all(futures).await?;
-        if streams.len() == 1 {
-            let stream = streams.into_iter().next().unwrap();
-            Ok(Arc::new(OneShotExec::new(stream)))
-        } else {
-            let stream_execs = streams
-                .into_iter()
-                .map(|stream| Arc::new(OneShotExec::new(stream)) as Arc<dyn ExecutionPlan>)
-                .collect();
-            Table::multi_vector_plan(stream_execs)
```
westonpace, line 480: Can we get rid of the `multi_vector_plan` method too?  Or is it used somewhere else?

westonpace (approved):
Just a few questions but this seems fine to me

rust/lancedb/src/remote/table.rs
BubbleCal: it's still used for native table

rust/lancedb/src/remote/table.rs
BubbleCal: before this, batch queries for remote table is handled at SDK side, it sends multiple requests to the server.
now this just sends all vectors to the server in one request and let the server to handle all the cases.

PR #2049, opened 2025-01-20
feat(python): add search() method to async API

### LINKED ISSUE #1938: feature: add `search()` method in Async Python
Make it similar to the sync API's `search()` method.

wjones127:
Reviving #1966.

Closes #1938

The `search()` method can apply embeddings for the user. This simplifies hybrid search, so instead of writing:

```python
vector_query = embeddings.compute_query_embeddings("flower moon")[0]
await (
    async_tbl.query()
    .nearest_to(vector_query)
    .nearest_to_text("flower moon")
    .to_pandas()
)
```

You can write:

```python
await (await async_tbl.search("flower moon", query_type="hybrid")).to_pandas()
```

Unfortunately, we had to do a double-await here because `search()` needs to be async. This is because it often needs to do IO to retrieve and run an embedding function.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -30,6 +30,7 @@ from .dependencies import _check_for_pandas
 import pyarrow as pa
 import pyarrow.compute as pc
 import pyarrow.fs as pa_fs
+import numpy as np
 from lance import LanceDataset
 from lance.dependencies import _check_for_hugging_face
 
@@ -2944,7 +2945,7 @@ class AsyncTable:
 
         return LanceMergeInsertBuilder(self, on)
 
-    def search(
+    async def search(
         self,
         query: Optional[Union[VEC, str, "PIL.Image.Image", Tuple]] = None,
         vector_column_name: Optional[str] = None,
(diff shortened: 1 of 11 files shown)


python/python/lancedb/table.py
```diff
@@ -2931,6 +2948,236 @@ def merge_insert(self, on: Union[str, Iterable[str]]) -> LanceMergeInsertBuilder
...
+                loop = asyncio.get_running_loop()
+                # This function is likely to block, since it either calls an expensive
+                # function or makes an HTTP request to an embeddings REST API.
+                return (
+                    await loop.run_in_executor(
+                        None,
+                        embedding.function.compute_query_embeddings_with_retry,
+                        query,
+                    )
+                )[0]
```
westonpace, line 3112: I guess at some point in the future we can support async embedding functions?

python/python/lancedb/table.py
```diff
@@ -2931,6 +2948,236 @@ def merge_insert(self, on: Union[str, Iterable[str]]) -> LanceMergeInsertBuilder
...
+                return None
+
+        if query_type == "auto":
+            # Infer the query type.
+            if is_embedding(query):
+                vector_query = query
+                query_type = "vector"
+            elif isinstance(query, str):
+                try:
+                    indices, (vector_column_name, conf) = await asyncio.gather(
```
westonpace, line 3123: What is `conf` here?  The embedding function?

python/python/tests/test_query.py
```diff
@@ -651,3 +661,98 @@ async def test_query_with_f16(tmp_path: Path):
...
+    )
+    table = await mem_db_async.create_table("test", data)
+    await table.create_index("text", config=FTS())
+
+    # Validate that specifying fts, vector or hybrid gets the right query.
+    q = [0.1] * ndims
+    query = await table.search(q, query_type="vector")
+    assert isinstance(query, AsyncVectorQuery)
+
+    query = await table.search("0.1", query_type="fts")
```
westonpace, line 748: Maybe an error case for `query_type="blah"`?

westonpace (approved):
Some minor questions / comments but looks good

python/python/lancedb/table.py
```diff
@@ -2931,6 +2948,236 @@ def merge_insert(self, on: Union[str, Iterable[str]]) -> LanceMergeInsertBuilder
...
+        vector_column_name: Optional[str] = None,
+        query_type: Literal["auto"] = ...,
+        ordering_field_name: Optional[str] = None,
+        fts_columns: Optional[Union[str, List[str]]] = None,
+    ) -> AsyncVectorQuery: ...
+
+    @overload
+    async def search(
+        self,
+        query: Optional[Union[VEC, str, "PIL.Image.Image", Tuple]] = None,
```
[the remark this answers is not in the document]
wjones127: This overload is more trying to capture the idea that if you set `query_type="fts"` it will be FTS query. Though you are right that it's not valid with non str `query` values, and we could have python type check that.

python/python/lancedb/table.py
```diff
@@ -2931,6 +2948,236 @@ def merge_insert(self, on: Union[str, Iterable[str]]) -> LanceMergeInsertBuilder
...
+
+        Returns
+        -------
+        LanceQueryBuilder
+            A query builder object representing the query.
+            Once executed, the query returns
+
+            - selected columns
+
+            - the vector
```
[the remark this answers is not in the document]
wjones127: Hmm this is copied from the sync API. I'll remove the out-dated parts and then we can clean this up in follow up PRs.

python/python/lancedb/table.py
wjones127: Yeah I think that would make sense. Especially since a lot of them call out to APIs.

python/python/lancedb/table.py
wjones127: Yeah, the embedding function config.

python/python/lancedb/table.py
```diff
@@ -2931,6 +2948,236 @@ def merge_insert(self, on: Union[str, Iterable[str]]) -> LanceMergeInsertBuilder
...
+                except ValueError as e:
+                    if "Column" in str(
+                        e
+                    ) and "has no registered embedding function" in str(e):
+                        # If the column has no registered embedding function,
+                        # then it's an FTS query.
+                        query_type = "fts"
+                    else:
+                        raise e
+                else:
```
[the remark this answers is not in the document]
wjones127: `else` is attached to the try. It's what to do if there is no exception.

PR #2050, opened 2025-01-21
VoyageAI multimodal embedding documentation/example

fzowl:
VoyageAI multimodal embedding documentation/example


docs/src/embeddings/available_embedding_models/multimodal_embedding_functions/voyageai_multimodal_embedding.md
```diff
@@ -0,0 +1,97 @@
...
+
+Because we're using a multi-modal embedding function, we can also search using images
+
+```python
+# image search
+query_image_uri = "http://farm1.staticflickr.com/200/467715466_ed4a31801f_z.jpg"
+image_bytes = requests.get(query_image_uri).content
+query_image = Image.open(BytesIO(image_bytes))
+actual = table.search(query_image, "vec_from_bytes").limit(1).to_pydantic(Images)[0]
+print(actual.label == "dog")
```
westonpace, line 87:
```suggestion
print(actual.label) # "dog"
```

For consistency with the above example

docs/src/embeddings/available_embedding_models/multimodal_embedding_functions/voyageai_multimodal_embedding.md
```diff
@@ -0,0 +1,97 @@
...
+
+# image search using a custom vector column
+other = (
+    table.search(query_image, vector_column_name="vec_from_bytes")
+    .limit(1)
+    .to_pydantic(Images)[0]
+)
```
westonpace, line 94: How is this different than the above?  You have to list `vec_from_bytes` in both cases

docs/src/embeddings/available_embedding_models/multimodal_embedding_functions/voyageai_multimodal_embedding.md
```diff
@@ -0,0 +1,97 @@
...
+os.environ['VOYAGE_API_KEY'] = 'YOUR_VOYAGE_API_KEY'
+
+db = lancedb.connect(".lancedb")
+func = get_registry().get("voyageai").create(name="voyage-multimodal-3")
+
+
+def image_to_base64(image_bytes: bytes):
+    buffered = BytesIO(image_bytes)
+    img_str = base64.b64encode(buffered.getvalue())
+    return img_str.decode("utf-8")
```
westonpace, line 34: Why base64 encode the bytes?  Is that something the embedding function expects?

I would expect the column to be a regular `binary` column and, if the remote API wanted base64 encoded bytes, then the embedding function would do the base64 work.  This way, if the user wants to fetch an image out of the database later, they don't have to base64 decode it.  It also keeps us from storing the less efficient representation on disk.

docs/src/embeddings/available_embedding_models/multimodal_embedding_functions/voyageai_multimodal_embedding.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+class Images(LanceModel):
+    label: str
+    image_uri: str = func.SourceField()  # image uri as the source
+    image_bytes: str = func.SourceField()  # image bytes base64 encoded as the source
+    vector: Vector(func.ndims()) = func.VectorField()  # vector column
+    vec_from_bytes: Vector(func.ndims()) = func.VectorField()  # Another vector column
```
westonpace, line 42: I think I see what you're after but it can be a little confusing with both `image_uri` and `image_bytes` on one table.  In real world cases you would pick one approach or the other right?  It might be clearer to do everything using `bytes` and then make a completely separate example below that shows you can also use URIs.

westonpace (commented):
I haven't actually run the examples but scanned through the code and have a few suggestions.

PR #2054, opened 2025-01-22
feat!: split lancedb rust crate into multiple crates

westonpace:
This is partly just a necessary step as lancedb continues to grow and partly because I want to be able to break out some logic for catalogs.  I can probably achieve the latter without splitting things into multiple crates if there is strong opposition.

This PR should make very little functional change.  Also, it shouldn't be changing the public API much because I added re-exports to restore the original API as much as possible.

A few things that were `pub(crate)` have now become `pub` so some things were added.

BREAKING CHANGE: The `NoData` struct became an internal detail so at least one API was removed.  Also, moving something to a different crate and re-export it is still technically a breaking change so I think we will want to consider this a breaking change.


westonpace:
After some offline discussion I am abandoning this effort.  At the moment the crate is not complex enough to really need this.  There are also unlikely to be much compiler benefits since the core crate already needs arrow, lance, and DF.

Issue #2055, opened 2025-01-22
bug(python): table.merge_insert does not accept a list of columns [duplicated #1303]


dantetemplar:
### What happened?

The very same problem as in falsy closed issue #1303

dantetemplar:
I've attached a PR

Xuanwo:
Hi, this issue is a duplicate of #1303. The current code already accepts `Iterable[str]` for the `on` parameter in `merge_insert`, and there is no `.iter()` call that causes the AttributeError. I'm closing this now. Feel free to create a new issue if the problem still exists.

PR #2056, opened 2025-01-22
fix(python): use iter() function instead of .iter() method (#2055)

dantetemplar:


dantetemplar:
Solves: #2055

wjones127:
Hi @dantetemplar . Thanks for attempting a fix here.

In general, we don't accept any fixes without any unit tests.

And I think if you do try to test this, you'll find that the underlying implementation doesn't even support multiple keys yet: https://github.com/lancedb/lance/issues/3124

If you want, you can make change this PR to provide a more useful error.

Or you are welcome to create a PR upstream to support multiple keys in Lance.

dantetemplar:
> Hi @dantetemplar . Thanks for attempting a fix here.
> 
> In general, we don't accept any fixes without any unit tests.
> 
> And I think if you do try to test this, you'll find that the underlying implementation doesn't even support multiple keys yet: [lancedb/lance#3124](https://github.com/lancedb/lance/issues/3124)
> 
> If you want, you can make change this PR to provide a more useful error.
> 
> Or you are welcome to create a PR upstream to support multiple keys in Lance.

Okay, the type-hint and .item should be removed then

dantetemplar:
i'm not the only one who said it :3

PR #2065, opened 2025-01-23
fix(node): `createTable()` should save embeddings, and `mergeInsert` should use them

### LINKED ISSUE #2066: bug(node): mergeInsert and add don't seem to have access to embedding functions.
### LanceDB version

_No response_

### What happened?

See failing test in https://github.com/lancedb/lancedb/pull/2065

_No response_

wjones127:
* `createTable()` now saves embeddings in the schema metadata. Previously, it would drop them. (`createEmptyTable()` was already tested and worked.)
* `mergeInsert()` now uses embeddings.

Fixes #2066

### DIFF
--- a/nodejs/lancedb/merge.ts
+++ b/nodejs/lancedb/merge.ts
@@ -1,13 +1,20 @@
-import { Data, fromDataToBuffer } from "./arrow";
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+import { Data, Schema, fromDataToBuffer } from "./arrow";
 import { NativeMergeInsertBuilder } from "./native";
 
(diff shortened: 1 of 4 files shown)


nodejs/lancedb/merge.ts
```diff
@@ -1,13 +1,20 @@
-import { Data, fromDataToBuffer } from "./arrow";
+// SPDX-License-Identifier: Apache-2.0
```
westonpace, line 1: Hmmm, are we missing SPDX checks?

westonpace (approved):

nodejs/lancedb/merge.ts
wjones127: I don't see any license check in our CI jobs. I can copy over the Lance one later today.

PR #2067, opened 2025-01-24
feat!: refactor ConnectionInternal into a Database trait

westonpace:
This opens up the door for more custom database implementations than the two we have today.  The biggest change should be inivisble:  `ConnectionInternal` has been renamed to `Database`, made public, and refactored

However, there are a few breaking changes.  `data_storage_version` and `enable_v2_manifest_paths` have been moved from options on `create_table` to options for the database which are now set via `storage_options`.

Before:
```
db = connect(uri)
tbl = db.create_table("my_table", data, data_storage_version="legacy", enable_v2_manifest_paths=True)
```

After:
```
db = connect(uri, storage_options={
  "new_table_enable_v2_manifest_paths": "true",
  "new_table_data_storage_version": "legacy"
})
tbl = db.create_table("my_table", data)
```

BREAKING CHANGE: the data_storage_version, enable_v2_manifest_paths options have moved from options to create_table to storage_options.
BREAKING CHANGE: the use_legacy_format option has been removed, data_storage_version has replaced it for some time now

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2036,6 +2036,8 @@ class LanceTable(Table):
         embedding_functions: Optional[List[EmbeddingFunctionConfig]] = None,
         *,
         storage_options: Optional[Dict[str, str]] = None,
+        data_storage_version: Optional[str] = None,
+        enable_v2_manifest_paths: Optional[bool] = None,
     ):
         """
         Create a new table.
@@ -2083,9 +2085,36 @@ class LanceTable(Table):
             The value to use when filling vectors. Only used if on_bad_vectors="fill".
         embedding_functions: list of EmbeddingFunctionModel, default None
             The embedding functions to use when creating the table.
+        data_storage_version: optional, str, default "stable"
+            Deprecated.  Set `storage_options` when connecting to the database and set
+            `new_table_data_storage_version` in the options.
+        enable_v2_manifest_paths: optional, bool, default False
+            Deprecated.  Set `storage_options` when connecting to the database and set
+            `new_table_enable_v2_manifest_paths` in the options.
         """
         self = cls.__new__(cls)
         self._conn = db
+
+        if data_storage_version is not None:
+            warnings.warn(
+                "setting data_storage_version directly on create_table is deprecated. ",
+                "Use database_options instead.",
+                DeprecationWarning,
+            )
+            if storage_options is None:
+                storage_options = {}
+            storage_options["new_table_data_storage_version"] = data_storage_version
+        if enable_v2_manifest_paths is not None:
+            warnings.warn(
+                "setting enable_v2_manifest_paths directly on create_table is deprecated. ",
+                "Use database_options instead.",
+                DeprecationWarning,
+            )
+            if storage_options is None:
+                storage_options = {}
+            storage_options["new_table_enable_v2_manifest_paths"] = (
+                enable_v2_manifest_paths
+            )
 
         self._table = LOOP.run(
             self._conn._conn.create_table(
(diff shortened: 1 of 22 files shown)


westonpace:
Closes #249

westonpace:
I left the embedding registry out of the catalog trait for now.  This has the curious side effect that embeddings will be applied to data going into the remote table.  However, the embedding calculation runs locally.  I'm not sure if this is a good thing or not.  Would a remote table want to run the embeddings itself?

However, I don't know if it's necessarily a bad thing either.  I can easily make it a listing-catalog only feature if we'd rather.

rust/lancedb/src/database.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub mode: CreateTableMode,
+    pub write_options: WriteOptions,
+    pub table_definition: Option<TableDefinition>,
+}
+
+/// The `Database` trait defines the interface for database implementations.
+///
+/// A database is responsible for managing tables and their metadata.
+#[async_trait::async_trait]
+pub trait Database: Send + Sync + std::fmt::Debug + std::fmt::Display + 'static {
```
wjones127, line 96: I wonder if we want to have them all implement `std::any::Any`, so we could do downcasting? Or maybe an `as_any()` as is done in DataFusion. (I forget why they do that...)

python/python/tests/test_db.py
```diff
@@ -610,10 +614,12 @@ async def is_in_v2_mode(tbl):
...
 
     # Create table in v2 mode
-    tbl = await mem_db_async.create_table(
-        "test_v2", data=make_data(), schema=schema, use_legacy_format=False
+    v2_db = await lancedb.connect_async(
+        "memory://", storage_options={"new_table_data_storage_version": "stable"}
     )
 
+    tbl = await v2_db.create_table("test_v2", data=make_data(), schema=schema)
+
```
wjones127, line 622: Out of curiosity, what happens if you pass `new_table_data_storage_version` in the `storage_options` parameter of `create_table()`? Does that override what is configured at the connection level?

wjones127 (approved):

rust/lancedb/src/table.rs
```diff
@@ -230,6 +230,21 @@ pub struct OptimizeStats {
...
+enum BadVectorHandling {
+    /// An error is returned
+    #[default]
+    Error,
+    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
+    /// The offending row is droppped
+    Drop,
+    #[allow(dead_code)] // https://github.com/lancedb/lancedb/issues/992
+    /// The invalid/missing items are replaced by fill_value
+    Fill(f32),
```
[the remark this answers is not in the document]
westonpace: Added

python/python/lancedb/table.py
```diff
@@ -2036,8 +2036,6 @@ def create(
         embedding_functions: Optional[List[EmbeddingFunctionConfig]] = None,
         *,
         storage_options: Optional[Dict[str, str]] = None,
-        data_storage_version: Optional[str] = None,
-        enable_v2_manifest_paths: Optional[bool] = None,
```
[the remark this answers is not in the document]
westonpace: I added these back in with a deprecation cycle.

rust/lancedb/src/database.rs
westonpace: Fair enough.  I've added `std::any::Any` and `as_any` (I don't know exactly why but I think `as_any` is a required part of the pattern).

python/python/tests/test_db.py
westonpace: Connection level storage options override any passed to `create_table` (I didn't change this, it's just the way it was)

PR #2073, opened 2025-01-27
feat: add `distance_type()` parameter to python sync query builders and `metric()` as an alias

### LINKED ISSUE #2047: Feature: consistently move towards distance type name over metric type
### SDK

Python

### Description

In Python sync, we use "metric type", but in Python async we use "distance type". We are moving toward "distance type", since it seems to be more standard in the industry.

We should:

1. Add a `distance_type()` parameter to the sync query builders.
2. Make `metric()` an alias to `distance_type()`.
3. Convert all code examples to use `distance_type()`.

vaifai:
This PR aims to fix #2047 by doing the following things:
- Add a distance_type parameter to the sync query builders of Python SDK.
- Make metric an alias to distance_type.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -505,7 +505,7 @@ class LanceQueryBuilder(ABC):
                 "column": self._vector_column,
                 "q": self._query,
                 "k": self._limit,
-                "metric": self._metric,
+                "metric": self._distance_type,
                 "nprobes": self._nprobes,
                 "refine_factor": self._refine_factor,
                 "use_index": self._use_index,
@@ -596,7 +596,7 @@ class LanceVectorQueryBuilder(LanceQueryBuilder):
     ):
         super().__init__(table)
         self._query = query
-        self._metric = "L2"
+        self._distance_type = "L2"
         self._nprobes = 20
         self._lower_bound = None
         self._upper_bound = None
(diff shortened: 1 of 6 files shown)


python/python/lancedb/query.py
```diff
@@ -620,7 +623,32 @@ def metric(self, metric: Literal["L2", "cosine", "dot"]) -> LanceVectorQueryBuil
...
+        ----------
+        distance_type: "L2" or "cosine" or "dot"
+            The distance metric to use. By default "L2" is used.
+
+        Returns
+        -------
+        LanceVectorQueryBuilder
+            The LanceQueryBuilder object.
+        """
+        self._metric = distance_type.lower()
```
wjones127, line 651: It would be nice to change the internal name to `distance_type` as well:

```suggestion
        self._distance_type = distance_type.lower()
```

(This will require other changes as well.)

wjones127 (changes requested):
Thanks for working on this. Looks like you've gotten all the examples. Could you also change the unit tests to use `distance_type` instead of `metric`? (Search for `.metric(` within the folder `python/python/tests`.)

Also the example here:

https://github.com/lancedb/lancedb/blob/d999d72c8d99acedbabb769344edb949d24e7994/python/python/lancedb/query.py#L579

python/python/lancedb/query.py
vaifai: Hi Will, I have made the appropriate changes. Kindly, do have a look again.

wjones127 (approved):
This is great work, @vaifai. Thank you!

PR #2081, opened 2025-01-29
ci: fix arm64 windows cross compile build

wjones127:
* Adds a CI job to check the cross compiled Windows ARM build.
   * Didn't replace the test build because we need native build to run tests. But for some reason (I forget why) we need cross compiled for nodejs.
* Pinned crunchy to workaround https://github.com/eira-fransham/crunchy/issues/13

This is needed to fix failure from https://github.com/lancedb/lancedb/actions/runs/13020773184/job/36320719331


westonpace (approved):
At what point do we buy you a trophy?  Thanks for venturing into Windows-land :tada:

PR #2083, opened 2025-01-30
docs: user guide for merge insert

### LINKED ISSUE #2062: Document merge-insert / upsert
### Description

We are missing a guide on how to do upsert, which is a common operation.

### Link

https://lancedb.github.io/lancedb/guides/tables/

wjones127:
Closes #2062

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -604,7 +604,7 @@ After a table has been created, you can always add more data to it using the `ad
 ## Upserting into a table
 
 Upserting lets you insert new rows or update existing rows in a table. To upsert
-in LanceDB, use the `merge insert` command.
+in LanceDB, use the merge insert API.
 
 === "Python"
 
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -518,7 +518,7 @@ After a table has been created, you can always add more data to it using the `ad
         --8<-- "python/python/tests/docs/test_guide_tables.py:add_table_from_polars"
         ```
     === "Async API"
-    
+
         ```python
         --8<-- "python/python/tests/docs/test_guide_tables.py:add_table_async_from_polars"
         ```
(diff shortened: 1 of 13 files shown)


docs/src/guides/tables.md
```diff
@@ -601,6 +601,38 @@ After a table has been created, you can always add more data to it using the `ad
     )
     ```
 
+## Upserting into a table
+
+Upserting lets you insert new rows or update existing rows in a table. To upsert
+in LanceDB, use the `merge insert` command.
```
QianZhu, line 607: `merge insert` command ->`merge_insert` API?

docs/src/guides/tables/merge_insert.md
```diff
@@ -0,0 +1,126 @@
+The merge insert command is a flexible API that can be used to perform:
+
+1. Upsert
+2. Insert-if-not-exists
+3. Replace range
```
QianZhu, line 5: do we also want to talk about `conditional update` and `replace range without condition`

docs/src/guides/tables/merge_insert.md
```diff
@@ -0,0 +1,126 @@
...
+!!! tip "Use scalar indices to speed up merge insert"
+
+    The merge insert command needs to perform a join between the input data and the
+    target table on the `on` key you provide. This requires scanning that entire
+    column, which can be expensive for large tables. To speed up this operation,
+    you can create a scalar index on the `on` column, which will allow LanceDB to
+    find matches without having to scan the whole tables.
+
+    Read more about scalar indices in [Building a Scalar Index](../scalar_index.md)
+    guide.
```
QianZhu, line 30: `merge_insert` now can use embedding functions. Can we add an example on that?

wjones127: I can add a note about that. I'm not sure it warrants a whole example.

docs/src/guides/tables/merge_insert.md
wjones127: I think they could be added in a future update if there's a particular pattern that users commonly find useful. I think three examples is likely enough for users to get some ideas of how they use it. But if we get more questions from users on a particular pattern, we should add it. By far, the most common question we get is how to upsert, which is why that's the first example. And the second most common is how to insert but avoid duplicates.

docs/src/guides/tables.md
wjones127:
```suggestion
in LanceDB, use the merge insert API.
```

I say "merge insert" just to talk about in english. In Python it's `merge_insert`, in Node it's `mergeInsert`.  But we can use the word API instead here, I'm good with that.

QianZhu:
Looks good overall. Thank you @wjones127 !

Issue #1581, opened 2024-08-30
How to add embeding_registry to a table which is returned by "open_table" function?


laonger:
### Description

I get the error 
`EmbeddingFunctionNotFound { name: "openai", reason: "Table was defined with an embedding column 'openai' but no embedding function was found with that name within the registry." }`

and my code like

```
    let tempdir = tempfile::tempdir().unwrap();
    let tempdir = tempdir.path().to_str().unwrap();
    let db = connect(tempdir).execute().await?;
    
    db.embedding_registry()
        .register("openai", embedding.clone())?;

    db.create_table("test", create_some_records()?)
        .add_embedding(EmbeddingDefinition::new(
            "content",
            "openai",
            Some("embeddings"),
        ))?
        .execute()
        .await?;

    let db = connect(tempdir).execute().await?;

    let tbl = db.open_table("test").execute().await?;
    let res = tbl.add(create_some_records()?).execute().await;
```

I readed lancedb's source code, and I find out there is no "embeding_registry" in a table object which is returned by "open_table" function.

So, how do I add a registry?

thanks very much.

jeff1010322:
Is there any update on this?

I tried to follow the rust sentence_transformers example, but adding an `open_table` followed by `add` gets the same error described here.

Can replicate by adding the following two lines to attempt to add data to the sentence_transformers example:
https://github.com/lancedb/lancedb/blob/main/rust/lancedb/examples/sentence_transformers.rs#L30

```
let open_table = db.open_table("vectors").execute().await?;

open_table.add(make_data()).execute().await?;
```

```
Error: EmbeddingFunctionNotFound { name: "sentence-transformers", reason: "Table was defined with an embedding column `sentence-transformers` but no embedding function was found with that name within the registry." }
```

jeff1010322:
Just tested locally and it appears that this PR fixes this issue: https://github.com/lancedb/lancedb/pull/1781

Would be great if this could get updated and merged.

PR #2086, opened 2025-01-31
fix(rust): add embedding_registry on open_table

jeff1010322:
# Description

Fix for: https://github.com/lancedb/lancedb/issues/1581

This is the same implementation as https://github.com/lancedb/lancedb/pull/1781 but with the addition of a unit test and rustfmt.


wjones127 (approved):
Excellent work. Thank you for fixing this!

PR #2094, opened 2025-02-04
feat(python): streaming larger-than-memory writes

### LINKED ISSUE #2082: Feature: support streaming data with `create_table` and `add`
### SDK

Python

### Description

In Python, we support passing in `RecordBatchReader`s and iterators, but we collect them all in memory to do preprocessing in Python:

https://github.com/lancedb/lancedb/blob/e6b4f14c1fcf9c6af4cf8bcc29915af15ff8291f/python/python/lancedb/table.py#L209

We should transform `_santize_data` to convert all inputs to a `RecordBatchReader` and then do all subsequent steps as a stream. This will enable doing larger-than-memory writes.

wjones127:
Makes our preprocessing pipeline do transforms in streaming fashion, so users can do larger-then-memory writes.

Closes #2082

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -132,7 +132,7 @@ def _into_pyarrow_reader(data) -> pa.RecordBatchReader:
     ):
         return data.collect().to_arrow().to_reader()
     elif isinstance(data, Iterable):
-        return _iterator_to_table(data)
+        return _iterator_to_reader(data)
     else:
         raise TypeError(
             f"Unknown data type {type(data)}. "
@@ -142,14 +142,16 @@ def _into_pyarrow_reader(data) -> pa.RecordBatchReader:
         )
 
 
-def _iterator_to_table(data: Iterable) -> pa.RecordBatchReader:
+def _iterator_to_reader(data: Iterable) -> pa.RecordBatchReader:
+    # Each batch is treated as it's own reader, mainly so we can
+    # re-use the _into_pyarrow_reader logic.
     first = _into_pyarrow_reader(next(data))
     schema = first.schema
 
     def gen():
         yield from first
         for batch in data:
-            table = _into_pyarrow_reader(batch).read_all()
+            table: pa.Table = _into_pyarrow_reader(batch).read_all()
             if table.schema != schema:
                 try:
                     table = table.cast(schema)
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -23,6 +23,7 @@ from typing import (
 from urllib.parse import urlparse
 
 import lance
+from lancedb.arrow import peek_reader
 from lancedb.background_loop import LOOP
 from .dependencies import _check_for_pandas
 import pyarrow as pa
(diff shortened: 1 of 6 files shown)


wjones127:
Using this script, I can write 10GB of data with peak memory usage of ~450MB:

```python
import lancedb
import pyarrow as pa
import pyarrow.compute as pc
import shutil

ndims = 1536
batch_size = 1_000
total_data = 1 * 1024 * 1024 * 1024 # 10 GB

schema = pa.schema({
    "id": pa.int64(),
    "vector": pa.list_(pa.float32(), ndims)
})

def make_batch() -> pa.RecordBatch:
    values = pc.random(ndims * batch_size).cast(pa.float32())
    vectors = pa.FixedSizeListArray.from_arrays(values, ndims)
    ids = pa.array(range(batch_size), type=pa.int64())
    return pa.RecordBatch.from_arrays([ids, vectors], schema)

def gen():
    data_written = 0
    while data_written < total_data:
        batch = make_batch()
        data_written += batch.get_total_buffer_size()
        yield batch

reader = pa.RecordBatchReader.from_batches(schema, gen())

shutil.rmtree(".data", ignore_errors=True)
db = lancedb.connect(".data")
db.create_table("table", reader)

shutil.rmtree(".data", ignore_errors=True)
```

Measured with `memray`:

```bash
python -m memray run -o output.bin example_streaming_write.py
python -m memray flamegraph output.bin
```

Screenshot 2025-02-04 at 2 40 24 PM

python/python/lancedb/table.py
```diff
@@ -139,30 +142,26 @@ def _into_pyarrow_table(data) -> pa.Table:
...
 
-def _iterator_to_table(data: Iterable) -> pa.Table:
-    batches = []
-    schema = None  # Will get schema from first batch
-    for batch in data:
-        batch_table = _into_pyarrow_table(batch)
-        if schema is not None:
-            if batch_table.schema != schema:
+def _iterator_to_table(data: Iterable) -> pa.RecordBatchReader:
+    first = _into_pyarrow_reader(next(data))
```
westonpace, line 146: Minor nit: Slightly odd that this is named `first` but it is an entire iterator.

python/python/lancedb/table.py
```diff
@@ -139,30 +142,26 @@ def _into_pyarrow_table(data) -> pa.Table:
...
 
 
-def _iterator_to_table(data: Iterable) -> pa.Table:
-    batches = []
-    schema = None  # Will get schema from first batch
-    for batch in data:
-        batch_table = _into_pyarrow_table(batch)
-        if schema is not None:
-            if batch_table.schema != schema:
+def _iterator_to_table(data: Iterable) -> pa.RecordBatchReader:
```
westonpace, line 145: Should this method be renamed?

python/python/lancedb/table.py
```diff
@@ -139,30 +142,26 @@ def _into_pyarrow_table(data) -> pa.Table:
...
+    schema = first.schema
+
+    def gen():
+        yield from first
+        for batch in data:
+            table = _into_pyarrow_reader(batch).read_all()
+            if table.schema != schema:
                 try:
-                    batch_table = batch_table.cast(schema)
+                    table = table.cast(schema)
```
westonpace, line 155: Is `cast` a valid thing to call on a reader?

python/python/lancedb/table.py
```diff
@@ -139,30 +142,26 @@ def _into_pyarrow_table(data) -> pa.Table:
...
-        if schema is not None:
-            if batch_table.schema != schema:
+def _iterator_to_table(data: Iterable) -> pa.RecordBatchReader:
+    first = _into_pyarrow_reader(next(data))
+    schema = first.schema
+
+    def gen():
+        yield from first
+        for batch in data:
+            table = _into_pyarrow_reader(batch).read_all()
```
westonpace, line 152: Is `table` a table or a reader?

wjones127: We call `read_all()` on it, so it's a `Table`.

python/python/lancedb/table.py
wjones127: We called `read_all()` to make it a table.

PR #2108, opened 2025-02-10
fix: fix index out of bound in load indices

LuQQiu:
panicked at 'index out of bounds: the len is 24 but the index is 25':Lancedb/rust/lancedb/src/index/vector.rs:26\n

load_indices() on the old manifest while use the newer manifest to get column names could result in index out of bound if some columns are removed from the new version.
This change reduce the possibility of index out of bound operation but does not fully remove it.
Better that lance can directly provide column name info so no need extra calls to get column name but that require modify the public APIs

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -1380,10 +1380,11 @@ impl NativeTable {
 
     pub async fn load_indices(&self) -> Result<Vec<VectorIndex>> {
         let dataset = self.dataset.get().await?;
-        let (indices, mf) = futures::try_join!(dataset.load_indices(), dataset.latest_manifest())?;
+        let mf = dataset.manifest();
+        let indices = dataset.load_indices().await?;
         Ok(indices
             .iter()
-            .map(|i| VectorIndex::new_from_format(&(mf.0), i))
+            .map(|i| VectorIndex::new_from_format(&mf, i))
             .collect())
     }
 
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -1384,7 +1384,7 @@ impl NativeTable {
         let indices = dataset.load_indices().await?;
         Ok(indices
             .iter()
-            .map(|i| VectorIndex::new_from_format(&mf, i))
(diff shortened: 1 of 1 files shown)


wjones127:
> This change reduce the possibility of index out of bound operation but does not fully remove it.

What do you mean? Now they are both reading from the same manifest, so they should always have consistent state.

> Better that lance can directly provide column name info so no need extra calls to get column name but that require modify the public APIs

I agree it would be better if Lance's public facing APIs did the translation from `field_id` to `field_name` for the user.

wjones127 (approved):

LuQQiu:
@wjones127 

> This change reduce the possibility of index out of bound operation but does not fully remove it.
> 
> What do you mean? Now they are both reading from the same manifest, so they should always have consistent state.
> 

under multithreading, the dataset checkout version may be changed anytime, so during load_indices() the version may have been updated?

```
    async fn load_indices(&self) -> Result<Arc<Vec<IndexMetadata>>> {
        let dataset_dir = self.base.to_string();
        if let Some(indices) = self
            .session
            .index_cache
            .get_metadata(&dataset_dir, self.version().version)
        {
            return Ok(indices);
        }
        // Some race condition happens?
        let manifest_file = self.manifest_file().await?;
        let loaded_indices: Arc<Vec<IndexMetadata>> =
            read_manifest_indexes(&self.object_store, &manifest_file, &self.manifest)
                .await?
                .into();

        self.session.index_cache.insert_metadata(
            &dataset_dir,
            self.version().version,          // Some race condition happens?
            loaded_indices.clone(),
        );
        Ok(loaded_indices)
    }
```

rpgreen (approved):

wjones127:
> under multithreading, the dataset checkout version may be changed anytime

That shouldn't be true. The underlying dataset is protected by a `RwLock`, which prevents mutation when there are one or more readers. The `.get()` method on the dataset handle creates a read guard, which is alive during the lifetime of this function:

https://github.com/lancedb/lancedb/blob/a7755cb3137b53aa139620ab033a1cc065c8af79/rust/lancedb/src/table/dataset.rs#L136-L142

PR #2112, opened 2025-02-11
feat!: add variable store to embeddings registry

### LINKED ISSUE #521: EmbeddingFunction should work across device
Currently the embedding function does not explicitly move the input to device or the output back to CPU. So if created with device != cpu (the default), then an exception is encountered complaining that the weights and the data must be on the same device

### LINKED ISSUE #2110: Credential handling in embedding functions
Much of the time, embedding API credentials are handled via environment variables. But they can also be passed as part of the embedding function.

## Typescript

Right now, we skip API keys when serializing to JSON, so they credentials are lost easily:

https://github.com/lancedb/lancedb/blob/a7755cb3137b53aa139620ab033a1cc065c8af79/nodejs/lancedb/embedding/openai.ts#L55-L59

## Python

Python **doesn't** skip API keys when serializing to JSON:

```python
from lancedb.embeddings.registry import get_registry
import lancedb

registry = get_registry()
func = registry.get('openai').create(api_key='I AM SECRET')

func.safe_model_dump()
```

```
{'max_retries': 7, 'name': 'text-embedding-ada-002', 'dim': None, 'base_url': None, 'default_headers': None, 'organization': None, 'api_key': 'I AM SECRET', 'use_azure': False}
```

```python
class Schema(LanceModel):

db = lancedb.connect("./creds_test")
table = db.create_table("test", schema=Schema)
table.schema.metadata
```

```
{b'embedding_functions': b'[\n  {\n    "name": "openai",\n    "model": {\n      "max_retries": 7,\n      "name": "text-embedding-ada-002",\n      "dim": null,\n      "base_url": null,\n      "default_headers": null,\n      "organization": null,\n      "api_key": "I AM SECRET",\n      "use_azure": false\n    },\n    "source_column": "text",\n    "vector_column": "vector"\n  }\n]'}
```

This means: 
1. We leak credentials into the table metadata
2. We don't provide any easy way to rotate credentials, since it's not obvious how to update the embedding function config

## Solution

We probably want a way to provide secrets in a session, but not serialize those into the metadata. Perhaps we can give the secrets names within a session?

```python
from lancedb.embeddings.registry import get_registry
import lancedb

registry = get_registry()
registry.set_secret('openai_api_key', 'MY SECRET')

func = registry.get('openai').create(api_key='$secret:open_api_key')

func.safe_model_dump()
```

```
{'max_retries': 7, 'name': 'text-embedding-ada-002', 'dim': None, 'base_url': None, 'default_headers': None, 'organization': None, 'use_azure': False}
```

This secret would exist for the in-memory session.

And when you load the table without having set the secret, you could provide an error

```
LanceDBError: missing secret 'open_api_key'. Could not load embedding function.
```

wjones127:
BREAKING CHANGE: embedding function implementations in Node need to now call `resolveVariables()` in their constructors and should **not** implement `toJSON()`.

This tries to address the handling of secrets. In Node, they are currently lost. In Python, they are currently leaked into the table schema metadata.

This PR introduces an in-memory variable store on the function registry. It also allows embedding function definitions to label certain config values as "sensitive", and the preprocessing logic will raise an error if users try to pass in hard-coded values.

Closes #2110
Closes #521

### DIFF
--- a/docs/src/embeddings/custom_embedding_function.md
+++ b/docs/src/embeddings/custom_embedding_function.md
@@ -56,8 +56,8 @@ Let's implement `SentenceTransformerEmbeddings` class. All you need to do is imp
 This is a stripped down version of our implementation of `SentenceTransformerEmbeddings` that removes certain optimizations and default settings.
 
 !!! danger "Use sensitive keys to prevent leaking secrets"
-    To prevent leaking secrets, such as API keys, you should add an configuration
-    parameteres of embedding function to the output of the
+    To prevent leaking secrets, such as API keys, you should add any sensitive
+    parameters of an embedding function to the output of the
     [sensitive_keys()][lancedb.embeddings.base.EmbeddingFunction.sensitive_keys] /
     [getSensitiveKeys()](../../js/namespaces/embedding/classes/EmbeddingFunction/#getsensitivekeys)
     method. This prevents users from accidentally instantiating the embedding
--- a/docs/src/embeddings/custom_embedding_function.md
+++ b/docs/src/embeddings/custom_embedding_function.md
@@ -55,6 +55,14 @@ Let's implement `SentenceTransformerEmbeddings` class. All you need to do is imp
 
 This is a stripped down version of our implementation of `SentenceTransformerEmbeddings` that removes certain optimizations and default settings.
 
+!!! danger "Use sensitive keys to prevent leaking secrets"
+    To prevent leaking secrets, such as API keys, you should add an configuration
+    parameteres of embedding function to the output of the
+    [sensitive_keys()][lancedb.embeddings.base.EmbeddingFunction.sensitive_keys] /
+    [getSensitiveKeys()](../../js/namespaces/embedding/classes/EmbeddingFunction/#getsensitivekeys)
+    method. This prevents users from accidentally instantiating the embedding
+    function with hard-coded secrets.
+
 Now you can use this embedding function to create your table schema and that's it! you can then ingest data and run queries without manually vectorizing the inputs.
 
 === "Python"
--- a/docs/src/embeddings/variables_and_secrets.md
+++ b/docs/src/embeddings/variables_and_secrets.md
@@ -1,0 +1,58 @@
+# Variable and Secrets
+
+Most embedding configuration options are saved in the table's metadata. However,
+this isn't always appropriate. For example, API keys should never be stored in the
+metadata. Additionally, other configuration options might be best set at runtime,
+such as the `device` configuration that controls whether to use GPU or CPU for
+inference. If you hardcoded this to GPU, you wouldn't be able to run the code on
+server without one.
+
+To handle these cases, you can set variables on the embedding registry and
+reference them in the embedding configuration. These variables will be available
+during the runtime of your program, but not saved in the table's metadata. When
+the table is loaded from a different process, the variables must be set again
+for the embedding functions to be able to be loaded.
+
+To set a variable, use the `set_var()` / `setVar()` method on the embedding registry.
+To reference a variable, use the syntax `$env:VARIABLE_NAME`. If there is a default
+value, you can use the syntax `$env:VARIABLE_NAME:DEFAULT_VALUE`.
+
+## Using variables to set secrets
+
+Sensitive configuration, such as API keys, must either be set as environment
+variables or using variables on the embedding registry. If you pass in a hardcoded
+value, LanceDB will raise an error.
+
+=== "Python"
+
+    ```shell
+    --8<-- "python/python/tests/docs/test_embeddings_optional.py:register_secret"
+    ```
+
+=== "Typescript"
+
+    ```shell
+    --8<-- "docs/src/basic_legacy.ts:create_table"
+    ```
+
+## Using variables to set device
+
+Because not all computers have a GPU, it's helpful to be able to set the `device`
+parameter at runtime, rather than have it hard coded in the embedding configuration.
+To make it work even if the variable isn't set, you could provide a default value
+of `cpu` in the embedding configuration.
+
+Some embedding libraries have a method to detect which devices are available,
+which could also be used to dynamically set the device at runtime. For example, in Python you can check if a CUDA GPU is available using `torch.cuda.is_available()`.
+
+=== "Python"
+
+    ```shell
+    --8<-- "python/python/tests/docs/test_embeddings_optional.py:register_device"
+    ```
+
+=== "Typescript"
+
+    ```shell
+    --8<-- "docs/src/basic_legacy.ts:create_table"
+    ```
--- a/docs/src/embeddings/variables_and_secrets.md
+++ b/docs/src/embeddings/variables_and_secrets.md
@@ -5,13 +5,12 @@ this isn't always appropriate. For example, API keys should never be stored in t
 metadata. Additionally, other configuration options might be best set at runtime,
 such as the `device` configuration that controls whether to use GPU or CPU for
 inference. If you hardcoded this to GPU, you wouldn't be able to run the code on
-server without one.
+a server without one.
 
 To handle these cases, you can set variables on the embedding registry and
 reference them in the embedding configuration. These variables will be available
 during the runtime of your program, but not saved in the table's metadata. When
-the table is loaded from a different process, the variables must be set again
-for the embedding functions to be able to be loaded.
+the table is loaded from a different process, the variables must be set again.
 
 To set a variable, use the `set_var()` / `setVar()` method on the embedding registry.
 To reference a variable, use the syntax `$env:VARIABLE_NAME`. If there is a default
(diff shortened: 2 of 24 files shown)


docs/src/embeddings/custom_embedding_function.md
```diff
@@ -55,6 +55,14 @@ Let's implement `SentenceTransformerEmbeddings` class. All you need to do is imp
 
 This is a stripped down version of our implementation of `SentenceTransformerEmbeddings` that removes certain optimizations and default settings.
 
+!!! danger "Use sensitive keys to prevent leaking secrets"
+    To prevent leaking secrets, such as API keys, you should add an configuration
+    parameteres of embedding function to the output of the
```
westonpace, line 60:
```suggestion
    To prevent leaking secrets, such as API keys, you should add any sensitive
    parameters of an embedding function to the output of the
```

docs/src/js/namespaces/embedding/classes/EmbeddingFunctionRegistry.md
```diff
@@ -145,3 +167,27 @@ reset the registry to the initial state
...
+
+```ts
+setVar(name, value): void
+```
+
+Set a variable. These can be accessed in the embedding function
+configuration using the syntax `$var:variable_name`. If they are not
+set, an error will be thrown letting you know which key is unset. If you
+want to supply a default value, you can add an additional part in the
+configuration like so: `$var:variable_name:default_value`.
```
westonpace, line 183: Why would I set a default value in the `setVar` function?

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -44,33 +61,75 @@ export abstract class EmbeddingFunction<
...
+   * Get the original arguments to the constructor, to serialize them so they
+   * can be used to recreate the embedding function later.
+   */
+  // biome-ignore lint/suspicious/noExplicitAny :
+  toJSON(): Record<string, any> {
+    return JSON.parse(JSON.stringify(this.#config));
+  }
+
+  constructor(args: Partial<M> = {}) {
+    this.#config = args;
```
westonpace, line 77: Why store `this.config` here AND also require it to be passed into `resolveVariables`?

Wouldn't it be simpler to skip passing it in here?

westonpace (commented):
Overall I think this strategy works well.  I think my only feedback would be that it seems slightly odd to me that the embedding function author is the one that chooses which keys are set permanently and which keys have to be supplied each time.

I think for sensitive keys this can kind of make sense as the embedding function author will probably have a good idea which options are sensitive.

Maybe this is true for runtime-configurable keys like gpu/cpu too.  I guess an alternative would be to just always allow the var syntax to be recognized even if the key is not listed as a sensitive key?

wjones127:
> I think my only feedback would be that it seems slightly odd to me that the embedding function author is the one that chooses which keys are set permanently and which keys have to be supplied each time.
> ...
> I guess an alternative would be to just always allow the var syntax to be recognized even if the key is not listed as a sensitive key?

The var syntax indeed can be used by any key, not just sensitive ones. That's why it works with `device`. The difference is sensitive keys **must** be supplied as variables. I think embedding function authors should know which keys are sensitive enough that they shouldn't be serialized as metadata.

docs/src/js/namespaces/embedding/classes/EmbeddingFunctionRegistry.md
wjones127: It probably doesn't make sense for secrets, but for runtime configuration like `device` I feel like it's nice to provide a fallback value. For the `device` example, I show using `cpu` as a fallback default value.

docs/src/embeddings/variables_and_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+a server without one.
+
+To handle these cases, you can set variables on the embedding registry and
+reference them in the embedding configuration. These variables will be available
+during the runtime of your program, but not saved in the table's metadata. When
+the table is loaded from a different process, the variables must be set again.
+
+To set a variable, use the `set_var()` / `setVar()` method on the embedding registry.
+To reference a variable, use the syntax `$env:VARIABLE_NAME`. If there is a default
+value, you can use the syntax `$env:VARIABLE_NAME:DEFAULT_VALUE`.
```
westonpace, line 17: Can the colon be escaped?  If not, we can leave for a todo

docs/src/embeddings/variables_and_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+Many embedding functions that run locally have a `device` parameter that controls
+whether to use GPU or CPU for inference. Because not all computers have a GPU,
+it's helpful to be able to set the `device` parameter at runtime, rather than
+have it hard coded in the embedding configuration. To make it work even if the
+variable isn't set, you could provide a default value of `cpu` in the embedding
+configuration.
+
+Some embedding libraries even have a method to detect which devices are available,
+which could be used to dynamically set the device at runtime. For example, in Python
+you can check if a CUDA GPU is available using `torch.cuda.is_available()`.
```
westonpace, line 49: Are you arguing that users should automatically call `set_var` based on the result of this function?  Or are you arguing that embedding function authors should use this function to avoid the need for the variable in the first place?

nodejs/__test__/registry.test.ts
```diff
@@ -155,3 +145,79 @@ describe.each([arrow15, arrow16, arrow17, arrow18])("Registry", (arrow) => {
     expect(schema.metadata).toEqual(expectedMetadata);
   });
 });
+
+describe("Registry.setVar", () => {
+  const registry = getRegistry();
+
+  beforeEach(() => {
+    @register("mock-embedding")
+    // biome-ignore lint/correctness/noUnusedVariables :
```
westonpace, line 154: What's the unused variable here?

nodejs/__test__/registry.test.ts
```diff
@@ -155,3 +145,79 @@ describe.each([arrow15, arrow16, arrow17, arrow18])("Registry", (arrow) => {
...
+    class MockEmbeddingFunction extends EmbeddingFunction<string> {
+      constructor(optionsRaw: FunctionOptions = {}) {
+        super();
+        const options = this.resolveVariables(optionsRaw);
+
+        expect(optionsRaw["someKey"].startsWith("$var:someName")).toBe(true);
+        expect(options["someKey"]).toBe("someValue");
+
+        if (options["secretKey"]) {
```
westonpace, line 163: Why the if here?  Shouldn't this always assert the value is present?

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -44,33 +61,74 @@ export abstract class EmbeddingFunction<
...
-   *       model: this.model,
-   *       timeout: this.timeout,
-   *     };
-   * }
-   * ```
+   * Get the original arguments to the constructor, to serialize them so they
+   * can be used to recreate the embedding function later.
+   */
+  // biome-ignore lint/suspicious/noExplicitAny :
+  toJSON(): Record<string, any> {
```
westonpace, line 72: Minor nit: you can maybe use `unknown` instead of `any`

nodejs/lancedb/embedding/registry.ts
```diff
@@ -164,6 +162,35 @@ export class EmbeddingFunctionRegistry {
...
+
+  /**
+   * Set a variable. These can be accessed in the embedding function
+   * configuration using the syntax `$var:variable_name`. If they are not
+   * set, an error will be thrown letting you know which key is unset. If you
+   * want to supply a default value, you can add an additional part in the
+   * configuration like so: `$var:variable_name:default_value`. Default values
+   * can be used for runtime configurations that are not sensitive, such as
+   * whether to use a GPU for inference.
+   *
```
westonpace, line 174: Mention that variable names cannot contain colons?

nodejs/lancedb/embedding/transformers.ts
```diff
@@ -44,11 +44,12 @@ export class TransformersEmbeddingFunction extends EmbeddingFunction<
   #ndims?: number;
 
   constructor(
-    options: Partial<XenovaTransformerOptions> = {
+    optionsRaw: Partial<XenovaTransformerOptions> = {
       model: "Xenova/all-MiniLM-L6-v2",
     },
   ) {
     super();
+    const options = this.resolveVariables(optionsRaw);
```
westonpace, line 52: Is this a breaking change for existing JS embedding functions?

python/python/lancedb/embeddings/registry.py
```diff
@@ -156,6 +157,26 @@ def get_table_metadata(self, func_list):
...
+    def set_var(self, name: str, value: str) -> None:
+        """
+        Set a variable. These can be accessed in embedding configuration using
+        the syntax `$var:variable_name`. If they are not set, an error will be
+        thrown letting you know which variable is missing. If you want to supply
+        a default value, you can add an additional part in the configuration
+        like so: `$var:variable_name:default_value`. Default values can be
+        used for runtime configurations that are not sensitive, such as
+        whether to use a GPU for inference.
+        """
```
westonpace, line 169: Mention that colons are not allowed in keys?

westonpace (approved):
A few minor questions but nothing major.  This was a good idea, thanks!

docs/src/embeddings/variables_and_secrets.md
wjones127: What do you mean? We disallow colons in the variable name. Any colon after the second one is assumed to be part of the `DEFAULT_VALUE`.

docs/src/embeddings/variables_and_secrets.md
wjones127: I'm suggesting they could use `set_var` based on this.

nodejs/__test__/registry.test.ts
wjones127: The embedding function. The `@register()` decorator adds it to the registry, and we use it implicitly. Typescript doesn't seem to understand that.

nodejs/__test__/registry.test.ts
wjones127: Maybe. I didn't want to write a new embedding function impl for each test. But I guess it could be more robust if I did.

nodejs/lancedb/embedding/embedding_function.ts
wjones127: I'll consider that for a follow up. Unfortunately it seems like it would affect a lot of code.

PR #2122, opened 2025-02-16
fix: metric type inconsistency

### LINKED ISSUE #2113: bug(python):  metric typehint is inconsistent
### LanceDB version

_No response_

### What happened?

def metric(self, metric: Literal["L2", "cosine", "dot"]) -> LanceHybridQueryBuilder:

@dataclass
class HnswPq:
distance_type: Literal["l2", "cosine", "dot"] = "l2"

So, should it be standard lower l2?

_No response_

gagan-bhullar-tech:
PR fixes #2113

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1095,16 +1095,15 @@ dependencies = [
 
 [[package]]
 name = "blake3"
-version = "1.6.0"
+version = "1.6.1"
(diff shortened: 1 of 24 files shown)


wjones127 (approved):
Looks good. I've just rebased this on top of newer type checking changes.

PR #2123, opened 2025-02-16
docs: passing env variables for azure blob connection is mandatory

shuklaji28:
Necessary to add env variables even if passing them as parameters in storage_options


wjones127:
If the storage options aren't working for azure blob storage, this is a bug. Could you open an issue and describe how to reproduce?

PR #2147, opened 2025-02-24
feat: record the server version for remote table

BubbleCal:


python/pyproject.toml
```diff
@@ -54,6 +54,7 @@ tests = [
     "polars>=0.19, <=1.3.0",
     "tantivy",
     "pyarrow-stubs",
+    "semver",
```
wjones127, line 57: Is another dependency worthwhile?

rust/lancedb/src/remote/db.rs
```diff
@@ -22,9 +22,16 @@ use crate::table::BaseTable;
...
-use super::util::batches_to_ipc_bytes;
+use super::util::{batches_to_ipc_bytes, parse_server_version};
 use super::ARROW_STREAM_CONTENT_TYPE;
 
+// the versions of the server that we support
+// for any new feature that we need to change the SDK behavior, we should bump the server version,
+// and add a feature flag here, so use `self.server_version >= FEATURE_VERSION` to check if the server
+// supports the feature.
+pub const DEFAULT_SERVER_VERSION: semver::Version = semver::Version::new(0, 1, 0);
+pub const MULTIVECTOR_VERSION: semver::Version = semver::Version::new(0, 2, 0);
```
wjones127, line 33: I think these belong alongside the `parse_server_version`. You might also consider consolidating into an easy to read struct.

```rust
struct ServerVersion(pub semver::Version);

impl ServerVersion {
    pub fn supports_multivector_endpoint() -> bool { ... }
}
```

rust/lancedb/src/remote/table.rs
```diff
@@ -334,9 +364,10 @@ mod test_utils {
     use super::*;
     use crate::remote::client::test_utils::client_with_handler;
     use crate::remote::client::test_utils::MockSender;
+    use crate::remote::db::DEFAULT_SERVER_VERSION;
 
     impl RemoteTable<MockSender> {
-        pub fn new_mock<F, T>(name: String, handler: F) -> Self
+        pub fn new_mock<F, T>(name: String, handler: F, version: Option<semver::Version>) -> Self
```
wjones127, line 370: I appreciate the new test utils!

python/pyproject.toml
BubbleCal: switched to `packaging.version.Version`

wjones127 (approved):
Nice work!

PR #2148, opened 2025-02-25
feat(rust): introduce Catalog trait and implement ListingCatalog

yanghua:


yanghua:
This PR is not ready for review.

rust/lancedb/src/catalog.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub start_after: Option<String>,
+    /// Maximum number of names to return
+    pub limit: Option<u32>,
+}
+
+/// Request to open an existing database
+#[derive(Clone, Debug)]
+pub struct OpenDatabaseRequest {
+    pub name: String,
+    pub database_options: HashMap<String, String>,
```
westonpace, line 28:
```suggestion
    /// The name of the database to open
    pub name: String,
    /// A map of database-specific options
    ///
    /// Consult the catalog / database implementation to determine which options are available
    pub database_options: HashMap<String, String>,
```

rust/lancedb/src/catalog.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn default() -> Self {
+        Self::Create
+    }
+}
+
+/// Request to create a new database
+pub struct CreateDatabaseRequest {
+    pub name: String,
+    pub mode: CreateDatabaseMode,
+    pub options: HashMap<String, String>,
```
westonpace, line 51:
```suggestion
    /// The name of the database to create
    pub name: String,
    /// The creation mode
    pub mode: CreateDatabaseMode,
    /// A map of catalog-specific options, consult your catalog implementation to determine what's available
    pub options: HashMap<String, String>,
```

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+use crate::error::{CreateDirSnafu, Error, Result};
+use async_trait::async_trait;
+use lance::io::{ObjectStore, ObjectStoreParams, ObjectStoreRegistry};
+use lance_io::local::to_local_path;
+use lance_io::object_store::WrappingObjectStore;
+use object_store::path::Path as ObjectStorePath;
+use snafu::ResultExt;
+
+#[derive(Debug)]
+pub struct ListingCatalog {
```
westonpace, line 26:
```suggestion
/// A catalog implementation that works by listing subfolders in a directory
///
/// The listing catalog will be created with a base folder specified by the URI.  Every subfolder
/// in this base folder will be considered a database.  These will be opened as a
/// [`crate::database::listing::ListingDatabase`]
#[derive(Debug)]
pub struct ListingCatalog {
```

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+    object_store: ObjectStore,
+
+    query_string: Option<String>,
+
+    pub(crate) uri: String,
```
westonpace, line 31: Why `pub(crate)`?

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+
+    base_path: ObjectStorePath,
+
+    pub(crate) store_wrapper: Option<Arc<dyn WrappingObjectStore>>,
```
westonpace, line 35: Why `pub(crate)`?

rust/lancedb/src/catalog.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// Request to open an existing database
+#[derive(Clone, Debug)]
+pub struct OpenDatabaseRequest {
+    pub name: String,
+    pub database_options: HashMap<String, String>,
+}
+
+/// Database creation mode
```
westonpace, line 31:
```suggestion
/// Database creation mode
///
/// The default behavior is Create
```

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+#[async_trait]
+impl Catalog for ListingCatalog {
+    async fn database_names(&self, request: DatabaseNamesRequest) -> Result<Vec<String>> {
+        let mut f = self
+            .object_store
+            .read_dir(self.base_path.clone())
+            .await?
+            .iter()
+            .map(Path::new)
+            .filter_map(|p| p.file_stem().and_then(|s| s.to_str().map(String::from)))
```
westonpace, line 114: You are using `file_stem` here which means it will throw away any extension.  However, you are not using `file_stem` in the `open_database` method.  What should happen if someone creates a database named `foo.bar`?  Right now I think `database_names` will return `foo` but if we pass `foo` to `open_database` it will fail.

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+        if let Some(limit) = request.limit {
+            f.truncate(limit as usize);
+        }
+        Ok(f)
+    }
+
+    async fn create_database(&self, request: CreateDatabaseRequest) -> Result<Arc<dyn Database>> {
+        let db_path = self.database_path(&request.name);
+        let db_path_str = to_local_path(&db_path);
+        let exists = Path::new(&db_path_str).exists();
```
westonpace, line 134: I'm not sure we should use `to_local_path` here.  It will not work for remote storage.  The `lance::io::ObjectStore` (`object_store`) has an `exists` method which takes a `Path`.  Can you use that instead?

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        match request.mode {
+            CreateDatabaseMode::Create if exists => {
+                return Err(Error::DatabaseAlreadyExists { name: request.name })
+            }
+            CreateDatabaseMode::Create => {
+                let db_local_path = to_local_path(&db_path);
```
westonpace, line 141: Same, let's not use local path here.  We should only call `create_dir_all` if it is a local path.  If it is remote storage then there is no need to create the directory.

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+                create_dir_all(&db_local_path).unwrap();
+            }
+            CreateDatabaseMode::ExistOk => {
+                if !exists {
+                    let db_local_path = to_local_path(&db_path);
+                    create_dir_all(&db_local_path).unwrap();
+                }
```
westonpace, line 148: Same as above, only call `create_dir_all` if local.

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+            }
+            CreateDatabaseMode::Overwrite => {
+                if exists {
+                    self.drop_database(&request.name).await?;
+                }
+                let db_local_path = to_local_path(&db_path);
+                create_dir_all(&db_local_path).unwrap();
```
westonpace, line 155: Same as above, only call `create_dir_all` if local

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        let connect_request = ConnectRequest {
+            uri: db_uri,
+            api_key: None,
+            region: None,
+            host_override: None,
+            #[cfg(feature = "remote")]
+            client_config: Default::default(),
+            read_consistency_interval: None,
+            storage_options: HashMap::new(),
```
westonpace, line 169: How should we populate `storage_options` for a database?  Should we copy the `storage_options` from the catalog?  Or should we pass in the `database_options` from the request?

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    async fn open_database(&self, request: OpenDatabaseRequest) -> Result<Arc<dyn Database>> {
+        let db_path = self.database_path(&request.name);
+        let db_path_str = to_local_path(&db_path);
+
+        let exists = Path::new(&db_path_str).exists();
+        if !exists {
+            return Err(Error::DatabaseNotFound { name: request.name });
+        }
```
westonpace, line 184: Same thought as above regarding local paths

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        let db_uri = format!("file:///{}/{}", self.base_path, request.name);
```
westonpace, line 186: We should do `self.database_path`?

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+        let connect_request = ConnectRequest {
+            uri: db_uri,
+            api_key: None,
+            region: None,
+            host_override: None,
+            #[cfg(feature = "remote")]
+            client_config: Default::default(),
+            read_consistency_interval: None,
+            storage_options: HashMap::new(),
```
westonpace, line 195: Same question as above.  Do we use `storage_options` from the database?  Or do we use `database_options` from open request?

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+        };
+
+        Ok(Arc::new(
+            ListingDatabase::connect_with_options(&connect_request).await?,
+        ))
+    }
+
+    async fn rename_database(&self, _old_name: &str, _new_name: &str) -> Result<()> {
+        Err(Error::NotSupported {
+            message: "rename_database is not supported in LanceDB OSS".to_string(),
```
westonpace, line 205:
```suggestion
            message: "rename_database is not supported in LanceDB OSS yet".to_string(),
```

We can probably add this in the future.  The `object_store::ObjectStore` has a `rename` method (defaults to copy-then-delete but is specialized for most object stores)

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,560 @@
...
+
+    query_string: Option<String>,
+
+    pub(crate) uri: String,
+
+    base_path: ObjectStorePath,
+
+    pub(crate) store_wrapper: Option<Arc<dyn WrappingObjectStore>>,
+
+    storage_options: HashMap<String, String>,
```
westonpace, line 37: The `storage_options` that get passed in through `connect` are already used to create `object_store` but not used elsewhere.  Do we want to also pass these `storage_options` in to the database when we open / create a database?

Same question for `store_wrapper`.

westonpace (commented):
This looks great.  I left a few minor suggestions.  I think they summarize into three categories.

* At some point (doesn't have to be this PR) we need to figure out a better story for `catalog_options`, `database_options`, and `storage_options`.  I think we might want to start phasing out `storage_options` and say that `storage_options` are just a specific set of `database_options` and `catalog_options` that are used by the listing dataset / listing catalog.
* There's a few things in here that won't work for remote storage.  We can address them later and add remote storage tests at the same time if you'd rather (and just say remote storage isn't supported yet) or you can address them now and hopefully reduce work when we do add remote storage tests.
* The `file_stem` issue could lead to bugs later on I think

rust/lancedb/src/catalog/listing.rs
yanghua: Good catch. To be unified, do not use `file_stem`. Is that OK?

rust/lancedb/src/catalog/listing.rs
yanghua: filed a ticket: https://github.com/lancedb/lancedb/issues/2166

rust/lancedb/src/catalog/listing.rs
yanghua: If I change:

```
let db_path_str = to_local_path(&db_path);
let exists = Path::new(&db_path_str).exists();
```

to 

```
let exists = self.object_store.exists(&db_path).await.unwrap();
```

The `exists` variable is not correct. It's false for `CreateDatabaseMode::Overwrite`

rust/lancedb/src/catalog/listing.rs
yanghua: > Should we copy the storage_options from the catalog?
I'd prefer this option.

rust/lancedb/src/catalog/listing.rs
yanghua: we use `storage_options`.

rust/lancedb/src/catalog/listing.rs
yanghua: Actually, I am not sure if we get an instance of `ObjectStore`, do we still need `storage_options` and `store_wrapper` in known scenario?

rust/lancedb/src/catalog.rs
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+//! Catalog implementation for managing databases
+
+mod listing;
```
westonpace, line 6:
```suggestion
pub mod listing;
```

rust/lancedb/src/catalog/listing.rs
```diff
@@ -0,0 +1,561 @@
...
+impl ListingCatalog {
+    /// Try to create a local directory to store the lancedb dataset
+    pub fn try_create_dir(path: &str) -> core::result::Result<(), std::io::Error> {
+        let path = Path::new(path);
+        if !path.try_exists()? {
+            create_dir_all(path)?;
+        }
+        Ok(())
+    }
+
```
westonpace, line 54:
```suggestion

pub fn uri(&self) -> &str {
  &self.uri
}

```

westonpace (commented):
Some thoughts.

1. exists not working

This was surprising and I had to investigate.  Since `object_store` is emulating a cloud object store it only returns `true` if a file exists (directories do not exist in cloud object stores).  So using `exists` with `object_store` won't work.  One possibility is to call remove_dir_all if overwrite is specified and then ignore all errors.

2. query_string

If the user is using [`dynamodb`](https://lancedb.github.io/lance/read_and_write.html#concurrent-writer-on-s3-using-dynamodb) then we need to open the table with a query string:

```
lance.dataset("s3+ddb://my-bucket/mydataset?ddbTableName=mytable")
```

In order to make this work in `listing_database` we propagate any query string from the database.  So if the user opens the database...

```
db = lancedb.connect("s3+ddb://my-bucket/my-db?ddbTableName=mytable")
db.open_table("foo")
```

This will open the table `s3+ddb://my-bucket/my-db/foo?ddbTableName=mytable`

If we want this to work with listing catalog then we need to utilize query string when opening the database.

This feature is only needed for dynamodb which is only needed for S3 with concurrent writers.

We could choose to add later if you prefer.

3. store_wrapper

This feature is used when customers want to wrap the object store.  We use this for traceability purposes in our enterprise software.  We wrap the object store with a proxy that tracks how many bytes have been read/written from object store and also logs the reads & writes for debugging.

If we want this feature to work there would need to be some way to set the wrapper on a catalog.  I'm not actually sure how we set the wrapper on a listing database :laughing: .  I think maybe it requires using `patch_with_store_wrapper` which seems complicated.

4. storage_options / database_options / catalog_options

There are three kinds of options.

catalog_options -> options to control the catalog (e.g. username/password to login to catalog, what port is catalog using if it is a remote catalog, what storage options to use if the catalog is a listing catalog, etc.)

database_options -> options to control the database (e.g. what storage version should the database use when creating tables, should the database create tables with advanced features like stable row ids, what storage options should a listing database use when connecting to object store)

storage_options -> These are a subset of catalog_options and database_options that only apply when using the listing catalog / listing database (i.e. these are not needed for an HTTP catalog).  These options control the connection to the object store.  So if the URI is s3:// then storage_options are needed to know the username password.

Also, there are options that should be set when the database/catalog is created and options that need to be set every time the database/catalog is opened.

I don't think we have a way right now, for example, to set database options that are not storage options (e.g. `new_table_data_storage_version`)

Let's not worry about it today.  I will think on this and fix in a follow-up PR.  I need to clarify all of this anyways in documentation.

## Finishing this PR

Let's do the following.  Let's only support local storage for now.  We can add remote storage later.

* Make the listing catalog `pub` (see suggestion)
* Remove query_string
* Remove store_wrapper
* Add a public accessor function for uri (see suggestion)

Once this is done we will merge.  We need the next follow-ups:

* Add "connect_catalog" method or add some way to connect to a catalog via a URI (@yanghua)
* Add object storage support (not urgent, can be added when someone needs it)
* Clarify options (@westonpace)

Does this sound good?

yanghua:
> This was surprising and I had to investigate. Since object_store is emulating a cloud object store it only returns true if a file exists (directories do not exist in cloud object stores). So using exists with object_store won't work. One possibility is to call remove_dir_all if overwrite is specified and then ignore all errors.

Oh.. It's true, last year I was working for a project to transfer object store's API semantic to file system API semantic. "DIR" semantic needs to use a `/a/b/c/`(end with a `/`) object to mock. And related apis, e.g. `exicts`, `isdir`, `rm recurisively` all need special adaptation.

yanghua:
Hi @westonpace I just fixed a lint issue, please help to re-trigger the full CI again. thanks~

yanghua:
@westonpace I tried to fix the windows related CI issue. Please help to re-trigger it again.

BTW, why does CI need to take permission control? Because my first PR in this repo causes this?

I'm really sorry to keep bothering you because of this.

westonpace:
Yeah, it's just a first PR thing.  I can chase down any remaining Windows issues today if needed.

yanghua:
> Yeah, it's just a first PR thing. I can chase down any remaining Windows issues today if needed.

OK, then, please, I'm just afraid that my failed attempts will keep bothering you. If it's convenient for you, please help fix it, thank you very much.

westonpace:
> Yeah, it's just a first PR thing. I can chase down any remaining Windows issues today if needed.

Hah, I give up for now.  I'm just going to disable the tests on windows so we can move forwards.  I'll boot up my windows desktop soon so I don't have to test in CI.

westonpace (approved):

westonpace:
Thanks for all the help!

PR #2151, opened 2025-02-25
feat!: revert query limit to be unbounded for scans

### LINKED ISSUE #1852: bug(python): Query limit([-1 | 0 | None]) not being respected

### LINKED ISSUE #1983: bug(python): plain search limited to 10 by default (regression)

### LINKED ISSUE #2141: bug(node): `query` returns `10` rows w/o `limit`
## Related

- https://github.com/lancedb/lancedb/issues/1852#issuecomment-2489837804
- #1983

### LanceDB version

v0.16.0

## Problem

Documentation for `.query()` says:

> Queries allow you to search your existing data.  By default the query will return all the data in the table in no particular order.

Notice: `all the data`.

However, it returns `10` rows by default

1. Run `table.query()` on a table w/ more than 10 rows.
2. Observe the response; the response for the table has 10 rows

wjones127:
In earlier PRs (#1886, #1191) we made the default limit 10 regardless of the query type. This was confusing for users and in many cases a breaking change. Users would have queries that used to return all results, but instead only returned the first 10, causing silent bugs.

Part of the cause was consistency: the Python sync API seems to have always had a limit of 10, while newer APIs (Python async and Nodejs) didn't.

This PR sets the default limit only for searches (vector search, FTS), while letting scans (even with filters) be unbounded. It does this consistently for all SDKs.

Fixes #1983
Fixes #1852
Fixes #2141

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -666,11 +666,11 @@ describe("When creating an index", () => {
     expect(fs.readdirSync(indexDir)).toHaveLength(1);
 
     for await (const r of tbl.query().where("id > 1").select(["id"])) {
-      expect(r.numRows).toBe(10);
+      expect(r.numRows).toBe(298);
     }
     // should also work with 'filter' alias
     for await (const r of tbl.query().filter("id > 1").select(["id"])) {
-      expect(r.numRows).toBe(10);
+      expect(r.numRows).toBe(298);
     }
   });
 
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -110,7 +110,7 @@ class Query(pydantic.BaseModel):
     full_text_query: Optional[Union[str, dict]] = None
 
     # top k results to return
-    k: int
+    k: Optional[int] = None
 
     # # metrics
     metric: str = "L2"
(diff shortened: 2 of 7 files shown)


westonpace (approved):
Do we have a test case verifying the default limit is 10 for FTS?

wjones127:
> Do we have a test case verifying the default limit is 10 for FTS?

We do for remote: https://github.com/lancedb/lancedb/blob/ecda0d8f2867a3b4c48fc78f9f81ddd03ef3fce5/rust/lancedb/src/remote/table.rs#L1389-L1390

I've added one for local.

PR #2165, opened 2025-02-28
chore: fix types and remove outdated pyi file

msu-reevo:


python/python/lancedb/_lancedb.pyi
wjones127, line 1: Why are you removing these? They correspond to the classes written in Rust and exported here: https://github.com/lancedb/lancedb/blob/dfe4ba8dad580232e4450903a5d1b239b19f8173/python/src/lib.rs#L30-L40

msu-reevo: @wjones127 sorry didn't mean to open a PR against upstream yet, this was just a temporary work branch to PR against my fork for easier viewing. I was planning on adding the rust ones back. Was removing the Python ones like `connect` first where the definitions don't match up in a test to see how many Python issues there were from this file.

I'll close this PR and reopen once the cleanup is in a better spot. But yeah there's a lot of pyright issues coming up when I run it.

PR #2167, opened 2025-02-28
fix(python): typing

msu-reevo:
@wjones127 is there a standard way you guys setup your virtualenv? I can either relist all the dependencies in the pyright precommit section, or specify a venv, or the user has to be in the virtual environment when they run git commit. If the venv location was standardized or a python manager like `uv` was used it would be easier to avoid duplicating the pyright dependency list.

Per your suggestion, in `pyproject.toml` I added in all the passing files to the `includes` section.

For ruff I upgraded the version and removed "TCH" which doesn't exist as an option.

I added a `pyright_report.csv` which contains a list of all files sorted by pyright errors ascending as a todo list to work on.

I fixed about 30 issues in `table.py` stemming from str's being passed into methods that required a string within a set of string Literals by extracting them into `types.py`

Can you verify in the rust bridge that the schema should be a property and not a method here? If it's a method, then there's another place in the code where `inner.schema` should be `inner.schema()`
``` python
class RecordBatchStream:
    @property
    def schema(self) -> pa.Schema: ...
```

Also unless the `_lancedb.pyi` file is wrong, then there is no `__anext__` here for `__inner` when it's not an `AsyncGenerator` and only `next` is defined:
``` python
    async def __anext__(self) -> pa.RecordBatch:
        return await self._inner.__anext__()
        if isinstance(self._inner, AsyncGenerator):
            batch = await self._inner.__anext__()
        else:
            batch = await self._inner.next()
        if batch is None:
            raise StopAsyncIteration
        return batch
```
in the else statement, `_inner` is a `RecordBatchStream`
```python
class RecordBatchStream:
    @property
    def schema(self) -> pa.Schema: ...
    async def next(self) -> Optional[pa.RecordBatch]: ...
```


wjones127:
> Can you verify in the rust bridge that the schema should be a property and not a method here? If it's a method, then there's another place in the code where inner.schema should be inner.schema()

`#[getter]` means property

https://github.com/lancedb/lancedb/blob/dfe4ba8dad580232e4450903a5d1b239b19f8173/python/src/arrow.rs#L38-L41

wjones127:
> Also unless the _lancedb.pyi file is wrong, then there is no __anext__ here for __inner when it's not an AsyncGenerator and only next is defined:

They are defined here:

https://github.com/lancedb/lancedb/blob/dfe4ba8dad580232e4450903a5d1b239b19f8173/python/src/arrow.rs#L43-L58

msu-reevo:
> > @wjones127 is there a standard way you guys setup your virtualenv? I can either relist all the dependencies in the pyright precommit section, or specify a venv, or the user has to be in the virtual environment when they run git commit. If the venv location was standardized or a python manager like `uv` was used it would be easier to avoid duplicating the pyright dependency list.
> 
> We don't have a standard place. Can we instead just start with pyright being run in CI instead of as precommit?
> 
> And also add the pyright command into the [Makefile](https://github.com/lancedb/lancedb/blob/main/python/Makefile) and mention it in [development docs](https://github.com/lancedb/lancedb/blob/main/python/CONTRIBUTING.md)?

First thing, thanks for the quick replies! Appreciate it.

Yes I will switch it to CI first and add it to the Makefile and docs

msu-reevo:
> > Also unless the _lancedb.pyi file is wrong, then there is no **anext** here for __inner when it's not an AsyncGenerator and only next is defined:
> 
> They are defined here:
> 
> https://github.com/lancedb/lancedb/blob/dfe4ba8dad580232e4450903a5d1b239b19f8173/python/src/arrow.rs#L43-L58

Got it thanks! Then the `@property` I added will fix the `pyi` file. Thanks for confirming, I'll try to read through more of the Rust code as I get used to the codebase a bit more.

msu-reevo:
> > Also unless the _lancedb.pyi file is wrong, then there is no **anext** here for __inner when it's not an AsyncGenerator and only next is defined:
> 
> They are defined here:
> 
> https://github.com/lancedb/lancedb/blob/dfe4ba8dad580232e4450903a5d1b239b19f8173/python/src/arrow.rs#L43-L58

Great! Will update the pyi file to reflect.

msu-reevo:
Finished the items you requested above. It looks like you'll have to approve the workflow run.

Let me know if you want me to change anything.

msu-reevo:
@wjones127 Had some time to update this today. It ran here and typechecks work but didn't see a way to pip install the dependencies without it building lance and taking 20mins. https://github.com/ReevoAI/lancedb/actions/runs/13690247173/job/38282084983

Not sure if you know a better way to do this without manually listing out the dependencies which seems prone to error.

wjones127:
I couldn't push to your branch, so I created a PR here: 

https://github.com/ReevoAI/lancedb/pull/3

There's a quick script in there that parses out the requirements. That should run a lot more quickly.

msu-reevo:
@wjones127 Thanks I pulled that in and cleaned it up a little bit. The pyright run passes here: https://github.com/ReevoAI/lancedb/actions/runs/13742435044. It doesn't provide a warning but I think that's just because we're now not building the rust package.
```
/home/runner/work/lancedb/lancedb/python/python/lancedb/index.py
  /home/runner/work/lancedb/lancedb/python/python/lancedb/index.py:7:6 - warning: Import "._lancedb" could not be resolved from source (reportMissingModuleSource)
```

let me know if you're good to merge this in to freeze the typing baseline or if you want something else changed.

PR #2170, opened 2025-03-03
feat: add ColPali embedding support with MultiVector type

MagnusS0:
This PR adds ColPali support with ColPaliEmbeddings class (tagged "colpali") using ColQwen2.5 for multi-vector text/image embeddings. Also added MultiVector Pydantic type to handle the vector lists. 

I've added some integration test for the embedding model and some unit test for the new Pydantic type. Could be a template for other ColPali variants as well. or until transformers🤗 starts supporting it.  

Still `TODO`:

- [ ] Documentation
- [ ] Add an example

_Could also allow Image as query, but didn't work well when testing it._

[ColPali-Engine](https://github.com/illuin-tech/colpali) version:
0.3.9.dev17+g3faee24

## Summary by CodeRabbit

- **New Features**
  - Introduced support for ColPali-based multimodal multi-vector embeddings for both text and images.
  - Added a new embedding class for generating multi-vector embeddings, configurable for various model and processing options.
  - Added a new Pydantic type for multi-vector embeddings, supporting validation and schema generation for lists of fixed-dimension vectors.

- **Bug Fixes**
  - Ensured proper asynchronous index creation in query tests for improved reliability.

- **Tests**
  - Added integration tests for ColPali embeddings, including text-to-image search and validation of multi-vector fields.
  - Added comprehensive tests for the new multi-vector Pydantic type, covering schema, validation, and default value behavior.

- **Chores**
  - Updated optional dependencies to include the ColPali engine.
  - Added utility to check for availability of flash attention support.

### DIFF
--- a/python/python/lancedb/embeddings/colpali.py
+++ b/python/python/lancedb/embeddings/colpali.py
@@ -1,0 +1,217 @@
+from functools import lru_cache
+from typing import List, Union, TYPE_CHECKING
+import numpy as np
+from PIL import Image
+import torch
+import requests
+import io
+
+from ..util import attempt_import_or_raise
+from .base import EmbeddingFunction
+from .registry import register
+from .utils import TEXT, IMAGES, is_flash_attn_2_available
+
+
+@register("colpali")
+class ColPaliEmbeddings(EmbeddingFunction):
+    """
+    An embedding function that uses the ColPali engine for multimodal multi-vector embeddings.
+
+    This embedding function supports ColQwen2.5 models, producing multivector outputs
+    for both text and image inputs. The output embeddings are lists of vectors, each
+    vector being 128-dimensional by default, represented as List[List[float]].
+
+    Parameters
+    ----------
+    model_name : str
+        The name of the model to use (e.g., "Metric-AI/ColQwen2.5-3b-multilingual-v1.0").
+    device : str
+        The device for inference (default "cuda:0").
+    dtype : str
+        Data type for model weights (default "bfloat16").
+    use_token_pooling : bool
+        Whether to use token pooling to reduce embedding size (default True).
+    pool_factor : int
+        Factor to reduce sequence length if token pooling is enabled (default 2).
+    """
+
+    model_name: str = "Metric-AI/ColQwen2.5-3b-multilingual-v1.0"
+    device: str = "cuda:0"
+    dtype: str = "bfloat16"
+    use_token_pooling: bool = True
+    pool_factor: int = 2
+
+    _model = None
+    _processor = None
+    _token_pooler = None
+    _vector_dim = None
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        (
+            self._model,
+            self._processor,
+            self._token_pooler,
+        ) = self._load_model(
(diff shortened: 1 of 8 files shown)


changhiskhan:
@BubbleCal is this using XTR?

BubbleCal:
> @BubbleCal is this using XTR?

I think this is just for generating embeddings, it can benefit from XTR

MagnusS0:
I had not read the XTR paper but from a quick look, looks like it require to retrain the ColPali model with a new objective function.  It looks really promising though—definitely worth exploring further! 

For this PR, I’m focused on generating embeddings with the current ColPali setup. Hoping to wrap it up and get it passing checks over the next weekend.

MagnusS0:
@BubbleCal This is ready for a review now. The failing test on Windows seems unrelated to my changes. Please take a look when you get a chance, if it looks good, I’ll start on the documentation next.

Also related to XTR, looking at this lanceDB [blog post](https://lancedb.github.io/lancedb/notebooks/Multivector_on_LanceDB/)  it mentions XTR is used with ColPali, so is this how all multi-vector retrievals are fetched? 
 
The max-sim function in the [docs](https://lancedb.github.io/lancedb/search/#multivector-type) seems to reference the standard function and not the XTR approach.

BubbleCal:
@MagnusS0 sorry for missing this, reviewing it

python/python/lancedb/embeddings/colpali.py
```diff
@@ -0,0 +1,256 @@
...
+        if quantization_config is not None:
+            if not isinstance(quantization_config, transformers.BitsAndBytesConfig):
+                raise ValueError("quantization_config must be a BitsAndBytesConfig")
+
+        if dtype == "bfloat16":
+            torch_dtype = torch.bfloat16
+        elif dtype == "float16":
+            torch_dtype = torch.float16
+        else:
+            torch_dtype = torch.float32
```
BubbleCal, line 99: can float64 be supported?

python/python/lancedb/pydantic.py
```diff
@@ -152,6 +152,104 @@ def __modify_schema__(cls, field_schema: Dict[str, Any]):
     return FixedSizeList
 
 
+def MultiVector(
```
BubbleCal, line 155: this is nice!

BubbleCal (approved):

python/python/lancedb/embeddings/colpali.py
MagnusS0: Yes, technically float64 can be supported. I wouldn't recommend running it in that high precision though, would likely very quickly lead to OOM. Or is there a specific use-case?

BubbleCal: let's add it if it's very easy, i don't know whether there's a real world use case, just for covering any possible use cases

BubbleCal:
@MagnusS0 need to resolve the conflicts, I will merge this after that, let me know if you need help

MagnusS0:
@BubbleCal conflicts resolved and also added some of the improvements suggested by @coderabbitai

BubbleCal:
Thanks for the great work!

PR #2177, opened 2025-03-04
feat(rust): add connect_catalog method to support connect catalog via url

yanghua:


yanghua:
@westonpace Please take a look when you have time.

rust/lancedb/src/connection.rs
```diff
@@ -764,6 +765,110 @@ pub fn connect(uri: &str) -> ConnectBuilder {
...
+
+    pub fn region(mut self, region: &str) -> Self {
+        self.request.region = Some(region.to_string());
+        self
+    }
+
+    pub fn host_override(mut self, host_override: &str) -> Self {
+        self.request.host_override = Some(host_override.to_string());
+        self
+    }
```
westonpace, line 804: Let's not add these until we need them.  I'd prefer the only options here be options that are likely to apply to all catalogs.  If I could go back in time we probably wouldn't have them on the DB connect.

rust/lancedb/src/connection.rs
```diff
@@ -764,6 +765,110 @@ pub fn connect(uri: &str) -> ConnectBuilder {
...
+        mut self,
+        pairs: impl IntoIterator<Item = (impl Into<String>, impl Into<String>)>,
+    ) -> Self {
+        for (key, value) in pairs {
+            self.request
+                .storage_options
+                .insert(key.into(), value.into());
+        }
+        self
+    }
```
westonpace, line 829: Instead of these two can we just have one method `catalog_options` that works like `ConnectBuilder::database_options`

We will need new trait `CatalogOptions` that can look just like `DatabaseOptions`.  We don't need any implementations of `CatalogOptions` today.  Don't worry about `storage_options` today.  I will add these soon (I have an idea for what it should look like).

rust/lancedb/src/connection.rs
```diff
@@ -764,6 +765,110 @@ pub fn connect(uri: &str) -> ConnectBuilder {
...
+        self.request
+            .storage_options
+            .insert("aws_secret_access_key".into(), aws_creds.secret_key.clone());
+        if let Some(token) = &aws_creds.token {
+            self.request
+                .storage_options
+                .insert("aws_session_token".into(), token.clone());
+        }
+        self
+    }
```
westonpace, line 852: Let's leave these two out as well.

westonpace (changes requested):
I think you did more work than needed :)

A lot of the things on `ConnectBuilder` I don't think we want to do in `CatalogConnectBuilder`.  They made sense when we got started because there were only ever going to be two implementations.

However, if there are going to be many implementations of catalogs and databases then we can't list all the options on the builders.  Instead, the builders should just take flat options objects.

The actual connect part looks good!

PR #2186, opened 2025-03-06
feat: upgrade Lance to v0.24.0-beta.3

wjones127:
https://github.com/lancedb/lance/releases/tag/v0.24.0-beta.3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,7 +2571,7 @@ dependencies = [
 [[package]]
 name = "fsst"
 version = "0.24.0"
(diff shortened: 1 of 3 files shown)


LuQQiu (approved):

westonpace (commented):
How should we handle the `pylance` version in `pyproject.toml` now?  Since we are mostly split from `pylance` we probably don't _need_ to keep them in lockstep.  Should we?  Or should we just bump on major changes?

wjones127:
> How should we handle the `pylance` version in `pyproject.toml` now? Since we are mostly split from `pylance` we probably don't _need_ to keep them in lockstep. Should we? Or should we just bump on major changes?

Let's loosen to a simple >= pin. We are pretty decoupled, so unless we have a serious breaking change in the table format, I don't think we need to upgrade the pin.

PR #2194, opened 2025-03-08
fix: delete tables from DDB on drop_all_tables

wkalt:
Prior to this commit, issuing drop_all_tables on a listing database with an external manifest store would delete physical tables but leave references behind in the manifest store. The table drop would succeed, but subsequent creation of a table with the same name would fail with a conflict.

With this patch, the external manifest store is updated to account for the dropped tables so that dropped table names can be reused.

### DIFF
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -336,6 +336,9 @@ impl ListingDatabase {
         for name in names {
             let dir_name = format!("{}.{}", name, LANCE_EXTENSION);
             let full_path = self.base_path.child(dir_name.clone());
+
+            commit_handler.delete(&full_path).await?;
+
             self.object_store
                 .remove_dir_all(full_path.clone())
                 .await
@@ -347,8 +350,6 @@ impl ListingDatabase {
                     },
                     _ => Error::from(err),
                 })?;
-
-            commit_handler.delete(&full_path).await?;
         }
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


rust/lancedb/src/database/listing.rs
```diff
@@ -322,6 +322,36 @@ impl ListingDatabase {
...
+                .map_err(|err| match err {
+                    // this error is not lance::Error::DatasetNotFound, as the method
+                    // `remove_dir_all` may be used to remove something not be a dataset
+                    lance::Error::NotFound { .. } => Error::TableNotFound {
+                        name: name.to_owned(),
+                    },
+                    _ => Error::from(err),
+                })?;
+
+            commit_handler.delete(&full_path).await?;
```
wkalt, line 351: is the ordering right?

westonpace: It does seem a little backwards.  The default delete impl is a no-op.  The only commit handler with a custom delete handler is dynamodb and it drops the table.  So the only thing this could be doing is deleting records from dynamodb.  I think it would be safer to do this _before_ deleting the files.

If the file op succeeds but the dynamo op fails we will be in a weird corrupt state with failures.
If the dynamo op succeeds but the file op fails we will have wasted files in S3

Neither are great but the latter seems more harmless and could be manually repaired (by deleting the data).

westonpace (approved):
I agree about the order.  Might be nicer to swap those.  No concerns otherwise.

rust/lancedb/src/database/listing.rs
wkalt: thanks, will update.

PR #2203, opened 2025-03-11
fix: make pylance a core dependency

erikgaas:
Lance is used as a normal dependency in embeddings/utils.py https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/embeddings/utils.py#L19

This is causing some of our CI tests to fail, here I am making it a core dependency. I pinned it to the same version as the tests.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -9,6 +9,7 @@ dependencies = [
     "pydantic>=1.10",
     "packaging",
     "overrides>=0.7",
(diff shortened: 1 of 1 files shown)


wjones127:
Because of its size, we’d like to make pylance optional. I’ll make a PR tomorrow to fix this.

wjones127:
Duplicate of #2204

PR #2208, opened 2025-03-11
feat(node): parse arrow types in `alterColumns()`

### LINKED ISSUE #2185: Feature: accept arrow data type in `alterColumns()`
### SDK

Node

### Description

Right now, we accept a string. This works fine for primitive types, but it doesn't work for casting nested types. We should support casting things like vectors to new types.

wjones127:
Previously, users could only specify new data types in `alterColumns` as strings:

```ts
await tbl.alterColumns([
  path: "price",
  dataType: "float"
]);
```

But this has some problems:

1. It wasn't clear what were valid types
2. It was impossible to specify nested types, like lists and vector columns.

This PR changes it to take an Arrow data type, similar to how the Python API works. This allows casting vector types:

```ts
await tbl.alterColumns([
  {
    path: "vector",
    dataType: new arrow.FixedSizeList(
      2,
      new arrow.Field("item", new arrow.Float16(), false),
    ),
  },
]);
```

Closes #2185

### DIFF
--- a/rust/lancedb/src/utils.rs
+++ b/rust/lancedb/src/utils.rs
@@ -166,13 +166,16 @@ pub fn supported_vector_data_type(dtype: &DataType) -> bool {
 
 /// Note: this is temporary until we get a proper datatype conversion in Lance.
 pub fn string_to_datatype(s: &str) -> Option<DataType> {
-    let mut data_type = serde_json::Value::String(s.to_string());
-    dbg!(&data_type);
+    let mut data_type: serde_json::Value = {
+        if let Ok(data_type) = serde_json::from_str(s) {
+            data_type
+        } else {
+            serde_json::Value::String(s.to_string())
+        }
+    };
     if data_type.get("type").is_none() {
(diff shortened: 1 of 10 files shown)


rust/lancedb/src/utils.rs
```diff
@@ -166,10 +166,18 @@ pub fn supported_vector_data_type(dtype: &DataType) -> bool {
...
 pub fn string_to_datatype(s: &str) -> Option<DataType> {
-    let data_type = serde_json::Value::String(s.to_string());
-    let json_type =
-        serde_json::Value::Object([("type".to_string(), data_type)].iter().cloned().collect());
-    let json_type: JsonDataType = serde_json::from_value(json_type).ok()?;
+    let mut data_type: serde_json::Value = {
+        if let Ok(data_type) = serde_json::from_str(s) {
+            data_type
+        } else {
+            serde_json::Value::String(s.to_string())
```
BubbleCal, line 173: Nit: make a map with key "type" here then no need to check this below

BubbleCal (approved):

PR #2209, opened 2025-03-11
fix: make pylance optional again

wjones127:
The two remaining blockers were:

* A method `with_embeddings` that was deprecated a year ago
* A typecheck for `LanceDataset`


westonpace (commented):
Can we add a CI check that runs tests (or some basic subset of tests) without having pylance installed?

westonpace (approved):
Thanks for adding the test!

PR #2213, opened 2025-03-12
refactor: rework catalog/database options

westonpace:
The `ConnectRequest` has a set of properties that only make sense for listing databases / catalogs and a set of properties that only make sense for remote databases.

This PR reduces all options to a single `HashMap<String, String>`.  This makes it easier to add new database / catalog implementations and makes it clearer to users which options are applicable in which situations.

I don't believe there are any breaking changes here.  The closest thing is that I placed the `ConnectBuilder` methods `api_key`, `region`, and `host_override` behind a `remote` feature gate.  This is not strictly needed and I could remove the feature gate but it seemed appropriate.  Since using these methods without the remote feature would have been meaningless I don't feel this counts as a breaking change.

We could look at removing these methods entirely from the `ConnectBuilder` (and encouraging users to use `RemoteDatabaseOptions` instead) but I'm not sure how I feel about that.

Another approach we could take is to move these methods into a `RemoteConnectBuilderExt` trait (and there could be a similar `ListingConnectBuilderExt` trait to add methods for the listing database / catalog).

For now though my main goal is to simplify `ConnectRequest` as much as possible (I see this being part of the key public API for database / catalog integrations, similar to the `BaseTable`, `Catalog`, and `Database` traits and I'd like it to be simple).

### DIFF
--- a/rust/lancedb/src/remote/db.rs
+++ b/rust/lancedb/src/remote/db.rs
@@ -14,8 +14,8 @@ use serde::Deserialize;
 use tokio::task::spawn_blocking;
 
 use crate::database::{
-    CreateTableData, CreateTableMode, CreateTableRequest, Database, OpenTableRequest,
-    TableNamesRequest,
+    CreateTableData, CreateTableMode, CreateTableRequest, Database, DatabaseOptions,
+    OpenTableRequest, TableNamesRequest,
 };
 use crate::error::Result;
 use crate::table::BaseTable;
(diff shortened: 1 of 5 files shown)


westonpace:
@yanghua do you want to take a look?

yanghua:
> @yanghua do you want to take a look?

OK, I will take a look soon.

rust/lancedb/src/remote/db.rs
```diff
@@ -54,6 +54,115 @@ impl ServerVersion {
     }
 }
 
+pub const OPT_REMOTE_PREFIX: &str = "remote_database_";
+pub const OPT_REMOTE_API_KEY: &str = "remote_database_api_key";
+pub const OPT_REMOTE_REGION: &str = "remote_database_region";
+pub const OPT_REMOTE_HOST_OVERRIDE: &str = "remote_database_host_override";
```
yanghua, line 60: use `OPT_REMOTE_PREFIX ` to connect "api_key", "region"?  Not sure if it's a good practise in RUST.

yanghua (approved):
LGTM

rust/lancedb/src/remote/db.rs
wjones127: Like this?
```suggestion
pub const OPT_REMOTE_API_KEY: &str = concat!(OPT_REMOTE_PREFIX, "key");
pub const OPT_REMOTE_REGION: &str = concat!(OPT_REMOTE_PREFIX, "region");
pub const OPT_REMOTE_HOST_OVERRIDE: &str = concat!(OPT_REMOTE_PREFIX, "host_override");
```

wjones127 (approved):

PR #2216, opened 2025-03-12
feat(rust): design openapi for remote catalog

yanghua:


jackye1995:
Have we thought about using OpenAPI vs other protocols like gRPC? Because Lance is defined using protobuf, it seems like protocols would be an easier choice compared to Restful OpenAPI. Also you get some perf benefits in gRPC.

yanghua:
> Have we thought about using OpenAPI vs other protocols like gRPC?

No, we have not thought. But, it's really a good point of view.

> Because Lance is defined using protobuf, it seems like gRPC would be an easier choice compared to Restful OpenAPI.

ASAIK, Lance uses protobuf, which is mainly used for ser/deserialization. This information is mainly related to some metadata stored by Lance. The catalog is higher up. And it seems that the way Catalog works has little to do with the use of protobuf.

> Also you get some perf benefits in gRPC.

In terms of performance advantages, the advantages of gRPC are obvious. This is also a classic topic for selecting binary protocols and text protocols.

Although there is no in-depth research and comparison, we can also talk about the original intention of choosing the restful protocol.

* performance: the catalog interface is lightweight, and the request frequency is not high;
* maintainability and complexity: Text protocol is more friendly to analysis and localization problems;
* popularity(in same field): We have seen like iceberg, unity catalog, grivatino, polaris, these catalog frameworks are chosen restful spec(I don't know if this is a coincidence, or if they saw some potential problems with HMS choosing the Thrift protocol?)

jackye1995:
I cannot speak about Gravitino. Polaris == Iceberg, Unity also exposes HMS & Iceberg REST APIs as the main way for engine integration, so technically it's just Iceberg and Gravitino at this point. I guess I can first talk about why I started to think about gRPC.

If we look into the Iceberg REST catalog (IRC) spec, its most important API, LoadTable, is basically returning a response that is exactly the same schema as the Iceberg metadata JSON file. This is probably one of the few APIs that performance actually matters (the other one is probably UpdateTable).

That means by going through that REST API, you can directly get the full Iceberg metadata from the IRC server, and potentially be faster than reading a JSON file on disk. For example, in cases like streaming, the JSON file on disk could be very large and slow to read, and the IRC server could do smart things like caching and returning just the subset of the information while still have it being a valid Iceberg metadata. So JSON-based text protocol definitely makes sense for Iceberg from that perspective to exchange JSON-based Iceberg metadata and convert Iceberg's original JSON based metadata definition to become a wire protocol. And with that, REST/OpenAPI is probably the easiest go-to option for a JSON-based protocol.

But if we apply the same logic here, then we have basically lost the same optimization path in Lance, at least in the current design. The table metadata is just

```
    TableMetadata:
      type: object
      properties:
        schema: {type: object}
        version: {type: integer}
        location: {type: string}
```

which means we always have to jump one more hook to read the file at the specific location to get the actual full table metadata. We could potentially continue to use JSON-based text protocol, and include a binary field to return the actual Lance manifest content as serialized protobuf binary in the future to have the same optimization as IRC LoadTable, but (in addition to it being a bit ugly...) that would be quite inefficient, and at that point it seems like defining a gRPC start to make sense to also turn Lance from a on-disk format to a wire protocol.

But I agree with the point about "maintainability and complexity: Text protocol is more friendly to analysis and localization problems". Curious what others think @westonpace @SaintBacchus

jackye1995:
Thank you so much for the detailed analysis! @yanghua In that case I don't have any other questions!

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+          enum: [Create, ExistOk, Overwrite]
+        options:
+          type: object
+          additionalProperties: { type: string }
+
+    DatabaseMetadata:
+      type: object
+      properties:
+        name: {type: string}
+        tableCount: {type: integer}
```
jackye1995, line 238: why returning table count here? This is not commonly seen elsewhere, is there a specific use case?

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+        mode:
+          type: string
+          enum: [Create, ExistOk, Overwrite]
+        options:
+          type: object
+          additionalProperties: { type: string }
+
+    DatabaseMetadata:
+      type: object
+      properties:
```
jackye1995, line 236: I see in CreateDatabase you are returning a location field for the database, but it seems missing here?

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+        name: {type: string}
+        tableCount: {type: integer}
+
```
jackye1995, line 239: Typically there are free-form `properties` field for a database, do we plan to also add it here?

```
properties:
          type: object
          description:
            Properties stored on the database, if supported by the server.
            If the server does not support database properties, it should return null for this field.
            If database properties are supported, but none are set, it should return an empty object.
          additionalProperties:
            type: string
          example: { "owner": "Jack", "last_update_timestamp": "1234567890000" }
```

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+    delete:
+      tags: [Database]
+      summary: Delete database
+      parameters:
+        - $ref: '#/components/parameters/dbName'
+      responses:
+        204:
+          description: Database deleted
+
+    put:
```
jackye1995, line 70: regarding the verb here, I remember IRC uses POST but Gravitino uses PUT to express an update. Personally I am always in favor of POST because every system supports POST, GET & DELETE, and other verbs like PUT, PATCH gets strange behavior definitions in some systems. But I am not too strongly opinionated on this.

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+      summary: Delete all databases
+      responses:
+        204:
+          description: All databases deleted
+
+  /v1/databases/{dbName}:
+    get:
+      tags: [Database]
+      summary: Get database metadata
+      parameters:
```
jackye1995, line 51: I think this parameters section can be put before all the verb definitions so you don't have to repeat it in every verb:

```
/v1/databases/{dbName}:
    parameters:
        - $ref: '#/components/parameters/dbName'
    get:
        ...
```

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+      summary: Delete database
+      parameters:
+        - $ref: '#/components/parameters/dbName'
+      responses:
+        204:
+          description: Database deleted
+
+    put:
+      tags: [Database]
+      summary: Rename database
```
jackye1995, line 72: why is a put only used to express rename? Can't I update things like database URI?

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DatabaseMetadata'
+
+    delete:
+      tags: [Database]
+      summary: Delete database
+      parameters:
+        - $ref: '#/components/parameters/dbName'
```
jackye1995, line 65: I think we are missing a parameter `dropBehavior` to describe drop behavior like cascade, restrict, etc.

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/RenameRequest'
+      responses:
+        204:
+          description: Database renamed
```
jackye1995, line 83: For rename, we might also want to think a bit more about the route. 

IRC made the intentional choice of having a dedicated rename route, like `/v1/{prefix}/tables/rename` to rename tables, but this does not work for database because `/v1/{prefix}/databases/rename` can also suggest a database of name `rename`...

I also see other systems that do renamae as creating a destination database like

```
POST /v1/databases?from=fromDatabase
{ "name": "toDatabase" }
```

This approach requires making all the other fields optional.

Another way is to use the destination database route with POST verb:

```
POST /v1/databases/toDatabase?from=fromDatabase
{ }
```

That might be the most clean way, and that means we should use PUT for updating an existing database to differentiate the behavior.

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+        201:
+          description: Database created
+          headers:
+            Location:
+              description: URI of created database
+              schema: {type: string}
+
+    delete:
+      tags: [ Database ]
+      summary: Delete all databases
```
jackye1995, line 42: One thing I like about Gravitino's naming convention is that they use SQL verbs instead of REST verbs to describe the operation, for example use `AlterDatabase` instead of `UpdateDatabase`, `DropDatabase` instead of `DeleteDatabase`, I think that helps people with data background to understand the behavior better. What do you think?

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+      schema: {type: string}
+
+    tableName:
+      name: tableName
+      in: path
+      required: true
+      schema: {type: string}
+
+    startAfter:
+      name: startAfter
```
jackye1995, line 204: I think we should describe more about the behavior of this parameter. If we use this parameter, this means the response returned by the server must be in increasing lexicographical order. If the server cannot support such feature, it should fail the operation with unsupported error code.

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+          items: {type: string}
+        nextToken: {type: string}
+
+    CreateTableRequest:
+      type: object
+      required: [name, definition]
+      properties:
+        name: {type: string}
+        definition:
+          $ref: '#/components/schemas/TableDefinition'
```
jackye1995, line 260: It seems like the table related schema is not fully defined yet? I see a lot of things that are just generic object at this moment.

rust/lancedb/resources/remote-catalog-open-api.yaml
```diff
@@ -0,0 +1,284 @@
...
+        mode:
+          type: string
+          enum: [Create, ExistOk, Overwrite]
+
+    TableMetadata:
+      type: object
+      properties:
+        schema: {type: object}
+        version: {type: integer}
```
jackye1995, line 269: why having a version in the table metadata? If we know the location, we will figure out the version based on the manifests in the location.

PR #2225, opened 2025-03-14
fix: make `DatasetConsistencyWrapper` public so that we can implement `BaseTable`

BubbleCal:
tried locally we can import `DatasetConsistencyWrapper` with this


wjones127 (approved):

BubbleCal:
#2226 will resolve this

PR #2226, opened 2025-03-14
refactor: remove dataset reference from base table

westonpace:


westonpace:
@Akagi201 @wjones127 can we do this instead?  I'd like to avoid a reference from `BaseTable` to `Dataset` because there is no guarantee that `BaseTable` is backed by a dataset (it could be a remote table and maybe in the future a view or some other non-lance entity that acts like a table).

We can keep the helper method in `Table` but change it to return an `Option` instead of panic when a table is not a `NativeTable`?

wjones127 (approved):

Akagi201:
LGTM

Issue #2229, opened 2025-03-16
bug(node): `offset` (pagination) in `vectorSearch` doesn't work


o-alexandrov:
### LanceDB version

v0.18.1 (latest on Mar 16, 2025)

### What happened?

`offset` is broken when working with `vectorSearch`.
`offset` currently works **within the** `limit`, so when `offset` reaches the `limit`, it stops returning more rows.

### Are there known steps to reproduce?

1. Create table with a vector column

2. Add 2 rows

3. Run

```ts
const search = (offset: number) =>
  table.vectorSearch(vectorValue[]).distanceType(type).limit(1).offset(offset)
const data1 = await search(0) // returns 1st row
const data2 = await search(1) // returns undefined
```

o-alexandrov:
@wjones127 in #2232, added reproduction to this bug, as we don't have tests for `vectorSearch`'s `offset`.

o-alexandrov:
Edited the title & description. `offset` also doesn't work in `fts` search

urbien:
On the topic of pagination for a different method: the token-based pagination feature with page_token is only available in Python binding

o-alexandrov:
@urbien please clarify: what your comment adds to this issue?
- this issue is for `node`, not `python`. And this issue is about `offset` pagination, not `cursor` (token)

I think your comment is more suited for #2201

Xuanwo:
Hi, this issue is invalid because it was already fixed by PR #2592 (merged 2025-08-15), which implements limit+offset pagination through FTS and vector search results and closed the same root-cause issue #2459; main now fetches top_k = limit + offset. I'm closing this now. Feel free to create a new issue if the problem still exists.

PR #2232, opened 2025-03-18
test: add tests for `offset`, `mergeInsert`, `search().fastSearch()`

o-alexandrov:
error reproduction for #2229

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -448,6 +448,21 @@ describe("When creating an index", () => {
   });
   afterEach(() => tmpDir.removeCallback());
 
+  it('tests "offset" for "vectorSearch"', async () => {
+    // to trigger an error, change the default limit to 1 on the line below
+    const vectorSearch = (offset: number, limit = 2) =>
+      tbl
+        .vectorSearch(queryVec)
+        .limit(limit)
+        .offset(offset)
+        .select("id")
+        .toArray();
+    const rowsFirstTwo = await vectorSearch(0, 2);
+    const rowsFirst = await vectorSearch(0);
+    const rowsSecond = await vectorSearch(1);
+    expect(rowsFirstTwo).toEqual([rowsFirst[0], rowsSecond[0]]);
+  });
+
   it("should create a vector index on vector columns", async () => {
     await tbl.createIndex("vec");
 
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -474,21 +474,50 @@ describe("When creating an index", () => {
     );
   });
 
-  it("runs full text search ", async () => {
-    const pathDb = path.join(tmpDir.name, ".lancedb");
-    const db = await connect(pathDb);
+  it("runs full text search w/ initially empty table", async () => {
+    const db = await connect(tmpDir.name);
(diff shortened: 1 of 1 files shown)


nodejs/__test__/table.test.ts
```diff
@@ -448,6 +448,21 @@ describe("When creating an index", () => {
   });
   afterEach(() => tmpDir.removeCallback());
 
+  it('tests "offset" for "vectorSearch"', async () => {
+    // to trigger an error, change the default limit to 1 on the line below
+    const vectorSearch = (offset: number, limit = 2) =>
```
o-alexandrov, line 453: Cannot be merged as-is, because `offset` doesn't work in `vectorSearch`.
When someone decides to fix `offset`, to trigger the error, change default `limit` from 2 to 1.

nodejs/__test__/table.test.ts
```diff
@@ -274,6 +278,9 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
           { id: 4, z: 5 },
         ]);
 
+      const indices2 = await table.listIndices();
+      expect(indices2.length).toBe(1);
```
o-alexandrov, line 282: triggers error, because `.mergeInsert` removes an index

nodejs/__test__/table.test.ts
```diff
@@ -448,6 +455,43 @@ describe("When creating an index", () => {
...
+    );
+  });
+
+  it("runs full text search ", async () => {
+    const pathDb = path.join(tmpDir.name, ".lancedb");
+    const db = await connect(pathDb);
+    const schema = new arrow.Schema([
+      new arrow.Field("text", new arrow.Utf8()),
+    ]);
+    const table = await db.createEmptyTable("full-text-search", schema, {
```
o-alexandrov, line 484: if a table is created using `createEmptyTable`, then `fts` search won't work afterwords.
- returns empty array, when should return 2 rows

wjones127: Hm does the index exist on the table? If there are no rows to index, it might not.

I would create the index after inserting data instead.

o-alexandrov: The .listIndices() a couple of lines below returns an index

o-alexandrov: If an index doesn’t exist, LanceDB returns an error that you must create an inverted index

nodejs/__test__/table.test.ts
```diff
@@ -448,6 +455,71 @@ describe("When creating an index", () => {
...
+    await table.createIndex("text", { config: Index.fts() }); // index must be created after there's at least one row; otherwise, an index is automatically removed when the first row is added
+    await table.add([data[1]]); // add multiple rows (i.e. do a batch) in real life; adding rows one by one is just for testing here
+
+    const runSearch = () =>
+      table.search("world", "fts", "text").fastSearch().toArray();
+
+    const results1 = await runSearch();
+    /** @todo the commented out test must pass */
+    // expect(results1.length).toBe(1); // since we use ".fastSearch", the "data[1]" row doesn't exist
+    /** @todo this must fail */ expect(
```
o-alexandrov, line 504: @wjones127 `.fastSearch()` doesn't work. It doesn't search through indexed data only.

PR #2239, opened 2025-03-19
feat: add to_query_object method

westonpace:
This PR adds a `to_query_object` method to the various query builders (except not hybrid queries yet).  This makes it possible to inspect the query that is built.

In addition this PR does some normalization between the sync and async query paths.  A few custom defaults were removed in favor of None (with the default getting set once, in rust).

Also, the synchronous to_batches method will now actually stream results

Also, the remote API now defaults to prefiltering


python/python/tests/test_remote_db.py
```diff
@@ -315,7 +315,7 @@ def handler(body):
         assert body == {
             "distance_type": "l2",
             "k": 10,
-            "prefilter": False,
```
westonpace, line 318: It appears the remote API was still defaulting to `prefilter=False`.  Is there any particular reason we wanted this behavior?

wjones127: The intention was to keep consistent with the local sync API. Looks like we switched the default in https://github.com/lancedb/lancedb/pull/2000, though. So we can switch it for remote.

python/python/lancedb/table.py
```diff
@@ -3320,8 +3329,23 @@ async def _execute_query(
             fts_columns = query.full_text_query.get("columns", []) or []
             async_query = async_query.nearest_to_text(fts_query, columns=fts_columns)
 
-        table = await async_query.to_arrow()
```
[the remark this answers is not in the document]
wjones127: Honestly I forget why. If the test pass with this, I am good with streaming 👍

wjones127 (approved):

PR #2242, opened 2025-03-19
ci: remove OpenSSL as dependency in favor of rustls

wjones127:
`object_store` already hard codes `rustls` as the TLS implementation, so we have been shipping a mix of `rustls` and `openssl`. For simplicity of builds, we should consolidate to one, and that has to be `rustls`.


wjones127:
CI failure is unrelated and will be fixed by https://github.com/lancedb/lancedb/pull/2243

BubbleCal (approved):

Issue #2249, opened 2025-03-20
bug(python): Hybrid search crashes with UnboundLocalError if FTS returned no results


Odessit007:
### LanceDB version

v0.20.0

### What happened?

My hybrid search app crashes when there are vector results but no FTS results.

I'm instantiating reranker as
```python
reranker = RRFReranker(return_score="all")
```
and the actual query looks like this:
```python
table.search(query, query_type="hybrid").limit(30).rerank(reranker).to_list()
```

I intended to use scores for downstream analysis so that I can have addition filtering in the app but the `return_score="all"` parameter is the root cause of an error. Everything is fine with the default "relevance" option.

The error occurs here:
https://github.com/lancedb/lancedb/blob/b595d8a5794229062abf51748a31636575e0c5bc/python/python/lancedb/query.py#L1261

What happen is that "_score" is in the `results.column_names`, because I chose "all" option for reranker.
But `original_score_row_ids` variable is only instantiated if FTS returned some results:
https://github.com/lancedb/lancedb/blob/b595d8a5794229062abf51748a31636575e0c5bc/python/python/lancedb/query.py#L1241
So if there are no results the code crashes with the following error:
```
UnboundLocalError: cannot access local variable 'original_score_row_ids' where it is not associated with a value
```

PR #2252, opened 2025-03-21
fix: add better check for empty results in hybrid search

AyushExel:
fixes: https://github.com/lancedb/lancedb/issues/2249


AyushExel:
not sure why js test is failing. This PR doesn't touch any JS

BubbleCal (approved):

PR #2258, opened 2025-03-21
feat(python): warn on fork

### LINKED ISSUE #768: Add warning when multiprocessing is using `fork`
### Description

GPU IVF training requires `spawn`, we should add a warning and let user know that they need to set `start_method("spawn")`

[This can also cause problem in webservers](https://discord.com/channels/1030247538198061086/1197630499926057021/1344762583743463646).

* [ ] Use [register_at_fork](https://docs.python.org/3/library/os.html#os.register_at_fork) to issue a warning saying to use `spawn` instead.
   * Note: there are caveats for some web servers: https://stackoverflow.com/questions/72505167/os-register-at-fork-after-in-child-not-called
* [ ] Add note in documentation that `spawn` should be used instead of fork. This could be in the [troubleshooting guide](https://lancedb.github.io/lancedb/troubleshooting/) under a new "common issues" header.

### Link

_No response_

wjones127:
Closes #768

### DIFF
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -8,6 +8,10 @@ For trouble shooting, the best place to ask is in our Discord, under the relevan
 language channel. By asking in the language-specific channel, it makes it more
 likely that someone who knows the answer will see your question.
 
+## Common issues
+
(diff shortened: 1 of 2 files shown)


westonpace (approved):
Same windows issue

PR #2263, opened 2025-03-22
feat: add explain plan remote api

LuQQiu:
Add explain plan remote api


rust/lancedb/src/remote/table.rs
```diff
@@ -559,6 +562,52 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
             )?))
         }
     }
+
+    async fn explain_plan(&self, query: &AnyQuery, verbose: bool) -> Result<String> {
+        let base_request = self
+            .client
+            .post(&format!("/v1/table/{}/explain_plan/", self.name));
```
wkalt, line 569: is the trailing slash required?

wjones127 (approved):

rust/lancedb/src/remote/table.rs
```diff
@@ -559,6 +562,52 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+                status_code: None,
+            })
+        });
+
+        let plan_texts = futures::future::try_join_all(futures).await?;
+        let final_plan = if plan_texts.len() > 1 {
+            plan_texts
+                .into_iter()
+                .enumerate()
+                .map(|(i, plan)| format!("--- Plan #{} ---\n{}", i + 1, plan))
```
wkalt, line 601: Is this because when we do a vector query we create two plans? Can you show an example of the output?

wkalt: should the API instead accept two plans in a request?

wkalt (approved):

rust/lancedb/src/remote/table.rs
LuQQiu: follows the other existing apis

rust/lancedb/src/remote/table.rs
LuQQiu: Double checked, two plans (or two client requests) are only happen for new multivector client to work with old server that does not support multivector. So instead of sending one multivector request to the new server, it sends multiple single vector request to the old server.
Normal query just go with one query and only 1 plan

LuQQiu:
The build failure is addressed by https://github.com/lancedb/lance/commit/32c99af804dee8c1b44c2c144e3f0baa000ea47b

PR #2269, opened 2025-03-25
fix(java): make test work for jdk8

yanghua:


yanghua:
cc @LuQQiu

LuQQiu (approved):

eddyxu:
@yanghua qq, do you folks care jdk8 at all? 

We can remove jdk8 support if that makes things easier.

yanghua:
> @yanghua qq, do you folks care jdk8 at all?

Yes, currently, we still depend on jdk8.

This is a bit out of date, but it's the current status quo.

PR #2274, opened 2025-03-25
fix: allow re-creating tables

wjones127:
This adds some better error messages catching cases where data files are missing, due to either:

1. `optimize` / `cleanup_old_version` has been called and deleted the version we are reading from
2. Another process has dropped and recreated the table.

I was thinking I could make it so we automatically catch this and retry for queries, but because we return a stream, it's difficult to do that. At the very least, we can return an error so downstream applications can do this retry if they wish.


wjones127:
Closing in favor of #2298

PR #2278, opened 2025-03-26
chore(java): introduce spotless plugin

yanghua:


java/core/src/main/java/com/lancedb/lancedb/Connection.java
```diff
@@ -1,31 +1,37 @@
-// SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: Copyright The LanceDB Authors
-
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
LuQQiu, line 2: Sorry i am too familiar with the header and license stuff, but 
the header should remain unchanged for all files under lancedb/lancedb repo

eddyxu: @westonpace any concerns.

yanghua: I make it align with lancedb/lance and the project's license is Apache 2.0.

I am open about it and let you guys judge.

eddyxu: I am fine either way. This is aligned with what I saw in other java projects.

eddyxu (approved):

java/core/src/main/java/com/lancedb/lancedb/Connection.java
westonpace: No concerns.  I don't love the verbosity but I don't work with Java that much so :shrug: 

I agree it matches what we have in lance Java so that is good.

On looking at it with fresh eyes though I do notice this version (here and in Lance) is technically missing a copyright.

PR #2280, opened 2025-03-26
feat: add analyze_plan api

LuQQiu:
add analyze plan api to allow executing the queries and see runtime metrics.
Which help identify the query IO overhead and help identify query slowness


LuQQiu:
fixing nodejs and adding nodejs test

docs/src/js/classes/Query.md
```diff
@@ -30,6 +30,37 @@ protected inner: Query | Promise<Query>;
...
+
+#### Example
+
+```ts
+import * as lancedb from "@lancedb/lancedb"
+const db = await lancedb.connect("./.lancedb");
+const table = await db.createTable("my_table", [
+  { vector: [1.1, 0.9], id: "1" },
+]);
+const plan = await table.query().nearestTo([0.5, 0.2]).analyzePlan();
```
wkalt, line 55: let's print the plan in the example and show the output

docs/src/js/classes/VectorQuery.md
```diff
@@ -48,6 +48,37 @@ addQueryVector(vector): VectorQuery
...
+
+#### Example
+
+```ts
+import * as lancedb from "@lancedb/lancedb"
+const db = await lancedb.connect("./.lancedb");
+const table = await db.createTable("my_table", [
+  { vector: [1.1, 0.9], id: "1" },
+]);
+const plan = await table.query().nearestTo([0.5, 0.2]).analyzePlan();
```
wkalt, line 73: same comment applies in these other two places. I wonder if we could get docs automation to handle the inheritance for us at some point.

docs/src/troubleshooting.md
```diff
@@ -35,3 +35,9 @@ print the resolved query plan. You can use the `explain_plan` method to do this:
 * Python Sync: [LanceQueryBuilder.explain_plan][lancedb.query.LanceQueryBuilder.explain_plan]
 * Python Async: [AsyncQueryBase.explain_plan][lancedb.query.AsyncQueryBase.explain_plan]
 * Node @lancedb/lancedb: [LanceQueryBuilder.explainPlan](/lancedb/js/classes/QueryBase/#explainplan)
+
+It can also be helpful to executes the query and view the detailed IO metrics. You can use the analyze_plan method for this:
```
wkalt, line 39:
```suggestion
It can also be helpful to execute the query and view detailed IO metrics. You can use the `analyze_plan` method for this:
```

python/python/lancedb/query.py
```diff
@@ -2519,3 +2537,17 @@ async def explain_plan(self, verbose: Optional[bool] = False):
         results.append(await self._inner.to_fts_query().explain_plan(verbose))
 
         return "\n".join(results)
+
+    async def analyze_plan(self):
+        """Execute both queries and display with runtime metrics.
+
+        Returns
+        -------
+        plan"
```
wkalt, line 2546: plan : str

python/python/tests/test_hybrid_query.py
```diff
@@ -114,6 +114,17 @@ async def test_explain_plan(table: AsyncTable):
     assert "LanceScan" in plan
 
 
+@pytest.mark.asyncio
+async def test_analyze_plan(table: AsyncTable):
+    res = await (
+        table.query().nearest_to_text("dog").nearest_to([0.1, 0.1]).analyze_plan()
+    )
+    print(res)
```
wkalt, line 122: is this left in intentionally?

docs/src/js/classes/Query.md
```diff
@@ -30,6 +30,37 @@ protected inner: Query | Promise<Query>;
 
 ## Methods
 
+### analyzePlan()
+
+```ts
+analyzePlan(): Promise<string>
+```
+
+Executes the query and returns the runtime metrics.
```
wkalt, line 39:
```suggestion
Executes the query and returns the query plan annotated with runtime metrics.
```

wkalt (approved):
Looks good, just a few minor comments.

docs/src/js/classes/VectorQuery.md
LuQQiu: Those are auto generated docs, let me see fix the original place

docs/src/js/classes/Query.md
LuQQiu: addressed in nodejs/lancedb/query.ts

python/python/tests/test_hybrid_query.py
LuQQiu: removed, thanks

python/python/lancedb/query.py
LuQQiu: fixed

docs/src/troubleshooting.md
LuQQiu: fixed

PR #2281, opened 2025-03-26
feat!: change default read_consistency_interval=5s

wjones127:
Previously, when we loaded the next version of the table, we would block all reads with a write lock. Now, we only do that if `read_consistency_interval=0`. Otherwise, we load the next version asynchronously in the background. This should mean that `read_consistency_interval > 0` won't have a meaningful impact on latency.

Along with this change, I felt it was safe to change the default consistency interval to 5 seconds. The current default is `None`, which means we will **never** check for a new version by default. I think that default is contrary to most users expectations.

### DIFF
--- a/docs/src/guides/tables.md
+++ b/docs/src/guides/tables.md
@@ -1001,9 +1001,11 @@ In LanceDB OSS, users can set the `read_consistency_interval` parameter on conne
 
 There are three possible settings for `read_consistency_interval`:
 
-1. **Unset (default)**: The database does not check for updates to tables made by other processes. This provides the best query performance, but means that clients may not see the most up-to-date data. This setting is suitable for applications where the data does not change during the lifetime of the table reference.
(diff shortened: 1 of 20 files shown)


wjones127:
Test failure unrelated

LuQQiu:
Where can we show this consistency related information clearly to end user? I would expect the current read_consistency_interval = None and your change to read_consistency_interval=5s both seems unexpected to user. If i am the user, i would expect strong consistency by default with long latency for OSS.

wjones127:
> Where can we show this consistency related information clearly to end user? I would expect the current read_consistency_interval = None and your change to read_consistency_interval=5s both seems unexpected to user. If i am the user, i would expect strong consistency by default with long latency for OSS.

We discuss this in our documentation here: https://lancedb.github.io/lancedb/guides/tables/#consistency

I will update this part of the docs.

LuQQiu (approved):

PR #2284, opened 2025-03-27
feat: add columns using pyarrow schema

eddyxu:


jmhsieh:
lgtm.  My one question is if the add_columns_with_schema case handles if there is a conflict on a column name but it uses the same add_columns and would have the same behavior as that (which I assume checks).

jmhsieh (approved):
lgtm.  (see previous comment).

PR #2286, opened 2025-03-28
Remote catalog code gen via openapi

yanghua:


yanghua:
This is a very early PR and not ready for review

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
```diff
@@ -0,0 +1,333 @@
...
+              "Failed to create dataset (can not load): " + dbName + "." + tableName);
+        }
+      } catch (TException e) {
+        throw new RuntimeException(e);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e);
+      }
+    }
+
+    public Dataset register() {
```
jackye1995, line 240: looks like we don't really have a registerTable API yet. Do you plan to introduce one?

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    public HMSTableBuilder withProperty(String key, String value) {
+      this.storageOption.put(key, value);
+      return this;
+    }
+
+    public Dataset create() {
+      try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)) {
+        Table hmsTable = newHmsTable("admin", dbName, tableName, null);
```
jackye1995, line 210: I think this is an important part we need to decide what experience we go with for creating and getting a table.

There are 2 ways I see:

## option 1: only track table location for the table

This is the logic you have here right now. HMS stores the following info:

1. table type = LANCE
2. table location

When loading a Lance table from HMS, that means you only get information about the table location, and at client side we use Lance core library to derive the latest version by listing the `_versions` directory under the table location.

## option 2

HMS tracks the following information:

1. table type = LANCE
2. table location
3. latest table version (or equivalently you can also track the latest manifest file name)

When loading a Lance table from HMS, that means you can directly load the latest version of the table.

# Why this is important

This is important because any time an update happens to the table, there is a difference between these 2 options:

- in option 1, HMS will not be aware of the change, because the Lance dataset would increment to the new version manifest directly in storage, and HMS is not involved in the code path
- in option 2, HMS is aware of the change, because after the new manifest is created in storage, the update is not completed until the definition in HMS is also updated to reflect that.

I think we should go with option 2, because without the catalog knowing changes to the table, it becomes difficult to perform optimizations against it reactively, and this is very important for Lance with all the indexes it maintains.

Iceberg naturally worked in this way because the Iceberg commit protocol requires this architecture. But we also see formats like Delta evolving in a similar way. As they move to support multi-table transaction, they have moved from file system-based commit to a [coordinated commit](https://docs.google.com/document/d/1ZzNwKWrMVrNHc9qfhm7CZEn-AXcnD7NnosL0j6l5Iko/edit?tab=t.0#heading=h.dd2cc57oc5wk) where the commit coordinator keeps track of the latest version of the table, and [the commit coordinator is basically Unity Catalog](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/coordinatedcommits/UCCommitCoordinatorBuilder.scala).

# Implications

If we agree with option 2, there is an important implication to it. We need to change the Lance spec to update its commit semantics. So instead of saying the success of a commit is determined by committing to the new version manifest file location, if the Lance table is managed by a catalog (e.g. HMS), the catalog is responsible for determining the latest version of the table to be used.

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/controller/DatabaseApiController.java
```diff
@@ -0,0 +1,75 @@
...
+    this.request = request;
+  }
+
+  @Override
+  public Optional<NativeWebRequest> getRequest() {
+    return Optional.ofNullable(request);
+  }
+
+  @Override
+  public ResponseEntity<List<String>> v1DatabasesGet(
```
jackye1995, line 56: I did some experiments locally, looks like if I specify the operation ID in the spec:

```
  /v1/databases:
    get:
      tags: [Database]
      summary: List databases
      operationId: ListDatabase
```

Then I can get a nicer method name here:

```
@Override
  public ResponseEntity<List<String>> listDatabase(
      Optional<String> startAfter, Optional<Integer> limit) {
```

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
```diff
@@ -0,0 +1,333 @@
...
+      Database database = clients.run(client -> client.getDatabase(dbName));
+      String tbLocation;
+      if (location == null || location.isEmpty()) {
+        String dbLocation = StringUtils.stripEnd(database.getLocationUri(), "/");
+        tbLocation = String.format("%s/%s%s", dbLocation, tableName, LANCE_DATASET_SUFFIX);
+      } else {
+        tbLocation = location;
+      }
+
+      StorageDescriptor sd =
```
jackye1995, line 294: do we really need a dummy storage descriptor? I think it can be null in a Hive table

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
```diff
@@ -0,0 +1,333 @@
...
+              sd,
+              Collections.emptyList(),
+              tableParams,
+              "",
+              "",
+              TableType.EXTERNAL_TABLE.name());
+
+      newTable
+          .getParameters()
+          .put("EXTERNAL", "TRUE"); // using the external table type also requires this
```
jackye1995, line 327: why can't this be set in L307?

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
```diff
@@ -0,0 +1,333 @@
...
+      newTable.getParameters().put("storage_format", LANCE_TYPE);
```
jackye1995, line 328: why do we need to set both the table and another storage type?

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
```diff
@@ -0,0 +1,333 @@
...
+              Integer.MAX_VALUE,
+              sd,
+              Collections.emptyList(),
+              tableParams,
+              "",
+              "",
+              TableType.EXTERNAL_TABLE.name());
+
+      newTable
+          .getParameters()
```
jackye1995, line 326: why can't this be set in L307?

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
yanghua: Yes, I believe this is an excellent question that we've also considered internally, including aspects like versioning information you mentioned and whether to maintain schemas in the catalog.

To answer this, it ultimately depends on another design trade-off: whether we enforce the Catalog interface as the exclusive entry point for operating Lance tables. Currently, the answer is no. This means Catalog-based operations are just one of multiple access pathways to Lance tables, implying coexistence of multiple "operational sources." If we retain external metadata copies, this could introduce consistency issues with external metadata (or impose substantial maintenance costs to preserve consistency).

While I personally favor Option 2 from a design perspective, we haven't yet identified an optimal solution to the aforementioned challenge (at least within our current internal discussions).

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
yanghua: Yes, we already have. I just have not exposed it right now.

java/catalog-adapter-hms/src/main/java/com/lancedb/catalog/adapter/hms/service/HmsCatalogService.java
jackye1995: Yes, there were also many discussions internally we had about this topic. Here is my take:

Traditionally in database systems, there are 2 types of tables, the so-called "managed" and "external" tables. If the catalog manages the table, then it is natural for the catalog to control the versioning of the table, that means the catalog should be the exclusive entry point to the table. If the table is external, then the table definition is static, and it is expected that the user needs to manually update the table definition if there is an update happening to the table, so basically catalog is not directly involved in the table's activity. That results in well-known external table behaviors like read only, schema on read, drop table does not delete data, etc.

When the whole data lake movement started with Hive back then, the definition was technically broken. All Hive tables on data lake are external (excluding Hive ACID), but the table is technically dynamic because you can add files to a directory bypassing the catalog, and write is allowed against an external table. This problem continued to exist for Hudi (at least in the past, I did not follow the latest development there) and Delta (without coordinated commits), and Iceberg and Delta (with coordinated commits) basically made the table managed. 

What I was proposing is basically the managed table route, but that will as you said result in losing functionality to access the object bypassing catalog.

I think your point about "enabling multiple operational sources" is completely valid, but the key here is that it should  be another type of table, not managed nor external. In my previous company, I gave it the name "federated". This term "federated table" is typically used to describe tables in a catalog that is in another system. But here, a Lance table in a storage is really just another system, so I think the definition actually matches pretty well.

If the table type is federated, its definition is just the table location. The actual definition of the table is really federated to the storage layer and handled by the Lance library and spec. That means an UpdateTable would simply pass the operation to be executed by the Lance library against the storage layer, and an GetTable would, again, be a facade and pass the operation to be executed by the Lance library to figure out the latest version.

In this setup, we would allow all the behaviors to co-exist with clear trade offs and rigorous definition:

- All activities of a managed table must go through its corresponding catalog. The catalog is thus aware of all the activities happening to the table.
- All activities of a federated table could go through a catalog (or even multiple) or be done directly to the storage. That means the catalog might not have full knowledge about all activities to the table. And the catalog simply serves as a facade to perform all the actions against the storage layer.
- All activities of an external table go to the storage directly. The catalog only holds a static definition of the table at a specific version and is read only, and users are expected to update the version manually to reflect any changes to the table.

And we can start with federated as the first table type to be implemented if that is the one we desire from behavior perspective. But with this, I would be curious how you would track all activities happening in the table which is critical for table optimization and maintenance. Do you then use more like a pull based model to check periodically? Do you do it based on storage-level notification features?

yanghua: I believe categorizing tables into distinct types is an excellent approach. Several considerations I am thinking:

* The necessity of differentiating external tables from federated tables
* Whether table type attributes should be managed solely at the catalog level without requiring awareness from Lance/LanceDB
* Clarifying if LanceDB's existing "native table" and "remote table" concepts are currently code-level abstractions not be tracked at the metadata-level

In scenarios involving legacy or historical Lance tables, introducing a catalog system would entail transitional challenges:

* Managing pre-existing tables
* Coordinating catalog integration pace with access interfaces (Spark/Ray + PyLance)

Current context in our inner: Nearly all tables align with your defined "federated table" concept. For interim catalog metadata management, we are exploring mechanisms like hooks or periodic catalog backfilling.

jackye1995: I think "external" is just for completeness, we can probably ignore that for practical purposes. 

But when you say "Nearly all tables align with your defined "federated table" concept" I feel you might want to think internally a bit more about the implications. Clearly most of the vendors are moving towards a world of managed tables. I agree that limits functionality, but this makes governance features easy to implement because you have only a single entry point to guard.

jackye1995: For remote and native table, it almost feels like we have the mapping of 

Managed = remote
Federated = native?

yanghua: > this makes governance features easy to implement because you have only a single entry point to guard.

Absolutely, from a governance point of view.

Our current implementation remains in the early stages of catalog adoption (due to divergence from the community version, our internal Java implementation precedes the Python version). We retain flexibility for polyglot access across languages.

jackye1995: >Whether table type attributes should be managed solely at the catalog level without requiring awareness from Lance/LanceDB

I think it should not require awareness from Lance. But considering our integration model of LanceDB -> LanceCatalog -> Lance, LanceDB should be able to have the awareness of it through the catalog level information.

> Clarifying if LanceDB's existing "native table" and "remote table" concepts are currently code-level abstractions not be tracked at the metadata-level

I think they are code level abstractions. To some extent, the concept of a "managed table" at catalog level should help with removing some things that RemoteTable needs to differ from NativeTable, since the metadata will be provided by the specific catalog implementation, and LanceDB Cloud will give an implementation.

yanghua: +1 , clear

PR #2288, opened 2025-03-29
feat: add timeout to query execution options

### LINKED ISSUE #2287: Feature: query timeout parameter
### SDK

Python, Node, Rust

### Description

Users might want to cancel queries if they take too long. We should pass down a `timeout` parameter.

* In remote, this can be passed down as a header to the server, who can handle the time
* For local, we should pass then down, and cancel the future it we hit the timeout.

It's important that the query actually stops. For example, Javascript promises don't have built-in cancellation.

wjones127:
Closes #2287

## Summary by CodeRabbit

- **New Features**
	- Added configurable timeout support for query executions. Users can now specify maximum wait times for queries, enhancing control over long-running operations across various integrations.
- **Tests**
	- Expanded test coverage to validate timeout behavior in both synchronous and asynchronous query flows, ensuring timely error responses when query execution exceeds the specified limit.
	- Introduced a new test suite to verify query operations when a timeout is reached, checking for appropriate error handling.


wjones127:
@coderabbitai full review

rust/lancedb/src/remote/table.rs
```diff
@@ -44,6 +44,8 @@ use super::client::{HttpSend, RestfulLanceDbClient, Sender};
 use super::db::ServerVersion;
 use super::ARROW_STREAM_CONTENT_TYPE;
 
+const REQUEST_TIMEOUT_HEADER: HeaderName = HeaderName::from_static("x-request-timeout-ms");
```
wjones127, line 47: @albertlockett Does this header name sound good?

albertlockett: Yep looks good to me

rpgreen (approved):
LGTM. We should assume that we might want timeouts for other operations as well.
Should timeout be part of connection options or query options? Should we have separate read / write timeouts?

PR #2296, opened 2025-04-01
ci: fix lancedb musl builds

### LINKED ISSUE #2255: Get x64 musl builds of lancedb working again
Previous attempts consistently caused the GHA runner to freeze up, even with LTO disabled. Need to figure out why and fix this, so we can re-enable it.

Follow up to #2223

wjones127:
Fixes #2255

## Summary by CodeRabbit

- **Chores**
  - Enhanced the build process to improve performance and reliability across Linux platforms.
  - Updated environment settings for more accurate compiler integration.
  - Activated previously inactive build configurations to support advanced feature support.
  - Added support for the x86_64 architecture on Linux systems utilizing the musl C library.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -96,51 +96,46 @@ jobs:
       fail-fast: false
       matrix:
         settings:
-          # - target: x86_64-apple-darwin
-          #   host: macos-latest
-          #   features: ","
-          #   pre_build: |-
-          #       brew install protobuf
-          #       rustup target add x86_64-apple-darwin
-          # - target: aarch64-apple-darwin
-          #   host: macos-latest
-          #   features: fp16kernels
-          #   pre_build: brew install protobuf
-          # - target: x86_64-pc-windows-msvc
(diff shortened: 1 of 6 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -170,8 +165,8 @@ jobs:
               set -e &&
               apk add protobuf-dev &&
               rustup target add aarch64-unknown-linux-musl &&
-              export CC="/aarch64-linux-musl-cross/bin/aarch64-linux-musl-gcc" &&
-              export CXX="/aarch64-linux-musl-cross/bin/aarch64-linux-musl-g++"
+              export CC_aarch64_unknown_linux_musl=aarch64-linux-musl-gcc &&
+              export CXX_aarch64_unknown_linux_musl=aarch64-linux-musl-g++
```
wjones127, line 169: `psm` was using the `CC` in a build script, but the build script target is the host, not the final target. Therefore, we need to make sure to only set the cross compilation for the specific host.

westonpace (approved):

PR #2313, opened 2025-04-07
fix: robust handling of empty result when reranking

AyushExel:
I found some edge cases while running experiments that - depending on the base reranking libraries, some of them don't handle empty lists well. This PR manually checks if the result set to be reranked is empty

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced search result processing by ensuring that reordering only occurs when valid, non-empty results are available, thereby preventing unnecessary operations and potential errors.

- **Tests**
	- Added automated tests to verify that empty search result sets are handled correctly, ensuring consistent behavior across various rerankers.


BubbleCal (approved):
can we just move the logic in `_handle_empty_results` into `_rerank`?

PR #2314, opened 2025-04-07
fix: bugs for new FTS APIs

BubbleCal:


python/python/tests/docs/test_search.py
```diff
@@ -154,6 +156,41 @@ async def test_vector_search_async():
...
+                    "foul",
+                ]
+            }
+        ),
+        mode="overwrite",
+    )
+    table.create_fts_index("text", use_tantivy=False, replace=True)
+
+    results = table.search(
+        MatchQuery("foo", "text", fuzziness=1), query_type="fts"
```
QianZhu, line 183: is `query_type="fts"` required here?

python/python/tests/docs/test_search.py
```diff
@@ -154,6 +156,41 @@ async def test_vector_search_async():
     # --8<-- [end:search_result_async_as_list]
 
 
+def test_fts_fuzzy_query():
```
QianZhu, line 159: can we also add examples for boosting

python/python/tests/docs/test_search.py
BubbleCal: If there's a vector column, then no.

This sounds strange but it's true for now:
1. The SDK would try to infer the query type if it's not specified
2. If the query is not a vector, then it first tries getting a vector column and check whether there's a embedding function registered on the column, obviously it would fail if there's no vector column

We should change the behavior, but IIRC there's some limitations that blocking us to change it

python/python/tests/docs/test_search.py
BubbleCal: added

docs/src/js/classes/MatchQuery.md
```diff
@@ -37,14 +35,17 @@ Creates an instance of MatchQuery.
 * **column**: `string`
     The name of the column to search within.
 
-* **boost**: `number` = `1.0`
-    (Optional) The boost factor to influence the relevance score of this query. Default is `1.0`.
+* **options?**
+    Optional parameters for the match query.
+    - `boost`: The boost factor for the query (default is 1.0).
```
LuQQiu, line 40: is the boost factor only negative boost factor?
searched over the network, this explaination is more clear "A boosting query returns documents that match a positive query. Among those documents, the ones that also match the negative query are scored lower in relevance (their relevance score is multiplied by the negative boosting factor)."

docs/src/js/classes/BoostQuery.md
```diff
@@ -35,8 +35,11 @@ Creates an instance of BoostQuery.
 * **negative**: [`FullTextQuery`](../interfaces/FullTextQuery.md)
     The negative query that reduces the relevance score.
 
-* **negativeBoost**: `number`
-    The factor by which the negative query reduces the score.
+* **options?**
+    Optional parameters for the boost query.
+    - `negativeBoost`: The boost factor for the negative query (default is 0.0).
```
LuQQiu, line 40: why sometimes it's boost and sometimes it's negative boost?

nodejs/lancedb/query.ts
```diff
@@ -772,130 +770,138 @@ export enum FullTextQueryType {
...
+  /** @ignore */
+  public readonly inner: JsFullTextQuery;
   /**
    * Creates an instance of BoostQuery.
    *
    * @param positive - The positive query that boosts the relevance score.
    * @param negative - The negative query that reduces the relevance score.
-   * @param negativeBoost - The factor by which the negative query reduces the score.
+   * @param options - Optional parameters for the boost query.
+   *  - `negativeBoost`: The boost factor for the negative query (default is 0.0).
```
LuQQiu, line 858: could we add more detailed description for what negative boost does? e.g. A boosting query returns documents that match a positive query. Among those documents, the ones that also match the negative query are scored lower in relevance (their relevance score is multiplied by the negative boosting factor).

LuQQiu (approved):
LGTM, left some small comments about doc for sb don't know boost / fuzzy concept without look at tests and examples

docs/src/js/classes/BoostQuery.md
QianZhu: boost is pretty much the positive boost :)

python/python/tests/docs/test_search.py
BubbleCal: make sense, changed it.

nodejs/lancedb/query.ts
BubbleCal: added

BubbleCal:
> @BubbleCal Will the Rust SDK get these changes?

Yeah will do

PR #2338, opened 2025-04-15
feat: add new table API to wait for async indexing

rpgreen:
* Add new wait_for_index() table operation that polls until indices are created/fully indexed
* Add an optional wait timeout parameter to all create_index operations
* Python and NodeJS interfaces

## Summary by CodeRabbit

- **New Features**
  - Added optional waiting for index creation completion with configurable timeout.
  - Introduced methods to poll and wait for indices to be fully built across sync and async tables.
  - Extended index creation APIs to accept a wait timeout parameter.
- **Bug Fixes**
  - Added a new timeout error variant for improved error reporting on index operations.
- **Tests**
  - Added tests covering successful index readiness waiting, timeout scenarios, and missing index cases.

### DIFF
--- a/rust/lancedb/src/index/waiter.rs
+++ b/rust/lancedb/src/index/waiter.rs
@@ -36,7 +36,7 @@ pub async fn wait_for_index(
         let mut completed = vec![];
         let indices = table.list_indices().await?;
 
-        for idx in remaining {
+        for &idx in &remaining {
             if !indices.iter().any(|i| i.name == *idx) {
                 debug!("still waiting for new index '{}'", idx);
                 continue;
@@ -73,13 +73,12 @@ pub async fn wait_for_index(
         sleep(Duration::from_millis(DEFAULT_SLEEP_MS)).await;
     }
 
-    // debug log some diagnostics
-    for r in &remaining {
+    // debug log index diagnostics
+    for &r in &remaining {
         let stats = table.index_stats(r.as_ref()).await?;
-        if let Some(s) = stats {
-            debug!("index {} not fully indexed after {:?}. stats: {:?}", r, timeout, s)
-        } else {
-            debug!("index {} not found after {:?}", r, timeout)
+        match stats {
+            Some(s) => debug!("index '{}' not fully indexed after {:?}. stats: {:?}", r, timeout, s),
+            None => debug!("index '{}' not found after {:?}", r, timeout),
         }
     }
 
--- a/rust/lancedb/src/index/waiter.rs
+++ b/rust/lancedb/src/index/waiter.rs
@@ -36,7 +36,7 @@ pub async fn wait_for_index(
         let mut completed = vec![];
         let indices = table.list_indices().await?;
 
-        for idx in index_names {
+        for idx in remaining {
             if !indices.iter().any(|i| i.name == *idx) {
                 debug!("still waiting for new index '{}'", idx);
                 continue;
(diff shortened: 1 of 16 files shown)


rust/lancedb/src/table.rs
```diff
@@ -1086,6 +1094,16 @@ impl Table {
         self.inner.drop_index(name).await
     }
 
+    /// Poll until the columns are fully indexed. Will return Error::Timeout if the columns
+    /// are not fully indexed within the timeout.
+    pub async fn wait_for_index(
+        &self,
+        index_names: &[&str],
```
rpgreen, line 1101: One design question is if this should take index names or columns as parameters. Index names seems more appropriate, and consistent with index_stats(), but the "_idx" convention is a bit funky and not very explicit.

wjones127: I agree the convention is funky. In Lance, the name is a parameter users would request.

One thing we should consider is having `create_index` return an object with information about the index, including it's name.

I think for now taking an index name makes the most sense, and we can work on making that name more discoverable for users later.

wjones127 (approved):

wjones127 (approved):

rust/lancedb/src/table.rs
westonpace: We use the index name for `drop` (I seem to recall a similar concern)

rust/lancedb/src/index/waiter.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            let stats = table.index_stats(idx.as_ref()).await?;
+            match stats {
+                None => {
+                    debug!("still waiting for new index '{}'", idx);
+                    continue;
+                }
+                Some(s) => {
+                    if s.num_unindexed_rows == 0 {
```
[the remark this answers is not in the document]
rpgreen: Yeah I think we can address this as part of a new index/job data model.

rust/lancedb/src/index/waiter.rs
```diff
@@ -0,0 +1,66 @@
...
+) -> Result<()> {
+    let start = Instant::now();
+    let mut remaining = index_names.to_vec();
+
+    // poll via list_indices() and index_stats() until all indices are created and fully indexed
+    while start.elapsed() < timeout {
+        let mut completed = vec![];
+        let indices = table.list_indices().await?;
+
+        for idx in index_names {
```
westonpace, line 25: Should this be `for idx in remaining`?

rust/lancedb/src/table.rs
```diff
@@ -1974,6 +1992,12 @@ impl BaseTable for NativeTable {
                 message: "Multi-column (composite) indices are not yet supported".to_string(),
             });
         }
+        if opts.wait_timeout.is_some() {
+            return Err(Error::InvalidInput {
+                message: "wait_timeout must be None for NativeTable".to_string(),
+            });
+        }
```
westonpace, line 1999: I think it would be useful to allow a `wait_timeout` on `NativeTable` and just return immediately.  Partly because users can still create their own background indexing processes.  But mostly because this will make it easier for users to design code that can switch between cloud / native seamlessly.

rust/lancedb/src/index/waiter.rs
rpgreen: yes good catch! this wouldn't affect the result but would make more calls than needed.

rust/lancedb/src/table.rs
rpgreen: yeah that makes sense. I was trying to guard against the case where a user sets the timeout but it does nothing, but I guess we can't really avoid that unless the local/remote APIs diverge.

albertlockett (approved):

PR #2345, opened 2025-04-19
chore: bump Python 3.12 to 3.13 in actions

reneleonhardt:
### Updates
- Bump Python 3.12 to 3.13 in GitHub Actions

### Cosmetics
- Bump Python 3.8 to 3.9 in comments

### Future updates
Many things are outdated, can they be updated too?
- protobuf compilers
- NodeJS 18
- Java dependencies
- No tests with Java 8 and 21 (publish uses 8)
- Ubuntu 22.04 runners
- Docker and docker-compose

## Summary by CodeRabbit

- **Chores**
  - Updated CI workflows and project metadata to support Python 3.13.
  - Removed the `python-minor-version` input parameter from several GitHub Actions workflows.
  - Adjusted Python versions used in CI jobs for Linux, macOS, and Windows.
  - Improved package installation command in doctest job for better binary preference.
  - Added a reference comment for the Docker container used in Linux builds.
  - Enhanced package installation step with pip upgrade and environment variable for compatibility in test workflows.


wjones127:
> ### Versions in Cargo.lock were outdated
> A bit strange, when I built the python package, Cargo.lock was bumped from beta.7 to 8, it seems your release job yesterday forgot that after bumping Cargo.toml.

Yes, that something we need to fix in our release process. We bump versions there but don't update the lockfiles.

wjones127:
> ### Future updates
> Many things are outdated, can they be updated too?
> 
> * protobuf compilers
> * NodeJS 18
> * Java dependencies
> * No tests with Java 8 and 21 (publish uses 8)
> * Ubuntu 22.04 runners
> * Docker and docker-compose

Sure, other things can be updated in future PRs. 👍

wjones127 (approved):

wjones127:
Upgrading to 3.13 seems to make Python want to compile `pandas` from source, which isn't working out well.

reneleonhardt:
> Upgrading to 3.13 seems to make Python want to compile `pandas` from source, which isn't working out well.

Good catch! But a bit strange, there are wheels for 3.12 and 3.13...
https://pypi.org/project/pandas/#files
pandas-2.2.3-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
pandas-2.2.3-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl

When I run `pip install pandas` on Apple wheels are being used on 3.12 and 3.13 too... let me try --prefer-binary.

On the other hand I didn't really understand why pandas is in the test group only when python/python/lancedb/query.py is importing it not only in doctests... 😅

reneleonhardt:
> Upgrading to 3.13 seems to make Python want to compile `pandas` from source, which isn't working out well.

And I didn't expect to see pandas 2.1.4, it should've been 2.2.3 (latest).

And Windows shows an old bug? 😞
https://github.com/lancedb/lancedb/actions/runs/14551182796/job/40883701637?pr=2345#step:6:308
> error: the configured Python interpreter version (3.13) is newer than PyO3's maximum supported version (3.12)

reneleonhardt:
Is it intentional that doctest compiles with gcc? In most places I saw clang-18 configured...

wjones127:
> Is it intentional that doctest compiles with gcc? In most places I saw clang-18 configured...

Yes. (Clang 18 is used in some places to test fp16 kernels feature, which requires a newer compile. Only very new versions of GCC support Float16, so we use clang there.)

wjones127:
> Great, it's logging the same 2 times! First resolved to 2.2.3 binary, then 2.1.4 source

I'm guessing there is some other test dependency that pins an older version of pandas.

reneleonhardt:
Could you explain what
```
pip3 install $(ls target/wheels/lancedb-*.whl)[tests,dev]
```
is intended to do? What if there are multiple wheels with different requirement versions?

I always just use `.` (current directory), so wouldn't
```
pip3 install "python[tests,dev]"
```
be enough to install those groups from the current checkout/source in the python/ folder?

reneleonhardt:
> > Great, it's logging the same 2 times! First resolved to 2.2.3 binary, then 2.1.4 source
> 
> I'm guessing there is some other test dependency that pins an older version of pandas.

I guessed the same, but pip isn't allowed to resolve and install 2 different versions, right? 😅

wjones127:
Moved dependabot to its own PR: https://github.com/lancedb/lancedb/pull/2361

reneleonhardt:
Rebased

wjones127:
Superseded by https://github.com/lancedb/lancedb/pull/2955

Issue #2307, opened 2025-04-04
Feature: Better retry handling in remote client


wjones127:
### SDK

Python

### Description

Right now, whether we retry or not is a boolean per endpoint. For example, `create_table` and `add` have retries entirely disabled. But it might make sense to be more nuanced.

Some response codes are likely not safe to retry for write operations, such as 500.

Others, we should retry, like 409 or 429.

https://github.com/lancedb/lancedb/blob/647dee4e949b7ccbd55b3d39468be750d6b8bab5/rust/lancedb/src/remote/client.rs#L116-L122

Passing `false` here means no retries:

https://github.com/lancedb/lancedb/blob/647dee4e949b7ccbd55b3d39468be750d6b8bab5/rust/lancedb/src/remote/table.rs#L534

PR #2349, opened 2025-04-22
feat: add retries to remote client for requests with stream bodies

rpgreen:
Closes https://github.com/lancedb/lancedb/issues/2307
* Adds retries to remote operations with stream bodies (add, merge_insert)
* Change default retryable status codes to 409, 429, 500, 502, 503, 504
* Don't retry add or merge_insert operations on 5xx responses

Notes:
* Supporting retries on stream bodies means we have to buffer the body into memory so it can be cloned on retry. This will impact memory use patterns for the remote client. This buffering can be disabled by disabling retries (i.e. setting retries to 0 in RetryConfig)
* It does not seem that retry config can be specified by env vars as the documentation suggests. I added a follow-up issue [here](https://github.com/lancedb/lancedb/issues/2350)

## Summary by CodeRabbit

- **New Features**
  - Enhanced retry support for remote requests with configurable limits and exponential backoff with jitter.
  - Added robust retry logic for streaming data uploads, enabling retries with buffered data to ensure reliability.

- **Bug Fixes**
  - Improved error handling and retry behavior for HTTP status codes 409 and 504.

- **Refactor**
  - Centralized and modularized HTTP request sending and retry logic across remote database and table operations.
  - Streamlined request ID management for improved traceability.
  - Simplified error message construction in index waiting functionality.

- **Tests**
  - Added a test verifying merge-insert retries on HTTP 409 responses.

### DIFF
--- a/rust/lancedb/src/remote/db.rs
+++ b/rust/lancedb/src/remote/db.rs
@@ -10,7 +10,6 @@ use http::StatusCode;
 use lance_io::object_store::StorageOptions;
 use moka::future::Cache;
 use reqwest::header::CONTENT_TYPE;
-use reqwest::{RequestBuilder, Response};
 use serde::Deserialize;
 use tokio::task::spawn_blocking;
 
@@ -256,7 +255,7 @@ impl<S: HttpSend> Database for RemoteDatabase<S> {
         if let Some(start_after) = request.start_after {
             req = req.query(&[("page_token", start_after)]);
         }
-        let (request_id, rsp) = self.send(req).await?;
+        let (request_id, rsp) = self.client.send_with_retry(req, None, true).await?;
         let rsp = self.client.check_response(&request_id, rsp).await?;
         let version = parse_server_version(&request_id, &rsp)?;
         let tables = rsp
@@ -303,7 +302,7 @@ impl<S: HttpSend> Database for RemoteDatabase<S> {
             .body(data_buffer)
             .header(CONTENT_TYPE, ARROW_STREAM_CONTENT_TYPE);
 
-        let (request_id, rsp) = self.send(req).await?;
+        let (request_id, rsp) = self.client.send(req).await?;
 
         if rsp.status() == StatusCode::BAD_REQUEST {
             let body = rsp.text().await.err_to_http(request_id.clone())?;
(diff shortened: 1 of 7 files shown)


rust/lancedb/src/remote/db.rs
```diff
@@ -411,6 +412,12 @@ impl<S: HttpSend> Database for RemoteDatabase<S> {
     }
 }
 
+impl<S: HttpSend> RemoteDatabase<S> {
+    async fn send(&self, req: RequestBuilder) -> Result<(String, Response)> {
+        RemoteTable::send_with_retry(&self.client, req, None, None, true).await
+    }
+}
+
```
wjones127, line 420: > This moves retry logic from the HTTP client to the remote table implementation. The reasoning for this is that retries require Arrow serialization which should not be a responsibility of the HTTP client. This aims to keep the REST client concerned with reqwest/HTTP dependencies only.

I'm a little skeptical of this, and it becomes even more awkward now that the database client has to call into the table client API to make an HTTP request.

Maybe keep the `send*` methods in the HTTP client, but can make `send_stream_with_retry` take a closure `FnMut() -> StreamingBody` or something?

rust/lancedb/src/remote/table.rs
```diff
@@ -127,6 +130,150 @@ impl<S: HttpSend> RemoteTable<S> {
...
+        mut data: Box<dyn RecordBatchReader + Send>,
+        with_retry: bool,
+    ) -> Result<(String, Response)> {
+        if !with_retry || self.client.retry_config.retries == 0 {
+            let body = Self::reader_as_body(data)?;
+            return self.client.send(req.body(body)).await;
+        }
+
+        // to support retries, we need to buffer into cloneable batches here
+        let (schema, batches) = Self::buffer_reader(&mut *data).await?;
```
wjones127, line 175: Perhaps when we pull the other retries PR up, we can do the same "spill to a temporary file if it's too much memory" thing here on the client side.

rust/lancedb/src/remote/db.rs
rpgreen: Yeah, I don't love it, but it definitely doesn't feel right to add RecordBatch stuff to the HTTP client. I also considered the closure idea but thought it might be too complicated. I can give it a try.

rpgreen: Done

wjones127 (approved):

Issue #2134, opened 2025-02-19
bug(node): FixedSizeList schema causes 'Found field not in schema' error in v0.16


clebert:
### LanceDB version

v0.16

### What happened?

When creating a new table using the following (simplified) schema:

```
new Arrow.Schema([
  new Arrow.Field("text", new Arrow.Utf8()),

  new Arrow.Field(
    "vector",
    new Arrow.FixedSizeList(dimensions, new Arrow.Field("item", new Arrow.Float32())),
  ),
])
```

We encounter the following error: `Error: Found field not in schema: vector.0 at row 0`

This error only occurs in version 0.16, while version 0.15 works as expected.

I believe the issue lies within the `getFieldForPath` function. It appears that FixedSizeList fields are not properly supported.
I'm curious why the `inferSchema` function is called at all when I explicitly provide a schema.

If I replace `let inferredSchema = inferSchema(data, schema, opt);` with `let inferredSchema = schema;`, the code works correctly.

wjones127:
Thanks for reporting. We'll look into this soon.

> I'm curious why the inferSchema function is called at all when I explicitly provide a schema.

`inferSchema`'s job is to determine the schema of the data, which is allowed to be a subset of fields of the full schema you supplied.

Xuanwo:
Hi, this issue is invalid because it was already fixed in main: the schema-inference path now excludes TypedArrays from object traversal (PR #3119, merged), and FixedSizeList schema table-creation tests are covered, so the v0.16 'Found field not in schema: vector.0' regression no longer occurs. I'm closing this now. Feel free to create a new issue if the problem still exists.

PR #2355, opened 2025-04-25
fix: prevent embedded objects from returning null in all of their fields

acoliver:
metadata{filename=xyz} filename would be there structurally, but ALWAYS null.

I didn't include this as a file but it may be useful for understanding the problem for people searching on this issue so I'm including it here as documentation. Before this patch any field that is more than 1 deep is accepted but returns null values for subfields when queried.

```js
const lancedb = require('@lancedb/lancedb');

// Debug logger
function debug(message, data) {
  console.log(`[TEST] ${message}`, data !== undefined ? data : '');
}

// Log when our unwrapArrowObject is called
const kParent = Symbol.for("parent");
const kRowIndex = Symbol.for("rowIndex");

// Override console.log for our test
const originalConsoleLog = console.log;
console.log = function() {
  // Filter out noisy logs
  if (arguments[0] && typeof arguments[0] === 'string' && arguments[0].includes('[INFO] [LanceDB]')) {
    originalConsoleLog.apply(console, arguments);
  }
  originalConsoleLog.apply(console, arguments);
};

async function main() {
  debug('Starting test...');
  
  // Connect to the database
  debug('Connecting to database...');
  const db = await lancedb.connect('./.lancedb');
  
  // Try to open an existing table, or create a new one if it doesn't exist
  let table;
  try {
    table = await db.openTable('test_nested_fields');
    debug('Opened existing table');
  } catch (e) {
    debug('Creating new table...');
    
    // Create test data with nested metadata structure
    const data = [
      {
        id: 'test1',
        vector: [1, 2, 3],
        metadata: {
          filePath: "/path/to/file1.ts",
          startLine: 10,
          endLine: 20,
          text: "function test() { return true; }"
        }
      },
      {
        id: 'test2',
        vector: [4, 5, 6],
        metadata: {
          filePath: "/path/to/file2.ts",
          startLine: 30,
          endLine: 40,
          text: "function test2() { return false; }"
        }
      }
    ];
    
    debug('Data to be inserted:', JSON.stringify(data, null, 2));
    
    // Create the table
    table = await db.createTable('test_nested_fields', data);
    debug('Table created successfully');
  }
  
  // Query the table and get results
  debug('Querying table...');
  const results = await table.search([1, 2, 3]).limit(10).toArray();
  
  // Log the results
  debug('Number of results:', results.length);
  
  if (results.length > 0) {
    const firstResult = results[0];
    debug('First result properties:', Object.keys(firstResult));
    
    // Check if metadata is accessible and what properties it has
    if (firstResult.metadata) {
      debug('Metadata properties:', Object.keys(firstResult.metadata));
      debug('Metadata filePath:', firstResult.metadata.filePath);
      debug('Metadata startLine:', firstResult.metadata.startLine);
      
      // Destructure to see if that helps
      const { filePath, startLine, endLine, text } = firstResult.metadata;
      debug('Destructured values:', { filePath, startLine, endLine, text });
      
      // Check if it's a proxy object
      debug('Result is proxy?', Object.getPrototypeOf(firstResult) === Object.prototype ? false : true);
      debug('Metadata is proxy?', Object.getPrototypeOf(firstResult.metadata) === Object.prototype ? false : true);
    } else {
      debug('Metadata is not accessible!');
    }
  }
  
  // Close the database
  await db.close();
}

main().catch(e => {
  console.error('Error:', e);
}); 
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of nested struct fields to ensure accurate preservation of values during serialization and deserialization.
  - Enhanced robustness when accessing nested object properties, reducing errors with missing or null values.

- **Tests**
  - Added tests to verify correct handling of nested struct fields through serialization and deserialization.


nodejs/__test__/table.test.ts
```diff
@@ -225,11 +225,23 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
       await table.add([data3]);
 
       let res = await table.query().limit(10).toArray();
-      const resVector = res.map((r) => r.get("vector").toArray());
+      const resVector = res.map((r) =>
+        (r as unknown as { get: (col: string) => { toArray: () => unknown } })
+          .get("vector")
+          .toArray(),
+      );
```
wjones127, line 232: Why is this necessary? It worries me that this has changed as it suggests this is a breaking change in the API and will affect users as well.

nodejs/__test__/table.test.ts
```diff
@@ -1201,73 +1215,6 @@ describe("when dealing with versioning", () => {
   });
 });
 
-describe("when dealing with tags", () => {
-  let tmpDir: tmp.DirResult;
-  beforeEach(() => {
-    tmpDir = tmp.dirSync({ unsafeCleanup: true });
-  });
```
wjones127, line 1208: I think on of your merges / rebases wasn't clean, and has accidentally deleted some concurrent changes.

nodejs/lancedb/query.ts
```diff
@@ -337,10 +337,19 @@ export class QueryBase<NativeQueryType extends NativeQuery | NativeVectorQuery>
...
-  // biome-ignore lint/suspicious/noExplicitAny: arrow.toArrow() returns any[]
-  async toArray(options?: Partial<QueryExecutionOptions>): Promise<any[]> {
-    const tbl = await this.toArrow(options);
-    return tbl.toArray();
+  async toArray(
+    options?: Partial<QueryExecutionOptions>,
+  ): Promise<Record<string, unknown>[]> {
+    // Use Arrow's built-in toJSON() on each row to get a plain JS object
+    const arrowTable = await this.toArrow(options);
+    const result: Record<string, unknown>[] = [];
```
wjones127, line 345: Ah yeah this is the breaking change here. Let's keep this change out of this PR, so we can focus on the bug in `transposeData`. If you want, we can consider doing a separate PR where we add a new `toJsArray()` or something that does this.

wjones127 (changes requested):
Thanks for making a PR. This looks like a very important fix. I wonder if it will also fix https://github.com/lancedb/lancedb/issues/2134, or if that is not quite the same issue.

There are some unrelated changes here. Could you clean those up?

nodejs/__test__/table.test.ts
acoliver: yeah I'm in the midst of cleaning that up. one moment.

nodejs/lancedb/query.ts
acoliver: Sure...the lint check prevented me from committing with Any so I'll just mark this as ignore (I didn't MAKE it any) -- 

One thing we HAVE to do because if we write the embedded correctly, we have to read it correctly, but we don't have to swap to unknown)

nodejs/__test__/table.test.ts
acoliver: so right now when we write "flattened" we basically just truncate the embedded values. the fields are there...but everything comes back null. We DO have to map the values back to an array (which is how we wrote them)... I mean the API is broken now so if anyone is using it...they are getting back an array like this:

metadata {filename:null, size: null, bla:null} -- even if they wrote values... so we do have to break it...a little.

acoliver: OK I cleaned it up. Turned out some of that was mis-included from a previous pass at fixing it. Auto-merging is always a risky plan :)

nodejs/lancedb/query.ts
acoliver: this is gone.  The test still passes

nodejs/__test__/table.test.ts
acoliver: I've now taken just the important part, merged it into head and clobbered my branch with head+this so all is clean the tests are passing... hopefully they pass here.

wjones127 (approved):
Thanks for doing this. This looks good now.

PR #2362, opened 2025-04-28
docs: update indexing and compaction docs

rpgreen:
## Summary by CodeRabbit

- **Documentation**
  - Clarified and expanded explanations of data management concepts in LanceDB.
  - Added notes on automatic background fragment compaction and incremental reindexing support in LanceDB Cloud/Enterprise.
  - Updated details on disabling interim exhaustive kNN search during background reindexing.
  - Improved formatting and removed outdated FTS reindexing subsection.

### DIFF
--- a/docs/src/concepts/data_management.md
+++ b/docs/src/concepts/data_management.md
@@ -38,7 +38,8 @@ Depending on the use case and dataset, optimal compaction will have different re
 - Keep the number of fragments under 100, which is suitable for most use cases (for *really* large datasets of >500M rows, more fragments might be needed)
 
 !!! note
-LanceDB Cloud/Enterprise supports [auto-compaction](https://docs.lancedb.com/enterprise/architecture/architecture#write-path) which automatically optimizes fragments in the background as data changes.
+
+    LanceDB Cloud/Enterprise supports [auto-compaction](https://docs.lancedb.com/enterprise/architecture/architecture#write-path) which automatically optimizes fragments in the background as data changes.
 
 ## Deletion
 


docs/src/concepts/data_management.md
```diff
@@ -50,13 +53,7 @@ Reindexing is the process of updating the index to account for new data, keeping
...
 
-When a reindex job is triggered in the background, the entire data is reindexed, but in the interim as new queries come in, LanceDB will combine results from the existing index with exhaustive kNN search on the new data. This is done to ensure that you're still searching on all your data, but it does come at a performance cost. The more data that you add without reindexing, the impact on latency (due to exhaustive search) can be noticeable.
-
-### Vector reindex
+When a reindex job is triggered in the background, the entire data is reindexed, but in the interim as new queries come in, LanceDB will combine results from the existing index with exhaustive kNN search on the new data. This is done to ensure that you're still searching on all your data, but it does come at a performance cost. The more data that you add without reindexing, the impact on latency (due to exhaustive search) can be noticeable. This can be disabled by setting the [fast_search](https://lancedb.github.io/lancedb/python/python/#lancedb.query.AsyncQuery.fast_search) parameter which will instruct the query to ignore un-indexed data.
```
wjones127, line 56: This document covers both OSS and Cloud, so when we talk about background jobs, we should make it clear that is just cloud.
```suggestion
In LanceDB OSS, re-indexing happens synchronously when you call either `create_index` or `optimize` on a table. In LanceDB Cloud, re-indexing happens asynchronously as you add and update data in your table.

By default, queries will search new data even if it has yet to be indexed. This is done using brute-force methods, such as kNN for vector search, and combined with the fast index search results. This is done to ensure that you're always searching over all your data, but it does come at a performance cost. Without reindexing, adding more data to a table will make queries slower and more expensive. This behavior can be disabled by setting the [fast_search](https://lancedb.github.io/lancedb/python/python/#lancedb.query.AsyncQuery.fast_search) parameter which will instruct the query to ignore un-indexed data.
```

docs/src/concepts/data_management.md
```diff
@@ -37,6 +37,9 @@ Depending on the use case and dataset, optimal compaction will have different re
 - It’s always better to use *batch* inserts rather than adding 1 row at a time (to avoid too small fragments). If single-row inserts are unavoidable, run compaction on a regular basis to merge them into larger fragments.
 - Keep the number of fragments under 100, which is suitable for most use cases (for *really* large datasets of >500M rows, more fragments might be needed)
 
+!!! note
+LanceDB Cloud/Enterprise supports [auto-compaction](https://docs.lancedb.com/enterprise/architecture/architecture#write-path) which automatically optimizes fragments in the background as data changes.
```
wjones127, line 41: Does this render correctly? I would think you need to indent this:

```suggestion
!!! note

    LanceDB Cloud/Enterprise supports [auto-compaction](https://docs.lancedb.com/enterprise/architecture/architecture#write-path) which automatically optimizes fragments in the background as data changes.
```

wjones127 (approved):
Thanks. This document could really use an update.

PR #2368, opened 2025-05-01
feat: return version for all write operations

LuQQiu:
return version info for all write operations (add, update, merge_insert and column modification operations)

## Summary by CodeRabbit

- **New Features**
  - Table modification operations (add, update, delete, merge, add/alter/drop columns) now return detailed result objects including version numbers and operation statistics.
  - Result objects provide clearer feedback such as rows affected and new table version after each operation.

- **Documentation**
  - Updated documentation to describe new result objects and their fields for all relevant table operations.
  - Added documentation for new result interfaces and updated method return types in Node.js and Python APIs.

- **Tests**
  - Enhanced test coverage to assert correctness of returned versioning and operation metadata after table modifications.


LuQQiu:
@wkalt @wjones127 PTAL thanks

python/python/tests/docs/test_merge_insert.py
```diff
@@ -18,19 +18,19 @@ def test_upsert(mem_db):
...
-    stats  # {'num_inserted_rows': 1, 'num_updated_rows': 1, 'num_deleted_rows': 0}
+    res  # {'num_inserted_rows': 1, 'num_updated_rows': 1, 'num_deleted_rows': 0}
     # --8<-- [end:upsert_basic]
     assert table.count_rows() == 3
-    assert stats["num_inserted_rows"] == 1
-    assert stats["num_updated_rows"] == 1
-    assert stats["num_deleted_rows"] == 0
+    assert res.num_inserted_rows == 1
+    assert res.num_deleted_rows == 0
+    assert res.num_updated_rows == 1
```
wjones127, line 33: So this is a breaking change, but since the existing API is unreleased, I think this is acceptable.

rust/lancedb/src/remote/table.rs
```diff
@@ -1035,11 +1083,28 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+        if body.trim().is_empty() || body == "{}" {
+            // Backward compatible with old servers
+            let version = self.version().await?;
+            return Ok(MergeResult {
+                version,
+                num_deleted_rows: 0,
+                num_inserted_rows: 0,
+                num_updated_rows: 0,
+            });
+        }
```
wjones127, line 1098: 🙌

rust/lancedb/src/remote/table.rs
wjones127, line 1: I really appreciate you testing the old and new server cases here.

The code duplication seems a little disappointing though. I wonder if we can use parameterized tests via `rstest` to consolidate them. Here is one example where we do this:

https://github.com/lancedb/lancedb/blob/e8c0c52315899ce214dd51b0d9abaa7f8f9397e7/rust/lancedb/src/remote/table.rs#L1934-L1938

rust/lancedb/src/remote/table.rs
```diff
@@ -1035,11 +1083,28 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+        let body = response.text().await.err_to_http(request_id.clone())?;
+
+        if body.trim().is_empty() || body == "{}" {
+            // Backward compatible with old servers
+            let version = self.version().await?;
+            return Ok(MergeResult {
+                version,
+                num_deleted_rows: 0,
+                num_inserted_rows: 0,
+                num_updated_rows: 0,
```
wjones127, line 1096: This is an unfortunate default, but I think it's okay since this code path will be unused as soon as we update the servers.

wjones127 (approved):
Excellent work.

If you have time and it's possible, it would be nice to consolidate some of those tests. Otherwise this looks great!

Eventually, I'd like to add more information do the responses. For example, `update()` and `delete()` can also report the number of rows affected. Also could provide performance statistics in responses. So I am glad we now have these result structs.

python/python/tests/docs/test_merge_insert.py
LuQQiu: Yep, it's a breaking change, good that that one is new and hasn't been released

rust/lancedb/src/remote/table.rs
LuQQiu: Let me update it, thanks for the advice, yep the duplication is annoying, i am duplicating it thinking it may be easier to remove the old tests as a whole in the future, but we don't know how long it last, so better to remove

LuQQiu:
@wjones127 I will have a separate PR consolidate the test cases and improve them. Thanks for the review!

PR #2373, opened 2025-05-05
chore: reduce unneeded API calls for return version for write operations and improve test

LuQQiu:
Reduce the duplicate code for remote write operation testing.
Avoid double call to remote to get version info, just return 0 instead of suddenly adding extra API calls for end users when they are using old servers. 

## Summary by CodeRabbit

- **New Features**
  - Added version tracking to table operation results, allowing users to see the commit version associated with add, update, delete, merge, and column modification operations.
- **Bug Fixes**
  - Improved compatibility with legacy servers by standardizing version information as zero when the server does not return a version.
- **Documentation**
  - Clarified the meaning of the version field in operation results, especially for cases involving legacy server responses.
- **Tests**
  - Enhanced test coverage to verify correct behavior with both legacy and modern server responses.

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -1617,6 +1617,16 @@ mod tests {
         let table = Table::new_with_handler("my_table", move |mut request| {
             if request.url().path() == "/v1/table/my_table/insert/" {
                 assert_eq!(request.method(), "POST");
+                assert!(request
+                    .url()
+                    .query_pairs()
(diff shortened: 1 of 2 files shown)


LuQQiu:
Solve the comment in https://github.com/lancedb/lancedb/pull/2368#pullrequestreview-2815578685
@wjones127 PTAL, thanks

wjones127 (approved):
Nice work! Thanks for cleaning this up.

PR #2374, opened 2025-05-06
fix: add restore with tag in python and nodejs API

LuQQiu:
add restore with tag API in python and nodejs API and add tests to guard them

## Summary by CodeRabbit

- **New Features**
  - The restore functionality now supports using version tags in addition to numeric version identifiers, allowing you to revert tables to a state marked by a tag.
- **Bug Fixes**
  - Restoring with an unknown tag now properly raises an error.
- **Documentation**
  - Updated documentation and examples to clarify that restore accepts both version numbers and tags.
- **Tests**
  - Added new tests to verify restore behavior with version tags and error handling for unknown tags.
  - Added tests for checkout and restore operations involving tags.


python/src/table.rs
```diff
@@ -551,15 +551,37 @@ impl Table {
...
+                    let num: u64 = i.extract()?;
+                    (true, num, String::new())
+                } else if let Ok(s) = version.downcast_bound::<PyString>(py) {
+                    let str_value = s.to_string();
+                    (false, 0, str_value)
+                } else {
+                    return Err(PyIOError::new_err(
+                        "version must be an integer or a string.",
+                    ));
+                };
```
wjones127, line 572: You could do something like:

```rust
#[derive(FromPyObject)]
enum LanceVersion {
    Version(u64),
    Tag(String),
}
```
And then you don't have to manually parse that stuff out:

```suggestion
        version: Option<LanceVersion>,
    ) -> PyResult<Bound<'_, PyAny>> {
        let inner = self_.inner_ref()?.clone();
        let py = self_.py();
        if let Some(version) = version {
```

wjones127 (approved):

python/src/table.rs
LuQQiu: Good suggestion! addressed comments

PR #2375, opened 2025-05-06
fix: fix backward compatibility with the add API

LuQQiu:
add API originally returns a struct with request_id, add backward compatibility for that

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of empty server responses for various data operations to ensure consistent behavior across server versions.
  - Added default values to version and numeric fields to prevent errors when response data is incomplete.

- **Tests**
  - Expanded tests to cover multiple server response scenarios, validating correct version handling in data operations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -223,7 +223,7 @@ dependencies = [
  "chrono",
  "chrono-tz 0.10.3",
  "half",
- "hashbrown 0.15.2",
+ "hashbrown 0.15.3",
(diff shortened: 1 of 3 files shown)


wkalt (approved):
I think it's odd that the API can return a "" body in addition to an empty "{}", necessitating the `body.trim().is_empty()` calls. But seems like it's a preexisting quirk.

PR #2394, opened 2025-05-20
fix: implement validity bitmap for struct rows in Arrow data processing

### LINKED ISSUE #2393: bug(node): `table.add` fails on nested `Struct` columns (unmasked nulls)
### LanceDB version

0.19.0

### What happened?

When using a nested `Struct` in a TypeScript schema and calling  
```ts
await table.add([
  { id: "a", …, metadata: { nested: { a: 1, b: 2 } } },
  …
]);
```  
LanceDB throws:

```
lance error: LanceError(Arrow): Arrow error: Invalid argument error:
Found unmasked nulls for non-nullable StructArray field "a", …
```

even though *all* struct fields are present and non-null.

**Steps to Reproduce**  
1. Define a schema with a nested struct, e.g.:  
   ```ts
   new Field("metadata", new Struct([
     new Field("nested", new Struct([
       new Field("a", new Int32()),
       new Field("b", new Int32()),
     ]))
   ]))
   ```  
2. `await db.createEmptyTable("demo", schema)`  
3. `await table.add([ … your two rows with metadata.nested.a/b … ])`  
4. Observe the above error.

**Expected Behavior**  
Rows are added successfully—no “unmasked nulls” error—since every nested field is supplied.

**Actual Behavior**  
`arrow2` (Rust) rejects the IPC buffer with an unmasked‐nulls error because the JS writer omitted the parent `Struct`’s `length` and `validity` bitmap.

**Root Cause & Fix**  
`transposeData` was calling `makeData({ type, children })` without setting `length`/`nullCount`/`nullBitmap`. Arrow-JS silently tolerates the mismatch, but arrow2 enforces:
> if a struct slot is “valid” then all non-nullable children must have a value.  

**Environment**  
- macOS darwin-arm64  
- Node.js ≥18  
- `@lancedb/lancedb` v0.19.1-beta.5  
- `apache-arrow` v18

```typescript
import { type Table, connect } from "@lancedb/lancedb";
import * as arrow from "apache-arrow";

async function main() {
  // 1. open (or create) a local DB directory
  const db = await connect("./tmp/example.lance");
  const tableName = "demo";
  let table: Table;
  // 2. create the schema
  const schema = new arrow.Schema([
    new arrow.Field("id", new arrow.Utf8()),
    new arrow.Field(
      "vector",
      new arrow.FixedSizeList(3, new arrow.Field("item", new arrow.Float32(), true)),
      false,
    ),
    new arrow.Field("text", new arrow.Utf8()),
    new arrow.Field("metadata", new arrow.Struct([
      new arrow.Field("nested", new arrow.Struct([
        new arrow.Field("a", new arrow.Int32()),
        new arrow.Field("b", new arrow.Int32()),
      ]))
    ]))
  ]);
  // 3. create table with schema
  table = await db.createEmptyTable(tableName, schema);
  // 4. add two example rows
  // this fails with an error [Error: lance error: LanceError(Arrow): Arrow error: Invalid argument error: Found unmasked nulls for non-nullable StructArray field "a", /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lance-datafusion-0.26.0/src/utils.rs:49:31  Caused by: LanceError(Arrow): Arrow error: Invalid argument error: Found unmasked nulls for non-nullable StructArray field "a", /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lance-datafusion-0.26.0/src/utils.rs:49:31]
  await table.add([
    { id: "a", vector: [0.1, 0.0, 0.9], text: "first", metadata: { nested: { a: 1, b: 2 } } },
    { id: "b", vector: [0.2, 0.1, 0.4], text: "second", metadata: { nested: { a: 3, b: 4 } } },
  ]);

  // 5. search
  const queryVector = [0.05, 0.0, 0.95];
  const results = await table.search(queryVector).limit(5).toArray();

  console.log("Search results:");
  console.log(results);

  db.close();
}

main().catch(console.error);
```

aryasaatvik:
- resolves #2393 

Previously, when serializing nested `Struct` columns in the JS SDK, we called:
```ts
makeData({ type: field.type, children: childVectors })
```
without supplying `length`, `nullCount`, or `nullBitmap`. Apache-Arrow-JS tolerated this omission, but Arrow2 (used by LanceDB’s Rust backend) enforces that:

> **If a Struct slot is valid, all non-nullable child fields must also be valid.**

Omitting the parent metadata caused Arrow2 to assume “all rows valid” while children carried their own nulls, triggering:

```
Found unmasked nulls for non-nullable StructArray field "a"
```

### Changes
- In `transposeData`, compute:
  - `numRows` = number of rows in the Struct (from child vectors).
  - A bit-packed `validityBitmap` where bit _i_ is `1` if **any** child in row _i_ is non-null.
  - `nullCount` = total rows where **all** children are null.
- Pass these into `makeData({ type, length, nullCount, nullBitmap, children })`.
- Omit `nullBitmap` entirely when `nullCount === 0` to preserve the “no nulls” fast path.
- add documentation on `makeArrowTable` to call out nested-Struct support and spec conformance.

### Impact
- **Bug fix only**: no breaking API changes.
- Nested `Struct` columns with any combination of null and non-null child fields now round-trip correctly through Arrow IPC and can be inserted via `table.add(...)` without error.
- Existing JS-only unit tests (which are permissive) continue to pass.

### Testing
- Verified the original reproduction script now succeeds.
- Ran existing Arrow-JS tests for nested struct retention.
- Manual smoke test: `await table.add([...])` on nested metadata schemas no longer throws.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of nested data fields to ensure more accurate detection and representation of missing or null values in structured data.

### DIFF
--- a/nodejs/lancedb/arrow.ts
+++ b/nodejs/lancedb/arrow.ts
@@ -643,8 +643,47 @@ function transposeData(
     const childVectors = childFields.map((child) => {
       return transposeData(data, child, fullPath);
     });
+
+    /**
+     * Compute the validity bitmap for the struct.  According to the Arrow
+     * specification a struct row is considered null iff *all* of its child
+     * fields are null for that row.  If at least one child value is non-null
+     * then the struct row itself is considered valid.  We create a validity
+     * bitmap which marks valid rows with a 1 bit and null rows with a 0 bit.
+     */
+    const numRows = childVectors.length === 0 ? 0 : childVectors[0].length;
+    let nullCount = 0;
+    // Only allocate the bitmap if we discover at least one null row.
+    const validityBitmap = new Uint8Array(Math.ceil(numRows / 8));
+
+    for (let i = 0; i < numRows; i++) {
+      // Determine if every child is null at this row.
+      let allChildrenNull = true;
+      for (const child of childVectors) {
+        // `get` returns undefined for an out-of-range index but that should
+        // not happen here since all child vectors have `numRows` elements.
+        const value = child.get(i);
+        if (value !== null && value !== undefined) {
+          allChildrenNull = false;
+          break;
+        }
+      }
+
+      if (!allChildrenNull) {
+        // Mark this row as valid.
+        validityBitmap[i >> 3] |= 1 << (i & 7);
+      } else {
+        nullCount++;
+      }
+    }
+
     const structData = makeData({
       type: field.type,
+      length: numRows,
+      nullCount,
+      // Only include the bitmap if there are any nulls.  Arrow interprets a
+      // missing bitmap as "no nulls".
+      ...(nullCount > 0 ? { nullBitmap: validityBitmap } : {}),
       children: childVectors as unknown as ArrowData<DataType>[],
     });
     return arrowMakeVector(structData);


entelligence-ai-reviews:
## Walkthrough
This update refines the handling of struct types within the transposeData function in the Arrow integration. It implements accurate computation of the validity bitmap for struct fields, ensuring that a struct row is marked as null only if all its child fields are null. The bitmap is allocated and included in the Arrow data only when necessary, aligning with the Apache Arrow specification. This change improves the correctness and compatibility of nullability representation for struct fields when working with Arrow data structures.

## Changes
| File(s) | Summary |
| --- | --- |
| nodejs/lancedb/arrow.ts | Enhanced transposeData to compute struct validity bitmaps per Apache Arrow spec, marking struct rows as null only if all child fields are null, and allocating the bitmap only when needed. |

## Sequence Diagram
This diagram shows the interactions between components:

```mermaid
sequenceDiagram
    title Struct Data Validity Bitmap Computation

    participant Client as "Client Code"
    participant Transpose as "transposeData()"
    participant StructProcessor as "Struct Processing"
    participant ChildVectors as "Child Vectors"
    participant ValidityBitmap as "Validity Bitmap"
    participant MakeData as "makeData()"
    participant ArrowVector as "arrowMakeVector()"

    Client->>Transpose: Call with data, field, path
    
    Note over Transpose: Process struct field
    
    Transpose->>StructProcessor: Process child fields
    loop For each child field
        StructProcessor->>Transpose: Recursive call to transposeData()
        Transpose-->>StructProcessor: Return child vector
    end
    
    Note over StructProcessor: New validity bitmap computation
    
    StructProcessor->>ChildVectors: Get number of rows
    ChildVectors-->>StructProcessor: Return numRows
    
    StructProcessor->>ValidityBitmap: Initialize bitmap array
    
    loop For each row (i = 0 to numRows-1)
        StructProcessor->>StructProcessor: Set allChildrenNull = true
        
        loop For each child vector
            StructProcessor->>ChildVectors: child.get(i)
            ChildVectors-->>StructProcessor: Return value
            
            alt value !== null && value !== undefined
                StructProcessor->>StructProcessor: Set allChildrenNull = false
                StructProcessor->>StructProcessor: Break inner loop
            end
        end
        
        alt !allChildrenNull
            StructProcessor->>ValidityBitmap: Mark row as valid
            Note over ValidityBitmap: validityBitmap[i >> 3] |= 1 << (i & 7)
        else
            StructProcessor->>StructProcessor: Increment nullCount
        end
    end
    
    StructProcessor->>MakeData: Call with type, length, nullCount, etc.
    
    alt nullCount > 0
        StructProcessor->>MakeData: Include nullBitmap in params
    end
    
    MakeData-->>StructProcessor: Return structData
    StructProcessor->>ArrowVector: Call with structData
    ArrowVector-->>Client: Return Arrow Vector
```


▶️ ⚡ **AI Code Reviews for VS Code, Cursor, Windsurf**  
[Install the extension](https://marketplace.visualstudio.com/items?itemName=EntelligenceAI.EntelligenceAI-PR-Reviewer)

<details>
<summary>Note for Windsurf</summary>
Please change the default marketplace provider to the following in the windsurf settings:

Marketplace Extension Gallery Service URL: https://marketplace.visualstudio.com/_apis/public/gallery

Marketplace Gallery Item URL: https://marketplace.visualstudio.com/items
</details>

<details>
<summary>Entelligence.ai can learn from your feedback. Simply add 👍 / 👎 emojis to teach it your preferences. More shortcuts below</summary>

### **Emoji Descriptions:**
- ⚠️ **Potential Issue** - May require further investigation.
- 🔒 **Security Vulnerability** - Fix to ensure system safety.
- 💻 **Code Improvement** - Suggestions to enhance code quality.
- 🔨 **Refactor Suggestion** - Recommendations for restructuring code.
- ℹ️ **Others** - General comments and information.

### **Interact with the Bot:**
- Send a message or request using the format:
`@entelligenceai + *your message*`
```plaintext
Example: @entelligenceai Can you suggest improvements for this code?
```
- Help the Bot learn by providing feedback on its responses.
`@entelligenceai + *feedback*`
```plaintext
Example: @entelligenceai Do not comment on `save_auth` function !
```

Also you can trigger various commands with the bot by doing
`@entelligenceai command`

The current supported commands are
1. `config` - shows the current config
2. `retrigger_review` - retriggers the review

More commands to be added soon.
</details>

westonpace (changes requested):
I'm not sure this is an accurate reading of the Arrow spec.  Both Arrow and JS consider `NULL`, `{"outer": null}`, and `{"outer": {"inner": null}}` to be three distinct concepts.

If I understand this PR correctly it would turn the array...

```
[
  {"outer": null},
  {"outer": null},
  {"outer": null}
]
```

...into the array...

```
[
  null,
  null,
  null
]
```

...which is a destructive transformation.

>  but Arrow2 (used by LanceDB’s Rust backend)

Lance does not use arrow2 in any way except possibly in conversion to polars nor do we consider any quirks of that library to be canonical.

> Found unmasked nulls for non-nullable StructArray field "a"

This is an error that can also be generated by `arrow-rs` so I wonder if the error was coming from there?  I do think a validity bitmap might be missing for the struct array.  However, I think that validity bitmap should be calculated from `data` (if the record itself is null) and not from "all child fields being null".

wjones127:
I created https://github.com/lancedb/lancedb/pull/3680, which should fix this issue and address Weston's feedback.

PR #2411, opened 2025-05-30
feat: support new FTS features in python SDK

BubbleCal:
- AND operator
- phrase query slop param
- boolean query

## Summary by CodeRabbit

- **New Features**
  - Added support for combining full-text search queries using AND/OR operators, enabling more flexible query composition.
  - Introduced new query types and parameters, including boolean queries, operator selection, occurrence constraints, and phrase slop for advanced search scenarios.
  - Enhanced asynchronous search to accept rich full-text query objects directly.

- **Bug Fixes**
  - Improved handling and validation of full-text search queries in both synchronous and asynchronous search operations.

- **Tests**
  - Updated and expanded tests to cover new full-text query types and their usage in search functions.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -108,19 +108,7 @@ class Occur(Enum):
 
 
 class FullTextQuery(abc.ABC, pydantic.BaseModel):
-    _inner: PyFullTextQuery
-
-    @property
-    def inner(self) -> PyFullTextQuery:
-        """
-        Get the inner query object.
-
-        Returns
-        -------
-        PyFullTextSearchQuery
-            The inner query object.
-        """
-        return self._inner
+    _inner: PyFullTextQuery = pydantic.PrivateAttr()
 
     @abc.abstractmethod
     def query_type(self) -> FullTextQueryType:
@@ -200,6 +188,7 @@ class MatchQuery(FullTextQuery):
             The maximum number of terms to consider for fuzzy matching.
             Defaults to 50.
         """
+        super().__init__()
         self._inner = PyFullTextQuery.match_query(
             query,
             column,
@@ -225,6 +214,7 @@ class PhraseQuery(FullTextQuery):
         column : str
             The name of the column to match against.
         """
+        super().__init__()
         self._inner = PyFullTextQuery.phrase_query(query, column, slop)
 
     def query_type(self) -> FullTextQueryType:
@@ -251,8 +241,9 @@ class BoostQuery(FullTextQuery):
         negative_boost : float, default 0.5
             The boost factor for the negative query.
         """
+        super().__init__()
         self._inner = PyFullTextQuery.boost_query(
-            positive.inner, negative.inner, negative_boost
+            positive._inner, negative._inner, negative_boost
         )
 
     def query_type(self) -> FullTextQueryType:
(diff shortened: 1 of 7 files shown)


python/python/lancedb/query.py
```diff
@@ -93,10 +93,41 @@ class FullTextQueryType(Enum):
...
+    OR = "OR"
+
+
+class Occur(Enum):
+    MUST = "MUST"
+    SHOULD = "SHOULD"
+
+
+class FullTextQuery(pydantic.BaseModel):
+    _inner: PyFullTextQuery = pydantic.PrivateAttr()
```
wjones127: This change seems to make the fields hidden. This is unfortunate because it makes it harder to introspect the query one has created. It's also a bit of a breaking change.

Is the reason you took away the fields because you weren't sure how to handle the nested conversions from Python -> Rust? If so, I had a similar issue in Lance with Transactions and Operations. What I ended up doing is keeping the Python objects and implementing `FromPyObject` to do the conversions automatically. That way I could keep the dataclasses in Python, and have them easily convert to Rust. You can see the conversions implemented here: https://github.com/lancedb/lance/blob/main/python/src/transaction.rs

You'll notice the conversions are for `FromPyObject<'_> for PyLance<T> `. The `T` is the Rust type, and the `PyLance` wrapper is a newtype wrapper to allow implementing `FromPyObject`.

https://github.com/lancedb/lance/blob/a94d0593383e59d5f4e287407374586ce73de44f/python/src/utils.rs#L252-L255

Then I can directly take a `PyLance<Transaction>` as a parameter to a PyO3 function here:
https://github.com/lancedb/lance/blob/a94d0593383e59d5f4e287407374586ce73de44f/python/src/dataset.rs#L1500

PyO3 handles calling the `FromPyObject` trait on input parameters for you, and will make nested calls as necessary.

What do you think? Maybe that could make this change cleaner?

BubbleCal: it's really helpful, thanks!
working on it

python/python/tests/test_query.py
```diff
@@ -1065,18 +1067,27 @@ async def test_query_serialization_async(table_async: AsyncTable):
...
 
     # FTS queries
-    q = (await table_async.search("foo")).limit(10).to_query_object()
+    match_query = MatchQuery("foo", "text")
+    q = (await table_async.search(match_query)).limit(10).to_query_object()
     check_set_props(
         q,
         limit=10,
-        full_text_query=FullTextSearchQuery(columns=[], query="foo"),
+        full_text_query=FullTextSearchQuery(columns=None, query=match_query),
```
wjones127, line 1075: Just checking: is there any breaking change here? Are we still allowed to pass `[]`?

wjones127 (approved):

python/python/tests/test_query.py
BubbleCal: i think no breaking change
the `FullTextSearchQuery` can be either `columns & query_text` or `FullTextQuery object`, the former will be converted to and stored as `MatchQuery` or `PhraseQuery` at rust side, and when we deser it back to python, it will become `FullTextQuery object`, so `columns` is `None`.

wjones127: I mostly asked because you changed `columns=[]` to `columns=None`. Whenever I see a change in the unit tests, I have to ask whether it was done because of a breaking change in the API, or if it was just an improvement.

PR #2413, opened 2025-05-30
fix: avoid failing list_indices for any unknown index

### LINKED ISSUE #2412: bug(python): list_indices should skip fragment reuse index
### LanceDB version

_No response_

### What happened?

In Lance we just introduced a fragment reuse index https://github.com/lancedb/lance/pull/3847, currently this will make the `list_indices` call fail in various places (index type not found, also we assume 1 column must exist for the index statistics, which is not the case for this one). 

Overall we should skip this system internal index.

In addition, we should also think about forward compatibility, maybe not fail the whole listing operation and just omit the unknown indices.

_No response_

jackye1995:
Closes #2412 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the reliability of listing indices by logging warnings for errors and skipping problematic entries, ensuring successful results are returned.
	- Internal indices used for optimization are now excluded from the visible list of indices.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -14,7 +14,7 @@ use datafusion_physical_plan::projection::ProjectionExec;
 use datafusion_physical_plan::repartition::RepartitionExec;
 use datafusion_physical_plan::union::UnionExec;
 use datafusion_physical_plan::ExecutionPlan;
-use futures::{FutureExt, StreamExt, TryFutureExt, TryStreamExt};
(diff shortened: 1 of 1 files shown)


wjones127:
> It's a bit difficult to test this at the moment, once the lance version is upgraded I can add a test for it.

What if we create an impl:

```rust
impl TryFrom<&serde_json::Value> for IndexConfig {
    fn try_from(...) { ... }
}
```

Then we can unit test that?

rpgreen (approved):

PR #2421, opened 2025-06-04
docs: developing LanceDB rust

wjones127:
Adds docs telling users how to use the script `ci/set_lance_version.py`.
Also adds a CI job that will keep a `lance-main` branch up-to-date. This can be used by downstream repos that want to use `lancedb` and `lance` as submodules in their Rust GitHub project.

### DIFF
--- a/.github/workflows/lance-main.yml
+++ b/.github/workflows/lance-main.yml
@@ -14,20 +14,33 @@ concurrency:
 jobs:
   sync:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - name: Checkout commit branch
         uses: actions/checkout@v3
+        with:
+          path: lancedb
       - name: Checkout lance main branch
         uses: actions/checkout@v3
         with:
           repository: lancedb/lance
+          path: lance
       - name: Catch lance-main up to date
+        working-directory: lancedb
         run: |
           git checkout lance-main
           git fetch origin main
-          git rebase origin/main --exec "python ci/set_lance_version.py local && git add Cargo.toml Cargo.lock && git commit -m 'use local lance'"
+          for commit in $(git rev-list lance-main..main --reverse); do
+            git cherry-pick $commit --strategy-option=theirs
+            python ci/set_lance_version.py local
+            git add Cargo.toml Cargo.lock
+            git commit -m "use local lance" --allow-empty --no-verify
+          done
       - name: Push changes to lance-main branch
-        run: |
-          git push origin lance-main --force
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: ad-m/github-push-action@master
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/lance-main.yml
```diff
@@ -0,0 +1,46 @@
...
+        working-directory: lancedb
+        run: |
+          git checkout lance-main
+          git fetch origin main
+          for commit in $(git rev-list lance-main..main --reverse); do
+            git cherry-pick $commit --strategy-option=theirs
+            python ci/set_lance_version.py local
+            git add Cargo.toml Cargo.lock
+            git commit -m "use local lance" --allow-empty --no-verify
+          done
```
westonpace, line 39: Is this for loop expected to only run once?  Also, won't this create a bunch of `use local lance` commits? (not sure if it matters)

Could you just do `git reset --hard origin/main` instead of the for loop (though it would have to be before you checked out lance?

.github/workflows/lance-main.yml
```diff
@@ -0,0 +1,46 @@
...
+          path: lance
+      - name: Catch lance-main up to date
+        working-directory: lancedb
+        run: |
+          git checkout lance-main
+          git fetch origin main
+          for commit in $(git rev-list lance-main..main --reverse); do
+            git cherry-pick $commit --strategy-option=theirs
+            python ci/set_lance_version.py local
+            git add Cargo.toml Cargo.lock
```
westonpace, line 37: You don't add `lance`?

.github/workflows/lance-main.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - name: Checkout commit branch
+        uses: actions/checkout@v3
+        with:
+          path: lancedb
+      - name: Checkout lance main branch
+        uses: actions/checkout@v3
+        with:
+          repository: lancedb/lance
+          path: lance
```
westonpace, line 28: Why checkout and not submodule?

westonpace (approved):
These look like good additions.  I'm still wrapping my head around how the `lance-main` branch works (have a few questions) but I don't see any issues with it.

.github/workflows/lance-main.yml
wjones127: I was thinking we want a full parallel history in `lance-main`, so that you can select from that branch any commit in `main` but with the patched lance dependency. So if `main` is

```
A -> B -> C
```

Then `lance-main` is:
```
A -> local patch -> B -> local patch -> C -> local path
```

I had thought about using `--amend`, which might look cleaner. But I wasn't sure git would handle that well when generating the commit diff, so I kept as separate commits.

.github/workflows/lance-main.yml
wjones127: Hmm I hadn't considered it. I was assuming downstream use cases would already have a submodule of their own.

Issue #672, opened 2023-11-30
Tantivy FTS support at the Rust level


prrao87:
One idea I had during my [experiments](https://github.com/prrao87/lancedb-study#summary-of-results-for-10000-random-queries) was that there's clearly some level of blocking by the Python GIL going on when passing a LanceDB FTS query down to the FTS index layer via `tantivy-py`. This will always make concurrency an issue when issuing FTS queries at the LanceDB level.

## Feature request

I think it makes sense to add a feature request here: In a future version, perhaps it makes sense to incorporate tantivy's Rust crate with LanceDB instead of `tantivy-py`? And expose a Rust API that allows users at the LanceDB level (regardless of whether they use Python or JS) to run FTS queries with true concurrency via Rust tantivy. This would massively improve FTS query performance in a way that blows Elasticsearch/Opensearch out of the water.

I wish I knew enough Rust to help with this, but I'm getting there!

wjones127:
Work ongoing upstream in https://github.com/lancedb/lance/issues/1195

BubbleCal:
lance has implemented its own FTS, and we are deprecating that tantivy-python functions, check: https://lancedb.com/docs/search/full-text-search/

PR #2428, opened 2025-06-06
feat!: switch default FTS to native lance FTS

BubbleCal:


python/python/lancedb/query.py
```diff
@@ -1400,10 +1400,13 @@ def to_arrow(self, *, timeout: Optional[timedelta] = None) -> pa.Table:
 
         query = self._query
         if self._phrase_query:
-            raise NotImplementedError(
-                "Phrase query is not yet supported in Lance FTS. "
-                "Use tantivy-based index instead for now."
-            )
+            if isinstance(query, str):
+                if not query.startswith('"') or not query.endswith('"'):
```
jackye1995, line 1404: what about `query.startswith("'") and query.endswith("'")`?

BubbleCal: only text wrapped by " is considered as phrase query

leventov:
@BubbleCal @jackye1995 this PR didn't update a number of places in the docs and code comments that could be found by searching for "use_tantivy" on the repo. Most importantly, this page: https://lancedb.github.io/lancedb/fts/.

leventov:
Since you are on it, maybe you could also close https://github.com/lancedb/lance/issues/1195 and https://github.com/lancedb/lancedb/issues/672 because these issues seem to be done/outdated but currently they create a confusing picture by being open.

BubbleCal:
@leventov thanks for figuring it out.
just closed that 2 issues.
for that docs, we are not maintaining that github docs, the docs have been moved to https://lancedb.com/docs/indexing/fts-index/

PR #2430, opened 2025-06-06
feat: add maximum and minimum nprobes properties

westonpace:
This exposes the maximum_nprobes and minimum_nprobes feature that was added in https://github.com/lancedb/lance/pull/3903

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying minimum and maximum probe counts in vector search queries, allowing finer control over search behavior.
  - Users can now independently set minimum and maximum probes for vector and hybrid queries via new methods and parameters in Python, Node.js, and Rust APIs.

- **Bug Fixes**
  - Improved parameter validation to ensure correct usage of minimum and maximum probe values.

- **Tests**
  - Expanded test coverage to validate correct handling, serialization, and error cases for the new probe parameters.


python/python/tests/test_remote_db.py
```diff
@@ -496,6 +496,8 @@ def handler(body):
             "ef": None,
             "vector": [1.0, 2.0, 3.0],
             "nprobes": 20,
+            "minimum_nprobes": 20,
+            "maximum_nprobes": 20,
```
wjones127, line 500: Will older server versions error or just ignore these parameters? I would hope ignore.

westonpace: Yes, I tested this.  It appears to ignore.

Issue #2183, opened 2025-03-05
bug(node): Embedding function errors when vector column already exists


wjones127:
### LanceDB version

v0.15.1-beta.3

### What happened?

If you try to add data that already includes the vector column, and you have a registered embedding function, then you will get an error:

```
Error: Attempt to apply embeddings to table failed because column vector already existed
    at applyEmbeddingsFromMetadata (/Users/willjones/Documents/notebooks/ts_saas/node_modules/@lancedb/lancedb/dist/arrow.js:454:19)
    at async fromTableToBuffer (/Users/willjones/Documents/notebooks/ts_saas/node_modules/@lancedb/lancedb/dist/arrow.js:631:33)
    at async LocalTable.add (/Users/willjones/Documents/notebooks/ts_saas/node_modules/@lancedb/lancedb/dist/table.js:67:24)
    at async recreate_table (/Users/willjones/Documents/notebooks/ts_saas/index.js:100:5)
```

In Python, we made is so that:

1. If the vector column exists, and it's all null, then we fill the column
2. If the vector column exists, and it's not null, then we leave it alone

PR #2433, opened 2025-06-10
fix: improve error handling and embedding logic in arrow.ts

satya-nutella:
- Enhanced error messages for schema inference failures to suggest providing an explicit schema.
- Updated embedding application logic to check for existing destination columns, allowing for filling embeddings in columns that are all null.
- Added comments for clarity on handling existing columns during embedding application.

Fixes https://github.com/lancedb/lancedb/issues/2183

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error messages for schema inference to enhance readability.
  - Prevented redundant embedding application by skipping columns that already contain data, avoiding unnecessary errors and computations.

### DIFF
--- a/nodejs/lancedb/arrow.ts
+++ b/nodejs/lancedb/arrow.ts
@@ -417,7 +417,9 @@ function inferSchema(
         } else {
           const inferredType = inferType(value, path, opts);
           if (inferredType === undefined) {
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Nice work!

PR #2438, opened 2025-06-12
Feat(python): improve `pydantic` support

### LINKED ISSUE #2436: bug(python): aliases in `LanceModel` are not recognised
### LanceDB version

v0.23.0

### What happened?

When a `LanceModel` definition contains field aliases, `to_pydantic` method fails to recognise them, raising a `ValidationError` instead. I'm reporting this as a bug since `FieldInfo` docstring does mention `alias`.

```python
import lancedb
from pydantic import Field, computed_field

uri = "data/sample-lancedb"
db = lancedb.connect(uri)

class Item(LanceModel):
    name: str = Field(alias='item')
    price: float
    distance: float = Field(alias='_distance')

    @computed_field
    def similarity(self) -> float:
        return 1 - self.distance

data = [
    {"vector": [3.1, 4.1], "item": "foo", "price": 10.0},
]

tbl = db.create_table("my_table_async", data=data, mode='overwrite')
result = tbl.search([100, 100]).distance_type('cosine').limit(2)

# this works
print([Item(**item) for item in result.to_list()])
# [Item(name='foo', price=10.0, distance=0.009507715702056885, similarity=0.9904922842979431), Item(name='bar', price=20.0, distance=0.15612411499023438, similarity=0.8438758850097656)]

# this raises a ValidationError
print(result.to_pydantic(Item))
# ValidationError: 2 validation errors for Item
# item
#   Field required [type=missing, input_value={'price': 10.0}, input_type=dict]
#     For further information visit https://errors.pydantic.dev/2.11/v/missing
# _distance
#   Field required [type=missing, input_value={'price': 10.0}, input_type=dict]
#     For further information visit https://errors.pydantic.dev/2.11/v/missing
```

### LINKED ISSUE #2437: Feature: support `to_pydantic` for `AsyncVectorQuery`
### SDK

Python

### Description

Currently, `to_pydantic` method is only implemented for an asynchronous API. It would be helpful to add the same method for async connections, i.e.:

```python
import lancedb

uri = "data/sample-lancedb"

class Item(LanceModel):
    item: str
    price: float

data = [
    {"vector": [3.1, 4.1], "item": "foo", "price": 10.0},
]

# in a sync connection, this works
db = lancedb.connect(uri)
tbl = db.create_table("my_table_async", data=data, mode='overwrite')
print(tbl.search([100, 100], query_type='vector').distance_type('cosine').limit(2).to_pydantic(Item))
# [Item(item='foo', price=10.0), Item(item='bar', price=20.0)]

# in an async connection, it is not implemented 
db = await lancedb.connect_async(uri)
tbl = await db.open_table("my_table_async")
print(await tbl.vector_search([100, 100]).distance_type('cosine').limit(2).to_pydantic(Item))
# AttributeError: 'AsyncVectorQuery' object has no attribute 'to_pydantic'
```

On a related note, I would also expect the `search` method to work the same way when using an asynchronous connection, and yet it fails:

```python
db = await lancedb.connect_async(uri)
tbl = await db.open_table("my_table_async")
print(await tbl.search([100, 100], query_type='vector').distance_type('cosine').limit(2).to_pydantic(Item))
# AttributeError: 'coroutine' object has no attribute 'distance_type'
```

Are there any design constraints that require async API to have `vector_search` specifically instead of overloading the existing `search` method?

mykolaskrynnyk:
This request improves support for `pydantic` integration by adding `to_pydantic` method to asynchronous queries and handling models that use `alias` in field definitions. Fixes #2436 and closes #2437 .

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -746,10 +746,7 @@ class LanceQueryBuilder(ABC):
         -------
         List[LanceModel]
         """
-        return [
-            model(**{k: v for k, v in row.items() if k in model.field_names()})
-            for row in self.to_arrow(timeout=timeout).to_pylist()
-        ]
+        return [model(**row) for row in self.to_arrow(timeout=timeout).to_pylist()]
 
     def to_polars(self, *, timeout: Optional[timedelta] = None) -> "pl.DataFrame":
         """
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -2327,6 +2327,28 @@ class AsyncQueryBase(object):
 
         return pl.from_arrow(await self.to_arrow(timeout=timeout))
 
+    async def to_pydantic(
+        self, model: Type[LanceModel], *, timeout: Optional[timedelta] = None
+    ) -> List[LanceModel]:
+        """
+        Convert results to a list of pydantic models.
+
+        Parameters
+        ----------
+        model : Type[LanceModel]
+            The pydantic model to use.
+        timeout : timedelta, optional
+            The maximum time to wait for the query to complete.
+            If None, wait indefinitely.
+
+        Returns
+        -------
+        list[LanceModel]
+        """
+        return [
+            model(**row) for row in (await self.to_arrow(timeout=timeout)).to_pylist()
+        ]
+
     async def explain_plan(self, verbose: Optional[bool] = False):
         """Return the execution plan for this query.
 
--- a/python/python/tests/test_pydantic.py
+++ b/python/python/tests/test_pydantic.py
@@ -440,3 +440,31 @@ def test_aliases_in_lance_model(mem_db):
     assert hasattr(model, "name")
     assert hasattr(model, "distance")
     assert model.similarity > 0.99
+
+
+@pytest.mark.asyncio
+async def test_aliases_in_lance_model_async(mem_db_async):
+    data = [
+        {"vector": [8.3, 2.5], "item": "foo", "price": 12.0},
+        {"vector": [7.7, 3.9], "item": "bar", "price": 11.2},
+    ]
+    tbl = await mem_db_async.create_table("items", data=data)
+
+    class TestModel(LanceModel):
+        name: str = Field(alias="item")
+        price: float
+        distance: float = Field(alias="_distance")
+
+        @computed_field
+        def similarity(self) -> float:
+            return 1 - self.distance
+
+    model = (
+        await tbl.vector_search([7.7, 3.9])
+        .distance_type("cosine")
+        .limit(1)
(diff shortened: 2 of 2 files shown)


mykolaskrynnyk:
Hi @wjones127 , I updated the tests to ensure they work with `pydantic` v1 too. Just checking if there is anything else that needs my attention before this PR is good to go. Thanks.

python/python/lancedb/query.py
```diff
@@ -746,10 +746,7 @@ def to_pydantic(
         -------
         List[LanceModel]
         """
-        return [
-            model(**{k: v for k, v in row.items() if k in model.field_names()})
-            for row in self.to_arrow(timeout=timeout).to_pylist()
-        ]
+        return [model(**row) for row in self.to_arrow(timeout=timeout).to_pylist()]
```
wjones127, line 767: This change seems concerning, but it didn't break any existing tests. I'm not sure why the filter on `model.field_names()` was there in the first place. Was it to ignore extra fields? What do you think?

wjones127 (commented):
Wouldn't mind some input on one change. otherwise look good.

python/python/lancedb/query.py
wjones127: @changhiskhan do you remember why we had that filter in the original PR? https://github.com/lancedb/lancedb/pull/384

mykolaskrynnyk: @wjones127 , this must have been added to ignore extra fields. But if I remember correctly, they were ignored by default even in v1 anyway. Besides, the original implementation basically ignores what the user might consider desirable – the original implementation would filter extra fields anyway. Consider this:

```python
import lancedb
from pydantic import ConfigDict

uri = "data/sample-lancedb"
db = lancedb.connect(uri)

class Item(LanceModel):
    model_config = ConfigDict(extra="allow")

    name: str

data = [
    {"vector": [3.1, 4.1], "name": "foo", "price": 10.0},
    {"vector": [5.9, 26.5], "name": "bar", "price": 20.0},
]

tbl = db.create_table("table", data=data, mode='overwrite')
tbl.search(None).limit(2).to_pydantic(Item)
```

In the original implementation, it does not matter if `extra` is set to `"allow"`, `"ignore"` or even `"forbid"` – whatever the value is, the conversion will just ignore extra fields. The proposed change respects the config setting in the model. So I think it's good to go.

mykolaskrynnyk: @wjones127 , just checking if there is still anything that needs resolution before this minor PR is merged.

mykolaskrynnyk: @wjones127 , can we move forward with this? It'd be really useful to have these `pydantic` features in the next LanceDB release.

PR #2441, opened 2025-06-17
feat: support prefix matching and must_not clause

BubbleCal:
## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a prefix length in fuzzy full-text search, allowing more precise control over which parts of a word are matched fuzzily.
  - Introduced the ability to perform boolean full-text search queries, including "must", "should", and "must not" conditions.

- **Tests**
  - Expanded test coverage for fuzzy search with prefix length and boolean query logic, ensuring correct results for various query types.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -101,8 +101,9 @@ class FullTextOperator(str, Enum):
 
 
 class Occur(str, Enum):
-    MUST = "MUST"
     SHOULD = "SHOULD"
+    MUST = "MUST"
+    MUST_NOT = "MUST_NOT"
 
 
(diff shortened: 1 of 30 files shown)


jackye1995 (approved):
looks good to me!

python/python/lancedb/query.py
```diff
@@ -101,8 +101,9 @@ class FullTextOperator(str, Enum):
 
 
 class Occur(str, Enum):
-    MUST = "MUST"
```
jackye1995, line 104: nit: might be better to keep the enum order the same since it changes the implicit ordinal value of it.

PR #2448, opened 2025-06-19
test: add weekly test

BubbleCal:
passed on main: https://github.com/BubbleCal/lancedb/actions/runs/15774996506
failed on 0.23.0 because of a known and fixed bug

## Summary by CodeRabbit

- **New Features**
  - Introduced a new suite of comprehensive weekly tests for large datasets, including randomized indexing, searching, and mutation operations.
  - Added a new workflow to automatically run weekly tests against the latest stable releases and the main branch, including upgrade scenarios.
- **Chores**
  - Updated test configuration to include a "weekly" marker, ensuring these tests are only run as scheduled or explicitly requested.
  - Adjusted existing test runs to exclude "weekly" tests from regular test executions.


rpgreen (commented):
Please add a PR description of the problem and proposed solution (not CodeRabbit generated description) and issue link if there is one.

I don't have any concerns with the tests per-se, but what is the process for reviewing the test results and following up?

Also we may want to give a more descriptive name to the tests instead of "weekly tests" since that doesn't say much and the cadence may change in the future.

wjones127:
Agreed that would like more description as to why we are adding this. What types of bugs will it catch?

In particular, I'm wondering why it's appropriate here (in LanceDB repo)? I would think most bugs in terms of sequence of operations are in Lance, not LanceDB. So maybe it makes more sense to do it in that repo.

Another question: if we get a failure, how easy is it to figure out which sequence of changes led to that failure?

BubbleCal:
agree it would be better to be in lance, so that we can control more operations, moved to https://github.com/lancedb/lance/pull/4190

PR #2453, opened 2025-06-21
feat: batch Ollama embed calls

wengh:
Other embedding integrations such as Cohere and OpenAI already send requests in batches. We should do that for Ollama too to improve throughput.


python/python/lancedb/embeddings/ollama.py
```diff
@@ -54,8 +53,8 @@ def generate_embeddings(
             The texts to embed
         """
         # TODO retry, rate limit, token limit
-        embeddings = [self._compute_embedding(text) for text in texts]
-        return embeddings
+        embeddings = self._compute_embedding(texts)
+        return list(embeddings)
```
wengh, line 57: This doesn't match the return type hint of `list[Union[np.array, None]]` but other embedding integrations seem to be returning list of lists of floats as well.

PR #2471, opened 2025-06-27
Improved Delete method with logging and error handling

### LINKED ISSUE #2434: bug(python): AsyncTable.delete deletes all the indices as a side effect
### LanceDB version

0.23.0

### What happened?

the `AsyncTable.delete` deletes also all the indices from the table

```python
import shutil
from pathlib import Path

import lancedb
import lancedb.pydantic
import pandas as pd
import pytest

from lancedb.index import BTree

class _CreateTableTestSchema(lancedb.pydantic.LanceModel):
    pk: str
    txt: str

@pytest.fixture(scope="function")
def lancedb_path_for_created_table(tmp_path_factory: pytest.TempPathFactory):
    tmp_directory = tmp_path_factory.mktemp("lancedb", numbered=True)

    lancedb_path = tmp_directory / "table"

    yield lancedb_path.absolute()

    shutil.rmtree(tmp_directory.absolute())

@pytest.fixture(scope="module")
def lancedb_table_name() -> str:
    return "test_table"


@pytest.mark.asyncio
async def test_create_empty_table_with_scalar_indices_with_delete_raw(
    lancedb_path_for_created_table: Path, lancedb_table_name: str
):
    db = await lancedb.connect_async(str(lancedb_path_for_created_table))
    tbl = await db.create_table(name=lancedb_table_name, mode="create", schema=_CreateTableTestSchema)
    await tbl.create_index(column="txt", config=BTree())

    indices = await tbl.list_indices()
    assert [i.columns for i in indices] == [["txt"]]
    assert [i.index_type for i in indices] == ["BTree"]

    data_list = [f"{i:00}" for i in range(10)]
    add_result = await tbl.add(data=pd.DataFrame({"pk": data_list, "txt": data_list}))
    indices = await tbl.list_indices()
    assert [i.columns for i in indices] == [["txt"]]
    assert [i.index_type for i in indices] == ["BTree"]
    n_rows = await tbl.count_rows()
    assert n_rows == 10
    print("ADD RESULT", add_result)

    delete_result = await tbl.delete(where="txt = '1'")
    n_rows = await tbl.count_rows()
    assert n_rows == 9
    indices = await tbl.list_indices()
    assert [i.columns for i in indices] == [["txt"]]
    assert [i.index_type for i in indices] == ["BTree"]
    print("DELETE RESULT:", delete_result)
```

The last two asserts fail

coder-aditi:
Fix #2434

Modifications ->

1) The delete method specified on the BaseTable implementation is now called by the Table::delete method.
2) This guarantees uniform behavior across all table backends, including remote and NativeTable.
3) Enhanced propagation of errors from lower-level dataset logic.

Testing confirmed by comparing to current integration and unit tests.

For a range of SQL predicates, manual tests verify proper deletion behavior.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3772,7 +3772,30 @@ class AsyncTable:
            x      vector
         0  3  [5.0, 6.0]
         """
-        return await self._inner.delete(where)
+
+        existing_indices = await self.list_indices()
+        print("🛠 Before delete, found indices:", existing_indices)
+
+        saved_index_info = []
+        for idx in existing_indices:
+            print("🛠 Saving index:", idx.columns, idx.index_type)
+            saved_index_info.append({
+                "columns": idx.columns,
+                "config": idx.config,
+                "name": idx.name,
+            })
+
+        result = await self._inner.delete(where)
+
+        for idx in saved_index_info:
+            column = idx["columns"][0]
+            config = idx["config"]
+            name = idx.get("name")
+            print(f"🛠 Recreating index on {column} with config={config} and name={name}")
+            await self.create_index(column=column, config=config, name=name)
+
+        return result
+
 
     async def update(
         self,
--- a/python/python/tests/docs/test_delete_preserves_indices.py
+++ b/python/python/tests/docs/test_delete_preserves_indices.py
@@ -1,0 +1,50 @@
+import shutil
+from pathlib import Path
+
+import lancedb
+import lancedb.pydantic
+import pandas as pd
+import pytest
+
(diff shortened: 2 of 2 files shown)


wjones127:
I'm actively working on this right now. I don't think re-creating the indices like this is what we want right now. Mainly because in between the delete and the create_index calls, you will have a temporary state where the indices are missing.

I'm working on this upstream in https://github.com/lancedb/lance/issues/3940

coder-aditi:
Ohh Great @wjones127 , will take up something else then. Thank you.

PR #2474, opened 2025-06-29
fix: handling empty result sets after hard filtering

### LINKED ISSUE #2425: bug(python): Graceful Handling of Empty Result Sets After Hard Filtering
### LanceDB version

_No response_

### What happened?

Description:
When applying hard filters via LanceDB and the result set is empty, a list index out of range exception is raised during subsequent rerank operations. This occurs because the downstream logic assumes at least one result is always returned after search/filter, which isn’t guaranteed.

```bash

../.venv/lib/python3.11/site-packages/lancedb/query.py:334: in to_list
../.venv/lib/python3.11/site-packages/lancedb/query.py:1199: in to_arrow
    return self._combine_hybrid_results(
../.venv/lib/python3.11/site-packages/lancedb/query.py:1248: in _combine_hybrid_results
    results = reranker.rerank_hybrid(fts_query, vector_results, fts_results)
../.venv/lib/python3.11/site-packages/lancedb/rerankers/answerdotai.py:68: in rerank_hybrid
../.venv/lib/python3.11/site-packages/lancedb/rerankers/answerdotai.py:50: in _rerank
    result = self.reranker.rank(query, docs, doc_ids=doc_ids)
../.venv/lib/python3.11/site-packages/rerankers/models/colbert_ranker.py:264: in rank
    docs = prep_docs(docs, doc_ids, metadata)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

docs = [], doc_ids = [], metadata = None

    def prep_docs(
        docs: Union[str, List[str], Document, List[Document]],
        doc_ids: Optional[Union[List[str], List[int]]] = None,
        metadata: Optional[List[dict]] = None,
    ):
        if isinstance(docs, Document) or (
>           isinstance(docs, List) and isinstance(docs[0], Document)
        ):
E       IndexError: list index out of range

```

_No response_

coder-aditi:
Fix #2425

This PR's modifications fix a bug that resulted in an IndexError when reranking was attempted on an empty result set following the application of hard filters. The AnswerDotAI._rerank method now has a safeguard to identify and return early if the input pyarrow.Because the table is empty, operations on empty data cannot be performed downstream. This increases the robustness of hybrid search workflows and guarantees the graceful handling of empty results. To confirm this behavior and avoid regressions, a matching unit test was also included.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1745,6 +1745,9 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
                 LanceHybridQueryBuilder._normalize_scores(original_scores),
             )
 
+        if vector_results.num_rows == 0 and fts_results.num_rows == 0:
+            return pa.table({})
+
         results = reranker.rerank_hybrid(fts_query, vector_results, fts_results)
 
         check_reranker_result(results)
--- a/python/python/tests/test_empty_results_handling.py
+++ b/python/python/tests/test_empty_results_handling.py
@@ -4,14 +4,17 @@ import pytest
 from lancedb.query import LanceHybridQueryBuilder
 from lancedb.rerankers.answerdotai import AnswerDotAIReranker
 
+
 @pytest.mark.asyncio
 async def test_combine_hybrid_results_empty():
-    schema = pa.schema([
-        pa.field("_rowid", pa.int64()),
-        pa.field("_distance", pa.float32()),
-        pa.field("_score", pa.float32()),
-        pa.field("content", pa.string())
-    ])
+    schema = pa.schema(
+        [
+            pa.field("_rowid", pa.int64()),
+            pa.field("_distance", pa.float32()),
+            pa.field("_score", pa.float32()),
+            pa.field("content", pa.string()),
+        ]
+    )
 
     empty_vector_results = pa.Table.from_batches([], schema=schema)
     empty_fts_results = pa.Table.from_batches([], schema=schema)
--- a/python/python/tests/test_empty_results_handling.py
+++ b/python/python/tests/test_empty_results_handling.py
@@ -1,0 +1,43 @@
+import pyarrow as pa
+import pytest
+
+from lancedb.query import LanceHybridQueryBuilder
+from lancedb.rerankers.answerdotai import AnswerDotAIReranker
+
+@pytest.mark.asyncio
+async def test_combine_hybrid_results_empty():
+    schema = pa.schema([
+        pa.field("_rowid", pa.int64()),
+        pa.field("_distance", pa.float32()),
+        pa.field("_score", pa.float32()),
+        pa.field("content", pa.string())
+    ])
+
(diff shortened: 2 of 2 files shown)


LuQQiu:
@coder-aditi Could you help fix the python format issue  https://github.com/lancedb/lancedb/actions/runs/15957356770/job/45081787746?pr=2474

BubbleCal (commented):
LGTM
@AyushExel I'd assume it's caused by calling `rerank_hybrid` with empty results, would you think we need to handle this case in reranker?

AyushExel:
Hmm I think this case should be handled here - https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/rerankers/answerdotai.py#L50 not sure why this still happens. @coder-aditi can you make sure you're on latest lancedb version and verify?

coder-aditi:
> @coder-aditi Could you help fix the python format issue https://github.com/lancedb/lancedb/actions/runs/15957356770/job/45081787746?pr=2474

@LuQQiu Done and committed the code, please review and let me know

coder-aditi:
Hello @AyushExel , I appreciate your feedback!

Yes, empty result sets are handled by the _rerank() method using _handle_empty_results(), and this is operating as planned. However, when both vector_results and fts_results are empty, the error we were experiencing occurs in _combine_hybrid_results() before _rerank() is ever called.

When the reranker is used with these empty tables, downstream problems (such as list index out of range) may arise because certain columns, such as _rowid, _score, or _distance, may not exist.

I included an early return in _combine_hybrid_results() to safely handle this:

```
if vector_results.num_rows == 0 and fts_results.num_rows == 0:
    return pa.table({})
```

AyushExel:
Ahh ok got it.. Looks good, thanks for catching that!

coder-aditi:
Thank you @AyushExel

AyushExel (approved):

AyushExel:
@coder-aditi please fix the tests. Otherwise looks good. Custom rerankers tests are to be marked as optional, these are only tested locally (see test_rerankers.py). You can use  the default reranker for testing - `RRFReranker` and mark AnswerDotAI reranker as optional. Just verify it works locally

coder-aditi:
Hi @AyushExel , as per your suggestion, I updated the tests. 
For the empty results test, I moved to using the RRFReranker, which allows it to function in any environment without the need for external dependencies. 
In keeping with the other custom rerankers in test_rerankers.py, I also ensured that the AnswerDotAI reranker test is marked as optional using pytest.importorskip(). 
I verified that the AnswerDotAI reranker functions as intended with empty results by running those optional tests locally. To me, everything appears to be in order now.

AyushExel:
@coder-aditi the tests are still failing. Maybe you forgot to push the local changes :) ? Can you double check? THanks!

wjones127:
This was closed by another PR.

PR #2479, opened 2025-06-30
fix: only monotonically update dataset

wjones127:
Make sure we only update the latest version if it's actually newer. This is important if there are concurrent queries, as they can take different amounts of time.

### DIFF
--- a/rust/lancedb/src/table/dataset.rs
+++ b/rust/lancedb/src/table/dataset.rs
@@ -129,7 +129,9 @@ impl DatasetRef {
                 dataset: ref mut ds,
                 ..
             } => {
-                *ds = dataset;
+                if dataset.manifest().version > ds.manifest().version {
+                    *ds = dataset;
+                }
             }
             _ => unreachable!("Dataset should be in latest mode at this point"),
         }


rust/lancedb/src/table/dataset.rs
```diff
@@ -129,7 +129,9 @@ impl DatasetRef {
                 dataset: ref mut ds,
                 ..
             } => {
-                *ds = dataset;
+                if dataset.manifest().version > ds.manifest().version {
+                    *ds = dataset;
```
BubbleCal, line 133: Is it possible that 2 threads are both trying to update this, and one updates it to version 3, but the other one then updates it to version 2?

Asking because here comparison and update is not atomic

westonpace: I think, since we have `&mut self`, we are guaranteed that two threads cannot be in this method.  There is a mutex higher up in `DatasetConsistencyWrapper`

westonpace (approved):
Is it worth updating the docs of `DatasetConsistencyWrapper::set_latest` just so future readers know they can call this method without concern?

PR #2484, opened 2025-07-06
fix: change 'return' to 'raise' for unimplemented remote table function

CyrusAttoun:
just noticed that we're doing a 'return' instead of a 'raise' while trying to get remote functionality working for my project. I went ahead and implemented tests for both of the unimplemented functions (to_pandas and to_arrow) while I was in there.


eddyxu (approved):

eddyxu:
@CyrusAttoun could you mind to fix test failures?

eddyxu:
Thanks @CyrusAttoun for your contribution!

PR #2488, opened 2025-07-08
feat: support to specify num_partitions and num_bits

BubbleCal:


rust/lancedb/src/remote/table.rs
```diff
@@ -997,23 +997,53 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
             // yet support them.
-            Index::IvfFlat(index) => ("IVF_FLAT", Some(index.distance_type)),
-            Index::IvfPq(index) => ("IVF_PQ", Some(index.distance_type)),
-            Index::IvfHnswSq(index) => ("IVF_HNSW_SQ", Some(index.distance_type)),
-            Index::BTree(_) => ("BTREE", None),
-            Index::Bitmap(_) => ("BITMAP", None),
-            Index::LabelList(_) => ("LABEL_LIST", None),
+            Index::IvfFlat(index) => {
+                body["index_type"] = serde_json::Value::String("IVF_FLAT".to_string());
+                body["metric_type"] =
```
westonpace: Pretty minor but since we repeat `index_type` and `metric_type` a lot maybe it would be good to have these as constants?

PR #2477, opened 2025-06-30
fix: can not build on python3.9 for dev

yihong0618:
This patch fix can not build on python3.9 dev

the reason is that for ibm-watsonx-ai the min version is py3.10

more can check on `pyoven` https://pyoven.org/package/ibm-watsonx-ai/

also fix tiny md lint

PR #2499, opened 2025-07-10
feat(embeddings): add siglip embedding support to lancedb

### LINKED ISSUE #2498: Feature:  Add support for SigLIP embedding model in the LanceDB embedding registry
### SDK

Python

###  Feature Request: Support SigLIP Embeddings

I'd like to add support for [SigLIP (Sigmoid Loss CLIP)](https://huggingface.co/google/siglip-base-patch16-224) in the LanceDB embedding registry.

### Why SigLIP
- Improved generalization: SigLIP replaces the contrastive loss in CLIP with a sigmoid loss, leading to better calibration and zero-shot performance.

- Better retrieval quality: In my internal benchmarks on an image-to-text retrieval task:

        - mAP: SigLIP scored 1.000 vs OpenCLIP's 0.917

        - R@1/5/10: All 1.000 for both

- Tradeoff: SigLIP uses a 768-dimensional embedding (vs OpenCLIP’s 512), which leads to slightly slower embedding generation awhile search latency is negligible, but significantly improves accuracy in retrieval tasks.

### Use Cases Where SigLIP Excels
- High-accuracy image-to-text retrieval (e.g., visual search engines, content moderation)
- Zero-shot classification (where labels may not be seen during training)
- Multilingual or multimodal datasets with diverse semantic content
- Personalized search or RAG pipelines, where embedding quality directly impacts response relevance

Heisenberg208:
###  Summary

This PR adds **SigLIP** (Sigmoid Loss Image Pretraining) as a new embedding model in the LanceDB embedding registry. SigLIP improves image-text alignment performance using sigmoid-based contrastive loss and offers robust zero-shot generalization.

Fixes #2498 

#### 1. `SigLIP` Embedding Class

* Added `SigLIP` support under `python/lancedb/embeddings/siglip.py`
* Implements:

  * `compute_source_embeddings`
  * `_batch_generate_embeddings`
  * Normalization logic
  * Batch-wise progress logging for image embedding

#### 2. Registry Integration

* Registered `SigLIP` in `embeddings/__init__.py`
* `SigLIP` now usable via `connect(..., embedding="siglip")`

#### 3. Evaluation Benchmark Support

* Added SigLIP to `test_embeddings_slow.py` for side-by-side benchmarking with OpenCLIP and ImageBind

####  `test_siglip`

* End-to-end test to verify embeddings table creation and vector shape for SigLIP
WhatsApp Image 2025-07-10 at 18 00 27_a3368163

####  `test_siglip_vs_openclip_vs_imagebind_benchmark_full`

* Benchmarks:

  * **Recall\@1 / 5 / 10**
  * **mAP (Mean Average Precision)**
  * **Embedding & Search Latency**
  * Dimensionality reporting
WhatsApp Image 2025-07-10 at 18 12 13_22c67a84

###  Notes

* SigLIP outputs 768D embeddings (vs 512D for OpenCLIP)
* Benchmark shows competitive performance despite higher dimensionality
* I'm still new to contributing to open-source and learning as I go.
Please feel free to suggest any improvements — I'm happy to make changes!

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -15,7 +15,7 @@ description = "lancedb"
 authors = [{ name = "LanceDB Devs", email = "dev@lancedb.com" }]
 license = { file = "LICENSE" }
 readme = "README.md"
-requires-python = ">=3.10,<3.13"
+requires-python = ">=3.9"
 keywords = [
     "data-format",
     "data-science",
@@ -66,7 +66,7 @@ dev = [
     "ruff",
     "pre-commit",
     "pyright",
-    'typing-extensions>=4.0.0; python_version < "3.13"',
+    'typing-extensions>=4.0.0; python_version < "3.11"',
 ]
 docs = ["mkdocs", "mkdocs-jupyter", "mkdocs-material", "mkdocstrings-python"]
 clip = ["torch", "pillow", "open-clip-torch"]
@@ -90,7 +90,6 @@ embeddings = [
     "ibm-watsonx-ai>=1.1.2",
     "transformers>=4.41.0",
     "sentencepiece",
-
 ]
 azure = ["adlfs>=2024.2.0"]
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -61,7 +61,6 @@ tests = [
     "pylance>=0.25",
     "requests",
     "datafusion",
-    "scikit-learn"
 ]
 dev = [
     "ruff",
@@ -71,7 +70,7 @@ dev = [
 ]
 docs = ["mkdocs", "mkdocs-jupyter", "mkdocs-material", "mkdocstrings-python"]
 clip = ["torch", "pillow", "open-clip-torch"]
-siglip = ["torch", "pillow", "transformers>=4.41.0"]
+siglip = ["torch", "pillow", "transformers>=4.41.0","sentencepiece"]
 embeddings = [
     "requests>=2.31.0",
     "openai>=1.6.1",
@@ -89,8 +88,7 @@ embeddings = [
     "botocore>=1.31.57",
     "ollama>=0.3.0",
     "ibm-watsonx-ai>=1.1.2",
-    "transformers>=4.41.0",
-    "sentencepiece",
+    "sentencepiece"
 ]
 azure = ["adlfs>=2024.2.0"]
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -61,6 +61,7 @@ tests = [
     "pylance>=0.25",
     "requests",
     "datafusion",
+    "scikit-learn"
 ]
 dev = [
     "ruff",
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -86,8 +86,8 @@ embeddings = [
     "boto3>=1.28.57",
     "awscli>=1.29.57",
     "botocore>=1.31.57",
+    'ibm-watsonx-ai>=1.1.2; python_version >= "3.10"',
     "ollama>=0.3.0",
-    "ibm-watsonx-ai>=1.1.2",
     "sentencepiece"
 ]
 azure = ["adlfs>=2024.2.0"]
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -15,7 +15,7 @@ description = "lancedb"
 authors = [{ name = "LanceDB Devs", email = "dev@lancedb.com" }]
 license = { file = "LICENSE" }
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10,<3.13"
 keywords = [
     "data-format",
     "data-science",
@@ -66,10 +66,11 @@ dev = [
     "ruff",
     "pre-commit",
     "pyright",
-    'typing-extensions>=4.0.0; python_version < "3.11"',
+    'typing-extensions>=4.0.0; python_version < "3.13"',
 ]
-docs = ["mkdocs", "mkdocs-jupyter", "mkdocs-material", "mkdocstrings[python]"]
+docs = ["mkdocs", "mkdocs-jupyter", "mkdocs-material", "mkdocstrings-python"]
 clip = ["torch", "pillow", "open-clip-torch"]
+siglip = ["torch", "pillow", "transformers>=4.41.0"]
 embeddings = [
     "requests>=2.31.0",
     "openai>=1.6.1",
@@ -87,6 +88,9 @@ embeddings = [
     "botocore>=1.31.57",
     "ollama>=0.3.0",
     "ibm-watsonx-ai>=1.1.2",
+    "transformers>=4.41.0",
+    "sentencepiece",
+
 ]
 azure = ["adlfs>=2024.2.0"]
 
@@ -113,7 +117,7 @@ markers = [
 include = [
     "python/lancedb/index.py",
     "python/lancedb/rerankers/util.py",
-    "python/lancedb/rerankers/__init__.py",
+    "python/lancedb/rerankers/_init_.py",
     "python/lancedb/rerankers/voyageai.py",
     "python/lancedb/rerankers/jinaai.py",
     "python/lancedb/rerankers/openai.py",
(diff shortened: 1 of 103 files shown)


Heisenberg208:
Hi @houjun @wujixiu @wjones127 @jackye1995 
Just following up to see if you could take a look at this PR when you have a moment.

Let me know if anything needs changes or clarification — happy to help!
Thanks again

python/pyproject.toml
```diff
@@ -113,7 +117,7 @@ markers = [
 include = [
     "python/lancedb/index.py",
     "python/lancedb/rerankers/util.py",
-    "python/lancedb/rerankers/__init__.py",
+    "python/lancedb/rerankers/_init_.py",
```
jackye1995: these looks like invalid changes?

python/pyproject.toml
```diff
@@ -15,7 +15,7 @@ description = "lancedb"
 authors = [{ name = "LanceDB Devs", email = "dev@lancedb.com" }]
 license = { file = "LICENSE" }
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10,<3.13"
```
jackye1995: can we avoid changing python version in the PR?

Heisenberg208: Thanks for pointing that out! @jackye1995 

Just to clarify, the change to `requires-python = ">=3.10,<3.13"` is not directly related to the SigLIP embeddings I implemented in this PR.

I had updated the Python version because `ibm-watsonx-ai >=1.1.2 `(used by the watsonx embedding function in the registry) requires Python ≥3.10. Since it’s part of the embedding registry, I assumed aligning the Python version range would help avoid downstream issues.

That said, if keeping broader compatibility is preferred, I’m happy to revert this.
I will raise a follow-up PR for the SigLIP documentation (similar to OpenCLIP’s doc structure) once this changes are valid.
Let me know what you'd recommend!

jackye1995: Thank you for being considerate about this! I would say let's have a dedicated PR for bumping if that is possible, so that if anyone else in the community has any concern that could be discussed specifically there.

python/pyproject.toml
```diff
@@ -87,6 +88,9 @@ embeddings = [
     "botocore>=1.31.57",
     "ollama>=0.3.0",
     "ibm-watsonx-ai>=1.1.2",
+    "transformers>=4.41.0",
+    "sentencepiece",
+
```
jackye1995: nit: unnecessary newline

python/python/lancedb/embeddings/siglip.py
```diff
@@ -0,0 +1,198 @@
+"""
```
jackye1995: missing the Apache license header

```
# SPDX-License-Identifier: Apache-2.0
# SPDX-FileCopyrightText: Copyright The LanceDB Authors
```

python/pyproject.toml
Heisenberg208: Sure!Reverted the Python version change as suggested. Found out that a PR is raised concerning the above #2477

python/python/lancedb/embeddings/siglip.py
Heisenberg208: Thanks! I've removed the extra newline and added the Apache license header to siglip.py

python/python/lancedb/embeddings/siglip.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+import io
+import os
+from typing import List, Union
+from urllib import parse as urlparse
+import numpy as np
+import PIL
+import pyarrow as pa
+import torch
```
wjones127: I think we want to lazily import `PIL` and `torch`. Otherwise, they will become required dependencies for anyone who imports `lancedb.embeddings`. See how we do this for open clip here: https://github.com/lancedb/lancedb/blob/f076bb41f4c94273260044a97887b8b66596c06e/python/python/lancedb/embeddings/open_clip.py#L148

Heisenberg208: Following Updates on changes:

- Lazily imported PIL and torch using attempt_import_or_raise, mirroring the approach in open_clip.py.

- Wrapped image embedding inference using ThreadPoolExecutor to improve CPU performance and mirror the    parallelism used in other embeddings (like OpenCLIP).

- Removed the top-level torch and PIL imports.

- Replaced direct calls with _encode_and_normalize_image() as in OpenCLIP for consistency.

Heisenberg208:
Hi @wjones127 , I've addressed the requested change and re-requested review.

All CI checks have passed except for one known doctest failure (which is also present in other merged PRs, due to the `mlx` dependency).

Let me know if anything else is needed. Thanks!

python/pyproject.toml
```diff
@@ -61,15 +61,17 @@ tests = [
     "pylance>=0.25",
     "requests",
     "datafusion",
+    "scikit-learn"
```
wjones127: Do we really need this? It seems like we use just one function from it. I'd prefer we didn't grab that dependency unless we really need it.

python/pyproject.toml
```diff
@@ -87,6 +89,8 @@ embeddings = [
     "botocore>=1.31.57",
     "ollama>=0.3.0",
     "ibm-watsonx-ai>=1.1.2",
+    "transformers>=4.41.0",
+    "sentencepiece",
```
wjones127: Do we use this at all?

python/python/tests/test_embeddings_slow.py
```diff
@@ -639,3 +643,177 @@ class MediaItems(LanceModel):
...
+        "http://farm9.staticflickr.com/8216/8434969557_d37882c42d_z.jpg",
+        "http://farm6.staticflickr.com/5142/5835678453_4f3a4edb45_z.jpg",
+    ]
+    headers = {
+        "User-Agent": (
+            "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
+            "AppleWebKit/537.36 (KHTML, like Gecko) "
+            "Chrome/114.0.0.0 Safari/537.36"
+        )
+    }
```
wjones127: I don't love the idea of impersonating a web browser to scrape images of flickr in a unit test.

We have huggingface available, so maybe a better thing to do is get a sample of the oxford pets dataset here: https://huggingface.co/datasets/enterprise-explorers/oxford-pets

Can load just a few rows using splits here: https://huggingface.co/docs/datasets/loading#slice-splits

wjones127 (changes requested):

python/pyproject.toml
Heisenberg208: Hi @WJones, 
- I added both `transformers` and `sentencepiece` because `transformers.AutoProcessor.from_pretrained()` throws an import error without `sentencepiece` when loading the `google/siglip-base-patch16-224` model. 
- Although they’re not imported at the top level and `transformers` are only lazy imported , they are required to successfully run the SigLIP embedding tests. 
- Happy to move them to optional dependencies if you’d prefer to keep the base install lighter.

python/pyproject.toml
Heisenberg208: Sure! We use it in the test to compare benchmarks between `open-clip`, `siglip`, and `imagebind`. I’ll remove the comparison test since it’s not essential.

python/python/tests/test_embeddings_slow.py
Heisenberg208: Got it !I followed the same pattern from OpenCLIP’s test where they loaded a few images from Flickr.

I’ll switch to using the Oxford Pets dataset from Hugging Face.

Thanks ! I’ll update the test accordingly.

wjones127: I guess if that's a pre-existing thing, then maybe that's fine.

wjones127: It would be nice if we could create a pytext fixture at the module level, so we don't have to re-download those for every test. https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

Heisenberg208: While OpenCLIP doesn’t set any impersonate when loading images from Flickr, it seems that Flickr has started blocking requests , resulting in HTTP 403 errors. To avoid this, I added a basic User-Agent header to mimic a web browser.

tristanz:
I believe I have fixed the doctest issue in #2536, although I'm not sure if that's the preferred approach.

python/pyproject.toml
Heisenberg208: I've moved `transformers` to the optional dependencies since other files also use it without declaring it explicitly in `pyproject.toml`, and it's only lazy imported.

However, I’ve kept `sentencepiece` as a required dependency because it’s necessary to successfully run the SigLIP embedding tests — `transformers.AutoProcessor.from_pretrained()` throws an import error without it when loading google/siglip-base-patch16-224.

python/python/tests/test_embeddings_slow.py
Heisenberg208: I’ve made the changes to use pytest fixtures to share logic between both the `open-clip` and `siglip` embedding tests.

- The fixtures help avoid re-downloading and re-processing the same test images and query image across tests, which improves test performance.

- They're scoped at the module level so the same setup (e.g., loading a query image or sample test set) is reused wherever needed without repeating code.

Heisenberg208:
Hi @wjones127 , just a quick follow-up — the PR was approved a few days ago and all checks have passed, but it hasn’t been merged yet. Please let me know if there’s anything else needed from my side. Thanks again!

PR #2502, opened 2025-07-13
feat(java): add create/drop table for java sdk

Hisoka-X:
This PR adds the implementation of three methods, createTable/createEmptyTable/dropTable, for the Java SDK. 

The original logic is used, the Java side responsible for provided interface, and Rust is responsible for calling the underlying method. The table data transmission between Java and Rust is implemented through Arrow.

### DIFF
--- a/java/core/src/main/java/com/lancedb/lancedb/Table.java
+++ b/java/core/src/main/java/com/lancedb/lancedb/Table.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


Hisoka-X:
cc @jackye1995 and @LuQQiu

java/core/src/main/java/com/lancedb/lancedb/Table.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lancedb.lancedb;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+public class Table implements Closeable {
```
Hisoka-X, line 19: We can implement the corresponding interface in the Table object later.

jackye1995:
Thank you very much for the contribution!

I discussed with a few folks about this, I think what we are trying to do now is to merge what is happening here vs what is going on in https://github.com/lancedb/lance-namespace. The sequence of operation would be something like: add LanceDB OpenAPIs to LanceNamespace and merge them as a single spec. import LanceNamespace codegen clients to LanceDB. Use the clients to build up the LanceDB SDK.

I will post more details later, but I think we will likely hold on this PR until this refactoring work has been completed.

Hisoka-X:
> I think we will likely hold on this PR until this refactoring work has been completed.

OK, thanks for your reply

wjones127:
Closing since this is out-dated.

PR #2503, opened 2025-07-13
chore: add PULL_REQUEST_TEMPLATE.md

Hisoka-X:
### purpose of this pull request
this PR add new PULL_REQUEST_TEMPLATE.md. Help PR creators clarify key issues when create PR.

### does this PR introduce _any_ user-facing change?
no

### how was this patch tested?
tested by this pr.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,8 @@
-### Purpose of this pull request
+### purpose of this pull request
 
 
-### Does this PR introduce _any_ user-facing change?
+### does this PR introduce _any_ user-facing change?
 
 
-### How was this patch tested?
+### how was this patch tested?
 


Xuanwo:
Thank you for this PR, but I don't think we're at a stage where we need a PULL_REQUEST_TEMPLATE, as it would just add more paperwork for our contributors. Feel free to raise a disucssion for this if you do think it's valuable.

Hisoka-X:
> Thank you for this PR, but I don't think we're at a stage where we need a PULL_REQUEST_TEMPLATE, as it would just add more paperwork for our contributors. Feel free to raise a disucssion for this if you do think it's valuable.

OK, maybe add more paperwork is not a bad thing, as it can help the PR reviewer and anyone else who wants to review the PR to quickly sort out the key issues. But not having it will not make a big difference.

PR #2504, opened 2025-07-13
chore: fix cleanupOlderThan docs

bmschmidt:
Thanks for all your work.

The docstring for `OptimizeOptions ` seems to reference a non-existent method on `Table`. I believe this is the correct example for `cleanupOlderThan`.

This also appears in the generated docs, but I assume they live downstream from this code?


Xuanwo (approved):
Thank you @bmschmidt for working on this!

PR #2506, opened 2025-07-15
feat: add utoipa toSchema for request/response structs 

LuQQiu:
Utoipa are used for auto generating open api.
All the structs in request and repsonses are needed to add "ToSchema" annotation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,9 +176,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "55.1.0"
+version = "55.2.0"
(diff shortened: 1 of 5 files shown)


jackye1995:
I am a bit -1 doing this at this moment. The main goal we want to have is a dependency tree of lance (storage layer format) -> lance namespace (catalog layer definition, generated code) -> lanceDB (user friendly abstraction + embedded compute engine)

If we do this, it creates a reverse dependency from lanceDB -> lance namespace, which will be problematic. But let me think a bit more about this and reply back.

LuQQiu:
Discussed with @jackye1995 for easy to copy struct, we copy in phalanx, 
For compilicated one, e.g. query and schema structs in lance, we do in place modification for now
will brainstorm on long term solution

PR #2509, opened 2025-07-15
feat: add support for returning all scores with rerankers

AyushExel:
Previously `return_score="all"` was supported only for the default reranker (RRF) and not the model based rerankers.
This adds support for keeping all scores in the base reranker so that all model based rerankers can use it. Its a slower path than keeping just the relevance score but can be useful in debugging

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1374,6 +1374,8 @@ class LanceVectorQueryBuilder(LanceQueryBuilder):
         if query_string is not None and not isinstance(query_string, str):
             raise ValueError("Reranking currently only supports string queries")
         self._str_query = query_string if query_string is not None else self._str_query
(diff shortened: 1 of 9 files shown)


python/python/lancedb/query.py
```diff
@@ -1374,6 +1374,8 @@ def rerank(
         if query_string is not None and not isinstance(query_string, str):
             raise ValueError("Reranking currently only supports string queries")
         self._str_query = query_string if query_string is not None else self._str_query
+        if reranker.score == "all":
+            self.with_row_id(True)
```
AyushExel, line 1378: I thought of adding a log/warn that this is being turned on automatically but then it would end up making it too verbose if queries run in a loop

BubbleCal (approved):

PR #2513, opened 2025-07-17
fix: enable streaming in remote add and merge_insert

### LINKED ISSUE #2339: Bug: HTTP2 can mask errors in a stream of data
## Error

Users can get an error like:

```
Http error: (request_id=XXX) error sending request for url (https://xxx.api.lancedb.com/v1/table/xxx/insert/)
lancedb.remote.errors.HttpError: stream error sent by user: unexpected internal error encountered
```

When forcing HTTP1, we find this error was masking an error producing while processing the input data stream on the client.

## Background

For both `add()` and `merge_insert()` in remote APIs, we stream the data in the body. We pass that stream from a `RecordBatchReader` into a streaming body here:

https://github.com/lancedb/lancedb/blob/24e42ccd4d9fe9644038d3dd77bd3206e9ccbdc2/rust/lancedb/src/remote/table.rs#L107-L128

There is a known issue that Hyper will not surface body errors for HTTP2, but will for HTTP1: https://github.com/hyperium/hyper/issues/2547

## Solution

When creating the stream, we send the error through a separate channel, and pass some generic error out of the stream. Then when handling errors from the request, we can first check if there was some error in the channel that we should report instead.

wjones127:
* Allows true streaming of data through append and merge_insert. (Spills to disk for retries on payloads > 20MB.)
* Will propagate errors that happen in the stream, even when using HTTP2.

Fixes #2339

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -196,12 +196,18 @@ impl<S: HttpSend + 'static> Tags for RemoteTags<'_, S> {
 }
 
 struct StreamingBodyGenerator {
-    read_task: tokio::task::JoinHandle<SpillSender>,
-    path: std::path::PathBuf,
+    handle: Arc<SpillHandle>,
     receiver: SpillReceiver,
 }
 
-impl Drop for StreamingBodyGenerator {
+/// A handle that keeps the reading task and the path alive.
+/// When it is dropped, it will cancel the reading task and clean up the temporary file.
+struct SpillHandle {
+    read_task: tokio::task::JoinHandle<SpillSender>,
+    path: std::path::PathBuf,
+}
+
+impl Drop for SpillHandle {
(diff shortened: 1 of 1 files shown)


wjones127:
## Tested with script
```python
import lancedb
import os
import numpy as np
import pyarrow as pa

# Create a generator that makes 5 10MB batches
# This is under the 100MB limit currently set on the server
# But will trigger the 20MB spilling threshold in the client

def generator():
    for _ in range(5):
        yield pa.table({'x': np.random.rand(1_000_000)})

db = lancedb.connect("db://test", api_key=os.environ.get("LANCEDB_API_KEY"))

table = db.create_table("test_table", schema=pa.schema([("x", pa.float64())]))

table.add(generator())
```

## Memory use

### Before
Screenshot 2025-07-23 at 1 54 02 PM


### After
Screenshot 2025-07-23 at 1 54 24 PM

PR #2519, opened 2025-07-21
feat(embeddings): add Azure AI Foundry embedding support to lancedb

### LINKED ISSUE #2518: Feature: Support Azure AI Foundry embeddings
### SDK

Python

### Description

Azure AI foundry offers a handful of embedding models that are  made available through their serverless API service.
https://learn.microsoft.com/en-us/python/api/overview/azure/ai-inference-readme?view=azure-python-preview#text-embeddings-example

I am unaware of the ideas thoughts behind adding more embedding provider support. But I know for my case it would be helpful and not that difficult to add.

I currently have a local derivation that works but could generalize it if interested.

1jamesthompson1:
Adds in a new embedding model that supports Azure AI foundry. Fixes #2518

*No multimodal support has been added due to my lack of experience with the gotchas and trip ups that come with dealing with images. If it is desired I can model something off voyageai or one of the other multimodal implementations*


1jamesthompson1:
Hey any chance this could be looked at @Xuanwo?

PR #2524, opened 2025-07-21
feat(java): integrate lance-namespace to lancedb Java SDK

jackye1995:


java/.mvn/wrapper/maven-wrapper.properties
jackye1995, line 1: current project seems missing maven wrapper, adding it along the way.

java/README.md
jackye1995, line 1: Add it in README for now, until the documentation website is stablized.

java/core/src/main/java/com/lancedb/lancedb/LanceDBRestNamespaces.java
```diff
@@ -0,0 +1,145 @@
...
+package com.lancedb.lancedb;
+
+import com.lancedb.lance.namespace.LanceRestNamespace;
+import com.lancedb.lance.namespace.client.apache.ApiClient;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class LanceDBRestNamespaces {
```
jackye1995: originally called `LanceRestNamespaceBuilder`. Given this is specific to LanceDB, and since this is a util class, use the convention `XXXXs`, which gives `LanceDBRestNamespaces`

LuQQiu: @jackye1995 This java package seems much heavier for enduser just to leverage this doc and the namespace helper. 
Can we not require user to install the whole lancedb java sdk??
(don't want they to have all the dependencies of lancedb-jni)

jackye1995: moved these to a separated module

PR #2530, opened 2025-07-23
feat: allow Python and Typescript users to create `Session`s

wjones127:
## Summary
- Exposes `Session` in Python and Typescript so users can set the `index_cache_size_bytes` and `metadata_cache_size_bytes`
     * The `Session` is attached to the `Connection`, and thus shared across all tables in that connection.
- Adds deprecation warnings for table-level cache configuration


nodejs/lancedb/index.ts
```diff
@@ -152,18 +156,33 @@ export async function connect(
  */
 export async function connect(
   options: Partial<ConnectionOptions> & { uri: string },
+  session?: Session,
 ): Promise<Connection>;
 export async function connect(
   uriOrOptions: string | (Partial<ConnectionOptions> & { uri: string }),
-  options: Partial<ConnectionOptions> = {},
+  optionsOrSession?: Partial<ConnectionOptions> | Session,
+  session?: Session,
```
wjones127: Could we actually move `session` into `ConnectionOptions`? Then a lot of these changes aren't necessary.

python/src/session.rs
```diff
@@ -0,0 +1,107 @@
...
+#[pymethods]
+impl Session {
+    /// Create a new session with custom cache sizes.
+    ///
+    /// Parameters
+    /// ----------
+    /// index_cache_size_bytes : int, optional
+    ///     The size of the index cache in bytes.
+    ///     The index cache stores decompressed vector indices in memory for faster access.
+    ///     A larger cache can improve performance for datasets with many indices, but uses more RAM.
```
wjones127:
```suggestion
    ///     The size of the index cache in bytes.
    ///     Index data is stored in memory in this cache to speed up queries.
```

python/src/session.rs
```diff
@@ -0,0 +1,107 @@
...
+    ///     Default: 6GB (6 * 1024 * 1024 * 1024 bytes)
+    /// metadata_cache_size_bytes : int, optional
+    ///     The size of the metadata cache in bytes.
+    ///     The metadata cache stores file metadata and schema information in memory.
+    ///     A larger cache can improve performance for datasets with many files or frequent schema access.
```
wjones127:
```suggestion
    ///     The metadata cache stores file metadata and schema information in memory.
    ///     This cache improves scan and write performance.
```

python/python/lancedb/db.py
```diff
@@ -37,6 +37,9 @@
     from ._lancedb import Connection as LanceDbConnection
     from .common import DATA, URI
     from .embeddings import EmbeddingFunctionConfig
+    from ._lancedb import Session as SESSION_TYPE
+else:
+    SESSION_TYPE = "Session"
```
wjones127: Can't this just be?
```suggestion
    from ._lancedb import Session
```

python/python/lancedb/db.py
```diff
@@ -354,6 +361,7 @@ def __init__(
         *,
         read_consistency_interval: Optional[timedelta] = None,
         storage_options: Optional[Dict[str, str]] = None,
+        session: Optional[SESSION_TYPE] = None,
```
wjones127:
```suggestion
        session: Optional[Session] = None,
```

docs/src/js/classes/Session.md
```diff
@@ -0,0 +1,84 @@
...
+***
+
+[@lancedb/lancedb](../globals.md) / Session
+
+# Class: Session
+
+A session for managing caches and object stores across LanceDB operations.
+
+Sessions allow you to configure cache sizes for index and metadata caches,
+which can significantly impact performance for large datasets.
```
wkalt, line 12: I would maybe mention here how they allow you to share cache across different lancedb connections. That's the main way they improve performance right?

Had a similar thought on the deprecation messages (repeated in a few places). We end with "are more efficient" but could say something like "they enable cache reuse across connections".

PR #2544, opened 2025-07-24
fix(python): expose register function in embeddings module

### LINKED ISSUE #2541: bug(python): 'register' couldn't be found in `lancedb.registry`
### LanceDB version

0.24.1

### What happened?

I tried to implement a custom embedding function according to the doc: https://www.lancedb.com/documentation/embeddings/custom_embedding_function.html

But it failed while importing the 'register' method:

```
from lancedb.registry import register  # -> Unresolved reference 'register' 
```

I found that the 'register' was removed in this commit: https://github.com/alexkohler/lancedb/commit/507f6087c2645126d75ac77809df26474314dd2b

Should we add it back? Has anyone successfully walked through this doc?

I would also like to help fix this issue if possible.

tristanz:
## Summary
Fixes #2541

**Problem**: The `register` function was not accessible via `from lancedb.embeddings import register` as documented, causing ImportError for users trying to create custom embedding functions.

**Solution**: Added `register` to the exports in `python/lancedb/embeddings/__init__.py` to match the documented API and follow the same pattern as other registry functions (`get_registry`, `EmbeddingFunctionRegistry`).

**Root Cause**: The function existed in `lancedb.embeddings.registry` but wasn't exposed through the main embeddings module interface.

## Changes
- Add `register` to imports in `/python/python/lancedb/embeddings/__init__.py`

## Test Plan
- [x] Verified `from lancedb.embeddings import register` works as documented
- [x] Confirmed existing embedding tests pass
- [x] Checked that the fix follows existing patterns (same as `get_registry`)
- [x] Validated linting and formatting passes

## References
Fixes #2541

### DIFF
--- a/python/python/lancedb/embeddings/__init__.py
+++ b/python/python/lancedb/embeddings/__init__.py
@@ -11,7 +11,7 @@ from .instructor import InstructorEmbeddingFunction
 from .ollama import OllamaEmbeddings
 from .open_clip import OpenClipEmbeddings
 from .openai import OpenAIEmbeddings
-from .registry import EmbeddingFunctionRegistry, get_registry
+from .registry import EmbeddingFunctionRegistry, get_registry, register
 from .sentence_transformers import SentenceTransformerEmbeddings
 from .gte import GteEmbeddings
 from .transformers import TransformersEmbeddingFunction, ColbertEmbeddings


wjones127 (approved):
Thank you!

PR #2545, opened 2025-07-24
Fix: FTS offset handling for Tantivy implementation

### LINKED ISSUE #2459: offset not working properly with FTS
### LanceDB version

0.19

### What happened?

I'm working on a Tauri application, and I'm really invested in lancedb at this point so I hope to God that this isn't a 'well the rust sdk can't do that' sort of thing, but I can't figure out why this full text search is returning an empty result when there are possibly a few dozen matches.

Here is where I set the index:

```rs
pub async fn set_mdx_note_body_index(db: &Connection) -> FlusterResult<()> {
    let tbl = db
        .open_table(DatabaseTables::MdxNote.to_string())
        .execute()
        .await
        .map_err(|e| {
            println!("Error: {:?}", e);
            FlusterError::FailToCreateIndex
        })?;
    tbl.create_index(
        &["raw_body"],
        lancedb::index::Index::FTS(FtsIndexBuilder {
            with_position: false,
            tokenizer_configs: TokenizerConfig::new(
                "simple".to_string(),
                tantivy::tokenizer::Language::English,
            ),
        }),
    )
    .execute()
    .await
    .map_err(|e| {
        println!("Error: {:?}", e);
        FlusterError::FailToCreateIndex
    })?;
    Ok(())
}
```

I've tried several approaches here with the `FtsIndexBuilder`, and so far none have been successful.

This is where I'm querying the data:

```rs
    pub async fn full_text_search(
        db: &FlusterDb<'_>,
        query: &String,
        pagination: &PaginationProps,
    ) -> FlusterResult<Vec<MdxNoteModel>> {
        let tbl = get_table(db, DatabaseTables::MdxNote).await?;

        let items_batch = tbl
            .query()
            .full_text_search(FullTextSearchQuery::new(query.to_string()))
            .limit(pagination.per_page as usize)
            .offset((pagination.per_page * (pagination.page_number - 1)) as usize)
            .execute()
            .await
            .map_err(|e| {
                println!("Error in MdxNoteEntity.full_text_search: {:?}", e);
                FlusterError::FailToConnect
            })?
            .try_collect::<Vec<_>>()
            .await
            .map_err(|e| {
                println!("Error in MdxNoteEntity.full_text_search: {:?}", e);
                FlusterError::FailToFind
            })?;

        if items_batch.is_empty() {
            return Ok(Vec::new());
        }

        let mut items: Vec<MdxNoteModel> = Vec::new();

        for batch in items_batch.iter() {
            let data: Vec<MdxNoteModel> =
                from_record_batch(batch).map_err(|_| FlusterError::FailToSerialize)?;
            items.extend(data);
        }
        Ok(items)
    }
```

Any help here is hugely appreciated it. This app is a really important piece of something that could turn out to be really impactful, and for that reason it's pretty time sensitive.

Thanks in advance.

The app is becoming pretty complex. I'm not sure if there's a single set of steps that I can think of that would reproduce this, but I've included the core pieces of relevant code.

tristanz:
## Summary
Fixes #2459

**Problem**: FTS queries with offset were applying limit before offset, causing `offset(2).limit(2)` to return empty results instead of the expected items.

**Root Cause**: The original implementation was using a fetch-and-slice approach, but tantivy-py natively supports offset parameters.

**Solution**: Use tantivy-py's native `offset` parameter in `searcher.search()` for cleaner and more efficient offset handling.

## Changes Made
- Added `offset` parameter to `search_index()` function in `fts.py` (backward compatible with default value)
- Updated `searcher.search()` calls to use native offset support  
- Fixed edge case handling when offset is beyond available results
- Added comprehensive test covering both Tantivy and non-Tantivy implementations

## Test Status
- ✅ **Tantivy FTS** (`use_tantivy=True`): Fixed with native offset support  
- ❌ **Non-Tantivy FTS** (`use_tantivy=False`): Still fails, requires fix in `lance` crate

## Test Plan
- [x] Reproduction test passes for Tantivy implementation
- [x] Test documents the bug for non-Tantivy (will pass when `lance` dependency is updated)
- [x] Edge cases handled (offset beyond results, boundary conditions)
- [x] Existing tests still pass (including direct `search_index()` calls)
- [x] Linting and formatting pass

## Immediate User Benefit
Users can get working offset behavior right now by using:
```python
table.create_fts_index("field", use_tantivy=True)
```

## Future Work
Once the `lance` crate is updated to fix the underlying offset bug:
1. Update `lance` dependency in `Cargo.toml`
2. Non-Tantivy test will automatically pass
3. Both FTS implementations will have working offset support

## Related Issues
- Original issue: #2459 
- Shows that both Python and Rust users are affected by this offset bug
- This PR provides immediate relief for Python users while the underlying lance issue gets resolved


tristanz:
This seems to document a bug in the upstream lance with limit + offset, but I have been able to look at it closely.

tristanz:
Related upstream: https://github.com/lancedb/lance/pull/4318

PR #2547, opened 2025-07-24
enhance: infer vector columns when name contains 'vector' or 'embedding'

### LINKED ISSUE #2546: Feature: infer vector column if "vector" or "embedding" in name
### SDK

Python

### Description

Vector columns must be fixed size list columns. Right now, we will detect a vector column and cast it to a fixed size list from a variable size list, but only if the name is exactly "vector". We could make this easier by changing the condition to `name.contains("vector") or name.contains("embedding")`.

Here's where the current Python implementation is:

https://github.com/lancedb/lancedb/blob/94fb9f364a77d327a4d0515ba4c6e2db3d3b1877/python/python/lancedb/table.py#L2928-L2932

Here's where the current Nodejs implementation is:

https://github.com/lancedb/lancedb/blob/94fb9f364a77d327a4d0515ba4c6e2db3d3b1877/nodejs/lancedb/arrow.ts#L593-L598

wjones127:
## Summary

- Enhanced vector column detection to use substring matching instead of exact matching
- Now detects columns with names containing "vector" or "embedding" (case-insensitive)
- Added integer vector support to Node.js implementation (matching Python)
- Comprehensive test coverage for both float and integer vector types

### Python (`python/python/lancedb/table.py`)
- Updated `_infer_target_schema()` to use substring matching with helper function `_is_vector_column()`
- Preserved original field names instead of forcing "vector"
- Consolidated duplicate logic for better maintainability

### Node.js (`nodejs/lancedb/arrow.ts`)
- Enhanced type inference with `nameSuggestsVectorColumn()` helper function
- Added `isAllIntegers()` function with performance optimization (checks first 10 elements)
- Implemented integer vector support using `Uint8` type (matching Python)
- Improved type safety by removing `any` usage

### Tests
- **Python**: Added `test_infer_target_schema_with_vector_embedding_names()` in `test_util.py`
- **Node.js**: Added comprehensive test case in `arrow.test.ts`
- Both test suites cover various naming patterns and integer/float vector types

## Examples of newly supported column names:
- `user_vector`, `text_embedding`, `doc_embeddings`
- `my_vector_field`, `embedding_model`
- `VECTOR_COL`, `Vector_Mixed` (case-insensitive)
- Both float and integer arrays are properly converted to fixed-size lists

## Test plan
- [x] All existing tests pass (backward compatibility maintained)
- [x] New tests pass for both Python and Node.js implementations
- [x] Integer vector detection works correctly in Node.js
- [x] Code passes linting and formatting checks
- [x] Performance optimized for large vector arrays

Fixes #2546

### DIFF
--- a/nodejs/__test__/arrow.test.ts
+++ b/nodejs/__test__/arrow.test.ts
@@ -1,7 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
-import { Bool, Field, Int32, List, Schema, Struct, Utf8 } from "apache-arrow";
+import {
+  Bool,
+  Field,
+  Int32,
+  List,
+  Schema,
+  Struct,
+  Uint8,
+  Utf8,
+} from "apache-arrow";
 
 import * as arrow15 from "apache-arrow-15";
 import * as arrow16 from "apache-arrow-16";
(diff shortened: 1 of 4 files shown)


nodejs/__test__/arrow.test.ts
```diff
@@ -255,6 +264,89 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
...
+          },
+        ]);
+
+        // Check that columns with 'vector' or 'embedding' in name are converted to FixedSizeList
+        expect(
+          DataType.isFixedSizeList(table.getChild("user_vector")?.type),
+        ).toBe(true);
+        expect(
+          DataType.isFixedSizeList(table.getChild("text_embedding")?.type),
+        ).toBe(true);
```
wjones127: Instead of having to write an assertion for each case, can we do a for-loop over the columns and assert the same thing?

We can separate the integer and normal list case into a different input.

nodejs/lancedb/arrow.ts
```diff
@@ -591,10 +617,19 @@ function inferType(
       return undefined;
     }
     // Try to automatically detect embedding columns.
-    if (valueType instanceof Float && path[path.length - 1] === "vector") {
-      // We default to Float32 for vectors.
-      const child = new Field("item", new Float32(), true);
-      return new FixedSizeList(value.length, child);
+    if (nameSuggestsVectorColumn(path[path.length - 1])) {
+      // Check if values are integers for type determination
+      const isIntegerVector = isAllIntegers(value);
```
wjones127: Because Javascript lacks real integers, I think this could get false positives. Especially if the user provides all zeros for one of the sample values. Instead, let's only assume an integer vector if:

1. `nameSuggestsVectorColumn()`
2. `value` is a `Uint8Array`

wkalt:
@wjones127 what's the reason for this change? IMO the current special casing of vector is a bit of a wart. Won't this make the behavior even more magical/difficult to explain?

Generally I want to avoid column names that would trigger this behavior, but this patch seems like it would make that more difficult.

wjones127:
> what's the reason for this change? IMO the current special casing of vector is a bit of a wart. Won't this make the behavior even more magical/difficult to explain?

Many users input data in a loosely typed fashion:

```python
db = lancedb.connect(...)
db.create_table("my_table", [
   { 'id': 1, 'vector': [1.0, 1.0, 1.0] },
   { 'id': 2, 'vector': [2.0, 2.0, 2.0] },
])
```

When we process this, we need to decide the type of the `vector` column: is it a variable size list? a fixed size list?

If we were a more generic database, a conservative choice would probably to always choose a variable size list. But we have a focus on vector search, so 99% of our user base intends for a list of floats to be fixed size list. Thus we make an educated guess based on the column name and data type.

It should be said: users always have the option of specifying their schema up front. **This inference is entirely optional.** It's just in these cases when the user provide data whose types are ambiguous.

> Generally I want to avoid column names that would trigger this behavior, but this patch seems like it would make that more difficult.

It's very easy to get around this behavior, and I suspect you already do. All you have to do is set the schema of your table up front.

I think you might be confusing this with a similar, but separate behavior: When you create a vector column called "vector" in LanceDB Cloud, we will automatically create an index for you and choose the indexing parameters (such as distance) on your behalf. For that behavior, there isn't an easy way to opt out, except calling your embedding column by a different name. This PR does **not** expand or change that behavior in any way.

wkalt:
> I think you might be confusing this with a similar, but separate behavior: When you create a vector column called "vector" in LanceDB Cloud, we will automatically create an index for you and choose the indexing parameters (such as distance) on your behalf. For that behavior, there isn't an easy way to opt out, except calling your embedding column by a different name. This PR does **not** expand or change that behavior in any way.

Yeah, I was. Thanks for the explanation and good to know this doesn't affect that behavior.

wkalt (approved):

Issue #1968, opened 2024-12-20
bug(python): create_table fails if no data is provided


qbit-:
### LanceDB version

v0.16.0

### What happened?

When using `create_table` without providing data (only providing schema), the function fails with 
```
  File "/usr/local/lib/python3.9/site-packages/lancedb/db.py", line 422, in create_table
    tbl = LanceTable.create(
  File "/usr/local/lib/python3.9/site-packages/lancedb/table.py", line 1888, in create
    new_table.add(data)
  File "/usr/local/lib/python3.9/site-packages/lancedb/table.py", line 1580, in add
    data, _ = _sanitize_data(
  File "/usr/local/lib/python3.9/site-packages/lancedb/table.py", line 172, in _sanitize_data
    data = _sanitize_schema(data, schema, on_bad_vectors, fill_value)
  File "/usr/local/lib/python3.9/site-packages/lancedb/table.py", line 2173, in _sanitize_schema
    data = _sanitize_vector_column(
  File "/usr/local/lib/python3.9/site-packages/lancedb/table.py", line 2228, in _sanitize_vector_column
    has_jagged_ndims = len(vec_arr.values) % len(data) != 0
ZeroDivisionError: integer division or modulo by zero
```

### Are there known steps to reproduce?

def test_add_table_with_empty_embeddings(tmp_path):
    db = lancedb.connect(tmp_path)

    class MySchema(LanceModel):
        text: str
        embedding: Vector(16)

    table = db.create_table("test", schema=MySchema)
    table.add(
        [{"text": "bar", "embedding": [0.1] * 16}],
        on_bad_vectors="drop",
    )

tristanz:
I cannot repro this on main.

Xuanwo:
Hi, this issue is invalid because the ZeroDivisionError when creating a table with schema and no data has been fixed by PR #2548 (merged 2025-07-25). The regression test `test_add_with_empty_fixed_size_list_drops_bad_rows` covers this scenario. I'm closing this now. Feel free to create a new issue if the problem still exists.

Issue #303, opened 2023-07-14
Searching on empty table produces scary error message


changhiskhan:
```
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', /Users/changshe/code/lance/rust/src/index/vector/flat.rs:75:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Neither friendly nor helpful

changhiskhan:
should this return empty or raise an error?

AlexanderTheLit:
It'd be nice if it returned empty.

tristanz:
I cannot repro this on main.

PR #2548, opened 2025-07-25
fix: handle empty list with schema in table creation

tristanz:
## Summary
Fixes IndexError when creating tables with empty list data and a provided schema. Previously, `_into_pyarrow_reader()` would attempt to access `data[0]` on empty lists, causing an IndexError. Now properly handles empty lists by using the provided schema.

Also adds regression tests for GitHub issues #1968 and #303 to prevent future regressions with empty table scenarios.

## Changes
- Fix IndexError in `_into_pyarrow_reader()` for empty list + schema case
- Add Optional[pa.Schema] parameter to handle empty data gracefully  
- Add `test_create_table_empty_list_with_schema` for the IndexError fix
- Add `test_create_empty_then_add_data` for issue #1968
- Add `test_search_empty_table` for issue #303

## Test plan
- [x] All new regression tests pass
- [x] Existing tests continue to pass
- [x] Code formatted with `make format`


python/python/lancedb/table.py
```diff
@@ -123,6 +125,11 @@ def _into_pyarrow_reader(data) -> pa.RecordBatchReader:
         raise ValueError("Cannot add a single dictionary to a table. Use a list.")
 
     if isinstance(data, list):
+        # Handle empty list case
+        if not data:
```
eddyxu, line 129: Should we check `schema is not None` here?

tristanz: Yes!

eddyxu (approved):
Thanks @tristanz !

PR #2550, opened 2025-07-25
feat: add overall timeout parameter to remote client

wjones127:
## Summary
- Adds an overall `timeout` parameter to `TimeoutConfig` that limits the total time for the entire request
- Can be set via config or `LANCE_CLIENT_TIMEOUT` environment variable
- Exposed in Python and Node.js bindings
- Includes comprehensive tests

## Test plan
- [x] Unit tests for Rust TimeoutConfig
- [x] Integration tests for Python bindings  
- [x] Integration tests for Node.js bindings
- [x] All existing tests pass

### DIFF
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -41,6 +41,16 @@ impl Default for ClientConfig {
 /// How to handle timeouts for HTTP requests.
 #[derive(Clone, Default, Debug)]
 pub struct TimeoutConfig {
+    /// The overall timeout for the entire request.
+    ///
+    /// This includes connection, send, and read time. If the entire request
+    /// doesn't complete within this time, it will fail.
+    ///
+    /// You can also set the `LANCE_CLIENT_TIMEOUT` environment variable
+    /// to set this value. Use an integer value in seconds.
+    ///
+    /// By default, no overall timeout is set.
+    pub timeout: Option<Duration>,
     /// The timeout for creating a connection to the server.
     ///
     /// You can also set the `LANCE_CLIENT_CONNECT_TIMEOUT` environment variable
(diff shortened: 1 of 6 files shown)


rust/lancedb/src/remote/client.rs
```diff
@@ -581,3 +597,51 @@ pub mod test_utils {
...
+        };
+
+        assert_eq!(config.timeout, Some(Duration::from_secs(60)));
+        assert_eq!(config.connect_timeout, Some(Duration::from_secs(10)));
+        assert_eq!(config.read_timeout, Some(Duration::from_secs(30)));
+        assert_eq!(config.pool_idle_timeout, Some(Duration::from_secs(300)));
+    }
+
+    #[test]
+    fn test_client_config_with_timeout() {
```
rpgreen, line 631: should we have a test that exercises the timeout? i.e. sets the timeout to 1s and sends a request and asserts that the reqwest error gets propogated

rpgreen (approved):

wjones127:
@rpgreen I was thinking: how do you think we should have `timeout` interact with retries? Right now the `timeout` is per attempt. So a timeout of 1s might take 5 or more seconds to run with 5 retries. Should it perhaps be made an overall timeout?

rpgreen:
> @rpgreen I was thinking: how do you think we should have `timeout` interact with retries? Right now the `timeout` is per attempt. So a timeout of 1s might take 5 or more seconds to run with 5 retries. Should it perhaps be made an overall timeout?

I think the current approach is the right way. The timeout should be per-request so it can be adjusted according to the server-side timeout. We already support configuring the number of retries, so we essentially support an overall timeout. We could consider adding another "overall timeout" parameter, but I don't think it's strictly necessary.

PR #2552, opened 2025-07-26
feat: a utility for creating "permutation views"

westonpace:
I'm working on a lancedb version of pytorch data loading (and hopefully addressing https://github.com/lancedb/lance/issues/3727).

However, rather than rely on pytorch for everything I'm moving some of the things that pytorch does into rust.  This gives us more control over data loading (e.g. using shards or a hash-based split) and it allows permutations to be persistent.  In particular I hope to be able to:

* Create a persistent permutation
* This permutation can handle splits, filtering, shuffling, and sharding
* Create a rust data loader that can read a permutation (one or more splits), or a subset of a permutation (for DDP)
* Create a python data loader that delegates to the rust data loader

Eventually create integrations for other data loading libraries, including rust & node

### DIFF
--- a/nodejs/src/lib.rs
+++ b/nodejs/src/lib.rs
@@ -12,7 +12,7 @@ mod header;
 mod index;
 mod iterator;
 pub mod merge;
-mod permutation;
+pub mod permutation;
 mod query;
 pub mod remote;
 mod rerankers;
--- a/nodejs/src/permutation.rs
+++ b/nodejs/src/permutation.rs
@@ -206,3 +206,17 @@ impl PermutationBuilder {
         Ok(Table::new(table))
     }
 }
+
+/// Create a permutation builder for the given table
+#[napi]
+pub fn permutation_builder(
+    table: &crate::table::Table,
+    dest_table_name: String,
+) -> napi::Result<PermutationBuilder> {
+    use lancedb::dataloader::permutation::PermutationBuilder as LancePermutationBuilder;
+
+    let inner_table = table.inner_ref()?.clone();
+    let inner_builder = LancePermutationBuilder::new(inner_table);
+
+    Ok(PermutationBuilder::new(inner_builder, dest_table_name))
+}


docs/random-data-loader-design.md
```diff
@@ -0,0 +1,1010 @@
+# LanceDB Random Data Loader Design Document
+
+## Overview
+
+This design document outlines the implementation of a data loader class for LanceDB in Rust that can load data from a LanceDB table in random order. The loader will be framework-agnostic and leverage a new `take` method in the `BaseTable` trait for efficient row-offset-based loading.
+
+The loader operates in two phases. First, a permutation of the base table is created. This permutation is persisted as a LanceDB table (can be persisted in memory or on disk). Next, a data loader is created from a permutation table. The data loader can be used to load data from the base table, applying the permutation.
```
wjones127: > This permutation is persisted as a LanceDB table (can be persisted in memory or on disk)

Can this be persisted in object storage? I imagine this is important if you want to do distributed training right?

wjones127 (commented):
One thing I wasn't sure of: which struct / method is the main entrypoint for this functionality? Perhaps we could use a guide in the rust docs showing a quick usage example.

docs/random-data-loader-design.md
westonpace: Oh, whoops...forgot I had this md still in here :laughing: 

I had originally started this PR as a vibe coding exercise before I derailed from the plan so I'm not sure how accurate it is anymore.  Let me remove it.

> Can this be persisted in object storage? I imagine this is important if you want to do distributed training right?

Yeah, it's just a table.

westonpace:
> One thing I wasn't sure of: which struct / method is the main entrypoint for this functionality? Perhaps we could use a guide in the rust docs showing a quick usage example.

The entry point will be `PermutationBuilder` but I want to wait to example it too much because I want to make sure the python bindings + pytorch stuff work correctly.

Then, I agree, we should have a rust data loading example.

rust/lancedb/src/dataloader/permutation.rs
```diff
@@ -0,0 +1,294 @@
...
+        split::{SplitStrategy, Splitter, SPLIT_ID_COLUMN},
+        util::{rename_column, TemporaryDirectory},
+    },
+    query::{ExecutableQuery, QueryBase},
+    Connection, Error, Result, Table,
+};
+
+/// Configuration for creating a permutation table
+#[derive(Debug, Default)]
+pub struct PermutationConfig {
```
westonpace, line 32: Don't we need project / select?

westonpace: Will do in follow-up as it may be a separate view type

westonpace:
> One thing I wasn't sure of: which struct / method is the main entrypoint for this functionality? Perhaps we could use a guide in the rust docs showing a quick usage example.


@wjones127 I've added python bindings and I guess I'm back to revisiting your question.  On the one hand, something like this is very natural:

```
db = ...
tbl = db.open_table(...)
permutation = tbl.create_permutation("my_permutation")...execute()
```

However, the current API is:

```
db = ...
tbl = db.open_table(...)
permutation = permutation_builder(tbl, "my_permutation")...execute()
```

The main reason is that I want views to be a plugin that composes on top of tables.  In other words, I don't want tables to have to know about every type of view that gets invented.

rust/lancedb/src/database.rs
```diff
@@ -213,13 +214,31 @@ impl CloneTableRequest {
     }
 }
 
+/// How long until a change is reflected from one Table instance to another
+///
+/// Tables are always internally consistent.  If a write method is called on
+/// a table instance it will be immediately visible in that same table instance.
+pub enum ReadConsistency {
```
wjones127, line 221: This is arguably a lot easier to understand than `read_consistency_interval` 😆

wjones127:
> The main reason is that I want views to be a plugin that composes on top of tables. In other words, I don't want tables to have to know about every type of view that gets invented.

Yeah that seems reasonable to me.

PR #2554, opened 2025-07-29
fix: add `name` to `IndexConfig` + fix type error in Python bindings

MarkMcCaskey:
Hello! I was upgrading our client and I noticed a few issues.

1. `name` in `IndexConfig` is not exposed in the python types (it exists on the type though https://github.com/lancedb/lancedb/blob/70d9b04ba595bc57d4a1e4ec5668a1a36608a2cd/python/src/index.rs#L183-L192)
2. `wait_timeout` on RemoteTable has a default value of `None` but the type is not optional. The linter on my side made this hard to workaround.

### DIFF
--- a/python/python/lancedb/remote/table.py
+++ b/python/python/lancedb/remote/table.py
@@ -148,7 +148,7 @@ class RemoteTable(Table):
         column: str,
         *,
         replace: bool = False,
-        wait_timeout: timedelta = None,
+        wait_timeout: Optional[timedelta] = None,
         with_position: bool = False,
         # tokenizer configs:
         base_tokenizer: str = "simple",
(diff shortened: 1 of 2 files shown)


python/python/lancedb/remote/table.py
```diff
@@ -114,7 +114,7 @@ def create_scalar_index(
         index_type: Literal["BTREE", "BITMAP", "LABEL_LIST", "scalar"] = "scalar",
         *,
         replace: bool = False,
-        wait_timeout: timedelta = None,
```
wkalt, line 117: could we do the same for create_fts_index as well?

wkalt (approved):
@MarkMcCaskey thanks a lot for this, just one minor comment.

python/python/lancedb/remote/table.py
MarkMcCaskey: Oh yeah, good catch! fixed

wkalt:
superseded by https://github.com/lancedb/lancedb/pull/2660

PR #2555, opened 2025-07-29
feat: add `status` to `IndexConfig` and Python bindings

MarkMcCaskey:
In our code we were making a direct http call and reading this value. It would be nice to expose it here but we can pretty easily work around not having this if you all don't want to change your API for this.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -2633,6 +2633,11 @@ impl BaseTable for NativeTable {
                 return None;
             };
 
+            let Some(status) = stats.get("status").and_then(|v| v.as_str()) else {
+                log::warn!("Index statistics was missing 'status' field for index {} ({})", idx.name, idx.uuid);
+                return None;
+            };
+
             let index_type: crate::index::IndexType = match index_type.parse() {
                 Ok(index_type) => index_type,
                 Err(e) => {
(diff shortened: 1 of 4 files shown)


wkalt:
@MarkMcCaskey I think we should consider whether we should do this or return the whole stats blob. We will discuss internally. If the stats is the only place to get the status from then, we there may be an argument for just doing that instead.

westonpace:
I agree with Wyatt that we might as well expose the entire statistics blob.  I think it's ok to still add status as a separate callout.  However, if we are doing this, we should at least open a ticket on `lance` requesting that `status` be made part of the list indices call (and also the index type).

rust/lancedb/src/table.rs
```diff
@@ -2633,6 +2633,11 @@ impl BaseTable for NativeTable {
                 return None;
             };
 
+            let Some(status) = stats.get("status").and_then(|v| v.as_str()) else {
+                log::warn!("Index statistics was missing 'status' field for index {} ({})", idx.name, idx.uuid);
```
westonpace, line 2637: We shouldn't warn here.  This field is only present on enterprise / lancedb cloud.  It will not be present on OSS lance.

wjones127:
I think we're moving away from returning status for now, so closing this.

PR #2556, opened 2025-07-29
fix: `.nprobes` method in python bindings, improve error messages

MarkMcCaskey:
`nprobes` with a value greater than 20 fails with the minimum error:

```
self = <lancedb.query.AsyncVectorQuery object at 0x10b749720>, minimum_nprobes = 30

    def minimum_nprobes(self, minimum_nprobes: int) -> Self:
        """Set the minimum number of probes to use.

        See `nprobes` for more details.

        These partitions will be searched on every indexed vector query and will
        increase recall at the expense of latency.
        """
>       self._inner.minimum_nprobes(minimum_nprobes)
E       ValueError: Invalid input, minimum_nprobes must be less than or equal to maximum_nprobes

python/lancedb/query.py:2744: ValueError
```

Putting the max set before the min seems reasonable but it causes this reasonable case to fail:
```
def test_nprobes_min_max_works_sync(table):
    LanceVectorQueryBuilder(table, [0, 0], "vector").minimum_nprobes(2).maximum_nprobes(4).to_list()
```

with

```
self = <lancedb.query.AsyncVectorQuery object at 0x1203f1c90>, maximum_nprobes = 4

    def maximum_nprobes(self, maximum_nprobes: int) -> Self:
        """Set the maximum number of probes to use.

        See `nprobes` for more details.

        If this value is greater than `minimum_nprobes` then the excess partitions
        will be searched only if we have not found enough results.

        This can be useful when there is a narrow filter to allow these queries to
        spend more time searching and avoid potential false negatives.

        If this value is 0 then no limit will be applied and all partitions could be
        searched if needed to satisfy the limit.
        """
>       self._inner.maximum_nprobes(maximum_nprobes)
E       ValueError: Invalid input, maximum_nprobes must be greater than or equal to minimum_nprobes

python/lancedb/query.py:2761: ValueError
```.

The case I care about is where min == max, but this solution handles it even if they're not. If both min and max exist, we set both to the minimum and then set the max. This isn't 100% the same as the minimum setter checks for 0 on the min and `.nprobes` does not do any sanity checking at all. But I figured this was the most reasonable and general solution without touching more of this code.

As part of this I noticed the error messages were a bit ambiguous so I made them symmetric and clarified them while I was here.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3674,7 +3674,7 @@ class AsyncTable:
             if query.distance_type is not None:
                 async_query = async_query.distance_type(query.distance_type)
             if query.minimum_nprobes is not None and query.maximum_nprobes is not None:
-                # Set both to the minimum to avoid min > max error.
+                # Set both to the minimum first to avoid min > max error.
                 async_query = async_query.nprobes(query.minimum_nprobes).maximum_nprobes(query.maximum_nprobes)
             elif query.minimum_nprobes is not None:
                 async_query = async_query.minimum_nprobes(query.minimum_nprobes)
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3675,7 +3675,9 @@ class AsyncTable:
                 async_query = async_query.distance_type(query.distance_type)
             if query.minimum_nprobes is not None and query.maximum_nprobes is not None:
                 # Set both to the minimum first to avoid min > max error.
-                async_query = async_query.nprobes(query.minimum_nprobes).maximum_nprobes(query.maximum_nprobes)
+                async_query = async_query.nprobes(
+                    query.minimum_nprobes
+                ).maximum_nprobes(query.maximum_nprobes)
             elif query.minimum_nprobes is not None:
                 async_query = async_query.minimum_nprobes(query.minimum_nprobes)
             elif query.maximum_nprobes is not None:
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3673,9 +3673,12 @@ class AsyncTable:
             )
             if query.distance_type is not None:
                 async_query = async_query.distance_type(query.distance_type)
-            if query.minimum_nprobes is not None:
+            if query.minimum_nprobes is not None and query.maximum_nprobes is not None:
+                # Set both to the minimum to avoid min > max error.
+                async_query = async_query.nprobes(query.minimum_nprobes).maximum_nprobes(query.maximum_nprobes)
+            elif query.minimum_nprobes is not None:
                 async_query = async_query.minimum_nprobes(query.minimum_nprobes)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: python/python/lancedb/table.py [hunk 1]
```diff
@@ -3673,9 +3673,12 @@ def _sync_query_to_async(
             )
             if query.distance_type is not None:
                 async_query = async_query.distance_type(query.distance_type)
-            if query.minimum_nprobes is not None:
+            if query.minimum_nprobes is not None and query.maximum_nprobes is not None:
```
BubbleCal, line 3676: iiuc we will still see error if `min_nprobes` is set to a value greater than default `min_nprobes` right?

i think we need to change the default,
let `min_probes = min(default_min, max.unwrap_or(default_min))` and `max_probes = max(default_max, max.unwrap_or(default_max))`

MarkMcCaskey: > iiuc we will still see error if min_nprobes is set to a value greater than default min_nprobes right?

Yeah the defaults are taken into account, so if you set min probes above the default for max probes, it will fail. So those edge cases do exist but given that an existing test assert that they throw, I thought maybe the default values were meant to work that way. i.e. there's an existing test like:

```python
def test_invalid_nprobes_sync(table):
    with pytest.raises(ValueError, match="minimum_nprobes must be greater than 0"):
        LanceVectorQueryBuilder(table, [0, 0], "vector").minimum_nprobes(0).to_list()
    with pytest.raises(
        ValueError, match="maximum_nprobes must be greater than or equal to minimum_nprobes"
    ):
        LanceVectorQueryBuilder(table, [0, 0], "vector").maximum_nprobes(5).to_list()
    with pytest.raises(
        ValueError, match="minimum_nprobes must be less than or equal to maximum_nprobes"
    ):
        LanceVectorQueryBuilder(table, [0, 0], "vector").minimum_nprobes(100).to_list()
```

So we could have max/min behave more naturally even if both aren't set with the logic you described above but my main concern was more so that there's no way to set it at all without the changes in the PR. The issue you're describing can be made to work by setting both min and max. That's not to say it's not important, but I'd say the fix in this PR is needed to make the API usable at all and making min/max not assume a default is really more about making the API intuitive.

What I mean is that as-is I can't even do `query.maximum_probes(75).minimum_probes(75)` -- I can't use nprobes at all. `.nprobes(N)` for N > 20 worked on the older version of the client library I'm upgrading from.

Happy to do more here and make the defaults be less unintuitive but perhaps it should be a follow up PR. It seems a bit more ambiguous to me what exactly the ideal logic should be here / where it should live. Should a max only API call always set the minimum to that value? I'm actually not sure what the semantics / implications of min/max actually are -- I just wanted the normal `.nprobes(N)` where min == max to work again.

wkalt (approved):

wkalt:
@MarkMcCaskey thanks!

PR #2568, opened 2025-08-01
fix: make cloud features optional (#2567)

### LINKED ISSUE #2567: lancedb pulls in aws-smithy / aws-config even if bedrock feature is disabled
LanceDB depends on lance which depends on aws by default which pulls in a lot of heavy dependencies even if AWS support isn't needed (same goes for azure, gcp, and oss features although I haven't looked at their dependency chains). Where it specifically caused me problems was that it enabled ring as the backend for rustls whereas everywhere else in my project I use aws-lc-sys which then caused the project to start failing at runtime because I have to now explicitly set the provider.

It would be nice if those were optional features that could be opted in to.

Is changing the lance dependency for lancedb to default-features=False a change that you'd be open for me to make? Or is it more involved?

vlovich:
This shrinks the size of a local embedded build that can disable all the default features. When combined with https://github.com/lancedb/lance/pull/4362 and the dependencies are updated to point to the fix, this resolves #2567 fully.

Verified by patching the workspace to redirect to my clone of lance with the PR applied.
```
cargo tree -p lancedb -e no-build -e no-dev --no-default-features -i aws-config | less
```

The reason that lance itself needs to change too is that many dependencies within that project depend on lance-io/default and lancedb depends on them which transitively ends up enabling the cloud regardless. The PR in lance removes the dependency on lance-io/default from all sibling crates.

### DIFF
--- a/nodejs/Cargo.toml
+++ b/nodejs/Cargo.toml
@@ -18,7 +18,7 @@ arrow-array.workspace = true
 arrow-schema.workspace = true
 env_logger.workspace = true
 futures.workspace = true
-lancedb = { path = "../rust/lancedb" }
+lancedb = { path = "../rust/lancedb", default-features = false }
 napi = { version = "2.16.8", default-features = false, features = [
     "napi9",
     "async"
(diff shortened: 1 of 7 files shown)


nodejs/Cargo.toml
wjones127, line 1: Should we also do this for python?

vlovich: Good catch.

wjones127 (approved):

vlovich:
Do I need to take any action to merge this?

PR #2569, opened 2025-08-01
fix: correct column name replacement in query_schemas()

xhwhis:


wjones127:
Hello. Could you please explain what you are trying to fix? It is not clear from your PR.

xhwhis:
@wjones127 When converting `fts_schema` to `vec_schema`, it should change from target=`SCORE_COL` to replacement=`DIST_COL`, not from `DIST_COL` to `SCORE_COL`.

PR #2575, opened 2025-08-04
fix: always uses slashes in table uris

### LINKED ISSUE #2574: bug(rust): can't create table on S3 from Windows
### LanceDB version

_No response_

### What happened?

Table URI is constructed with local file path handling, even when on S3. This means that when running on Windows against S3, backslashes will be put in S3 URI.

This particular line is the likely offender:

https://github.com/lancedb/lancedb/blob/9d683e4f0b1ddd3a2abed8f02dc13d252862a23c/rust/lancedb/src/database/listing.rs#L414-L415

Reported: https://github.com/lancedb/lance/issues/4387#issuecomment-3152482414

wjones127:
Closes #2574

### DIFF
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -411,17 +411,13 @@ impl ListingDatabase {
     fn table_uri(&self, name: &str) -> Result<String> {
         validate_table_name(name)?;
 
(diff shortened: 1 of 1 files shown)


westonpace (approved):
~~Pending CI~~ nevermind, CI passed :laughing:

PR #2577, opened 2025-08-07
docs: remove old Documentation

davidmyriel:
- I have unlinked the old docs. They are still in the docs folder, for a few days and then I will retire them altogether.
- Right now, the site should just build the SDKs - and those can remain with the core project.
- Let me know if you have any better ideas how we can do this.


prrao87:
This PR is totally out of date and unnecessary. Closing as it's been addressed in other more recent PRs.

PR #2583, opened 2025-08-11
feat: add missing rust examples

yuvalif:
all 3 example are running now with:
```
cargo run --example simple
cargo run --example full_text_search
cargo run --example ivf_pq
```


westonpace (approved):
Thanks for catching this!

PR #2584, opened 2025-08-12
feat: add take_offsets and take_row_ids

westonpace:
These operations have existed in lance for a long while and many users need to drop down to lance for this capability.  This PR adds the API and implements it using filters (e.g. `_rowid IN (...)`) so that in doesn't currently add any load to `BaseTable`.  I'm not sure that is sustainable as base table implementations may want to specialize how they handle this method.  However, I figure it is a good starting point.

In addition, unlike Lance, this API does not currently guarantee anything about the order of the take results.  This is necessary for the fallback filter approach to work (SQL filters cannot guarantee result order)


wjones127 (approved):
Needs a rebase, but otherwise looks good 👍

PR #2586, opened 2025-08-13
feat: allow setting `train=False` and `name` on indices

wjones127:
Enables two new parameters when building indices:

* `name`: Allows explicitly setting a name on the index. Default is `{col_name}_idx`.
* `train` (default `True`): When set to `False`, an empty index will be immediately created.

The upgrade of Lance means there are also additional behaviors from https://github.com/lancedb/lance/commit/cd76a993b8b65877135cb9cfb47448a761ac7f7b:

* When a scalar index is created on a Table, it will be kept around even if all rows are deleted or updated.
* Scalar indices can be created on empty tables. They will default to `train=False` if the table is empty.

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -999,6 +999,18 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
             "column": column
         });
 
+        // Add name parameter if provided (for backwards compatibility, only include if Some)
+        if let Some(ref name) = index.name {
+            body["name"] = serde_json::Value::String(name.clone());
+        }
+
+        // Warn if train=false is specified since it's not meaningful
+        if !index.train {
+            log::warn!(
+                "train=false has no effect remote tables. The index will be created empty and automatically populated in the background."
+            );
+        }
+
         match index.index {
             // TODO: Should we pass the actual index parameters? SaaS does not
             // yet support them.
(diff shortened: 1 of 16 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -999,6 +999,18 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+        // Add name parameter if provided (for backwards compatibility, only include if Some)
+        if let Some(ref name) = index.name {
+            body["name"] = serde_json::Value::String(name.clone());
+        }
+
+        // Warn if train=false is specified since it's not supported in remote tables
+        if !index.train {
+            log::warn!(
+                "train=false is not supported for remote tables. The index will be created empty and automatically updated in the background."
+            );
```
wjones127:
```suggestion
        // Warn if train=false is specified since it's not meaningful
        if !index.train {
            log::warn!(
                "train=false has no effect remote tables. The index will be created empty and automatically populated in the background."
            );
```

rust/lancedb/src/table.rs
```diff
@@ -1649,345 +1647,211 @@ impl NativeTable {
...
+                    // Use IvfPq as the default for auto vector indices
+                    let dim = Self::get_vector_dimension(field)?;
+                    let num_partitions = self.get_num_partitions(None, false, None).await?;
+                    let num_sub_vectors = Self::get_num_sub_vectors(None, dim);
+                    let lance_idx_params = lance::index::vector::VectorIndexParams::ivf_pq(
+                        num_partitions as usize,
+                        /*num_bits=*/ 8,
+                        num_sub_vectors as usize,
+                        lance_linalg::distance::MetricType::L2,
+                        50, // max_iterations
```
westonpace: Super minor nit but you're mixing conventions here.
```suggestion
                        /*num_bits=*/ 8,
                        num_sub_vectors as usize,
                        lance_linalg::distance::MetricType::L2,
                        /*max_iterations=*/ 50,
```

westonpace (approved):
Nice cleanup of the create index code

PR #2587, opened 2025-08-13
feat: add APIs to update metadata

wjones127:


rust/lancedb/src/table.rs
```diff
@@ -605,6 +606,30 @@ pub trait BaseTable: std::fmt::Display + std::fmt::Debug + Send + Sync {
...
     /// Get statistics on the table
     async fn stats(&self) -> Result<TableStatistics>;
+
+    /// Update table metadata
+    ///
+    /// Pass `None` for a value to remove that key.
+    /// Pass `replace=true` to replace the entire metadata map.
+    ///
+    /// Returns the updated metadata map after the operation.
+    async fn update_metadata(
```
cmccabe, line 572: does it make more sense to have a separate function to replace the entire metadata map, rather than having a boolean?

cmccabe:
general comment (about all the APIs) is that it feels better to have a separate function to replace the entire metadata map, rather than making it a boolean

rust/lancedb/src/table.rs
wjones127: I could do that. cc @wkalt who helped design this API earlier. What do you think?

wjones127: @cmccabe is your motivation that you really think this should be another method (such as `replace_metadata()`)? Or that it's annoying to have to pass the boolean each time you call the API?

If it's the latter, here's how I solved it in Lance:

```rust
// Update single key
dataset.update_config([("key", "value")]).await?;

// Remove a key
dataset.update_config([("to_delete", None)]).await?;

// Clear all config
dataset.update_config([] as [UpdateMapEntry; 0]).replace().await?;

// Replace full config
dataset.update_config([("k1", "v1"), ("k2", "v2")]).replace().await?;
```

You just add `.replace()` to the chain if you want to replace, but it's optional if you just want to update / delete keys.

What do you think of that?

cmccabe: sure, a "fluent api" like that would be nice.

PR #2588, opened 2025-08-14
fix(embeddings/azure): correct AzureOpenAI client args and support explicit api_version

### LINKED ISSUE #2589: bug(python): LanceDB fails with AzureOpenAI when using 'base_url' and passing None for 'dimensions'
### LanceDB version

v0.24.2

## Describe the bug

When configuring LanceDB embeddings for Azure OpenAI, the following issues arise:

- The client should use `azure_endpoint`, not `base_url`.
- Passing 'dimensions': None in the embeddings kwargs causes errors.
- Users cannot easily supply `api_version` as an argument—have to rely on environment variables, which is not always robust or transparent.

## Expected behavior

- Robust handling of Azure endpoints.
- Automatic omission of unnecessary/invalid keys for AzureOpenAI API.
- Ability to specify `api_version` as a keyword argument, with fallback to env if not set.

## Proposed Fix

- Add `api_version` attribute/parameter to the embedding class and registry.
- Use it when constructing the AzureOpenAI client, falling back to environment for backward compatibility.
- See PR #2588  for the implementation.

## Additional context

Tested with:
- lancedb==0.24.2
- openai==1.99.9

_No response_

dutchfarao:
Fixes two key issues with Azure OpenAI embeddings integration.

- Use azure_endpoint (not base_url) when creating AzureOpenAI client.
- Remove "dimensions" kwarg if its value is None, improving Azure compatibility.
- feat: add support for an explicit api_version parameter when creating the AzureOpenAI client. If not provided, will fall back to the OPENAI_API_VERSION environment variable.

Tested locally with lancedb==0.24.2 and openai==1.99.9. Now embedding tables work with Azure OpenAI by setting api_version as an argument or via environment variable.

Fixes #2589

### DIFF
--- a/python/python/lancedb/embeddings/openai.py
+++ b/python/python/lancedb/embeddings/openai.py
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
+import logging
 from functools import cached_property
 from typing import TYPE_CHECKING, List, Optional, Union
-import logging
 
 from ..util import attempt_import_or_raise
 from .base import TextEmbeddingFunction
@@ -50,6 +50,7 @@ class OpenAIEmbeddings(TextEmbeddingFunction):
 
     # Set true to use Azure OpenAI API
     use_azure: bool = False
+    api_version: Optional[str] = None
 
     def ndims(self):
         return self._ndims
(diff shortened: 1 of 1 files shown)


dutchfarao:
Breaking Change label was added because my original description had BREAKING CHANGE = None.

Cannot remove label, but this change is obviously not breaking.

PR #2595, opened 2025-08-15
ci: remove more mentions of node

wjones127:
I promise this time I tested it locally :)


westonpace (approved):
Pending CI (hmm...looks like not enough change to trigger any jobs.  If you tested locally I'll take your word for it :laughing: )

PR #2596, opened 2025-08-18
feat: add __getitems__ method impl for torch integration

westonpace:
This allows a lancedb Table to act as a torch dataset.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -1150,7 +1150,7 @@ class Table(ABC):
         is a record batch (and so users should take care not to pass in too many
         offsets)
 
-        Note: this method is primarily intended to fulfill the dataset contract
+        Note: this method is primarily intended to fulfill the Dataset contract
         for pytorch.
 
         Parameters
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -1113,7 +1113,9 @@ class Table(ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def take_offsets(self, offsets: list[int]) -> LanceTakeQueryBuilder:
+    def take_offsets(
+        self, offsets: list[int], *, with_row_id: bool = False
+    ) -> LanceTakeQueryBuilder:
         """
         Take a list of offsets from the table.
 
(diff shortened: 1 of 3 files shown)


westonpace:
Leaving in draft as this relies on https://github.com/lancedb/lance/pull/4478

python/python/lancedb/table.py
```diff
@@ -1139,8 +1141,60 @@ def take_offsets(self, offsets: list[int]) -> LanceTakeQueryBuilder:
...
+    def __getitems__(self, offsets: list[int]) -> pa.RecordBatch:
+        """
+        Take a list of offsets from the table and return as a record batch.
+
+        This method uses the `take_offsets` method to take the rows.  However, it
+        aligns the offsets to the passed in offsets.  This means the return type
+        is a record batch (and so users should take care not to pass in too many
+        offsets)
+
+        Note: this method is primarily intended to fulfill the dataset contract
```
wkalt: nit: `Dataset`

python/python/lancedb/table.py
```diff
@@ -1139,8 +1141,60 @@ def take_offsets(self, offsets: list[int]) -> LanceTakeQueryBuilder:
...
+            self.take_offsets(offsets)
+            .select(columns)
+            .to_arrow()
+            .sort_by("_rowoffset")
+            .take(permutation_inv)
+            .combine_chunks()
+            .drop_columns(["_rowoffset"])
+        )
+
+        return tbl
```
wkalt, line 1192: could you add a test that asserts the input/output ordering is correct? I managed to work it out on paper but it took a minute.

Also what part is the expensive part? I guess the sorting of tbl at the end?

wkalt: (if the test you added does assert that indirectly, feel free to disregard)

wkalt: I think it's possible to avoid the sorting with something like this:
```
let output_locations = defaultdict(list)
for offset, i in enumerate(offsets):
    output_locations[offset].append(i)

table = self.take_offsets(offsets).select(columns).to_arrow()
row_offsets = table.column("_rowoffsets")

permutation = [0] * len(offsets)
for offset, i in enumerate(row_offsets):
    idx = output_locations[offset].popleft()
    permutation[idx] = offset

table.take(permutation).combine_chunks().drop_columns(["_rowoffset"])
```

wkalt (approved):
couple minor comments

python/python/lancedb/table.py
westonpace: > Also what part is the expensive part? I guess the sorting of tbl at the end?

The expensive part is most likely the I/O unless the data is in memory.  The rest will be trivial (including the sort).

However, if there is a lot of data, then the `take` will be rather expensive since it will require a rewrite / copy of all the data in the new order.  E.g. sorting a batch of 1024 integers isn't so bad.  However, rewriting a batch of 1024 images in a new order is a bit bad.  Ideally we could avoid this in the future although I think the only way to truly avoid it is to push the take impl into the tables so they can use an optimized version that does guarantee order.

> I think it's possible to avoid the sorting with something like this:

This replaces a sort (N*LOG(N)) with O(N) hashmap lookups and insertions.  At the batch sizes we're likely to see in pytorch workloads (typically 8Ki or less I would guess) I'm not sure there is much difference performance wise.  Also, we can probably just use an int->int map instead of an int->list map.  The offsets should be unique.

I'm just going to stick with what I have for now (since we may push it down anyways) but will add a test case.

python/python/lancedb/table.py
westonpace: Fixed

PR #2599, opened 2025-08-20
feat(python): integrate with lance namespace

jackye1995:
This PR integrates `lancedb` with `lance-namespace` so that users can use LanceDB client to access Lance tables in any catalog services. In general, we expect most of the logic to be delegated to the existing `LanceDBConnection` and `LanceTable`, but the namespace implemenation will control how table is created, dropped, and describe where the table is stored with any related storage options like access credentials.

The implementation currently only supports a 1 level namespace that directly contains tables. We will introduce nested namespace support in a separated PR.

Users are expected to use it in the following way:

```python
>>> import lancedb
>>> import pyarrow as pa
>>> # Connect using GlueNamespace
>>> db = lancedb.connect_namespace("glue", {"catalog_id": "123456789012"})
>>> # Create a table with schema
>>> schema = pa.schema([
...     pa.field("id", pa.int64()),
...     pa.field("vector", pa.list_(pa.float32(), 2))
... ])
>>> table = db.create_table("my_table", schema=schema)
>>> # List tables
>>> db.table_names()
['my_table']
```

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -10,7 +10,7 @@ dependencies = [
     "pyarrow>=16",
     "pydantic>=1.10",
     "tqdm>=4.27.0",
-    "lance-namespace"
+    "lance-namespace==0.0.6"
 ]
 description = "lancedb"
 authors = [{ name = "LanceDB Devs", email = "dev@lancedb.com" }]
@@ -138,6 +138,3 @@ include = [
 ]
 exclude = ["python/tests/"]
 pythonVersion = "3.12"
-
-[tool.uv.sources]
-lance-namespace = { path = "../../lance-namespace/python/lance_namespace" }
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -10,6 +10,7 @@ dependencies = [
     "pyarrow>=16",
     "pydantic>=1.10",
     "tqdm>=4.27.0",
+    "lance-namespace"
 ]
 description = "lancedb"
 authors = [{ name = "LanceDB Devs", email = "dev@lancedb.com" }]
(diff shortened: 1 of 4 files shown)


python/pyproject.toml
```diff
@@ -137,3 +138,6 @@ include = [
 ]
 exclude = ["python/tests/"]
 pythonVersion = "3.12"
+
+[tool.uv.sources]
+lance-namespace = { path = "../../lance-namespace/python/lance_namespace" }
```
jackye1995: need to do a release, will remove after release completion

wjones127:
This adds support for Python. Eventually, we want this for Rust, Node, and Java too. What's are plan for multi-language support? Will we have one-off implementations for each?

jackye1995:
> This adds support for Python. Eventually, we want this for Rust, Node, and Java too. What's are plan for multi-language support? Will we have one-off implementations for each?

The namespace SDK is mostly codegen from OpenAPI, so there will be an integration per-language. Actually Java is already using that https://github.com/lancedb/lancedb/tree/main/java/lance-namespace (not fully completed work yet), the integration will in general be lightweight like this one, just to add additional layers on top of tables.

There is one question regarding rust, my understanding is that rust is mainly for python, I don't know if there are people actually using rust directly, and if it is worth doing an integration with rust at this point, what do you think? @wjones127

wjones127:
> I don't know if there are people actually using rust directly, and if it is worth doing an integration with rust at this point, what do you think?

We do have users that use Rust directly. I don't know if they need the catalog. I just ask to make sure we aren't making the maintenance burden too high. If it's mostly just codegen, then that's fine.

wjones127 (approved):
Seems reasonable. I suppose as a follow up we can generate the async version of connect as well?

PR #2603, opened 2025-08-21
feat!: support multi-level namespace

jackye1995:
This PR adds support of multi-level namespace in a LanceDB database, according to the Lance Namespace spec.

This allows users to create namespace inside a database connection, perform create, drop, list, list_tables in a namespace. (other operations like update, describe will be in a follow-up PR)

The 3 types of database connections behave like the following:
1 Local database connections will continue to have just a flat list of tables for backwards compatibility.
2. Remote database connections will make REST API calls according to the APIs in the Lance Namespace spec. 
3. Lance Namespace connections will invoke the corresponding operations against the specific namespace implementation which could have different behaviors regarding these APIs.

All the table APIs now take identifier instead of name, for example `/v1/table/{name}/create` is now `/v1/table/{id}/create`. If a table is directly in the root namespace, the API call is identical. If the table is in a namespace, then the full table ID should be used, with `$` as the default delimiter (`.` is a special character and creates issues with URL parsing so `$` is used), for example `/v1/table/ns1$table1/create`. If a different parameter needs to be passed in, user can configure the `id_delimiter` in client config and that becomes a query parameter, for example `/v1/table/ns1__table1/create?delimiter=__`

The Python and Typescript APIs are kept backwards compatible, but the following Rust APIs are not:
1. `Connection::drop_table(&self, name: impl AsRef<str>) -> Result<()>` is now `Connection::drop_table(&self, name: impl AsRef<str>, namespace: &[String]) -> Result<()>`
2. `Connection::drop_all_tables(&self) -> Result<()>` is now `Connection::drop_all_tables(&self, name: impl AsRef<str>) -> Result<()>`

### DIFF
--- a/nodejs/src/connection.rs
+++ b/nodejs/src/connection.rs
@@ -100,10 +100,12 @@ impl Connection {
     #[napi(catch_unwind)]
     pub async fn table_names(
         &self,
+        namespace: Vec<String>,
         start_after: Option<String>,
         limit: Option<u32>,
     ) -> napi::Result<Vec<String>> {
         let mut op = self.get_inner()?.table_names();
+        op = op.namespace(namespace);
         if let Some(start_after) = start_after {
             op = op.start_after(start_after);
         }
@@ -125,12 +127,15 @@ impl Connection {
         name: String,
         buf: Buffer,
         mode: String,
+        namespace: Vec<String>,
         storage_options: Option<HashMap<String, String>>,
     ) -> napi::Result<Table> {
         let batches = ipc_file_to_batches(buf.to_vec())
             .map_err(|e| napi::Error::from_reason(format!("Failed to read IPC file: {}", e)))?;
         let mode = Self::parse_create_mode_str(&mode)?;
         let mut builder = self.get_inner()?.create_table(&name, batches).mode(mode);
+
+        builder = builder.namespace(namespace);
 
         if let Some(storage_options) = storage_options {
             for (key, value) in storage_options {
@@ -147,6 +152,7 @@ impl Connection {
         name: String,
         schema_buf: Buffer,
         mode: String,
+        namespace: Vec<String>,
         storage_options: Option<HashMap<String, String>>,
     ) -> napi::Result<Table> {
         let schema = ipc_file_to_schema(schema_buf.to_vec()).map_err(|e| {
@@ -157,6 +163,9 @@ impl Connection {
             .get_inner()?
             .create_empty_table(&name, schema)
             .mode(mode);
+
+        builder = builder.namespace(namespace);
+
         if let Some(storage_options) = storage_options {
             for (key, value) in storage_options {
                 builder = builder.storage_option(key, value);
@@ -170,10 +179,14 @@ impl Connection {
     pub async fn open_table(
         &self,
         name: String,
+        namespace: Vec<String>,
         storage_options: Option<HashMap<String, String>>,
         index_cache_size: Option<u32>,
     ) -> napi::Result<Table> {
         let mut builder = self.get_inner()?.open_table(&name);
+
+        builder = builder.namespace(namespace);
+
         if let Some(storage_options) = storage_options {
             for (key, value) in storage_options {
                 builder = builder.storage_option(key, value);
@@ -188,12 +201,18 @@ impl Connection {
 
     /// Drop table with the name. Or raise an error if the table does not exist.
     #[napi(catch_unwind)]
-    pub async fn drop_table(&self, name: String) -> napi::Result<()> {
-        self.get_inner()?.drop_table(&name).await.default_error()
+    pub async fn drop_table(&self, name: String, namespace: Vec<String>) -> napi::Result<()> {
+        self.get_inner()?
+            .drop_table(&name, &namespace)
+            .await
+            .default_error()
     }
 
     #[napi(catch_unwind)]
-    pub async fn drop_all_tables(&self) -> napi::Result<()> {
-        self.get_inner()?.drop_all_tables().await.default_error()
+    pub async fn drop_all_tables(&self, namespace: Vec<String>) -> napi::Result<()> {
+        self.get_inner()?
+            .drop_all_tables(&namespace)
+            .await
+            .default_error()
     }
 }
--- a/python/python/tests/test_namespace.py
+++ b/python/python/tests/test_namespace.py
@@ -621,3 +621,78 @@ class TestNamespaceConnection:
 
         # Now dropping namespace should work
         db.drop_namespace(["test_namespace"])
+
+    def test_same_table_name_different_namespaces(self):
+        db = lancedb.connect_namespace("temp", {"root": self.temp_dir})
+
+        # Create two namespaces
+        db.create_namespace(["namespace_a"])
+        db.create_namespace(["namespace_b"])
+
+        # Define schema
+        schema = pa.schema(
+            [
+                pa.field("id", pa.int64()),
+                pa.field("vector", pa.list_(pa.float32(), 2)),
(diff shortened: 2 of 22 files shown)


wkalt:
@jackye1995 are existing table names with "$" handled, so that current remote API calls continue to work? Assuming $ is legal in tables.

nodejs/src/connection.rs
```diff
@@ -100,10 +100,12 @@ impl Connection {
     #[napi(catch_unwind)]
     pub async fn table_names(
         &self,
+        namespace: Vec<String>,
```
wkalt, line 103: I'm wondering if namespace interacts with this cache key: https://github.com/lancedb/lance/blob/main/rust/lance-io/src/object_store/providers.rs#L158

Is it part of the base_path and thus included in the key? Namespace seems analogous in some ways to "azure_storage_account_name", which is included in that key through params.

jackye1995:
> are existing table names with "$" handled, so that current remote API calls continue to work? Assuming $ is legal in tables.

$ is not legal currently in the table name: https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/utils.rs#L22

```rust
lazy_static! {
    static ref TABLE_NAME_REGEX: regex::Regex = regex::Regex::new(r"^[a-zA-Z0-9_\-\.]+$").unwrap();
}
```

Even if for some reason $ is in the name (e.g. the user created the table in lance and try to use in lancedb), it is okay because the code path will be using the local mode which does not support namespace. It does not go through the remote codepath.

For the use case of connecting to another external catalog, $ is just the default. It can be overridden by setting the delimiter to something else, and the API request becomes something like for example `POST /namespace/ns1@ns2/create?delimiter=@`.

python/python/tests/test_namespace.py
```diff
@@ -412,3 +546,78 @@ def test_storage_options(self):
...
+
+        # Try to drop namespace with tables - should fail
+        with pytest.raises(RuntimeError, match="contains tables"):
+            db.drop_namespace(["test_namespace"])
+
+        # Drop table first
+        db.drop_table("test_table", namespace=["test_namespace"])
+
+        # Now dropping namespace should work
+        db.drop_namespace(["test_namespace"])
```
wkalt, line 623: I think it would be useful to have a test that creates identically-named tables in two different namespaces, and verifies that the real data stored in each table can be read back. Sorry if that's here and I missed it.

nodejs/src/connection.rs
jackye1995: No, cache key of a dataset is using its dataset URL, it works outside the concept of that is the actual table name or table ID of the dataset.

python/python/tests/test_namespace.py
jackye1995: no it's not there, I can add it

nodejs/src/connection.rs
wkalt: thanks, that makes sense.

rust/lancedb/src/remote/db.rs
```diff
@@ -264,9 +293,13 @@ impl<S: HttpSend> Database for RemoteDatabase<S> {
             .err_to_http(request_id)?
             .tables;
         for table in &tables {
+            let table_identifier =
+                build_table_identifier(table, &request.namespace, &self.client.id_delimiter);
             let remote_table = Arc::new(RemoteTable::new(
                 self.client.clone(),
                 table.clone(),
+                request.namespace.clone(),
+                table_identifier,
```
wkalt: should the cache insert two lines below be for "table_identifier"?

rust/lancedb/src/remote/db.rs
```diff
@@ -245,10 +246,38 @@ impl From<&CreateTableMode> for &'static str {
     }
 }
 
+fn build_table_identifier(name: &str, namespace: &[String], delimiter: &str) -> String {
+    if !namespace.is_empty() {
+        let mut parts = namespace.to_vec();
+        parts.push(name.to_string());
+        parts.join(delimiter)
+    } else {
+        name.to_string()
```
wkalt, line 255: how do we guarantee that a namespaced table identifier and a non-namespaced table identifier can never collide?

wkalt: is this delimiter the user-controlled one?

rust/lancedb/src/remote/db.rs
jackye1995: oh nice catch!

rust/lancedb/src/remote/db.rs
jackye1995: yes, so user can choose a delimiter that never collides. Although the default should work in most cases right now because of the table name regex we put in place.

wkalt: I'm thinking about the multitenant or adversarial case though. If I have a table called "my_table" and you create a table called "table" with namespace "my" and delimiter "_", are you going to see my table data? Would you be able to do some operations but not others, because some stuff goes through this cache and other stuff doesn't?

I think if that's possible we probably still need to think about how to fix it.

python/python/tests/test_namespace.py
jackye1995: added test `test_same_table_name_different_namespaces`

rust/lancedb/src/remote/db.rs
wkalt: I'm not sure if we really need to use the user's identifier here? This is our internal cache right?

If we forced the delimiter in this cache key to be an illegal character for tables, we would have a guarantee.

I guess that might not be simple since you want to return the delimited name to the user.

wkalt:
```
multitenant or adversarial case
```
Just to flesh out the concern - if I build a server application around LanceDB and implement RBAC and permissions, and properly validate table and namespace ownership for requests before sending them to LanceDB, then I think this would become a security bug in my app and it would be hard for me to find out about it.

jackye1995: I see, that's a good point. I think we can make a unique cache key that is the table_identifier + delimiter to decode the identifier, so that there is no ambiguity - if the name comes with `my_table` and the delimiter is `_`, then it will try to get cache key like `table://my_table?delimiter=_` which will not be able to collide. What do you think? @wkalt

jackye1995: oh actually there is an even better way, we store the list version of the identifier as cache key, like `["my", "table"]` vs `["my_table"]`, so there is no ambiguity about the reference, and if it's actually referring to the same table with different delimiter, it can also hit the cache

jackye1995: to make the representation more compact, we can use a totally invalid character like `\0` as well

wkalt: Yeah, that sounds like it would work, and would also handle the cross-namespace version of this, i.e my_table with delimiter "_" vs delimiter "y".

jackye1995: I ended up using a more complicated encoding scheme so that it has no dependency on any special character or delimiter, and is guaranteed to be unique, and much more compact than doing the JSON list thing:

```
hex([u32_len][namespace1][u32_len][namespace2]...[u32_len][table_name])
```

@wkalt let me know if there is any other concern!

wkalt: @jackye1995 my only thought on that scheme is that it'll be less debuggable (e.g in the debugger) than the non-hex strings with an illegal delimiter. But I think it'll do the trick and it's not user-exposed so it's easy to change if we want.

jackye1995:
Will merge it first to continue testing, will likely have followups for bug fixes

PR #2620, opened 2025-09-02
fix: remote python sdk namespace typing

wkalt:
This changes the default values for some namespace parameters in the remote python SDK from None to [], to match the underlying code it calls.

Prior to this commit, failing to supply "namespace" with the remote SDK would cause an error because the underlying code it dispatches to does not consider None to be valid input.


wkalt:
there's a couple more instances of this, I'll tack another commit on shortly.

jackye1995 (approved):

PR #2627, opened 2025-09-04
feat: initial implementation of C bindings for lancedb

yuvalif:
including C++ example files


Xuanwo:
Is it a good idea for us to go with cpp directly? I know we'll expose a C ABI under hood anyway but maintaining a high level cpp API seems more sustainable for us.

WDYT?

yuvalif:
> Is it a good idea for us to go with cpp directly? I know we'll expose a C ABI under hood anyway but maintaining a high level cpp API seems more sustainable for us.
> 
> WDYT?

I think that in order to cross language barriers we have to use C (e.g. apache arrow is written in C++ but uses C ABIs to interact with rust).
OTOH, a C++ API could be much nicer and safer when the calling code is C++ as well.
So, in the future, we should probably create a C++ wrapper: `include/liblancedb.hpp` to the C API. for example, we can have a `Connection` class:
```cpp
namespace lancedb {
	class Connection {
		LanceDBConnectBuilder* const builder;
		LanceDBConnection* connection = nullptr;
	public:
		Connection(const std::string& uri) : builder(lancedb_connect(uri.c_str()) {}
		~Connection() {
			lancedb_connect_builder_free(builder);
			lancedb_connection_free(connection);
		}
		
		LanceDBConnection* execute() {
			if (builder) {
				connection = lancedb_connect_builder_execute(builder);
			}
			return connection;
		}
		// member functions for the rest of the connection function from lancedb.h
	};
}
```

c/examples/simple.cpp
```diff
@@ -0,0 +1,271 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright The LanceDB Authors
+ */
+
+#include <sys/stat.h>
+#include <iostream>
+#include <memory>
+#include <vector>
+#include <arrow/api.h>
+#include <arrow/c/bridge.h>
```
wjones127, line 11: Would be cool to have a simple example with nanoarrow: https://arrow.apache.org/nanoarrow/latest/index.html

c/src/table.rs
```diff
@@ -0,0 +1,541 @@
...
+) -> LanceDBError {
+    if table.is_null() || schema_out.is_null() {
+        set_invalid_argument_message(error_message);
+        return LanceDBError::InvalidArgument;
+    }
+
+    let tbl = &(*table).inner;
+    let runtime = get_runtime();
+
+    match runtime.block_on(tbl.schema()) {
```
wjones127, line 48: I think `.block_on` can only be called once (might depend on the flavor of runtime you are using[^1]). So as written I don't think any of these functions could be called concurrently, which is probably an issue.

You should take a look what I did for pylance on the runtime: https://github.com/lancedb/lance/blob/main/python/src/executor.rs

That runs `block_on` in a background thread, and calling threads will use `spawn()` to run the futures. It also has nice handling for interrupts for cancellation, which I think it a worthwhile feature.

[^1]: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on

wjones127 (commented):
Thanks for starting this. I think this could be helpful for other C users.

I think talking to other engineers, we don't think we want to include this in the LanceDB repo yet. We'd like for the C bindings to start in a separate repo, until we have a critical mass of maintainers who are working on C. At this time, I don't think the existing maintainers are experienced enough with C to maintain these bindings and update them as APIs change. Having this code in a separate repo would ease that.

c/examples/simple.cpp
yuvalif: as I'm adding "C bindings" I guess that it would make senmse to add C example and not just C++  :-)
would add `simple.c` and use nanoarrow there

c/src/table.rs
yuvalif: I will look into that. for now I left implementing async IO between the app the lancedb (so everything is a blocking call...)
but would probably try to copy from what you did with the python binding

yuvalif:
> Thanks for starting this. I think this could be helpful for other C users.
> 
> I think talking to other engineers, we don't think we want to include this in the LanceDB repo yet. We'd like for the C bindings to start in a separate repo, until we have a critical mass of maintainers who are working on C. At this time, I don't think the existing maintainers are experienced enough with C to maintain these bindings and update them as APIs change. Having this code in a separate repo would ease that.

sounds good. 
how would you think this should be structured (assuming it might be incoporated into lancedb at some point):
* separate repo under my user (or maybe the ceph org) that takes a stable lancedb as a git submodule
* keep the current structure as a "friendly fork" - just rebasing with no PRs

wjones127:
> how would you think this should be structured (assuming it might be incoporated into lancedb at some point):
> 
> * separate repo under my user (or maybe the ceph org) that takes a stable lancedb as a git submodule
> * keep the current structure as a "friendly fork" - just rebasing with no PRs

I would create a separate repo and just have `lancedb` as a dependency from cargo. I don't think there's any reason to use a submodule. If you want to reference a commit rather than a released version, you could also use the [git references in Cargo.toml](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).

If `ceph` wants to host it, they are welcome. Otherwise, I'd be open to creating a repo under `lancedb`.

yuvalif:
> > how would you think this should be structured (assuming it might be incoporated into lancedb at some point):
> > 
> > * separate repo under my user (or maybe the ceph org) that takes a stable lancedb as a git submodule
> > * keep the current structure as a "friendly fork" - just rebasing with no PRs
> 
> I would create a separate repo and just have `lancedb` as a dependency from cargo. I don't think there's any reason to use a submodule. If you want to reference a commit rather than a released version, you could also use the [git references in Cargo.toml](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).
> 
> If `ceph` wants to host it, they are welcome. Otherwise, I'd be open to creating a repo under `lancedb`.

thanks @wjones127, i think that it would make more sense to add a a repo under the lancedb org (as ceph is mainly a user for this).

yuvalif:
as discussed closing in favor of: https://github.com/yuvalif/lancedb-c

@wjones127 please let me know if you want to fork the above into the lancedb org.
i will continue work there as the main repo for lancedb-c

wjones127:
@yuvalif sorry for the delay. I created https://github.com/lancedb/lancedb-c and have given you write access.

I added the apache-2.0 license to match the existing license for LanceDB. Would you mind moving the initial commit over there?

yuvalif:
> @yuvalif sorry for the delay. I created https://github.com/lancedb/lancedb-c and have given you write access.
> 
> I added the apache-2.0 license to match the existing license for LanceDB. Would you mind moving the initial commit over there?

sure. will continue in there. thanks!

PR #2630, opened 2025-09-04
feat: support shallow_clone for native tables

jackye1995:
## Summary
This PR exposes the shallow clone feature from Lance to LanceDB for native tables.

The shallow clone operation creates a new table that shares the underlying data files with the source table. This is useful for creating table copies without duplicating data.

## Changes
- Added `shallow_clone` method to `BaseTable` trait and `Table` struct
- Implemented `shallow_clone` for `NativeTable` using Lance's shallow_clone functionality
- Added stub implementation for `RemoteTable` that returns "not supported" error
- Updated API to take target connection, table name, and namespace instead of raw URI
- Added comprehensive test for shallow_clone functionality
- Added Python bindings with placeholder implementation (needs more work to properly pass connection objects)
- Updated Python type hints and table classes

## API Design
The shallow clone API takes:
- `target_conn`: The target database connection
- `target_table_name`: The name of the table to create
- `target_namespace`: The namespace for the new table (defaults to empty list)
- `version`: Optional version to clone (defaults to latest)

## Example Usage
```rust
let cloned_table = table
    .shallow_clone(&target_conn, "cloned_table", &[], None)
    .await?;
```

## Test Plan
- [x] Added unit test that verifies:
  - Data integrity after cloning
  - Schema matching between source and cloned table
  - Version-specific cloning
- [x] Tests pass successfully

## Notes
- Only NativeTable implementation is complete (as requested)
- RemoteTable support can be added in a future PR
- Python bindings need additional work to properly pass connection objects between Rust and Python

🤖 Generated with Claude Code


wkalt:
@jackye1995 I don't quite understand this:
```
Unlike the create_table API that actually creates the table with at least metadata, this API only prepares the database to create an empty table without any metadata and data yet. The key reason for doing so is because the remote database (once implemented) might decide to produce tables in a different location than our current convention of s3://database-location/table-name.lance, and might need special credentials to access the location. 
```

How is that different from what the create_table API also needs to do?

edit: does that mean this won't show up in table listings until it gets written to?

jackye1995:
> How is that different from what the create_table API also needs to do?

Currently create_table always create at least 1 version of the table, either with or without data. You will end up with at least a `_versions` directory with the first manifest file. This `create_empty_table` API will create no metadata at all, you just know this table exists in this database, and the table is set up for write (e.g. you can get the credentials to write to it).

It mainly solves the chicken-and-egg problem that for example here for shallow clone, you need to know the location and how to write to the location before you clone and create the first version of the table. But without creating the table first, you cannot know that (not for the ListingDatabase case, but more for some more complicated remote use cases). If we really want, we can reuse the `create_table` API, but `create_table` API takes an Arrow IPC stream which enforces you must at least know the schema of the table, and here we want something even simpler than that. Also based on some feedback this API is semantically different from the create table experience, thus I decide to just have a separated API for it.

> does that mean this won't show up in table listings until it gets written to?

It should show up, for listing database, we can write something like a `.reserved` empty file in the table directory to mark the table as existing but without any metadata yet. (it does not do that right now, but I am a bit worried it expands the PR too big so if it's okay I will do another PR to fix that experience)

wkalt:
another question - is the user able to read from the cloned table before they write to it? Where is the metadata to enable that stored?

jackye1995:
> is the user able to read from the cloned table before they write to it? Where is the metadata to enable that stored?

No, before the first write, the table is not a valid lance table in the first place.

jackye1995:
@wkalt sorry I did a reply and I realized I totally misunderstood your question

> is the user able to read from the cloned table before they write to it? 

users can immediately read after the clone finishes.

Note that the clone is internally separated into 2 steps - create the empty table and then put the initial manifest version based on the source table manifest. In between these 2 steps, the table is not readable since it's not a lance table.

> Where is the metadata to enable that stored?

so with that above answer, I think the answer here is that there is no metadata needed to enable it.

jackye1995:
This approach has some security concerns, I have created a simpler version first: https://github.com/lancedb/lancedb/pull/2653

PR #2631, opened 2025-09-05
feat: support per-request header override

jackye1995:
## Summary

This PR introduces a `HeaderProvider` which is called for all remote HTTP calls to get the latest headers to inject. This is useful for features like adding the latest auth tokens where the header provider can auto-refresh tokens internally and each request always set the refreshed token.


python/python/lancedb/remote/header.py
```diff
@@ -0,0 +1,268 @@
...
+    Users can implement this interface to provide dynamic headers for various purposes
+    such as authentication (OAuth tokens, API keys), request tracking (correlation IDs),
+    custom metadata, or any other header-based requirements. The provider is called
+    before each request to ensure fresh header values are always used.
+
+    Thread Safety
+    -------------
+    Implementations should be thread-safe as get_headers() may be called
+    concurrently from multiple threads. Use locks or other synchronization
+    primitives if your implementation maintains mutable state.
```
westonpace: This would only be a problem if the header provider called into underlying C routines right?  A pure python header provider doesn't need to worry?

python/python/lancedb/remote/header.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Error Handling
+    --------------
+    If get_headers() raises an exception, the request will fail. Implementations
+    should handle recoverable errors internally (e.g., retry token refresh) and
+    only raise exceptions for unrecoverable errors.
+    Examples
```
westonpace:
```suggestion
    only raise exceptions for unrecoverable errors.
    
    Examples
```

python/python/lancedb/remote/header.py
```diff
@@ -0,0 +1,268 @@
...
+        self._current_token: Optional[str] = None
+        self._token_expires_at: Optional[float] = None
+        self._refresh_lock = threading.Lock()
+
+    def _refresh_token_if_needed(self) -> None:
+        """Refresh the token if it's expired or close to expiring."""
+        with self._refresh_lock:
+            # Check again inside the lock in case another thread refreshed
+            if self._needs_refresh():
+                token_data = self._token_fetcher()
```
westonpace, line 136: Is this going to do blocking I/O?

python/src/header.rs
```diff
@@ -0,0 +1,75 @@
...
+            }
+        })
+    }
+}
+
+#[cfg(feature = "remote")]
+#[async_trait]
+impl lancedb::remote::HeaderProvider for PyHeaderProvider {
+    async fn get_headers(&self) -> lancedb::error::Result<HashMap<String, String>> {
+        self.get_headers_internal()
```
westonpace, line 62: Should we spawn this call on a dedicated blocking task with tokio's spawn_blocking?

python/src/header.rs
```diff
@@ -0,0 +1,75 @@
...
+use pyo3::prelude::*;
+use pyo3::types::PyDict;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+/// A wrapper around a Python HeaderProvider that can be called from Rust
+#[derive(Clone)]
+pub struct PyHeaderProvider {
+    provider: Arc<RwLock<Option<Py<PyAny>>>>,
```
westonpace: Does this need to be an async mutex?  It looks like it only gets set once from the constructor.  Actually, is a mutex needed here at all?  It doesn't look like this is ever modified.

python/src/header.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl PyHeaderProvider {
+    pub fn new(provider: Py<PyAny>) -> Self {
+        Self {
+            provider: Arc::new(RwLock::new(Some(provider))),
+        }
+    }
+
+    /// Get headers from the Python provider (internal implementation)
+    async fn get_headers_internal(&self) -> Result<HashMap<String, String>, String> {
```
westonpace: This method could be sync if you get rid of the async mutex

python/src/header.rs
jackye1995: yes you are right, originally I implemented the python HeaderProvider async, let me remove this

python/python/lancedb/remote/header.py
jackye1995: yes

jackye1995:
looks like there is in general a problem that building the lancedb wheel for doc code testing take a long time, I will publish another PR to fix that, and merge this for a prerelease first.

PR #2636, opened 2025-09-08
feat: expose disableScoringAutoprojection to lancedb

Xuanwo:
Action for https://github.com/lancedb/lancedb/discussions/2629

---

**This PR was primarily authored with Codex using GPT-5 and then hand-reviewed by me. I AM responsible for every change made in this PR. I aimed to keep it aligned with our goals, though I may have missed minor issues. Please flag anything that feels off, I'll fix it quickly.**

### DIFF
--- a/nodejs/lancedb/query.ts
+++ b/nodejs/lancedb/query.ts
@@ -363,6 +363,23 @@ export class StandardQueryBase<
     return this.where(predicate);
   }
 
+  /**
+   * Disable autoprojection of scoring columns.
+   *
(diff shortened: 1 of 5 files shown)


westonpace (changes requested):
Instead of adding a new configuration parameter can we change `lancedb` to just specify a value?

Xuanwo:
> Instead of adding a new configuration parameter can we change `lancedb` to just specify a value?

Let me give it a try.

PR #2638, opened 2025-09-09
feat: support mTLS for remote database

jackye1995:
This PR adds mTLS (mutual TLS) configuration support for the LanceDB remote HTTP client, allowing users to authenticate with client certificates and configure custom CA certificates for server verification.


rust/lancedb/src/remote/client.rs
```diff
@@ -245,6 +261,49 @@ impl RestfulLanceDbClient<Sender> {
         if let Some(timeout) = timeout {
             client_builder = client_builder.timeout(timeout);
         }
+
+        // Configure mTLS if TlsConfig is provided
+        if let Some(tls_config) = &client_config.tls_config {
+            // Load client certificate and key for mTLS
+            if let (Some(cert_file), Some(key_file)) = (&tls_config.cert_file, &tls_config.key_file)
+            {
+                let cert = std::fs::read(cert_file).map_err(|err| Error::Other {
```
wjones127, line 270: You might want to make these async:


```suggestion
                let cert = tokio::fs::read(cert_file).await.map_err(|err| Error::Other {
```

wjones127 (approved):
Looks good. It's not critical, but I would consider using the async variants of `read` (`tokio::fs:read`) in Rust.

rust/lancedb/src/remote/client.rs
jackye1995: the issue is that the whole `try_new` is sync, if we change it to `tokio` it has to propagate all the way up and make all related functions aysnc. Since this is just a 1 time read when creating the client at initialization time, I felt it is okay to just make it sync. @wjones127 what do you think?

jackye1995:
@wjones127 I will merge it first for now to unblock testing, let me know if there is any better approach I can make another PR

PR #2653, opened 2025-09-12
feat: support shallow clone

jackye1995:
Support shallow cloning a dataset at a specific location to create a new dataset, using the shallow_clone feature in Lance. Also introduce remote `clone` API for remote tables for this functionality.

### DIFF
--- a/python/python/lancedb/remote/db.py
+++ b/python/python/lancedb/remote/db.py
@@ -212,6 +212,53 @@ class RemoteDBConnection(DBConnection):
         table = LOOP.run(self._conn.open_table(name, namespace=namespace))
         return RemoteTable(table, self.db_name)
 
+    def clone_table(
+        self,
+        target_table_name: str,
+        source_uri: str,
+        *,
+        target_namespace: List[str] = [],
+        source_version: Optional[int] = None,
+        source_tag: Optional[str] = None,
+        is_shallow: bool = True,
+    ) -> Table:
+        """Clone a table from a source table.
+
+        Parameters
+        ----------
+        target_table_name: str
+            The name of the target table to create.
+        source_uri: str
+            The URI of the source table to clone from.
+        target_namespace: List[str], optional
+            The namespace for the target table.
+            None or empty list represents root namespace.
+        source_version: int, optional
+            The version of the source table to clone.
+        source_tag: str, optional
+            The tag of the source table to clone.
+        is_shallow: bool, default True
+            Whether to perform a shallow clone (True) or deep clone (False).
+            Currently only shallow clone is supported.
+
+        Returns
+        -------
+        A RemoteTable object representing the cloned table.
+        """
+        from .table import RemoteTable
+
+        table = LOOP.run(
+            self._conn.clone_table(
+                target_table_name,
+                source_uri,
+                target_namespace=target_namespace,
+                source_version=source_version,
+                source_tag=source_tag,
+                is_shallow=is_shallow,
+            )
+        )
+        return RemoteTable(table, self.db_name)
+
     @override
     def create_table(
         self,
--- a/python/python/tests/test_db.py
+++ b/python/python/tests/test_db.py
@@ -831,3 +831,119 @@ def test_local_table_operations_with_namespace_raise_error(tmp_path):
     # Test table_names without namespace - should work normally
     tables_root = list(db.table_names())
     assert "test_table" in tables_root
+
+
+def test_clone_table_latest_version(tmp_path):
+    """Test cloning a table with the latest version (default behavior)"""
+    import os
+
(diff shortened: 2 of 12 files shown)


python/python/lancedb/remote/db.py
```diff
@@ -212,6 +212,53 @@ def open_table(
...
+        self,
+        target_table_name: str,
+        source_uri: str,
+        *,
+        target_namespace: List[str] = [],
+        source_version: Optional[int] = None,
+        source_tag: Optional[str] = None,
+        is_shallow: bool = True,
+    ) -> Table:
+        """Clone a table from a source table.
```
jmhsieh, line 225: At the api docs level can you better explain what a shallow clone is?  (is it basically a metadata copy of the table -- new copy of metadata but pointers to original datafiles?)

python/python/tests/test_db.py
```diff
@@ -830,3 +830,119 @@ def test_local_table_operations_with_namespace_raise_error(tmp_path):
...
+    more_data = [{"id": 3, "text": "test", "vector": [5.0, 6.0]}]
+    source_table.add(more_data)
+
+    # Clone the table (should get latest version with 3 rows)
+    source_uri = os.path.join(tmp_path, "source.lance")
+    cloned_table = db.clone_table("cloned", source_uri)
+
+    # Verify cloned table has all 3 rows
+    assert cloned_table.count_rows() == 3
+    assert "cloned" in db.table_names()
```
jmhsieh, line 859: what happens to the clone if I modify the source?  what happens to the source if I modify the clone?

jackye1995: Both evolves independently, see tests in `listing.rs`

python/python/lancedb/remote/db.py
jackye1995: Yes. All data files, deletion files, index files point to the original files. Let me add more info in the docs.

rust/lancedb/src/database/listing.rs
```diff
@@ -678,6 +679,65 @@ impl Database for ListingDatabase {
         }
     }
 
+    async fn clone_table(&self, request: CloneTableRequest) -> Result<Arc<dyn BaseTable>> {
+        if !request.target_namespace.is_empty() {
+            return Err(Error::NotSupported {
+                message: "Namespace parameter is not supported for listing database. Only root namespace is supported.".into(),
```
Xuanwo, line 691: This error message seems to be wrong. What's the relation of this API with `listing database`?

rust/lancedb/src/connection.rs
```diff
@@ -577,6 +633,30 @@ impl Connection {
...
 
+    /// Clone a table in the database
+    ///
+    /// Creates a new table by cloning from an existing source table.
+    /// By default, this performs a shallow clone where the new table shares
+    /// the underlying data files with the source table.
+    ///
+    /// # Parameters
+    /// - `target_table_name`: The name of the new table to create
+    /// - `source_uri`: The URI of the source table to clone from
```
Xuanwo, line 642: Can you elaborate more on why we are accepting a `source_uri` instead of `source_table_name`? It seems other APIs all use `name`.

rust/lancedb/src/database.rs
```diff
@@ -176,6 +176,42 @@ impl CreateTableRequest {
...
+/// both the source and cloned tables to evolve independently while initially
+/// sharing the same data, deletion, and index files.
+#[derive(Clone, Debug)]
+pub struct CloneTableRequest {
+    /// The name of the target table to create
+    pub target_table_name: String,
+    /// The namespace for the target table. Empty list represents root namespace.
+    pub target_namespace: Vec<String>,
+    /// The URI of the source table to clone from.
+    pub source_uri: String,
```
Xuanwo, line 192: Is it a good idea for us to have `source_namespace` and `source_name` to align with other APIs?

Xuanwo (commented):
Mostly LGTM, only have some questions over API.

rust/lancedb/src/database/listing.rs
jackye1995: oh the database struct is "ListingDatabase", that's what it means

rust/lancedb/src/connection.rs
jackye1995: yeah I went back and forth on this. But for the specific use cases we want to achieve, we need to be able to clone across database, so you can clone a table in database 1 to another table in database 2. We basically cannot achieve that since all APIs are within a database today, so my workaround is to expose source dataset through URI, so you can first get the URI and then feed into this API. The alternative is to create a whole new level above database, which seems to be a bit too much.

Xuanwo (approved):

PR #2656, opened 2025-09-15
fix(node): allow undefined/omitted values for nullable vector fields

### LINKED ISSUE #2654: bug(python): nullable struct field fails with AttributeError
### LanceDB version

0.24.3

### What happened?

When a struct field is marked as nullable but you try to insert None for the entire struct,
LanceDB raises: AttributeError: 'pyarrow.lib.DataType' object has no attribute 'fields'

```python

import pyarrow as pa
import lancedb
import tempfile

# Define schema with nullable struct field
my_struct = pa.struct([pa.field("x", pa.float32())])
schema = pa.schema([
    pa.field("id", pa.string()),
    pa.field("data", my_struct).with_nullable(True)  # This is nullable
])

print(f"Field 'data' is nullable: {schema.field('data').nullable}")

with tempfile.TemporaryDirectory() as tmpdir:
    db = lancedb.connect(f"{tmpdir}/test.lancedb")
    table = db.create_table("test", schema=schema)
    
    # This works fine
    print("✓ Inserting record with struct data...")
    table.add([{"id": "1", "data": {"x": 1.0}}])
    
    # This should work (field is nullable) but fails
    print("✗ Inserting record with None data (should work but fails)...")
    table.add([{"id": "2", "data": None}])

"""
STACK TRACE:
File "lancedb/table.py", line 323, in _align_field_types
    field.type.fields,
AttributeError: 'pyarrow.lib.DataType' object has no attribute 'fields'
"""
```

naaa760:
Fixes #2654

**Problem**: When a vector field is marked as nullable, users should be able to omit it or pass `undefined`, but this was throwing an error: "Table has embeddings: 'vector', but no embedding function was provided"

**Solution**: Modified `validateSchemaEmbeddings` to check `field.nullable` before treating `undefined` values as missing embedding fields.

**Changes**:
- Fixed validation logic in `nodejs/lancedb/arrow.ts`
- Enabled previously skipped test for nullable fields
- Added reproduction test case

**Behavior**:
- ✅ `{ vector: undefined }` now works for nullable fields
- ✅ `{}` (omitted field) now works for nullable fields  
- ✅ `{ vector: null }` still works (unchanged)
- ✅ Non-nullable fields still properly throw errors (unchanged)

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -65,7 +65,7 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
 
     it("be displayable", async () => {
       expect(table.display()).toMatch(
-        /NativeTable\(some_table, uri=.*, read_consistency_interval=None\)/,
+        /NativeTable\(some_table, uri=.*, read_consistency_interval=None\)/
       );
       table.close();
       expect(table.display()).toBe("ClosedTable(some_table)");
@@ -195,7 +195,7 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
           new arrow.Field(
             "vector",
             new arrow.FixedSizeList(512, new arrow.Field("item", floatType)),
-            true,
+            true
           ),
         ]);
 
@@ -209,20 +209,19 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
         } catch (e) {
           expect(e).toBeUndefined();
         }
-      },
+      }
     );
 
-    // TODO: https://github.com/lancedb/lancedb/issues/1832
-    it.skip("should be able to omit nullable fields", async () => {
+    it("should be able to omit nullable fields", async () => {
       const db = await connect(tmpDir.name);
       const schema = new arrow.Schema([
         new arrow.Field(
           "vector",
           new arrow.FixedSizeList(
             2,
-            new arrow.Field("item", new arrow.Float64()),
+            new arrow.Field("item", new arrow.Float64())
           ),
-          true,
+          true
         ),
         new arrow.Field("item", new arrow.Utf8(), true),
         new arrow.Field("price", new arrow.Float64(), false),
@@ -272,7 +271,7 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
 
       const data2 = { x: null, id: "bar" };
       await expect(table.add([data2])).rejects.toThrow(
-        "declared as non-nullable but contains null values",
+        "declared as non-nullable but contains null values"
       );
 
       // But we can alter columns to make them nullable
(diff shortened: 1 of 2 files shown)


nodejs/__test__/table.test.ts
```diff
@@ -331,7 +330,38 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
...
+      await table.add([{}]);
+
+      // Should still work for null
+      await table.add([{ vector: null }]);
+
+      // Should still work for actual vector
+      const testVector = new Array(32).fill(0.5);
+      await table.add([{ vector: testVector }]);
+
+      expect(await table.countRows()).toEqual(4);
```
wjones127, line 375: This test would be a lot better if it made an assertion about what the values in the `vector` column actually are at the end. See the test right below for an example of how to assert the values are equal to the expected ones.

nodejs/lancedb/arrow.ts
```diff
@@ -1286,19 +1286,24 @@ function validateSchemaEmbeddings(
...
-          );
-          if (
-            // biome-ignore lint/suspicious/noExplicitAny: we don't know the type of `f`
-            embeddings.find((f: any) => f["vectorColumn"] === field.name) ===
-            undefined
-          ) {
+        // If the field is nullable, allow undefined/omitted values
+        if (field.nullable) {
+          fields.push(field);
+        } else {
```
wjones127: I think this is in the wrong position. If it's nullable, but there's an embedding function registered, we still want to fill in nulls. It's only if there is no embedding function and it's nullable that we want to allow it to be empty. Could you move that conditional?

nehaaprasad:
@wjones127 
please let me know now

nehaaprasad:
@wjones127 
you still have not updated?

nodejs/__test__/table.test.ts
```diff
@@ -331,6 +343,43 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
...
+      const table = await db.createEmptyTable("test_table", schema);
+
+      // Should not throw error for undefined value
+      await table.add([{ id: 0, vector: undefined }]);
+
+      // Should not throw error for omitted field
+      await table.add([{ id: 1 }]);
+
+      // Should still work for null
+      await table.add([{ id: 2, vector: null }]);
```
wjones127, line 370: I found the problem here was that there were no non-null fields passed. So we ended up trying to pass a batch of data with no columns to Rust. That seemed to be interpreted instead as zero rows. I fixed this test by adding another field that isn't null, and it works as expected.

This does mean there is another bug if the user passes a batch with entirely empty objects or undefined/null fields. But that seems like an edge case we can solve in different PR.

Issue #2644, opened 2025-09-10
bug(node): can't pass `null` to nullable boolean field


wjones127:
### LanceDB version

v0.22.0

### What happened?

There seems to be a bug in the Arrow conversion. These should be solved to `null` in the Arrow arrays.

### Are there known steps to reproduce?

```js
const lancedb = require('@lancedb/lancedb');
const arrow = require('apache-arrow');

(async () => {
    const conn = await lancedb.connect("memory://");
    const schema = new arrow.Schema([
        new arrow.Field("test", new arrow.Bool(), true),
    ]);
    const table = await conn.createEmptyTable("test_table", schema);
    await table.add([{ test: null }])
})();
```
```
[Error: lance error: LanceError(Arrow): Arrow error: Invalid argument error: Need at least 1 bytes for bitmap in buffers[0] in array of type Boolean, but got 0, /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-datafusion-0.35.0/src/utils.rs:49:31  Caused by: LanceError(Arrow): Arrow error: Invalid argument error: Need at least 1 bytes for bitmap in buffers[0] in array of type Boolean, but got 0, /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-datafusion-0.35.0/src/utils.rs:49:31] {
  code: 'GenericFailure'
}
```

wjones127:
It looks like this one is an upstream bug: https://github.com/apache/arrow-js/issues/68

PR #2657, opened 2025-09-15
fix(node): handle null values in nullable boolean fields

nehaaprasad:
### Solution
Added special handling in `makeVector` function for boolean arrays where all values are null. The fix creates a proper null bitmap using `makeData` and `arrowMakeVector` instead of relying on Apache Arrow's `vectorFromArray` which doesn't handle this edge case correctly.

fixes: #2644

### Changes
- Added null value detection for boolean types in `makeVector` function
- Creates proper Arrow data structure with null bitmap when all boolean values are null
- Preserves existing behavior for non-null boolean values and other data types

- Fixes the boolean null value bug while maintaining backward compatibility.

### DIFF
--- a/nodejs/__test__/arrow.test.ts
+++ b/nodejs/__test__/arrow.test.ts
@@ -1008,41 +1008,34 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
         expect(result).toEqual(null);
       });
     });
-  },
-);
 
-describe.each([arrow15, arrow16, arrow17, arrow18])(
-  "Boolean null handling",
-  (arrow) => {
-    it("should handle null values in nullable boolean fields", () => {
-      const { makeArrowTable } = require("../lancedb/arrow");
-      const schema = new arrow.Schema([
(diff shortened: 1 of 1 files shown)


nodejs/lancedb/arrow.ts
```diff
@@ -751,6 +751,19 @@ function makeVector(
   // biome-ignore lint/suspicious/noExplicitAny: skip
 ): Vector<any> {
   if (type !== undefined) {
+    if (DataType.isBool(type)) {
```
wjones127, line 755: Could you add a comment saying this is a workaround for this issue in arrow? That way we can one day remove once this is fixed upstream.


```suggestion
    // Workaround for: https://github.com/apache/arrow-js/issues/68
    if (DataType.isBool(type)) {
```

wjones127 (changes requested):
This seems like it should work. Could you add a unit test for it though?

nehaaprasad:
@wjones127 
i have added the workaround comment with GitHub issue link and comprehensive unit tests for boolean null handling as requested.

Issue #2645, opened 2025-09-10
bug(node): passing undefined to nullable field gives coerced values


wjones127:
### LanceDB version

v0.22.0

### What happened?

Would instead expect to resolve to null values. Instead:

* `bool` -> `false`
* `f32` -> `NaN`
* `string` -> `""`

### Are there known steps to reproduce?

```js
const lancedb = require('@lancedb/lancedb');
const arrow = require('apache-arrow');

(async () => {
    const conn = await lancedb.connect("memory://");
    const schema = new arrow.Schema([
        new arrow.Field("text", new arrow.Utf8(), true),
        new arrow.Field("number", new arrow.Float32(), true),
        new arrow.Field("bool", new arrow.Bool(), true),
    ]);
    const table = await conn.createEmptyTable("test_table", schema);
    await table.add([{ text: undefined, number: undefined, bool: undefined }]);
    const result = await table.query().toArray();
    console.log(result);
})();
```
```
[ {"text": "", "number": NaN, "bool": false} ]
```

jeetsingh008:
Can you please assign me this issue, thanks.

nehaaprasad:
@wjones127 
undefined values are now properly converted to null for nullable fields instead of coerced defaults.

PR #2658, opened 2025-09-15
fix: undefined values should become null in nullable fields

nehaaprasad:
### Bug Fix: Undefined Values in Nullable Fields

**Issue**: When inserting data with `undefined` values into nullable fields, LanceDB was incorrectly coercing them to default values (`false` for booleans, `NaN` for numbers, `""` for strings) instead of `null`.

**Fix**: Modified the `makeVector()` function in `arrow.ts` to properly convert `undefined` values to `null` for nullable fields before passing data to Apache Arrow.

fixes: #2645

**Result**: Now `{ text: undefined, number: undefined, bool: undefined }` correctly becomes `{ text: null, number: null, bool: null }` when fields are marked as nullable in the schema.

**Files Changed**: 
- `nodejs/lancedb/arrow.ts` (core fix)
- `nodejs/__test__/arrow.test.ts` (test coverage)

- This ensures proper null handling for nullable fields as expected by users.

### DIFF
--- a/nodejs/__test__/arrow.test.ts
+++ b/nodejs/__test__/arrow.test.ts
@@ -1042,23 +1042,6 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
 
 // Test for the undefined values bug fix
 describe("undefined values handling", () => {
-  it("should convert undefined values to null for nullable fields", () => {
-    const schema = new Schema([
-      new Field("text", new Utf8(), true), // nullable
-      new Field("number", new Int32(), true), // nullable
-      new Field("bool", new Bool(), true), // nullable
-    ]);
-
-    const data = [{ text: undefined, number: undefined, bool: undefined }];
-    const table = makeArrowTable(data, { schema });
-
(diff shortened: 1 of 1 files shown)


nodejs/__test__/arrow.test.ts
```diff
@@ -1008,5 +1008,52 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
         expect(result).toEqual(null);
       });
     });
-  },
+  }
 );
+
+// Test for the undefined values bug fix
+describe("undefined values handling", () => {
```
wjones127: These are great tests!

wjones127 (commented):
Thanks for doing this. It looks like something went wrong with the formatter or linter and there are lot of irrelevant changes as well as a CI failure. Happy to merge once that is fixed.

nehaaprasad:
@wjones127 
fixed removed all formatting changes and CI issues. 
PR now contains only the essential bug fix with clean commits.

PR #2666, opened 2025-09-19
feat: add support for test_remote_connections

cmccabe:
Add a new test feature which allows for running the lancedb tests against a remote server. Convert over a few tests in src/connection.rs as a proof of concept.

To make local development easier, the remote tests can be run locally from a Makefile. This file can also be used to run the feature tests, with a single invocation of 'make'. (The feature tests require bringing up a docker compose environment.)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -96,6 +96,7 @@ jobs:
       # Need up-to-date compilers for kernels
       CC: clang-18
       CXX: clang++-18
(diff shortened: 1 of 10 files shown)


cmccabe:
failing test is `test_concurrent_dynamodb_commit` which I don't think is related.

cmccabe:
I have revised this so that it runs in the lancedb CI, before every commit. This will help us avoid compatibility breaks.

PR #2670, opened 2025-09-22
chore: increase pypi publish timeout and use warp runner for arm64

jackye1995:
Fix failures like: https://github.com/lancedb/lancedb/actions/runs/17840462235/job/50748940233

ARM64 build cannot succeed within 1 hour, x86-64 build sometimes cannot succeed within 1 hour.

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -56,7 +56,7 @@ jobs:
           pypi_token: ${{ secrets.LANCEDB_PYPI_API_TOKEN }}
           fury_token: ${{ secrets.FURY_TOKEN }}
   mac:
-    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-publish.yml
```diff
@@ -56,15 +56,15 @@ jobs:
...
           fury_token: ${{ secrets.FURY_TOKEN }}
   mac:
-    timeout-minutes: 60
+    timeout-minutes: 90
     runs-on: ${{ matrix.config.runner }}
     strategy:
       matrix:
         config:
           - target: x86_64-apple-darwin
             runner: macos-13
```
jackye1995, line 65: warp does not support x86-64 for mac, so we have to stick with the default runner...

Xuanwo (approved):

PR #2675, opened 2025-09-23
feat: merge insert `explain_plan` and `analyze_plan`

wjones127:
Adds new `explain_plan` and `analyze_plan` methods for merge insert builder, allowing users to inspect the plan and performance of the operation. This isn't currently supported for all plans, but support will expand to all merge insert queries eventually.

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -1212,9 +1212,9 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
         let (request_id, response) = self.send_streaming(request, empty_reader, false).await?;
 
         let response = self.check_table_response(&request_id, response).await?;
-        let body = response.text().await.err_to_http(request_id)?;
+        let plan_response: PlanResponse = response.json().await.err_to_http(request_id)?;
 
-        Ok(body)
+        Ok(plan_response.plan)
     }
 
     async fn merge_insert_analyze_plan(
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -1178,6 +1178,70 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+            ))
+            .query(&query)
+            .header(CONTENT_TYPE, ARROW_STREAM_CONTENT_TYPE);
+
+        let (request_id, response) = self.send_streaming(request, empty_reader, false).await?;
+
+        let response = self.check_table_response(&request_id, response).await?;
+        let body = response.text().await.err_to_http(request_id)?;
+
+        Ok(body)
```
wjones127: Instead of expecting just literal text, let's expect JSON response with a string field `plan` that has the plan text.

rust/lancedb/src/remote/table.rs
```diff
@@ -1178,6 +1178,70 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+            ))
+            .query(&query)
+            .header(CONTENT_TYPE, ARROW_STREAM_CONTENT_TYPE);
+
+        let (request_id, response) = self.send_streaming(request, new_data, false).await?;
+
+        let response = self.check_table_response(&request_id, response).await?;
+        let body = response.text().await.err_to_http(request_id)?;
+
+        Ok(body)
```
wjones127: Same thing here:  let's expect JSON response with a string field `plan` that has the plan text.

PR #2691, opened 2025-09-30
fix: inflated release size due to lance-namespace transitive dependency

jackye1995:
Fixed the issue on lance-namespace side to avoid pinning to a specific lance version. This should fix the issue of the increased release artifact size and build time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,14 +15,14 @@ categories = ["database-implementations"]
 rust-version = "1.78.0"
 
 [workspace.dependencies]
-lance = { "version" = "=0.37.0", default-features = false, "features" = ["dynamodb"] }
-lance-io = { "version" = "=0.37.0", default-features = false }
-lance-index = { "version" = "=0.37.0" }
-lance-linalg = { "version" = "=0.37.0" }
-lance-table = { "version" = "=0.37.0" }
-lance-testing = { "version" = "=0.37.0" }
-lance-datafusion = { "version" = "=0.37.0" }
-lance-encoding = { "version" = "=0.37.0" }
+lance = { "version" = "=0.37.0", default-features = false, "features" = ["dynamodb"], "tag" = "v0.37.1-beta.1", "git" = "https://github.com/lancedb/lance.git" }
+lance-io = { "version" = "=0.37.0", default-features = false, "tag" = "v0.37.1-beta.1", "git" = "https://github.com/lancedb/lance.git" }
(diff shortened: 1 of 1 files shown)


wjones127 (commented):
This script needs to be updated to handle `lance-namespace`: https://github.com/lancedb/lancedb/blob/main/ci/set_lance_version.py

Cargo.toml
```diff
@@ -52,15 +51,23 @@ pin-project = "1.0.7"
...
 # https://github.com/apache/arrow-rs/commit/2fddf85afcd20110ce783ed5b4cdeb82293da30b
 chrono = "=0.4.41"
-# https://github.com/RustCrypto/formats/issues/1684
-base64ct = "=1.6.0"
 # Workaround for: https://github.com/Lokathor/bytemuck/issues/306
 bytemuck_derive = ">=1.8.1, <1.9.0"
+
+[patch.crates-io]
+# Force to use the same lance version as the rest of the project to avoid duplicate dependencies
+lance = { "version" = "=0.37.0", "tag" = "v0.37.1-beta.1", "git" = "https://github.com/lancedb/lance.git" }
```
jackye1995, line 66: the existing line updater should be able to handle this correctly

Cargo.toml
```diff
@@ -23,15 +23,14 @@ lance-table = { "version" = "=0.37.0", "tag" = "v0.37.1-beta.1", "git" = "https:
...
-lance-namespace = "0.0.15"
+lance-namespace = "0.0.16"
 # Note that this one does not include pyarrow
 arrow = { version = "55.1", optional = false }
 arrow-array = "55.1"
 arrow-data = "55.1"
 arrow-ipc = "55.1"
 arrow-ord = "55.1"
 arrow-schema = "55.1"
-arrow-arith = "55.1"
```
jackye1995, line 34: a few unused dependencies automatically removed as a part of cargo update

Xuanwo:
The only place we need `lance` is the `CreateTable` call. Is it possible to make it purely rest call so we don't need it from namespace?

jackye1995:
> The only place we need lance is the CreateTable call. Is it possible to make it purely rest call so we don't need it from namespace?

I think that's only because it only implements create table for now, there are quite a few interfaces to be implemented that will require actually using the lance dependency.

Xuanwo:
> I think that's only because it only implements create table for now, there are quite a few interfaces to be implemented that will require actually using the lance dependency.

Ok, I got it.

jackye1995:
Known CI failure issue, merging to unblock release

PR #2692, opened 2025-09-30
fix: have CI download from ci-support-binaries

cmccabe:
Have CI download from ci-support-binaries to fix the build.


jackye1995 (approved):
looks good to me

PR #2700, opened 2025-10-06
ci: fix Python and Node CI on main

wjones127:
Example failure: https://github.com/lancedb/lancedb/actions/runs/18237024283/job/51932651993

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -1398,19 +1398,19 @@ impl Tags for NativeTags {
     }
 
     async fn create(&mut self, tag: &str, version: u64) -> Result<()> {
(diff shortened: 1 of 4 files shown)


rust/lancedb/src/table.rs
```diff
@@ -1398,19 +1398,19 @@ impl Tags for NativeTags {
     }
 
     async fn create(&mut self, tag: &str, version: u64) -> Result<()> {
-        let dataset = self.dataset.get_mut().await?;
+        let dataset = self.dataset.get().await?;
```
wjones127, line 1401: Tags don't mutate the table history itself, so we don't need to require that you have the latest table version checked out to create or modify tags.

westonpace (approved):

PR #2704, opened 2025-10-07
fix(node): Support specifying arrow field types by name

tlamarre91:
The [`FieldLike` type in arrow.ts](https://github.com/lancedb/lancedb/blob/5ec12c99713e28a8fa550547cc9e7536da063f6e/nodejs/lancedb/arrow.ts#L71-L78) can have a `type: string` property, but before this change, actually trying to create a table that has a schema that specifies field types by name results in an error:

```
Error: Expected a Type but object was null/undefined
```

This change adds support for mapping some type name strings to arrow `DataType`s. The type names that can be passed are upper/lowercase variations of the keys of the `constructorsByTypeName` object.

With this, it is possible to create empty tables from `SchemaLike` objects without instantiating arrow types, e.g.:

```
    import { SchemaLike } from "../lancedb/arrow"
    // ...
    const schemaLike = {
      fields: [
        {
          name: "id",
          type: "int64",
          nullable: true,
        },
        {
          name: "vector",
          type: "float64",
          nullable: true,
        },
      ],
    // ...
    } satisfies SchemaLike;
    const table = await con.createEmptyTable("test", schemaLike);
 ```

This change also makes `FieldLike.nullable` required since the `sanitizeField` function throws if it is undefined.

### CODE REVIEW

**Comment by tlamarre91** [comment]
Here is a simple project that reproduces the error when running `node index.mjs`: https://github.com/tlamarre91/lancedb-type-conversion-test

With this patch applied to `@lancedb/lancedb`, the script is able to create a table using a schema that specifies field types by name.

**Comment by tlamarre91** [comment]
I should also mention that this does not attempt to add mappings for the following Arrow types (updated PR description to mention): 
* Int
* Float
* Decimal
* Date
* Time
* Timestamp
* Interval
* List
* Struct
* Union
* FixedSizeBinary
* FixedSizeList
* Map
* Duration
* Dictionary
* TimestampNanosecond
* TimestampMicrosecond
* TimestampMillisecond
* TimestampSecond
* DenseUnion
* SparseUnion

That is because those types are expected to be passed in with properties like `bitWidth`, `timezone`, etc. and I don't see an obvious good way to support mapping directly to those types from strings.

File: nodejs/lancedb/sanitize.ts [hunk 1]
```diff
@@ -581,3 +592,46 @@ function sanitizeData(
     },
   );
 }
+
+const constructorsByTypeName = {
```
westonpace, line 596: What about the parameterized types like list, decimal, and struct?  I don't honestly remember how those are handled in node's arrow impl.

westonpace (approved):
Looks like a nice utility, thanks!

westonpace:
> This change also makes FieldLike.nullable required since the sanitizeField function throws if it is undefined.

I appreciate this catch.  We could also add a default in `sanitizeField`.  This would match python better (`nullable` defaults to true) but I don't feel too strongly about it either way.  Also, I think when we do field inference in typescript it actually infers `nullable` to be false so maybe we'd need to be consistent within the impl :thinking:

nodejs/lancedb/sanitize.ts
tlamarre91: Since those need parameters like bitWidth etc., they're just not supported by this :/ 

I'll keep this in mind... Maybe there could be a property on FieldLike for type parameters that should be passed to the constructor of the arrow DataType.

Issue #2703, opened 2025-10-06
bug(python): `client.create_namespace(namespace=...)` fails with NotImplemented error.


mkleinbort-wl:
### LanceDB version

0.25.1

### What happened?

Running 

```python
client.create_namespace(...)
```

results in 

```
NotImplementedError: LanceDBError: not supported: Namespace operations are not supported for listing database
```

### Are there known steps to reproduce?

```python

import lancedb
client = lancedb.connect('lancedb.db')
client.create_namespace(namespace=['test'])
```

jackye1995:
Thank you for trying this out, yeah this is indeed not implemented yet. I just added the APIs, but the implementation for local mode is still work in progress. 

Currently you need to use a Lance Namespace by connecting through `connect_namespace` and choose one of the implementations listed in https://lancedb.github.io/lance/format/namespace/impls/. Those (except for the directory namespace which is the same as the listing database) will be able to perform namespace operations.

for listing database, it is just a directory of tables so there is no namespace. We could just add folders after folders, but in general the folder-based approach is limiting since there are a lot of features like renaming, namespace properties, etc. that cannot be easily supported. I am working on a manifest-based solution that should better support the local mode.

mkleinbort-wl:
Makes sense - I'll close the issue as it's essentially a work in progress not a bug.

Issue #2701, opened 2025-10-06
Improve multi-level namespace tests


jackye1995:
Today we either have a REST namespace that is a client-side only thing without easily unit-testable server implementation, or a directory namespace (1 level). I am working on an implementation for a storage-based multi-level namespace, will bring it in so that we can more thoroughly test the namespace integration.

PR #2708, opened 2025-10-09
feat!: manifest-based listing database backed by Lance table

jackye1995:
Related to #2701 #2703

This PR is backwards compatible, but we mark it as breaking since it introduces new behavior that users need to be fully aware of.

This PR introduces manifest-based listing database, where the manifest is backed by a Lance table. After this change,  the `ListingDatabase` supports two modes of operation:

### 1. Directory-based Listing (Legacy)

Tables are discovered by listing directories with `.lance` extension in the object store. This is the original behavior and provides basic table management without namespace support.

**Limitations:**
- No namespace support (only a flat list of tables)
- No table renaming
- Requires scanning directories for each list operation
- Limited ability to offer more advanced database level features
- Layout not optimized for object store
- Table being created/deleted in progress is visible to readers
- ...

### 2. Manifest-based Listing (Recommended)

Tables and namespaces are tracked in a special Lance table called `__manifest` (without .lance extension). This enables advanced features and better performance.

**Features:**
- Full namespace support with hierarchical organization
- Table renaming capability
- Efficient indexed queries for listing operations
- Stores additional metadata (location, base objects, etc.), set up to support more objects like view and MV
- Creation and Deletion isolation from readers
- Layout optimized for object store

### Reading Old Databases

When `manifest_enabled=true` and `dir_listing_enabled=true` (migration mode), the database provides backwards compatibility with existing directory-based databases:

- **List operations** merge results from both manifest and directory listing
- **Open table** falls back to directory listing if table not found in manifest
- **Drop table** falls back to directory listing if table not found in manifest
- New tables are created in the manifest automatically

This allows gradual migration: old tables remain accessible while new tables use the manifest.

### Writing with Old Clients

**WARNING:** Old clients (without manifest support) can still write to the database, but:

- Tables created by old clients will NOT be registered in the manifest
- Migration mode (`dir_listing_enabled=true`) will discover these tables via directory listing
- Use the `migrate()` method to register these tables into the manifest

**Impact of mixing old and new clients:**
- Tables may appear inconsistently depending on which client created them
- Namespace operations will not see tables created by old clients
- After running `migrate()`, all tables will be consistent

### Reading New Databases with Old Clients

Old clients (without manifest support) have limited visibility when reading databases created by new clients:

**What old clients CAN see:**
- Tables in the root namespace with `.lance` extension in their location (via directory listing)

**What old clients CANNOT see:**
- Tables with hash-based locations (not discoverable via directory listing)
- Tables in nested namespaces (old clients only scan root namespace)
- Tables that have been renamed (location may no longer match discoverable name)
- The `__manifest` table itself (but it won't interfere with other operations)

**Recommendation:** If you need old client compatibility, ensure `parent_dir_listing_enabled=true` when creating the manifest database. This ensures new tables use `.lance` extension and remain discoverable via directory listing. Avoid using namespaces and rename operations if old clients need access.

### Phase 1: Enable Manifest (Migration Mode)

**This happens automatically when upgrading to the latest LanceDB version.**

The default configuration enables both manifest and directory listing. After upgrading, simply connect to your existing database:

```rust
// Connect to existing database - migration mode is enabled by default
let db = ListingDatabase::connect_with_options(&request).await?;

// Both old and new tables are accessible
let tables = db.table_names(Default::default()).await?;
```

**What happens:**
- Both old (directory-based) and new (manifest-based) tables are accessible
- New tables are automatically registered in manifest
- Old tables remain in directory-only mode until explicitly migrated
- No data is moved or modified

### Phase 2: Run Migration

Once you've verified that Phase 1 is working (both old and new tables are accessible), run the migration to register all existing tables in the manifest:

```rust
// Connect to the database
let db = ListingDatabase::connect_with_options(&request).await?;

// Run migration to register all existing tables in the manifest
let migrated_count = db.migrate().await?;
println!("Migrated {} tables to manifest", migrated_count);
```

**What happens:**
- Scans directory for unmigrated tables
- Registers each table in the manifest with its current location
- Safe to run multiple times (skips already-migrated tables)
- Tables remain in their original locations with `.lance` extension
- No data is moved or modified

### Phase 3: Disable Directory Fallback (Optional)

After confirming all tables are migrated and no old clients need directory fallback, you can improve performance by disabling directory listing:

```rust
use std::collections::HashMap;
use lancedb::database::listing::OPT_DIR_LISTING_ENABLED;

let mut options = HashMap::new();
options.insert(OPT_DIR_LISTING_ENABLED.to_string(), "false".to_string());

let request = ConnectRequest {
    uri: "/tmp/db".to_string(),
    options,
    read_consistency_interval: None,
    session: None,
};

// Connect with directory listing disabled
let db = ListingDatabase::connect_with_options(&request).await?;

// Now only the manifest is consulted for all operations
let tables = db.table_names(Default::default()).await?;
```

**What happens:**
- Only manifest is consulted for all operations (no directory scanning)
- Better performance for listing operations
- Old clients can still read existing tables but cannot discover them via listing
- Tables created by old clients (without manifest) will not be visible

## Configuration Options

- `manifest_enabled`: Enable manifest-based listing (default: true)
- `dir_listing_enabled`: Enable directory listing fallback (default: true)
- `manifest_inline_optimization_enabled`: Auto-optimize manifest after updates (default: true)
- `manifest_read_consistency_interval`: How often to refresh manifest (default: None)

## Best Practices

1. **New databases:** Use defaults (`manifest_enabled=true`, `dir_listing_enabled=true`)
2. **Migrating existing databases:**
   - Phase 1: Enable manifest with fallback
   - Phase 2: Run `migrate()`
   - Phase 3: Disable fallback after confirming migration
3. **Mixed client environments:** Keep `dir_listing_enabled=true` and `parent_dir_listing_enabled=true`
4. **Pure manifest environments:** Set `dir_listing_enabled=false` for best performance

### DIFF
--- a/python/python/tests/test_db.py
+++ b/python/python/tests/test_db.py
@@ -89,7 +89,7 @@ def test_ingest_pd(tmp_path):
     assert db.open_table("test").name == db["test"].name
 
 
-def test_ingest_iterator(mem_db: lancedb.DBConnection):
+def test_ingest_iterator(tmp_path):
     class PydanticSchema(LanceModel):
         vector: Vector(2)
         item: str
@@ -146,8 +146,8 @@ def test_ingest_iterator(mem_db: lancedb.DBConnection):
                 # name constraints
             ]
 
-    def run_tests(schema):
-        tbl = mem_db.create_table("table2", make_batches(), schema=schema)
+    def run_tests(db, schema):
+        tbl = db.create_table("table2", make_batches(), schema=schema)
         tbl.to_pandas()
         assert tbl.search([3.1, 4.1]).limit(1).to_pandas()["_distance"][0] == 0.0
         assert tbl.search([5.9, 26.5]).limit(1).to_pandas()["_distance"][0] == 0.0
@@ -156,10 +156,11 @@ def test_ingest_iterator(mem_db: lancedb.DBConnection):
         assert tbl_len == 50
         assert len(tbl) == tbl_len * 2
         assert len(tbl.list_versions()) == 2
-        mem_db.drop_database()
+        db.drop_all_tables()
 
-    run_tests(arrow_schema)
-    run_tests(PydanticSchema)
+    db = lancedb.connect(tmp_path)
+    run_tests(db, arrow_schema)
+    run_tests(db, PydanticSchema)
 
 
 def test_table_names(tmp_db: lancedb.DBConnection):
@@ -564,21 +565,21 @@ def test_drop_database(tmp_db: lancedb.DBConnection):
     assert tmp_db.table_names() == ["test"]
 
     tmp_db.create_table("new_test", data=new_data)
-    tmp_db.drop_database()
+    tmp_db.drop_all_tables()
     assert tmp_db.table_names() == []
 
     # it should pass when no tables are present
     tmp_db.create_table("test", data=new_data)
     tmp_db.drop_table("test")
     assert tmp_db.table_names() == []
-    tmp_db.drop_database()
+    tmp_db.drop_all_tables()
     assert tmp_db.table_names() == []
 
     # creating an empty database with schema
     schema = pa.schema([pa.field("vector", pa.list_(pa.float32(), list_size=2))])
     tmp_db.create_table("empty_table", schema=schema)
     # dropping a empty database should pass
-    tmp_db.drop_database()
+    tmp_db.drop_all_tables()
     assert tmp_db.table_names() == []
 
 
@@ -742,95 +743,84 @@ def test_bypass_vector_index_sync(tmp_db: lancedb.DBConnection):
     assert "KNN" in plan_without_index
 
 
-def test_local_namespace_operations(tmp_path):
-    """Test that local mode namespace operations behave as expected."""
-    # Create a local database connection
+def test_namespace_create_and_list(tmp_path):
+    """Test creating and listing namespaces with manifest enabled."""
     db = lancedb.connect(tmp_path)
 
-    # Test list_namespaces returns empty list for root namespace
+    # Initially no namespaces
     namespaces = list(db.list_namespaces())
     assert namespaces == []
 
-    # Test list_namespaces with non-empty namespace raises NotImplementedError
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace operations are not supported for listing database",
-    ):
-        list(db.list_namespaces(namespace=["test"]))
+    # Create a namespace
+    db.create_namespace(["ns1"])
+    namespaces = list(db.list_namespaces())
+    assert namespaces == ["ns1"]
+
+    # Create nested namespace
+    db.create_namespace(["ns1", "nested"])
+    namespaces = list(db.list_namespaces(namespace=["ns1"]))
+    assert namespaces == ["nested"]
 
 
-def test_local_create_namespace_not_supported(tmp_path):
-    """Test that create_namespace is not supported in local mode."""
+def test_namespace_table_operations(tmp_path):
+    """Test table operations with namespaces."""
     db = lancedb.connect(tmp_path)
-
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace operations are not supported for listing database",
-    ):
-        db.create_namespace(["test_namespace"])
-
-
-def test_local_drop_namespace_not_supported(tmp_path):
-    """Test that drop_namespace is not supported in local mode."""
-    db = lancedb.connect(tmp_path)
-
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace operations are not supported for listing database",
-    ):
-        db.drop_namespace(["test_namespace"])
-
-
-def test_local_table_operations_with_namespace_raise_error(tmp_path):
-    """
-    Test that table operations with namespace parameter
-    raise ValueError in local mode.
-    """
-    db = lancedb.connect(tmp_path)
-
-    # Create some test data
     data = [{"vector": [1.0, 2.0], "item": "test"}]
     schema = pa.schema(
         [pa.field("vector", pa.list_(pa.float32(), 2)), pa.field("item", pa.string())]
     )
 
-    # Test create_table with namespace - should raise ValueError
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace parameter is not supported for listing database",
-    ):
-        db.create_table(
-            "test_table_with_ns", data=data, schema=schema, namespace=["test_ns"]
-        )
+    # Create namespace
+    db.create_namespace(["test_ns"])
 
-    # Create table normally for other tests
-    db.create_table("test_table", data=data, schema=schema)
-    assert "test_table" in db.table_names()
+    # Create table in namespace
+    table = db.create_table(
+        "test_table", data=data, schema=schema, namespace=["test_ns"]
+    )
+    assert table.name == "test_table"
 
-    # Test open_table with namespace - should raise ValueError
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace parameter is not supported for listing database",
-    ):
-        db.open_table("test_table", namespace=["test_ns"])
+    # List tables in namespace
+    tables = list(db.table_names(namespace=["test_ns"]))
+    assert "test_table" in tables
 
-    # Test table_names with namespace - should raise ValueError
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace parameter is not supported for listing database",
-    ):
-        list(db.table_names(namespace=["test_ns"]))
+    # List tables in root namespace (should be empty)
+    root_tables = list(db.table_names())
+    assert "test_table" not in root_tables
 
-    # Test drop_table with namespace - should raise ValueError
-    with pytest.raises(
-        NotImplementedError,
-        match="Namespace parameter is not supported for listing database",
-    ):
-        db.drop_table("test_table", namespace=["test_ns"])
+    # Open table from namespace
+    opened_table = db.open_table("test_table", namespace=["test_ns"])
+    assert opened_table.count_rows() == 1
 
-    # Test table_names without namespace - should work normally
-    tables_root = list(db.table_names())
-    assert "test_table" in tables_root
+    # Drop table from namespace
+    db.drop_table("test_table", namespace=["test_ns"])
+    tables_after_drop = list(db.table_names(namespace=["test_ns"]))
+    assert "test_table" not in tables_after_drop
+
+
+def test_namespace_drop(tmp_path):
+    """Test dropping namespaces with manifest enabled."""
+    db = lancedb.connect(tmp_path)
+
+    # Create some namespaces
+    db.create_namespace(["ns1"])
+    db.create_namespace(["ns2"])
+    db.create_namespace(["ns1", "nested"])
+
+    # Verify they exist
+    namespaces = list(db.list_namespaces())
+    assert set(namespaces) == {"ns1", "ns2"}
+    nested_namespaces = list(db.list_namespaces(namespace=["ns1"]))
+    assert nested_namespaces == ["nested"]
+
+    # Drop nested namespace
+    db.drop_namespace(["ns1", "nested"])
+    nested_namespaces = list(db.list_namespaces(namespace=["ns1"]))
+    assert nested_namespaces == []
+
+    # Drop root namespace
+    db.drop_namespace(["ns2"])
+    namespaces = list(db.list_namespaces())
+    assert namespaces == ["ns1"]
 
 
 def test_clone_table_latest_version(tmp_path):
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -1,20 +1,22 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
-//! Provides the `ListingDatabase`, a simple database where tables are folders in a directory
-
-use std::fs::create_dir_all;
-use std::path::Path;
-use std::{collections::HashMap, sync::Arc};
+//! Provides the `ListingDatabase`, a simple storage-only database
+pub mod manifest;
 
 use lance::dataset::refs::Ref;
 use lance::dataset::{builder::DatasetBuilder, ReadParams, WriteMode};
 use lance::io::{ObjectStore, ObjectStoreParams, WrappingObjectStore};
+use lance::session::Session;
 use lance_datafusion::utils::StreamingWriteSource;
 use lance_encoding::version::LanceFileVersion;
 use lance_table::io::commit::commit_handler_from_url;
 use object_store::local::LocalFileSystem;
 use snafu::ResultExt;
+use std::collections::HashSet;
+use std::fs::create_dir_all;
+use std::path::Path;
+use std::{collections::HashMap, sync::Arc};
 
 use crate::connection::ConnectRequest;
 use crate::database::ReadConsistency;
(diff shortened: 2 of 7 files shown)


python/python/tests/test_db.py
```diff
@@ -797,7 +838,7 @@ def test_local_table_operations_with_namespace_raise_error(tmp_path):
     # Test create_table with namespace - should raise ValueError
     with pytest.raises(
         NotImplementedError,
-        match="Namespace parameter is not supported for listing database",
+        match="Namespace parameter is not supported",
```
westonpace: Why do these still fail?  Is this a todo?

rust/lancedb/src/database/listing.rs
```diff
@@ -179,29 +309,271 @@ impl ListingDatabaseOptionsBuilder {
...
+    ///
+    /// This controls how often the manifest checks for updates.
+    ///
+    /// # Arguments
+    ///
+    /// * `interval` - The read consistency interval
+    pub fn manifest_read_consistency_interval(mut self, interval: std::time::Duration) -> Self {
+        self.options.manifest_config.read_consistency_interval = Some(interval);
+        self
+    }
```
westonpace, line 335: What's the default here?

rust/lancedb/src/database/listing.rs
```diff
@@ -615,66 +1152,132 @@ impl Database for ListingDatabase {
...
 
-    async fn create_namespace(&self, _request: CreateNamespaceRequest) -> Result<()> {
+    async fn list_namespaces(&self, request: ListNamespacesRequest) -> Result<Vec<String>> {
+        if self.manifest_enabled {
+            return self.manifest_db.list_namespaces(request).await;
+        }
+
+        if !request.namespace.is_empty() {
+            return Err(Error::NotSupported {
+                message: "Namespace operations are not supported for listing database".into(),
```
westonpace:
```suggestion
                message: "Namespace operations are not supported for listing database unless manifest_enabled is true".into(),
```

westonpace: Pretty minor nit, but if you fix it here there are several other places it needs to be fixed as well

rust/lancedb/src/database/listing.rs
```diff
@@ -615,66 +1152,132 @@ impl Database for ListingDatabase {
...
+                if !manifest_locations.contains(&dir_name) {
+                    table_names.push(table_name);
+                }
+            }
+
+            table_names
+        } else if self.manifest_enabled {
+            let mut manifest_request = request.clone();
+            manifest_request.limit = None;
+            manifest_request.start_after = None;
```
westonpace: Do we need to blank these out in this case?  Would it be more efficient to keep them here and skip applying pagination below?  Admittedly it's probably a very minor diff but couldn't you use the limit / offset in the underlying query to the lance dataset?

rust/lancedb/src/database/listing.rs
```diff
@@ -615,66 +1152,132 @@ impl Database for ListingDatabase {
...
 
     async fn create_table(&self, request: CreateTableRequest) -> Result<Arc<dyn BaseTable>> {
+        if self.manifest_enabled {
+            return match self.manifest_db.create_table(request).await {
+                Ok(table) => Ok(table),
+                Err(Error::Manifest { table, message }) => {
+                    return if let Some(table) = table {
+                        log::warn!("Failure in manifest operation: {message}");
+                        Ok(table)
+                    } else {
```
westonpace: Can you comment what is going on here?  Why would the manifest_db return an error but we return ok?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+use std::cmp::PartialEq;
+use std::collections::HashMap;
+use std::fmt::{Display, Formatter};
+use std::hash::{DefaultHasher, Hash, Hasher};
+use std::sync::Arc;
+use std::time::Duration;
+
+/// Configuration for the manifest listing database
+#[derive(Debug, Clone)]
+pub struct ManifestListingDatabaseConfig {
```
westonpace: Minor nit but I wonder if we should just call it a `ManifestDatabase` instead of `ManifestListingDatabase`?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+        Self {
+            read_consistency_interval: None,
+            inline_optimization_enabled: true,
+            parent_dir_listing_enabled: true, // Default to true for backward compatibility
+        }
+    }
+}
+
+/// Delimiter used in object IDs to separate namespace components and table names
+const DELIMITER: &str = "$";
```
westonpace, line 64: ~~Is this a reserved character that we shouldn't use in table names now?~~

Nevermind, I see it was already a reserved character for tables names and namespaces!

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+
+    fn build_object_dir_name(
+        &self,
+        name: &str,
+        namespace: &[String],
+        object_type: ObjectType,
+        object_id: &str,
+    ) -> String {
+        if namespace.is_empty()
+            && self.manifest_config.parent_dir_listing_enabled
```
westonpace, line 285: I wonder if we want this to be just a configurable option `use_simple_naming_scheme` or something?  Some advanced users like to poke around at object storage themselves and having the folder names map to the table names can make it more obvious what is going on.

Though I suppose the folder name wouldn't track after rename, etc. and there would be duplicates issues with namespaces so maybe not worth the footgun.

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                e
+            );
+        }
+
+        if let Err(e) = Self::create_manifest_index_base_objects_label_list(&mut dataset).await {
+            log::warn!(
+                "Failed to create LabelList index on base_objects: {}. Query performance may be impacted.",
+                e
+            );
+        }
```
westonpace: Should we just fail the operation here instead of logging and continuing?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                true,
+            )
+            .await
+    }
+
+    async fn ensure_manifest_table_exists(&self) -> Result<()> {
+        let dataset_exists = {
+            let guard = self.manifest_dataset.lock().unwrap();
+            guard.is_some()
+        };
```
westonpace:
```suggestion
        let dataset_exists = self.manifest_dataset.lock().unwrap().is_some();
```

Would this also work?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                    return Ok(());
+                }
+                Err(e) => return Err(e),
+            }
+        }
+
+        let dataset_opt = self.manifest_dataset.lock().unwrap().clone();
+        let dataset = dataset_opt.ok_or_else(|| Error::Runtime {
+            message: "Manifest table not initialized".to_string(),
+        })?;
```
westonpace: Super minor nit: this seems unreachable, can probably just use `expect_ok` here.

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                &["base_objects"],
+                IndexType::LabelList,
+                Some("base_objects_label_list".to_string()),
+                &label_list_params,
+                true,
+            )
+            .await
+    }
+
+    async fn ensure_manifest_table_exists(&self) -> Result<()> {
```
westonpace: Not super important but why don't we just initialize the manifest dataset when the database is created?  It seems we could avoid the mutex in that case.

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+        }
+
+        Ok(())
+    }
+
+    async fn query_manifest(
+        &self,
+        filter: Option<&str>,
+        columns: Option<Vec<String>>,
+        limit: Option<usize>,
```
westonpace: Strange to see `limit` without `offset`?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+
+        Ok(stream
+            .map(|result| result.map_err(|e| Error::Lance { source: e }))
+            .boxed())
+    }
+
+    async fn manifest_contains_object(&self, object_id: &str) -> Result<bool> {
+        let filter = format!("object_id = '{}'", object_id);
+        let mut stream = self
+            .query_manifest(Some(&filter), Some(vec!["object_id".to_string()]), None)
```
westonpace: Why return `object_id`?  Couldn't you just return zero columns?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                None,
+            )
+            .await?;
+
+        let mut found_result: Option<TableInfo> = None;
+        let mut total_rows = 0;
+
+        while let Some(batch) = stream.next().await {
+            let batch = batch?;
+            if batch.num_rows() == 0 {
```
westonpace, line 414: Does this actually happen?  We should probably fix that in `lance`.  It seems like an annoying thing to have to deal with.

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+
+        Ok(())
+    }
+
+    async fn query_manifest(
+        &self,
+        filter: Option<&str>,
+        columns: Option<Vec<String>>,
+        limit: Option<usize>,
+    ) -> Result<BoxStream<'static, Result<RecordBatch>>> {
```
westonpace: Should we just collect the results into a vec as part of this method?  Or maybe have a variant of this method that does so?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+        drop(dataset_guard); // Drop read guard before merge insert
+
+        let mut merge_builder =
+            lance::dataset::MergeInsertBuilder::try_new(dataset_arc, vec!["object_id".to_string()])
+                .map_err(|e| Error::Lance { source: e })?;
+
+        // When matched: update all fields
+        merge_builder.when_matched(lance::dataset::WhenMatched::UpdateAll);
+        // When not matched: insert new row
+        merge_builder.when_not_matched(lance::dataset::WhenNotMatched::InsertAll);
```
westonpace, line 580: Hmm, merge_insert seems like the wrong way to have to solve this problem.  It's probably ok for now but it seems it would be a lot nicer if we could just add a "unique constraint" to the column.

Not something we need to fix right now.

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+        let random_num: u64 = rand::rng().random();
+
+        // Create hash from random number + object_id
+        let mut hasher = DefaultHasher::new();
+        random_num.hash(&mut hasher);
+        object_id.hash(&mut hasher);
+        let hash = hasher.finish();
+
+        // Format as lowercase hex (8 characters - sufficient entropy for uniqueness)
+        format!("{:08x}_{}", (hash & 0xFFFFFFFF) as u32, object_id)
```
westonpace, line 659: Ah, I see object id is part of the dir name.  That is nice.

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+            })?
+            .as_any()
+            .downcast_ref::<ListArray>()
+            .ok_or_else(|| Error::Runtime {
+                message: format!("Invalid {} type", column_name),
+            })
+    }
+
+    /// Register a table in the manifest without creating the physical table
+    /// TODO: this could be in public database API
```
westonpace, line 722: By "public database API" do you mean the `Database` trait?  Isn't this manifest-database specific?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                name: name.to_string(),
+            });
+        }
+
+        self.insert_into_manifest(object_id, ObjectType::Table, Some(location), None, None)
+            .await
+    }
+
+    /// List all tables in a namespace with their metadata
+    /// TODO: this could be in public database API
```
westonpace, line 745: What does this todo comment mean?  Don't we already have a list_tables in the `Database` trait?

rust/lancedb/src/database/listing/manifest.rs
```diff
@@ -0,0 +1,3991 @@
...
+                        &data_schema,
+                    )
+                    .await
+                }
+                Err(err) => Err(err),
+            }
+        } else {
+            // the table creation must succeed because concurrent creations
+            // writes to different directories with different hash names.
+            let table = create_result?;
```
westonpace, line 1089:
```suggestion
            // the table creation must succeed because concurrent creations
            // writes to different directories with different hash names and we
            // already ensured the object id didn't exist above.
            let table = create_result?;
```

rust/lancedb/src/database/listing/manifest.rs
jackye1995: this one contains table location information because the location is no longer hard coded from table name, the current one is just `table_names`.

python/python/tests/test_db.py
jackye1995: oh this is just testing behavior when this new feature is disabled and it continues to say not supported. But it probably is not meaningful test, I can remove it.

rust/lancedb/src/database/listing.rs
jackye1995: None

rust/lancedb/src/database/listing.rs
jackye1995: oh good point, updated

rust/lancedb/src/database/listing.rs
jackye1995: oh I commented in the inner class, let me add it also here. 

```
// To the ManifestListingDatabase, this table creation has failed.
// But to this caller ListingDatabase, the table has been successfully created
// in the right storage location, which means it has technically succeeded.
```

rust/lancedb/src/database/listing/manifest.rs
jackye1995: updated

rust/lancedb/src/database/listing/manifest.rs
jackye1995: > Though I suppose the folder name wouldn't track after rename, etc. 

yeah. the new naming scheme still keeps the table name, just appends a random enough hash prefix, which I think still allows people who want to poke around to do so.

rust/lancedb/src/database/listing/manifest.rs
jackye1995: I did warning in general for optimizations because the operation can technically still return correct result without these. Especially later for compaction and reindexing, if those fails while the operation still succeeds, it might appear as not atomic to users.

Is there any specific reason that failing is better?

rust/lancedb/src/database/listing/manifest.rs
jackye1995: The reason I did it this way is because it is possible that user only has read access and cannot even create the manifest table, and then the whole thing is not useable. But yeah it seems like we can check that and just turn the manifest_enabled to false in that case at listing database. let me update it in that way so we can remove the mutex.

rust/lancedb/src/database/listing/manifest.rs
jackye1995: no, this is just sanity check, can remove

rust/lancedb/src/database/listing/manifest.rs
jackye1995: > if we could just add a "unique constraint" to the column

even if we do that, isn't it basically a merge insert?

rust/lancedb/src/database/listing/manifest.rs
jackye1995: > By "public database API" do you mean the Database trait?

yes. I just mean it seems like an ability that database can in general have: someone has a table in some other location, and you can register it to become a table managed by the database. But maybe shallow clone is enough given we already have that.

rust/lancedb/src/database/listing/manifest.rs
jackye1995: yes good point

jackye1995:
Closed in favor of using the DirectoryNamespace that has this feature added

PR #2712, opened 2025-10-10
feat: add a permutation reader that can read a permutation view

westonpace:
This adds a rust permutation builder.  In the next PR I will have python bindings and integration with pytorch.

### DIFF
--- a/docs/src/js/functions/permutationBuilder.md
+++ b/docs/src/js/functions/permutationBuilder.md
@@ -7,7 +7,7 @@
 # Function: permutationBuilder()
 
 ```ts
-function permutationBuilder(table, destTableName): PermutationBuilder
+function permutationBuilder(table): PermutationBuilder
 ```
 
 Create a permutation builder for the given table.
@@ -16,9 +16,6 @@ Create a permutation builder for the given table.
 
 * **table**: [`Table`](../classes/Table.md)
     The source table to create a permutation from
-
-* **destTableName**: `string`
-    The name for the destination permutation table
 
 ## Returns
 


rust/lancedb/src/dataloader/permutation/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+impl Default for ShuffleStrategy {
+    fn default() -> Self {
+        Self::None
+    }
+}
+
+/// Builder for creating a permutation table.
+///
+/// A permutation table is a table that stores split assignments and a shuffled order of rows.  This
+/// can be used to create a
```
wkalt: comment trails off

rust/lancedb/src/dataloader/permutation.rs
```diff
@@ -10,285 +10,8 @@
 //! Building a permutation table should be fairly quick and memory efficient, even for billions or
 //! trillions of rows.
```
wkalt: we might want to quantify this better than "quick". How does the number of rows, size of split relate to the speed?

rust/lancedb/src/dataloader/permutation/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// A seed can be provided to make the shuffle deterministic.
+    ///
+    /// If a clump size is provided, then data will be shuffled in small blocks of contiguous rows.
+    /// This decreases the overall randomization but can improve I/O performance when reading from
+    /// cloud storage.
+    ///
+    /// For example, a clump size of 16 will means we will shuffle blocks of 16 contiguous rows.  This
+    /// will mean 16x fewer IOPS but these 16 rows will always be close together and this can influence
+    /// the performance of the model.  Note: shuffling within clumps can still be done at read time but
+    /// this will only provide a local shuffle and not a global shuffle.
```
wkalt, line 67: this comment is useful but I think the file would benefit from some similar level of detail about the whole permutation/split/clumping concept. Or else we should have that written up somewhere else in this code.
* Why do I want a permutation? How does this differ from a materialized view with a predicate?
* What is a typical case for more than one "split"? Does a single split of 10% of the data actually imply two splits (10% and 90%)? Calling this a single split makes it seem like the split is the dividing line, but separately we seem to be able to address splits in queries or operations which makes it seem like the split is the section of data.
* Do multiple splits overlap with each other?

rust/lancedb/src/dataloader/permutation/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+    ///
+    /// The default is a single split that contains all rows.
+    pub fn with_split_strategy(mut self, split_strategy: SplitStrategy) -> Self {
+        self.config.split_strategy = split_strategy;
+        self
+    }
+
+    /// Configures the strategy for shuffling the data.
+    ///
+    /// The default is to shuffle the data randomly at row-level granularity (no shard size) and
```
wkalt: is "shard size" the same as "clump size"?

rust/lancedb/src/dataloader/permutation/builder.rs
```diff
@@ -0,0 +1,322 @@
...
+        needs_sort |= !matches!(self.config.shuffle_strategy, ShuffleStrategy::None);
+
+        let sorted = if needs_sort {
+            self.sort_by_split_id(shuffled).await?
+        } else {
+            shuffled
+        };
+
+        // Rename _rowid to row_id
+        let renamed = rename_column(sorted, ROW_ID, SRC_ROW_ID_COL)?;
```
wkalt, line 250: what if the user has a column called "row_id"?

wkalt: same question for "split_id"

rust/lancedb/src/table.rs
```diff
@@ -511,6 +511,9 @@ pub trait BaseTable: std::fmt::Display + std::fmt::Debug + Send + Sync {
     /// Get the namespace of the table.
     fn namespace(&self) -> &[String];
     /// Get the id of the table
+    ///
+    /// This is the namespace of the table concatenated with the name
+    /// separated by a .
```
wkalt: nitpick - `separated by a dot (".")`

rust/lancedb/src/test_utils/connection.rs
```diff
@@ -0,0 +1,120 @@
...
+        uri: uri.to_string(),
+        connection,
+        _temp_dir: Some(temp_dir),
+        _process: Some(process),
+    })
+}
+
+fn read_process_port(mut stdout: BufReader<ChildStdout>) -> Result<String> {
+    let mut line = String::new();
+    let re = Regex::new(r"Query node now listening on 0.0.0.0:(.*)").unwrap();
```
wkalt, line 74: How does this get used? Parsing the log for this might break at some point. Does the caller of this set up the query node process? If so seems like we can know the port in advance

rust/lancedb/src/dataloader/permutation/builder.rs
westonpace: These are columns in the permutation table, not the base table, so it should be fine if the user has a column named `row_id` or `split_id`.  I can add a test for this.

rust/lancedb/src/test_utils/connection.rs
westonpace: Hmm, I dunno.  This was a copy/paste from `test_connection.rs` (moved it into `test_utils.connection`).  CC @cmccabe to field this question.

wkalt: it's not critical unless the user themselves might be running this code, in which case I think it would become important to address because the logs could well change

rust/lancedb/src/dataloader/permutation/builder.rs
wkalt: any reason not to stick with the leading prefix convention?

westonpace: These aren't internal / system columns.  This is just a regular table (think of it like a many-to-many join table).  I worry that `_row_id` would be too close to `_rowid` and cause confusion.  What if I named it `source_row_id`?

westonpace: I guess I think of this permutation stuff as an "extension on top of lancedb" and not a "feature of lancedb".  So the underscore prefix is something reserved for my database, not for tools that use the database.

wkalt: ok gotcha. I feel less strongly about it then. If there's no chance of these getting conflated with user columns in operation, and the naming won't seem unnatural, then it's fine.

rust/lancedb/src/dataloader/permutation.rs
westonpace: I added "it is an O(N) operation where N is the number of rows"

rust/lancedb/src/dataloader/permutation/builder.rs
westonpace: I'm going to address these questions in a "guide"

rust/lancedb/src/dataloader/permutation/builder.rs
westonpace: Changed to clump size

rust/lancedb/src/dataloader/permutation/builder.rs
westonpace: Expanded

wkalt (approved):

PR #2719, opened 2025-10-16
feat: expand support for multivector colpali models and enchancements

AyushExel:
We already had community integration for colpali engine. This PR adds a few improvements:
- Support more architectures
- Clamps nans
- Warns of nans when using mps
- Provide offloading folder 
- Support custom pooling functions
- expand tests

### DIFF
--- a/python/python/lancedb/embeddings/colpali.py
+++ b/python/python/lancedb/embeddings/colpali.py
@@ -22,13 +22,18 @@ class ColPaliEmbeddings(EmbeddingFunction):
     multimodal multi-vector embeddings.
 
     This embedding function supports ColPali models, producing multivector outputs
-    for both text and image inputs. The output embeddings are lists of vectors,
-    each vector being 128-dimensional by default, represented as List[List[float]].
+    for both text and image inputs.
 
     Parameters
     ----------
     model_name : str
         The name of the model to use (e.g., "Metric-AI/ColQwen2.5-3b-multilingual-v1.0")
+        Supports models based on these engines:
+        - ColPali: "vidore/colpali-v1.3" and others
+        - ColQwen2.5: "Metric-AI/ColQwen2.5-3b-multilingual-v1.0" and others
+        - ColQwen2: "vidore/colqwen2-v1.0" and others
+        - ColSmol: "vidore/colSmol-256M" and others
+
     device : str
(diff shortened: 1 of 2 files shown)


python/python/lancedb/embeddings/colpali.py
```diff
@@ -56,15 +75,43 @@ class ColPaliEmbeddings(EmbeddingFunction):
...
+            if torch.cuda.is_available():
+                device = "cuda"
+            elif torch.backends.mps.is_available():
+                device = "mps"
+            else:
+                device = "cpu"
+
+        dtype = self.dtype
+        if device == "mps" and dtype == "bfloat16":
+            dtype = "float32"  # Avoid NaNs on MPS
```
BubbleCal, line 103: is this because mps doesn't support bfloat16?

python/python/lancedb/embeddings/colpali.py
```diff
@@ -98,21 +155,45 @@ def _load_model(
...
+        model_class, processor_class = None, None
+        model_name_lower = model_name.lower()
+        if "colqwen2.5" in model_name_lower:
+            model_class = colpali_engine.models.ColQwen2_5
+            processor_class = colpali_engine.models.ColQwen2_5_Processor
+        elif "colsmol" in model_name_lower or "colidefics3" in model_name_lower:
+            model_class = colpali_engine.models.ColIdefics3
+            processor_class = colpali_engine.models.ColIdefics3Processor
+        elif "colqwen" in model_name_lower:
+            model_class = colpali_engine.models.ColQwen2
```
BubbleCal, line 167: should we make the default model of "colqwen" be the newer one (`ColQwen2_5`)?
and use "colqwen2_0" to specify `ColQwen2`

python/python/lancedb/embeddings/colpali.py
AyushExel: Yeah. And also there's a problem of getting nans with mps at any reduced precision with some late interaction models

python/python/lancedb/embeddings/colpali.py
AyushExel: This is actually trying to infer the architecture from the official model_id (from the hf hub) itself. To do this we'll have to provide name: arch mapping, I think?

PR #2725, opened 2025-10-17
feat: add python Permutation class to mimic hugging face dataset and provide pytorch dataloader

westonpace:


python/python/lancedb/permutation.py
```diff
@@ -70,3 +229,584 @@ async def do_execute():
...
+            try:
+                while True:
+                    batch = await async_iter.__anext__()
+                    if batch.num_rows == batch_size or not skip_last_batch:
+                        batches.append(batch)
+                    # else skip the partial last batch
+            except StopAsyncIteration:
+                # Reached the end of the stream
+                pass
+            return batches
```
Xuanwo: This function gathers all batches in `batches`, is that expected? It could cause OOM on large inputs.

python/python/lancedb/permutation.py
```diff
@@ -70,3 +229,584 @@ async def do_execute():
...
+    def arrow2python(batch: pa.RecordBatch) -> dict[str, list[Any]]:
+        return batch.to_pydict()
+
+    @staticmethod
+    def arrow2arrow(batch: pa.RecordBatch) -> pa.RecordBatch:
+        return batch
+
+    @staticmethod
+    def arrow2numpy(batch: pa.RecordBatch) -> dict[str, "np.ndarray"]:
+        return batch.to_pandas().to_numpy()
```
Xuanwo, line 345: The return type of this line seems to be `np.ndarray` instead of `dict[str, "np.ndarray"]`. Maybe `to_dict` is what we want.

Image

python/python/lancedb/permutation.py
```diff
@@ -22,8 +66,38 @@ def split_random(
         counts: Optional[list[int]] = None,
         fixed: Optional[int] = None,
         seed: Optional[int] = None,
+        split_names: Optional[list[str]] = None,
     ) -> "PermutationBuilder":
-        self._async.split_random(ratios=ratios, counts=counts, fixed=fixed, seed=seed)
+        """
+        Configure random splits for the permutation.
+
+        One of ratios, counts, or fixed must be provided.
```
jmhsieh, line 75: what happens if more than one of ratios, counts, of fixed is provided?

python/python/lancedb/permutation.py
```diff
@@ -42,25 +141,85 @@ def split_sequential(
         ratios: Optional[list[float]] = None,
         counts: Optional[list[int]] = None,
         fixed: Optional[int] = None,
+        split_names: Optional[list[str]] = None,
     ) -> "PermutationBuilder":
-        self._async.split_sequential(ratios=ratios, counts=counts, fixed=fixed)
+        """
+        Configure sequential splits for the permutation.
+
+        One of ratios, counts, or fixed must be provided.
```
jmhsieh, line 150: same - what happens if multiple specified?

python/python/tests/test_permutation.py
```diff
@@ -460,3 +477,357 @@ def test_filter_empty_result(mem_db):
...
+    with pytest.raises(ValueError, match="split `nonexistent` is not defined"):
+        Permutation.from_tables(some_table, some_perm_table, "nonexistent")
+
+    # Reference by ordinal and there are no rows
+    with pytest.raises(ValueError, match="No rows found"):
+        Permutation.from_tables(some_table, some_perm_table, 5)
+
+
+def test_permutations(some_table: Table, some_perm_table: Table):
+    permutations = Permutations(some_table, some_perm_table)
```
jmhsieh, line 542: Can I apply a permutation on a permutation?  e.g. could I create a subset permutation of this some_perm_table?  I think the api allows it.

jmhsieh (commented):
I did a first pass -- might need another to grok this. 

What happens if a user tries a write operation on a Permutation Table? 

Can permutation tables be used as the souce table for a another permutation?

python/python/lancedb/permutation.py
westonpace: An error is raised (Exactly one of 'ratios', 'counts', or 'fixed' must be provided).  This is verified in test_split_error_cases.

python/python/lancedb/permutation.py
westonpace: Ditto, the same error message is provided

python/python/tests/test_permutation.py
westonpace: If the goal is to read a portion of the permutation then creating a permutation reader with a limit / offset would be the easiest way to achieve it (this is my plan for distributed data loading).

If the goal is to further filter the permutation then you probably will need to create a fresh permutation from the original table.

If the goal is to re-shuffle the permutation then I hope to add a method to do this soon (this will be needed for training multiple epochs)

python/python/lancedb/permutation.py
westonpace: Good catch.  That was not the intention.  Fixed.

python/python/lancedb/permutation.py
westonpace: It returns a 2-dimensional ndarray.  I changed the return type to just np.ndarray.  I also created a test_transform method to test each of the possible formats (which is good because polars and torch had bugs :laughing: )

jmhsieh (approved):
the python tests and code looks good to me.  My rust was more of a skim so wait for @Xuanwo if you want more there.

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -11,58 +11,146 @@ use crate::arrow::{SendableRecordBatchStream, SimpleRecordBatchStream};
...
+            permutation_table
+                .count_rows(Some(Filter::Sql(format!(
+                    "{} = {}",
+                    SPLIT_ID_COLUMN, self.split
+                ))))
+                .await? as u64
+        } else {
+            self.base_table.count_rows(None).await? as u64
+        };
+        if let Some(offset) = offset {
```
Xuanwo: Just a small nit: is it easier to read by using `match (offset, limit)`?

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -230,26 +322,96 @@ impl PermutationReader {
...
+                    message: "Permutation table must contain a column named split_id".to_string(),
+                });
+            }
+        }
+        let avail_rows = if let Some(permutation_table) = &self.permutation_table {
+            permutation_table.count_rows(None).await? as u64
+        } else {
+            self.base_table.count_rows(None).await? as u64
+        };
+        if let Some(offset) = self.offset {
```
Xuanwo: The same. Btw, we have seen this pattern twice.

westonpace: Collapsed into helper function using match.  Thanks for the suggestion.

rust/lancedb/src/dataloader/permutation/reader.rs
westonpace: Done

PR #2755, opened 2025-10-29
feat: add fts udtf in sql

LuQQiu:
Support FTS feature parity in SQL to match current Python API capability.
Add `.to_json()` method to FTS query classes to enable usage with SQL `fts()` UDTF.
Related: https://github.com/lancedb/blog-lancedb/pull/147

query = MatchQuery("puppy", "text", fuzziness=2)
result = client.execute(f"SELECT * FROM fts('table', '{query.to_json()}')")

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -138,7 +138,7 @@ class FullTextQuery(ABC):
         >>> from lancedb.query import MatchQuery
         >>> query = MatchQuery("puppy", "text", fuzziness=2)
         >>> query.to_json()
-        '{"match":{"column":"text","terms":"puppy","fuzziness":2}}'
+        '{"match":{"column":"text","terms":"puppy","boost":1.0,"fuzziness":2,"max_expansions":50,"operator":"Or","prefix_length":0}}'
         """
         return fts_query_to_json(self)
 


python/python/tests/test_fts.py
```diff
@@ -727,3 +735,185 @@ def test_fts_ngram(mem_db: DBConnection):
...
     assert len(results) == 2
     assert set(r["text"] for r in results) == {"lance database", "lance is cool"}
+
+
+def test_fts_query_json_string_format_for_rust_parser():
+    """
+    Test that Python generates valid JSON strings that can be parsed by Rust.
+
+    Cross-reference: Rust tests that parse these JSON strings are located at:
+    /Users/lu/Projects/Github/Work/sophon/src/rust/sophon-sql/src/udtf.rs
```
jackye1995: should remove this

python/python/lancedb/query.py
```diff
@@ -202,6 +237,27 @@ class MatchQuery(FullTextQuery):
     def query_type(self) -> FullTextQueryType:
         return FullTextQueryType.MATCH
 
+    def to_dict(self) -> Dict[str, Any]:
```
jackye1995: does that mean we are using Lance FtsQuery for actual deserialization, and use LanceDB FullTextQuery for serialization, and we need to make sure they are equivalent? That feels a bit fragile

LuQQiu: Yep, that's the main problem with this approach.
The serializer and deserializer are not at the same place.

LuQQiu: Let me see how to put the serializer / deserializer in the same place to avoid breaking changes

rust/lancedb/src/table/datafusion/udtf/fts.rs
jackye1995, line 1: I think this should be placed within a new `/table/datafusion` folder, and also we should have `/table/datafusion/udtf/fts.rs` kind of setup to allow other udtfs.

PR #2770, opened 2025-11-06
feat: pare down docs to only show API refs

prrao87:
This PR does the following: 
- Pare down the docs to only what's needed (Python, JS/TS API docs and a pointer to Rust docs)
- Styling changes to be more in line with the main website theme

The relative URLs remain unchanged, so assuming CI passes, there should be no breaking changes from the main docs site that points back here.


prrao87:
Here's how it looks.
image

docs/mkdocs.yml
```diff
@@ -120,277 +119,17 @@ markdown_extensions:
       permalink: ""
 
 nav:
-  - Home:
```
Xuanwo, line 123: This PR seems to move a lot of pages. Maybe it’s better to use a dedicated branch for this refactor, so we can preview and push it to production as we get ready, instead of pushing everything all at once.

prrao87: Deleting the `docs_deprecated` files as they are outdated and do not reflect the current state of LanceDB. We have the [new docs](https://lancedb.com/docs) for that.

prrao87:
Alright, this seems good to go!

PR #2778, opened 2025-11-13
feat: support namespace credentials vending

jackye1995:
Based on https://github.com/lancedb/lance/pull/4984

1. Bump to 1.0.0-beta.2
2. Use DirectoryNamespace in lance to perform all testing in python and rust for much better coverage
3. Refactor `ListingDatabase` to be able to accept location and namespace. This is because we have to leverage listing database (local lancedb connection) for using namespace, namespace only resolves the location and storage options but we don't want to bind all the way to rust since user will plug-in namespace from python side. And thus `ListingDatabase` needs to be able to accept location and namespace that are created from namespace connection. 
4. For credentials vending, we also pass storage options provider all the way to rust layer, and the rust layer calls back to the python function to fetch next storage option. This is exactly the same thing we did in pylance.

### DIFF
--- a/python/python/lancedb/__init__.py
+++ b/python/python/lancedb/__init__.py
@@ -14,6 +14,7 @@ __version__ = importlib.metadata.version("lancedb")
 from ._lancedb import connect as lancedb_connect
 from .common import URI, sanitize_uri
 from .db import AsyncConnection, DBConnection, LanceDBConnection
+from .io import StorageOptionsProvider
 from .remote import ClientConfig
 from .remote.db import RemoteDBConnection
 from .schema import vector
@@ -233,6 +234,7 @@ __all__ = [
     "LanceNamespaceDBConnection",
     "RemoteDBConnection",
     "Session",
+    "StorageOptionsProvider",
     "Table",
     "__version__",
 ]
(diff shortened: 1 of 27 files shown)


python/python/lancedb/__init__.py
```diff
@@ -23,7 +23,7 @@
 
 
 def connect(
-    uri: URI,
+    uri: URI = None,
```
rpgreen: Should be Optional[URI]

python/python/lancedb/__init__.py
```diff
@@ -137,7 +171,7 @@ async def connect_async(
     storage_options: Optional[Dict[str, str]] = None,
     session: Optional[Session] = None,
 ) -> AsyncConnection:
-    """Connect to a LanceDB database.
+    """Connect to a LanceDB database. YAY!
```
rpgreen: 😆

jackye1995:
Failures same as main branch, merging for now to unblock beta release for geneva.

Pending TODOs:
1. fix windows issue with namespace
2. some parameters should be marked as internal (e.g. namespace and storage options provider for create_table and open_table)
3. use lance-namespace models directly instead of doing conversion with another layer of lancedb namespace API models
4. use None instead of [] as default for namespace for python, which seems like the recommended practice because python default argument value is globally shared and mutable
5. deprecate table_names in favor of list_tables which supports pagination

PR #2781, opened 2025-11-14
fix: .phrase_query() not working

jachewz:
The `self._query` value was not set when wrapping its copy `query` with quotation marks.

The test for phrase queries has been updated to test the `.phrase_query()` method as well, which will catch this bug.

### DIFF
--- a/python/python/tests/test_fts.py
+++ b/python/python/tests/test_fts.py
@@ -326,7 +326,7 @@ def test_search_fts_phrase_query(table):
     results = table.search("puppy").limit(100).to_list()
 
     # Test with quotation marks
-    phrase_results = table.search("puppy runs").phrase_query().limit(100).to_list()
+    phrase_results = table.search('"puppy runs"').limit(100).to_list()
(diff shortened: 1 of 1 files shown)


python/python/tests/test_fts.py
```diff
@@ -324,11 +324,18 @@ def test_search_fts_phrase_query(table):
         pass
     table.create_fts_index("text", use_tantivy=False, with_position=True, replace=True)
     results = table.search("puppy").limit(100).to_list()
-    phrase_results = table.search('"puppy runs"').limit(100).to_list()
+
+    # Test with quotation marks
+    phrase_results = table.search("puppy runs").phrase_query().limit(100).to_list()
```
wjones127: We should keep this like so, to match the comment, right?
```suggestion
    phrase_results = table.search('"puppy runs"').limit(100).to_list()
```

jachewz: 🤦 yeah whoops

wjones127 (approved):

PR #2788, opened 2025-11-18
feat: support async namespace connection

jackye1995:
Also fix 2 bugs:
1. make storage options provider serializable in ray
2. fix table.to_table() uri is wrong for namespace-backed tables


jackye1995:
Doctest failed due to OOM when installing pytorch, will fix in a separated PR

PR #2793, opened 2025-11-18
feat: implement head() for remote tables

cmccabe:
Implemnent the head() function for RemoteTable.


westonpace:
Is it possible to create a unit test in `test_remote_db.py`?

cmccabe:
Errors are the infamous "No space left on device" (for doctest) and an unrelated ruff error in a file that wasn't changed. Committing

PR #2804, opened 2025-11-20
ci: add support for lance-format fury index for downloading pylance

jackye1995:
Set the lance-format fury repo for most places that are downloading. For uploading, it is kept unchanged since lancedb is published to lancedb fury.


jackye1995:
doctest fixed in #2801

Xuanwo (approved):

PR #2806, opened 2025-11-20
refactor!: use namespace models directly for namespace operations

jackye1995:
1. Use generated models in lance-namespace for request response models to avoid multiple layers of conversions
2. Make sure the API is consistent with the namespace spec
3. Deprecate the table_names API in favor of the list_tables API in namespace that allows full pagination support without the need to have sorted table names
4. Add describe_namespace API which was a miss in the original implementation

### DIFF
--- a/rust/lancedb/src/database.rs
+++ b/rust/lancedb/src/database.rs
@@ -24,12 +24,12 @@ use datafusion_physical_plan::stream::RecordBatchStreamAdapter;
 use futures::stream;
 use lance::dataset::ReadParams;
 use lance_datafusion::utils::StreamingWriteSource;
-use lance_namespace::LanceNamespace;
 use lance_namespace::models::{
     CreateNamespaceRequest, CreateNamespaceResponse, DescribeNamespaceRequest,
     DescribeNamespaceResponse, DropNamespaceRequest, DropNamespaceResponse, ListNamespacesRequest,
     ListNamespacesResponse, ListTablesRequest, ListTablesResponse,
 };
+use lance_namespace::LanceNamespace;
 
 use crate::arrow::{SendableRecordBatchStream, SendableRecordBatchStreamExt};
 use crate::error::Result;
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -1812,6 +1812,7 @@ mod tests {
                 mode: CreateTableMode::Create,
                 write_options: Default::default(),
                 location: None,
+                namespace_client: None,
             })
             .await
             .unwrap();
@@ -1865,6 +1866,7 @@ mod tests {
                 mode: CreateTableMode::Create,
                 write_options,
                 location: None,
+                namespace_client: None,
             })
             .await
             .unwrap();
@@ -1938,6 +1940,7 @@ mod tests {
                 mode: CreateTableMode::Create,
                 write_options,
                 location: None,
+                namespace_client: None,
             })
             .await
             .unwrap();


rust/lancedb/src/database.rs
```diff
@@ -24,6 +24,12 @@ use datafusion_physical_plan::stream::RecordBatchStreamAdapter;
 use futures::stream;
 use lance::dataset::ReadParams;
 use lance_datafusion::utils::StreamingWriteSource;
+// Re-export lance_namespace types for public API
+pub use lance_namespace::models::{
+    CreateNamespaceRequest, CreateNamespaceResponse, DescribeNamespaceRequest,
+    DescribeNamespaceResponse, DropNamespaceRequest, DropNamespaceResponse, ListNamespacesRequest,
+    ListNamespacesResponse, ListTablesRequest, ListTablesResponse,
+};
```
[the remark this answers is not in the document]
jackye1995: thank you this is exactly what I am fixing

python/python/lancedb/db.py
```diff
@@ -48,6 +55,12 @@ class EnforceOverrides:
...
     from ._lancedb import Session
 
+from .namespace_utils import (
+    _normalize_create_namespace_mode,
+    _normalize_drop_namespace_mode,
+    _normalize_drop_namespace_behavior,
+)
+
 
 class DBConnection(EnforceOverrides):
```
westonpace, line 65: Not related but it would be really nice to phase out these abstract classes :cold_sweat:

python/python/lancedb/namespace.py
```diff
@@ -237,6 +249,19 @@ def table_names(
...
+        .. deprecated::
+            Use :meth:`list_tables` instead, which provides proper pagination support.
+        """
+        import warnings
+
+        warnings.warn(
+            "table_names() is deprecated, use list_tables() instead",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
westonpace, line 268: Do we have a migration guide for lancedb?  It would be good to mention this.

python/python/lancedb/remote/db.py
```diff
@@ -116,14 +123,15 @@ def list_namespaces(
             The parent namespace to list namespaces in.
             None or empty list represents root namespace.
         page_token: str, optional
-            The token to use for pagination. If not present, start from the beginning.
-        limit: int, default 10
-            The size of the page to return.
+            Token for pagination. Use the token from a previous response
+            to get the next page of results.
+        limit: int, optional
+            The maximum number of results to return.
```
westonpace, line 129: What's the default?  Is it unlimited or implementation defined?

python/python/lancedb/remote/db.py
```diff
@@ -134,26 +142,111 @@ def list_namespaces(
...
+        Parameters
+        ----------
+        namespace: List[str], optional
+            The namespace to list tables in.
+            None or empty list represents root namespace.
+        page_token: str, optional
+            Token for pagination. Use the token from a previous response
+            to get the next page of results.
+        limit: int, optional
+            The maximum number of results to return.
```
westonpace, line 236: Same question as above, can we document the default?

rust/lancedb/src/database/listing.rs
```diff
@@ -693,6 +719,57 @@ impl Database for ListingDatabase {
...
+            .map(Path::new)
+            .filter(|path| {
+                let is_lance = path
+                    .extension()
+                    .and_then(|e| e.to_str())
+                    .map(|e| e == LANCE_EXTENSION);
+                is_lance.unwrap_or(false)
+            })
+            .filter_map(|p| p.file_stem().and_then(|s| s.to_str().map(String::from)))
+            .collect::<Vec<String>>();
```
westonpace, line 785: I assume this is just copied from table_names?

rust/lancedb/src/database/listing.rs
```diff
@@ -693,6 +719,57 @@ impl Database for ListingDatabase {
...
+            if f.len() > limit as usize {
+                let token = f[limit as usize].clone();
+                f.truncate(limit as usize);
+                Some(token)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
westonpace, line 808: Can we create a follow-up task to use [`PaginatedListStore`](https://docs.rs/object_store/latest/object_store/list/trait.PaginatedListStore.html), if available, to avoid fetching all files in the list dir call?  Or should we not even bother since we hope to switch to a manifest-backed default anyways.

rust/lancedb/src/error.rs
```diff
@@ -28,6 +28,8 @@ pub enum Error {
 
     #[snafu(display("Table '{name}' already exists"))]
     TableAlreadyExists { name: String },
+    #[snafu(display("Namespace error: {message}"))]
+    Namespace { message: String },
```
westonpace: Is this useful?  Couldn't the user infer that the error has something to do with namespaces from the message itself?  Or do you envision a user may someday treat namespace errors differently than something like InvalidInput?

rust/lancedb/src/remote/db.rs
```diff
@@ -371,6 +369,44 @@ impl<S: HttpSend> Database for RemoteDatabase<S> {
         Ok(tables)
     }
 
+    async fn list_tables(&self, request: ListTablesRequest) -> Result<ListTablesResponse> {
+        let namespace_parts = request.id.as_deref().unwrap_or(&[]);
+        let namespace_id = build_namespace_identifier(namespace_parts, &self.client.id_delimiter);
+        let mut req = self
+            .client
+            .get(&format!("/v1/namespace/{}/table/list", namespace_id));
```
westonpace, line 377: Insider-knowledge question, do we have this implemented on our own rest-based catalog impl?

jackye1995: yes we do

rust/lancedb/src/database/listing.rs
jackye1995: Yeah I think that would be nice, and also useful in lance. Created https://github.com/lancedb/lancedb/issues/2839

rust/lancedb/src/database/listing.rs
jackye1995: yes

python/python/lancedb/remote/db.py
jackye1995: Actually I intentionally did not set a default. Basically if limit is not specified, the server can decide whatever number to return, it can return everything in one call, or do a few paginations.

python/python/lancedb/namespace.py
jackye1995: Created https://github.com/lancedb/lancedb/issues/2840 tracking this

python/python/lancedb/db.py
jackye1995: created https://github.com/lancedb/lancedb/issues/2841 tracking this

rust/lancedb/src/error.rs
jackye1995: yeah good point, removed it

PR #2829, opened 2025-12-01
feat: support embedding at search on remote table

BubbleCal:
### What is the bug?
Embedding-aware query strings worked for local tables but not for remote tables. On LanceDB Cloud, a query like:

```python
results = table.search("test example", query_type="vector").limit(5).to_pandas()
```

could fail with:

```
ValueError: No embedding function for None
```

If `query_type` was omitted, the same string query could instead be treated like full-text search and fail with:

```
400 Bad Request: Cannot perform full text search unless an INVERTED index has been created on at least one column
```

### Why did this happen?
`LanceTable.search` already normalizes string queries by inspecting the table schema and embedding metadata to infer the vector column and decide whether the query should be embedded client-side before sending the request.

`RemoteTable.search` did not go through the same inference path. It forwarded the raw query into `LanceQueryBuilder.create(...)` without first calling `infer_vector_column_name(...)` and without the same `FullTextQuery` normalization. As a result:

- explicit vector queries did not resolve the embedding function / vector column correctly
- implicit string queries could fall back to the wrong query mode and hit FTS behavior on the server

### What issues or incorrect behavior did this cause?
Users could not reliably run text-to-vector search against remote tables that have embedding metadata, even though the same table/query pattern worked locally.

That meant a valid remote workflow like `table.search("puppy")` or `table.search("puppy", query_type="vector")` could fail unless the user manually worked around the query mode or column selection.

This PR makes `RemoteTable.search` follow the same search setup path as `LanceTable.search`:

- normalize `FullTextQuery` inputs to `query_type="fts"`
- infer the vector column with `infer_vector_column_name(...)` using the remote table schema and embedding metadata
- pass the resolved `vector_column_name` into `LanceQueryBuilder.create(...)` so embedding-aware string queries are converted correctly before the request is sent

### Validation
Added a regression test in `python/python/tests/test_remote_db.py` that verifies a remote table with embedding metadata:

- accepts a string query
- computes the query embedding client-side
- sends the resolved vector and `vector_column` in the remote request body
- does not incorrectly send the query as full-text search


wjones127 (commented):
I'm actually surprised that this wouldn't work already. Could you describe what wasn't working before? And why this works?

BubbleCal:
full repro script:
```
from __future__ import annotations

import os
from datetime import datetime, timezone

import lancedb

def main() -> None:
    # Staging server connection (no secrets in code)
    uri = os.getenv("LANCEDB_URI")
    region = os.getenv("LANCEDB_REGION")
    api_key = os.getenv("LANCEDB_API_KEY")
    if uri.startswith("db://") and not api_key:
        raise RuntimeError("Missing LANCEDB_API_KEY for db:// URI")

    # Embedding function (same as official example)
    func = (
        get_registry()
        .get("sentence-transformers")
        .create(
            model=os.getenv("MODEL_NAME", "all-MiniLM-L6-v2"),
        )
    )

    class MySchema(LanceModel):

    if uri.startswith("db://"):
        db = lancedb.connect(uri, region=region, api_key=api_key)
    else:
        db = lancedb.connect(uri)

    table_name = os.getenv(
        "LANCEDB_TABLE",
        f"st_embed_ivf_pq_{datetime.now(timezone.utc).strftime('%Y%m%d_%H%M%S')}",
    )
    print(f"Creating table: {table_name}")
    table = db.create_table(table_name, schema=MySchema, mode="overwrite")

    # IVF_PQ needs enough rows to train; 256 is a safe minimum.
    print(f"Adding {256} rows to table")
    table.add([{"text": f"Doc {i}: This is a test example."} for i in range(256)])

    # Build IVF_PQ index after insert
    print(f"Creating IVF_PQ index")
    table.create_index(
        vector_column_name="vector",
        index_type="IVF_PQ",
        replace=True,
    )

    print(f"Searching for 'test example'")
    results = table.search("test example").limit(5).to_pandas()
    print(results)


if __name__ == "__main__":
    main()

```

BubbleCal:
> I'm actually surprised that this wouldn't work already. Could you describe what wasn't working before? And why this works?

added the results in description and see full repro script above

wjones127:
Closing this for now. Feel free to re-open if this is a real issue.

PR #2831, opened 2025-12-01
feat: support stable row IDs via storage_options

jmhsieh:
Add support for enabling stable row IDs when creating tables via the `new_table_enable_stable_row_ids` storage option.

Stable row IDs ensure that row identifiers remain constant after compaction, update, delete, and merge operations. This is useful for materialized views and other use cases that need to track source rows across these operations.

The option can be set at two levels:
- Connection level: applies to all tables created with that connection
- Table level: per-table override via create_table storage_options

### DIFF
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -606,18 +606,11 @@ impl ListingDatabase {
             write_params.enable_v2_manifest_paths = enable_v2_manifest_paths;
         }
 
-        // Apply enable_stable_row_ids: request param takes precedence over database config
-        // The request's lance_write_params.enable_stable_row_ids (set via builder) takes
-        // highest precedence since it's already in write_params. Here we only set it if
-        // the request didn't explicitly set it (i.e., it's still false).
-        if !write_params.enable_stable_row_ids {
-            if let Some(enable_stable_row_ids) = self
-                .new_table_config
-                .enable_stable_row_ids
-                .or(stable_row_ids_override)
-            {
-                write_params.enable_stable_row_ids = enable_stable_row_ids;
-            }
+        // Apply enable_stable_row_ids: table-level override takes precedence over connection config
+        if let Some(enable_stable_row_ids) =
+            stable_row_ids_override.or(self.new_table_config.enable_stable_row_ids)
+        {
+            write_params.enable_stable_row_ids = enable_stable_row_ids;
         }
 
         if matches!(&request.mode, CreateTableMode::Overwrite) {


jmhsieh:
All recent mac CI tests seem to fail.

jmhsieh:
> Can we enable this via storage options? If you look at data_storage_version and the manifest paths args you'll see they are deprecated pass throughs to storage options already.

Will look into it!

rust/lancedb/src/connection.rs
```diff
@@ -385,6 +385,22 @@ impl<const HAS_DATA: bool> CreateTableBuilder<HAS_DATA> {
...
     }
+
+    /// Enable stable row IDs for the table. (default: false)
+    ///
+    /// When enabled, row IDs remain stable after compaction operations,
+    /// though not after updates. This is useful for materialized views
+    /// and other use cases that need to track source rows across compaction.
+    ///
+    /// This is an experimental feature in Lance.
+    pub fn enable_stable_row_ids(mut self, enable: bool) -> Self {
```
jackye1995: I think in general the new table level configs should be added to NewTableConfig, you should be able to consult how the feature flag `enable_v2_manifest_paths` is passed in all the way, and do the same for this feature flag.

python/python/lancedb/db.py
```diff
@@ -1079,6 +1083,10 @@ async def create_table(
             See available options at
             <https://lancedb.com/docs/storage/>
 
+            To enable stable row IDs (row IDs remain stable after compaction),
```
jackye1995: nit: stable row ID would also remain unchanged after update/delete/merge.

rust/lancedb/src/database/listing.rs
```diff
@@ -118,6 +137,12 @@ impl DatabaseOptions for ListingDatabaseOptions {
                 enable_v2_manifest_paths.to_string(),
             );
         }
+        if let Some(enable_stable_row_ids) = self.new_table_config.enable_stable_row_ids {
```
jackye1995, line 137: can we also add a test at rust layer, instead of having to rely on python test to verify the functionality

python/python/lancedb/db.py
jmhsieh: updated

python/python/lancedb/db.py
```diff
@@ -149,6 +149,7 @@ def create_table(
         storage_options_provider: Optional["StorageOptionsProvider"] = None,
         data_storage_version: Optional[str] = None,
         enable_v2_manifest_paths: Optional[bool] = None,
+        enable_stable_row_ids: bool = False,
```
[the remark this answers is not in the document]
jmhsieh: legit concern but others (`new_table_enable_data_storage_version`, and `new_table_enable_v2_manifest_paths` )are not supported by remotes either

rust/lancedb/src/database/listing.rs
jmhsieh: done

rust/lancedb/src/connection.rs
jmhsieh: done

jackye1995 (approved):
Looks good to me!

jackye1995:
Test failures are fixed in main branch, feel free to merge directly

PR #2845, opened 2025-12-04
feat: use rest namespace for lancedb java sdk

jackye1995:
After the refactoring on both client and server side, we should have the ability to fully use lance REST namespace to call into LanceDB cloud and enterprise. We can avoid having a JNI implementation (which today does not really do anything except for vending a connection object), and just use lance-core's RestNamespace.

We will at this moment have a LanceDbRestNamespaceBuilder to allow users to more easily build the RestNamespace to talk to LanceDB Cloud or Enterprise endpoint. 

In the future, we could extend this further to also support the local mode through DirectoryNamespace. That will be a separated PR.

### DIFF
--- a/docs/src/java/java.md
+++ b/docs/src/java/java.md
@@ -14,7 +14,7 @@ Add the following dependency to your `pom.xml`:
 <dependency>
     <groupId>com.lancedb</groupId>
     <artifactId>lancedb-core</artifactId>
-    <version>0.22.4-beta.3</version>
+    <version>0.23.0-beta.0</version>
 </dependency>
 ```
 


docs/src/java/java.md
```diff
@@ -0,0 +1,495 @@
+# Java SDK
+
```
prrao87, line 2: I think it makes sense to add an admonition box here to indicate that this applies to only LanceDB Cloud and Enterprise? This would help bring attention to that fact in case OSS users stumble across it.

Here's how to add an [admonition box](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage) in mkdocs.

prrao87 (approved):
LGTM

PR #2852, opened 2025-12-05
feat: support polars version >=0.46

### LINKED ISSUE #2827: chore: bump polars
I found that polras interpolation are implemented through feature gate to offer seemingless conversion between lancedb and polars DataFrame. But the polars version (>=0.37,<0.40) is quite obselete.

I am using the latest polars 0.52 for some latest features. Is it ok to create a PR to bump polars to newer version? Or maybe add a feature gate like `polars_0_52` to indicate polars support for 0.52, in case the polars version are intentionally tagged to between 0.37 and 0.40?

fMeow:
The polars API seems relatively stabled after 0.46 and this change works from 0.46 until now. 

Maybe setting a upper version bound can be more robust. But this way we have to update the upper bound if any user wants to combine lancedb with latest polars.

This PR closes #2827

### DIFF
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -73,8 +73,8 @@ reqwest = { version = "0.12.0", default-features = false, features = [
 ], optional = true }
 http = { version = "1", optional = true } # Matching what is in reqwest
 uuid = { version = "1.7.0", features = ["v4"] }
-polars-arrow = { version = ">=0.37,<0.40.0", optional = true }
-polars = { version = ">=0.37,<0.40.0", optional = true }
+polars-arrow = { version = ">=0.46", optional = true }
+polars = { version = ">=0.46", optional = true }
 hf-hub = { version = "0.4.1", optional = true, default-features = false, features = [
     "rustls-tls",
     "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,7 +226,7 @@ dependencies = [
  "arrow-data",
  "arrow-schema",
  "chrono",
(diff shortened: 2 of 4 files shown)


rust/lancedb/Cargo.toml
```diff
@@ -73,8 +73,8 @@ reqwest = { version = "0.12.0", default-features = false, features = [
 ], optional = true }
 http = { version = "1", optional = true } # Matching what is in reqwest
 uuid = { version = "1.7.0", features = ["v4"] }
-polars-arrow = { version = ">=0.37,<0.40.0", optional = true }
-polars = { version = ">=0.37,<0.40.0", optional = true }
+polars-arrow = { version = ">=0.46", optional = true }
+polars = { version = ">=0.46", optional = true }
```
Xuanwo, line 77: What if we just support the latest version of polars? Is it normal for users to want to use lancedb with specific versions of polars?

fMeow: AFAIK, polars's API changes are quite frequent. And not everyone like to keep their code to latest polars version, especially when they are working on a quite complicate data analysis workflow and there are a lot to change when bump polars version. 

That's fine to me though. But why not support a range of polars version if possible. At lease the FFI API are somewhat stable since 0.46 until now (0.52).

fMeow:
I have fix the style and check if clippy is happy in my local env

fMeow:
This PR is ready for a review @Xuanwo

wjones127:
Polars integration reworked in merged #3584; linked issue #2827 closed NOT_PLANNED.

Issue #2329, opened 2025-04-12
Feature: Support Chinese/CJK text search in BM25 indexing


shellphy:
### SDK

Rust

### Description

Currently, BM25 fulltext search function does not support Chinese, Can Chinese support be added? There are some third-party plugins that can achieve this, such as tantivy-jieba or cang-jie

shellphy:
Are there any development plans?

bb2103:
any updates?

Issue #2168, opened 2025-03-01
Feature: Support Chinese Language in FTS


LastMonopoly:
### SDK

Python

### Description

Thank you for this excellent vector database! 

I would like to request support for the Chinese language in Full-text search (Native FTS) or Full-text search (Tantivy-based FTS). 

As per https://github.com/quickwit-oss/tantivy, it appears that tantivy has third-party support for Chinese (tantivy-jieba and cang-jie), Japanese (lindera, Vaporetto, and tantivy-tokenizer-tiny-segmenter), and Korean (lindera + lindera-ko-dic-builder). 

Could the development team consider providing such support in LanceDB? 

Thank you very much!

BubbleCal:
I think we support Chinese in lance, but not lancedb for now.

I will take a look how to integrate it into lancedb

LastMonopoly:
@BubbleCal Thank you guys

wzdavid:
+1, looking forward to this important enhancement.

bb2103:
+1, we need this enhancement

fierceX:
@BubbleCal @wjones127 How is the progress now? It is very urgent to support Chinese, or it would also be acceptable to be able to pass in custom sparse vectors for retrieval like Qdrant. Or support custom tokenizers.

BubbleCal:
This feature is on the roadmap but we don't get enough people to be working on this.

For now we can query over Chinese/Japanese text by [ngram tokenizer](https://lancedb.com/docs/search/full-text-search/#search-for-substring)

hpvd:
+1

sandsc:
+1, Any update?

naoya7076:
+1

w32zhong:
+1, this is the reason i drop lancedb in my options. (plus no FTS snippet or highlights)

Xuanwo:
Ok, let's do this.

mrwunderbar666:
> This feature is on the roadmap but we don't get enough people to be working on this.
> 
> For now we can query over Chinese/Japanese text by [ngram tokenizer](https://lancedb.com/docs/search/full-text-search/#search-for-substring)

I am not sure whether this really works well for Chinese (or any other language):

```python
import lancedb
from lancedb.query import MatchQuery

uri = "ldb/test"
db = lancedb.connect(uri)


table = db.create_table(
    "my_table_fts",
    data=[
        {
[...]

PR #2855, opened 2025-12-08
feat: enable CJK tokenizer features

YouthLab:
[#2168  Feature: Support Chinese Language in FTS](https://github.com/lancedb/lancedb/issues/2168)
[#2329 Feature: Support Chinese/CJK text search in BM25 indexing](https://github.com/lancedb/lancedb/issues/2329)

For python, the [create_fts_index](https://lancedb.github.io/lancedb/python/python/#lancedb.table.Table.create_fts_index) api will use the new full-text search implementation based on lance-index by default (when the parameter `use_tantivy` is false).
And for now the `lance-index` Lib already supports create FTS index on CJK (Chinese, Japanese, Korea) Characters.
The mandatory PR on lance-index is：[feat: support lindera for japanese and korea tokenization](https://github.com/lance-format/lance/pull/3218)
But unfortunately, for now, the Python SDK did not enable the feature flag for the lance-index when build artifacts.
So I made this PR for this scenario. enable the two feature flag (`tokenizer-lindera` and `tokenizer-jieba`)  will not take any effects on the existing biz logic/api behavior， but introduce the new `JiebaTokenizer ` and `LinderaTokenizer` of lance-index.

For the docs, please ref: 
[Language Models of Jieba](https://github.com/chenkovsky/lance/blob/main/docs/tokenizer.rst#language-models-of-jieba)
[Language Models of Lindera](https://github.com/chenkovsky/lance/blob/main/docs/tokenizer.rst#language-models-of-lindera)


YouthLab:
Test pass on local. Please review this pr. @BubbleCal ,Thanks.
CC: @fierceX @bb2103 @wzdavid @wjones127 @shellphy @LastMonopoly

BubbleCal:
@wjones127 do you happen to know what the concern is that we don't enable these features?
is it because we want to minimize the size?

YouthLab:
> @wjones127 do you happen to know what the concern is that we don't enable these features? is it because we want to minimize the size?

@BubbleCal @wjones127  Thank you for the review!

You're right to ask about the rationale. Enabling these two features (`tokenizer-lindera` and `tokenizer-jieba`) will **not** increase the artifact size. Here's why:

1. **No bundled models**: These features only introduce tokenizer classes written in Rust. The actual language models used by these tokenizers are downloaded by users and Specified with an ENV:`LANCE_LANGUAGE_MODEL_HOME`, not distributed with the artifacts.

2. **User flexibility**: Enabling these features gives users the option to use these tokenizers if needed, without imposing any size penalty on those who don't use them.

The current disabled state means users who want to use these tokenizers need to build from source with these features enabled, which creates unnecessary friction. Enabling them by default would improve the out-of-box experience without any downside in terms of artifact size.

Let me know if you'd like me to provide any benchmarks or measurements to confirm the size impact.

YouthLab:
FYI
The artifacts(python pkg) will only increase about 10M.
Before:
image
After:
image

And if the size is Critical, Maybe we can distribute with multi-Flavor? (i.e. multiple pypi distribute project: lancedb lancedb-cjk lancedb-balabala)

BubbleCal (commented):
overall good to me
could you add documents to the class `FTS` at 'python/python/lancedb/index.py'?

wjones127:
> The artifacts(python pkg) will only increase about 10M.

That seems surprisingly high for what it is, doesn't it? That's a 25% increase!

> And if the size is Critical, Maybe we can distribute with multi-Flavor? 

Yeah size is already a pain point for us, and I don't want to make it meaningfully worse.

It would be nice to make a [plugin system](https://www.arroyo.dev/blog/rust-plugin-systems/#implementing-a-c-interface) for the [tantivy_tokenizer_api](https://docs.rs/tantivy-tokenizer-api/latest/tantivy_tokenizer_api/index.html), then these could be loaded as dynamic libraries.

YouthLab:
> > The artifacts(python pkg) will only increase about 10M.
> 
> That seems surprisingly high for what it is, doesn't it? That's a 25% increase!
> 
> > And if the size is Critical, Maybe we can distribute with multi-Flavor?
> 
> Yeah size is already a pain point for us, and I don't want to make it meaningfully worse.
> 
> It would be nice to make a [plugin system](https://www.arroyo.dev/blog/rust-plugin-systems/#implementing-a-c-interface) for the [tantivy_tokenizer_api](https://docs.rs/tantivy-tokenizer-api/latest/tantivy_tokenizer_api/index.html), then these could be loaded as dynamic libraries.

I am not familiar with the rust lang, but I will try later.
And anyone who wants to contribute to this matter will be welcome. : )

KazuhitoT:
@YouthLab cc: @BubbleCal @wjones127 @chenkovsky

Can I take over the tokenizer plugin work discussed in https://github.com/lance-format/lance/issues/3222?
I have an implementation that seems to work.

wjones127:
> I have an implementation that seems to work.

If you have a draft implementation I'd love to see it.

(I'll be on vacation for the next week but will be back December 30.)

KazuhitoT:
@wjones127 Thanks for your reply.
I'll create a draft PR this week, so you can take a look whenever you settle back in.

wjones127:
jieba/lindera already enabled in rust/lancedb/Cargo.toml; FTS tokenization shipped via #3659.

PR #2864, opened 2025-12-14
feat(python): enhance Pydantic integration with `List[LanceModel]` support

### LINKED ISSUE #2683: bug(python): Can not use list[LanceModel] inside LanceModel
### LanceDB version

0.25.1

### What happened?

I don't know if it's a bug or not, but `list[LanceModel]` is not a valid field type

```

class SubFeature1(LanceModel):
    amount: int
    name: str


class RandomFeature1(LanceModel):
    email: str
    items: list[SubFeature1]

print(RandomFeature1.to_arrow_schema())
```

Result

```
TypeError: Converting Pydantic type to Arrow Type: unsupported type <class '__main__.SubFeature1'>.
```

Perhaps list of LanceModel containing Vector field is supported but I think list of LanceModel which doesn't contain any Vector field should be supported

_No response_

Kkartik14:
Fixes #2683

list[LanceModel] was throwing TypeError: unsupported type when converting to Arrow schema.

## Solution

Changed 2 function calls to use `_pydantic_type_to_arrow_type()` instead of `_py_type_to_arrow_type()` when processing list child elements. This ensures Pydantic models are properly converted to struct types.

### DIFF
--- a/python/python/lancedb/pydantic.py
+++ b/python/python/lancedb/pydantic.py
@@ -275,7 +275,12 @@ def _py_type_to_arrow_type(py_type: Type[Any], field: FieldInfo) -> pa.DataType:
         return pa.timestamp("us", tz=tz)
     elif getattr(py_type, "__origin__", None) in (list, tuple):
         child = py_type.__args__[0]
-        return pa.list_(_pydantic_type_to_arrow_type(child, field))
+        # Check if child is a Pydantic model (class that inherits from BaseModel)
+        if inspect.isclass(child) and issubclass(child, pydantic.BaseModel):
+            # Convert Pydantic model to struct type
+            fields = _pydantic_model_to_fields(child)
+            return pa.list_(pa.struct(fields))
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
This all seems reasonable. Thanks for the contribution :)

Kkartik14:
Thanks for the approval @wjones127  I noticed the Python tests are still failing. Should I fix the issue or are you planning to handle it? Happy to push a fix if needed!

wjones127:
> I noticed the Python tests are still failing. Should I fix the issue or are you planning to handle it?

The failure there looks like it is related to your changes. Could you look into it?

Kkartik14:
@wjones127  The CI failure appears to be a manylinux compatibility issue during Rust compilation (`__cxa_thread_atexit_impl` symbol), unrelated to my Python code changes. Could you please take a look?

wjones127:
Superceded by https://github.com/lancedb/lancedb/pull/2953.

PR #2873, opened 2025-12-17
fix: pass namespace storage options provider into native table

jackye1995:
Previously the native table is created with static credentials and could not auto-refresh credentials when expired.


jackye1995:
breaking glass for a release

PR #2892, opened 2025-12-24
fix(python): require explicit region for S3 buckets with dots

### LINKED ISSUE #1898: bug(python): S3 storage access fails
### LanceDB version

v0.17.0b0

### What happened?

Access to S3 database fails with 
```
RuntimeError: lance error: LanceError(IO): Generic S3 error: Failed to resolve region for bucket 'test_bucket', /root/.cargo/git/checkouts/lance-b31243ab5673a03e/920b191/rust/lance-io/src/object_store.rs:255:13
```

This seems to be related to https://github.com/duckdb/duckdb_delta/issues/14
May be fixed with updated duckdb version

Set up AWS bucket 
```python
import lancedb
uri = 's3://test_bucket'
db = lancedb.connect(uri)

for tab in db.table_names():
     print(tab)
```

Angryrou:
When region is not specific in the s3 path, `resolve_s3_region` from "lance-format" project (see [here][1]) will resolve the region by calling `resolve_bucket_region`, which is a function from the "arrow-rs-object-store" project expecting [virtual-hosted-style URLs][1]. When there are dot (".") in the virtual-hosted-style URLs, it breaks automatic region detection. See more details in the issue description: https://github.com/lancedb/lancedb/issues/1898#issuecomment-3690142427

This PR add early validation in connect() and connect_async() to raise a clear error with instructions when the region is not specified for such buckets.

[1]: https://github.com/lance-format/lance/blob/v2.0.0-beta.4/rust/lance-io/src/object_store/providers/aws.rs#L197
[2]: https://github.com/apache/arrow-rs-object-store/blob/eedbf3d7d8c347eb114eeebc96cbf5b718f7b59b/src/aws/resolve.rs#L52C5-L52C65
[3]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#virtual-hosted-style-access

Fixes #1898

### DIFF
--- a/python/python/lancedb/__init__.py
+++ b/python/python/lancedb/__init__.py
@@ -13,6 +13,7 @@ __version__ = importlib.metadata.version("lancedb")
 
 from ._lancedb import connect as lancedb_connect
 from .common import URI, sanitize_uri
+from urllib.parse import urlparse
 from .db import AsyncConnection, DBConnection, LanceDBConnection
 from .io import StorageOptionsProvider
 from .remote import ClientConfig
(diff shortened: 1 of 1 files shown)


Angryrou:
@wjones127 could you please help review or point someone to review this PR?

PR #2896, opened 2026-01-05
feat: parallelize embedding computations

ex172000:
Implement parallel execution of multiple embedding functions using std::thread::scope to improve performance when a table has multiple embedding columns.

Key changes:
- Add compute_embeddings_parallel() helper method to WithEmbeddings
- Use fast path for single embeddings (no threading overhead)
- Use scoped threads for parallel execution of multiple embeddings
- Add comprehensive tests including parallelization timing verification
- Update WithEmbeddings documentation

Performance improvements:
- I/O-bound embeddings (OpenAI, Bedrock): High benefit from concurrent API calls
- CPU-bound embeddings (sentence-transformers): Medium benefit from core utilization
- Single embedding: No overhead (fast path)

Closes TODO on line 266 in rust/lancedb/src/embeddings.rs

### DIFF
--- a/rust/lancedb/tests/embeddings_parallel_test.rs
+++ b/rust/lancedb/tests/embeddings_parallel_test.rs
@@ -7,7 +7,7 @@ use std::{
         atomic::{AtomicUsize, Ordering},
         Arc,
     },
-    time::{Duration},
+    time::Duration,
 };
 
 use arrow::buffer::NullBuffer;


ex172000:
> Thanks for the PR. This looks good. Will merge if CI tests pass.

Thanks Will, seems the CI is slow, adjusted the timeout and appreciate if you can take another look!

PR #2901, opened 2026-01-07
fix: support `exist_ok` in `RemoteDBConnection.create_table`

cmccabe:
RemoteDBConnection should support passing exist_ok to create_table, just like LanceDBConnection (the non-remote form) does. It can support this by passing 'exist_ok' as the mode parameter.

### DIFF
--- a/python/python/lancedb/remote/db.py
+++ b/python/python/lancedb/remote/db.py
@@ -417,8 +417,8 @@ class RemoteDBConnection(DBConnection):
             The mode to use when creating the table.
             Can be either "create", "overwrite", or "exist_ok".
         exist_ok: bool, default False
-            If True, return the existing table if it already exists. This
-            is equivalent to using mode="exist_ok".
+            If exist_ok is True, and mode is None or "create", mode will be changed
+            to "exist_ok".
         on_bad_vectors: str, default "error"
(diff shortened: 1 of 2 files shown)


python/python/tests/test_remote_db.py
```diff
@@ -168,6 +169,34 @@ def handler(request):
...
+
+def test_create_table_exist_ok_invalid_mode_sync():
+    def handler(request):
+        raise AssertionError("Unexpected request")
+
+    with mock_lancedb_connection(handler) as db:
+        with pytest.raises(
+            ValueError, match=re.escape("exist_ok to True")
+        ):
+            print(f"DEBUG2: type of db is {type(db)}")
```
wjones127:
```suggestion
```

python/python/lancedb/remote/db.py
```diff
@@ -483,6 +490,12 @@ def create_table(
         LanceTable(table4)
 
         """
+        if exist_ok:
+            if mode is not None:
+                if mode != "exist_ok":
+                    raise ValueError(f"You cannot a mode argument of {mode} when setting exist_ok to True.")
```
rpgreen: Typo?

rpgreen (approved):

Issue #2909, opened 2026-01-12
bug(python): LanceDB unable to scan from hugging face URI


prrao87:
### LanceDB version

master

### What happened?

When trying to scan data from Hugging Face via LanceDB, @AyushExel gets the following error. It works for the Lance dataset API, but LanceDB needs to be able to support `hf://` path specification for LanceDB users to benefit from this integration.

```py
import lancedb
import lance 

ds = lance.dataset("hf://datasets/lance-format/lerobot_xvla-soft-fold/data/train.lance")
print(ds.count_rows())

db = lancedb.connect("hf://datasets/lance-format/lerobot_xvla-soft-fold/data")
print(db.table_names())
tbl = db["train"]
```

```
(openvid-hf) ayushchaurasia@Ayushs-MacBook-Air lance-huggingface % python test.py
2852512
Traceback (most recent call last):
  File "/Users/ayushchaurasia/Documents/lance_datasets/lance-huggingface/test.py", line 7, in <module>
    db = lancedb.connect("hf://datasets/lance-format/lerobot_xvla-soft-fold/data")
  File "/Users/ayushchaurasia/Documents/lance_datasets/lance-huggingface/openvid_hf/.venv/lib/python3.13/site-packages/lancedb/__init__.py", line 127, in connect
    return LanceDBConnection(
        uri,
    ...<2 lines>...
        session=session,
    )
  File "/Users/ayushchaurasia/Documents/lance_datasets/lance-huggingface/openvid_hf/.venv/lib/python3.13/site-packages/lancedb/db.py", line 611, in __init__
    self._conn = AsyncConnection(LOOP.run(do_connect()))
                                 ~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/ayushchaurasia/Documents/lance_datasets/lance-huggingface/openvid_hf/.venv/lib/python3.13/site-packages/lancedb/background_loop.py", line 25, in run
    return asyncio.run_coroutine_threadsafe(future, self.loop).result()
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
[...]

PR #2910, opened 2026-01-13
feat: enable huggingface feature by default

Xuanwo:
Close https://github.com/lancedb/lancedb/issues/2909

Enable hf feature by default so our python users can use `hf://xxxx` directly.

---

**Parts of this PR were drafted with assistance from Codex (with `gpt-5.2`) and fully reviewed and edited by me. I take full responsibility for all changes.**

### DIFF
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -104,11 +104,16 @@ test-log = "0.2"
 
 
 [features]
-default = ["aws", "gcs", "azure", "dynamodb", "oss"]
+default = ["aws", "gcs", "azure", "dynamodb", "oss", "huggingface"]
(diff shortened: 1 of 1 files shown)


prrao87:
This could potentially enable Rust SDK users too, right @wjones127? There is a subset of users in the HF community who build data infra in Rust, not all of them are Python users.

wjones127:
> This could potentially enable Rust SDK users too, right @wjones127?

Yeah, I think it's good to add the feature to Rust, but I don't think it is good to enable it by default. Now everyone who enables default features in Rust gets a bunch more dependencies they didn't ask for. We can enable it by default in Python without having to enable it by default in Rust.

prrao87:
Ok, let's just test it for now and revert this change to what you suggested if what we test works!

PR #2912, opened 2026-01-13
feat(rust)!: remove default features

### LINKED ISSUE #2911: Enable Hugging Face integration as default in Python, not Rust
As per https://github.com/lancedb/lancedb/pull/2910, the fix we tested works for Hugging Face.

However, as @wjones127 pointed out:
> I think it's good to add the feature to Rust, but I don't think it is good to enable it by default. Now everyone who enables default features in Rust gets a bunch more dependencies they didn't ask for. We can enable it by default in Python without having to enable it by default in Rust.

How can we do this so that Python users can immediately benefit (no extra configuration), but Rust users "opt in" to the HF integration rather than enforcing it as a default?

wjones127:
BREAKING CHANGE: removes `aws`, `dynamodb`, `azure`, `gcs`, `oss`, `huggingface` from default Rust features. They can be enabled by users as needed.

They are still enabled for Python and NodeJS, since those users don't control the compilation of artifacts.

Closes #2911

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -167,13 +167,13 @@ jobs:
       - name: Build
         run: |
           $env:VCPKG_ROOT = $env:VCPKG_INSTALLATION_ROOT
-          cargo build --profile ci --features remote --tests --locked --target ${{ matrix.target }}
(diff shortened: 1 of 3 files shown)


prrao87:
What does "breaking change" and "2.0" mean here? Not sure I understand why it would be a breaking change.

PR #2915, opened 2026-01-16
chore: migrate Gemini embeddings to use `google-genai` SDK

markmcd:
As the `google-generativeai` SDK is deprecated, it will emit warnings and will not have support for all models going forward, so moving to our new SDK will ensure some forward-compatibility for your users.

I used this [test script here](https://gist.github.com/markmcd/d83ce548561ca043fe274dd98a0cde7a) to validate everything works.

This also adds support for batching embeddings over a single API request.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -83,7 +83,7 @@ embeddings = [
     "colpali-engine>=0.3.10",
     "huggingface_hub",
     "InstructorEmbedding",
-    "google.generativeai",
+    "google-genai",
     "boto3>=1.28.57",
     "awscli>=1.29.57",
     "botocore>=1.31.57",
--- a/python/python/lancedb/embeddings/gemini_text.py
+++ b/python/python/lancedb/embeddings/gemini_text.py
@@ -10,6 +10,7 @@ import numpy as np
 
 from lancedb.pydantic import PYDANTIC_VERSION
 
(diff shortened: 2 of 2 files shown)


wjones127 (approved):

markmcd:
Are these failing tests anything to do with this change? Let me know if you need anything from me to close this one out @wjones127

wjones127:
Thanks for this, @markmcd — and apologies it sat open long enough to be overtaken.

The same migration has since landed on main via #3250 (merged), which moves gemini-text from the legacy SDK to google-genai. As a result this PR now conflicts throughout and is effectively superseded, so I'm going to close it.

That said, this PR still has two things #3250 doesn't:

1. Partner-attribution header — you set x-goog-api-client: lancedb/{__version__} on the client (per goo.gle/gemini-partners). The merged version doesn't send this.
2. Batched embedding — you send all texts in a single embed_content call, whereas the merged version loops one call per text.

Both are worth keeping. If you'd like, I'm happy to open a small follow-up PR on top of the current main adding the attribution header (and the batching), and credit you — or you're very welcome to open it yourself. Just let me know which you'd prefer.

Thanks again for the contribution!

markmcd:
Thanks @wjones127 - I've sent https://github.com/lancedb/lancedb/pull/3618 for your perusal.

PR #2921, opened 2026-01-19
feat: support add_columns with embedding function

### LINKED ISSUE #2920: Feature: support add_columns with embedding function
### SDK

Python

### Description

Currently, LanceDB can only specify the embedding function when creating a table. This cannot meet the needs of column addition scenarios. Here, we extend the `add_columns` API to support adding an embedding function in the scenario where new columns are added and historical data is backfilled.

```
clip = registry.get("open-clip").create(
    name="ViT-B-32",
    pretrained="laion2b_s34b_b79k",
    device="gpu" if torch.cuda.is_available() else "cpu",
)

conf = EmbeddingFunctionConfig(source_column="image", vector_column="vector", function=clip)

table.add_columns(conf)
```

zhangyue19921010:
Closes: https://github.com/lancedb/lancedb/issues/2920

Wrap python embedding function as BatchUDF


python/python/tests/test_table.py
```diff
@@ -1678,6 +1678,45 @@ def test_add_columns(mem_db: DBConnection):
...
+    assert "vector" not in table.schema.names
+
+    registry = EmbeddingFunctionRegistry.get_instance()
+    func = registry.get("sentence-transformers").create(max_retries=0)
+    conf = EmbeddingFunctionConfig(
+        source_column="text",
+        vector_column="vector",
+        function=func,
+    )
+    table.add_columns(conf)
```
wjones127, line 1699: suggestion: I wonder if we can treat the embedding function as just a general UDF. The API I'm thinking of might be something like:

```python
table.add_columns(
    {'vector': 'sentence_transformer(text)'},
    udfs={'sentence_transformer': func}
)
```

That way you could also support other non-embedding UDFs here. Think like:

```python
def make_prompt(title: str, description: str) -> str:
    return f"{title}: {description}"

table.add_columns(
    {'vector': 'sentence_transformer(make_prompt(title, description))'},
    udfs={'sentence_transformer': func, 'make_prompt': make_prompt}
)
```

wjones127: Though I guess you are registering it as an embedding that can be used later, so maybe it makes sense as-is.

python/src/table.rs
```diff
@@ -752,6 +748,81 @@ impl Table {
...
+            let mut dataset = dataset.get_mut().await.infer_error()?;
+
+            let mut manifest = (*dataset.manifest).clone();
+            let metadata: HashMap<String, String> = schema_metadata_pairs.into_iter().collect();
+            let new_schema = lance_core::datatypes::Schema {
+                fields: manifest.schema.fields.clone(),
+                metadata,
+            };
+            manifest.schema = new_schema;
+            dataset.manifest = Arc::new(manifest);
```
wjones127, line 821: issue(blocking): This mutates the manifest in-place. I don't see anywhere that you are updating the schema on-disk.

python/python/tests/test_table.py
```diff
@@ -1678,6 +1678,45 @@ def test_add_columns(mem_db: DBConnection):
     assert table.schema.field("null_int").type == pa.int64()
 
 
+@pytest.mark.slow
+def test_add_columns_embedding_function_config_sentence_transformers(
```
wjones127, line 1682: issue(blocking): marking this as a slow test means it won't be checked by default. Could you use the MockEmbeddingFunction so we can test it as part of the main test group? You can keep an integration tests with sentence transformers if you want, but I don't think it's necesssary.

python/src/table.rs
```diff
@@ -752,6 +748,81 @@ impl Table {
...
+                let out_batch = Python::with_gil(|py| {
+                    let batch_py = batch.to_pyarrow(py)?;
+                    let batch_py = handler.bind(py).call1((batch_py,))?;
+                    let out_batch = RecordBatch::from_pyarrow_bound(&batch_py)?;
+                    Ok::<_, pyo3::PyErr>(out_batch)
+                })
+                .map_err(|_| lance_core::Error::InvalidInput {
+                    source: "computing embedding failed".into(),
+                    location: snafu::location!(),
+                })?;
```
wjones127, line 790: issue(non-blocking): This swallows the Python error. The user won't be able to see the original traceback, which could make debugging harder. See if you can find a way to pass that through and roundtrip. If you can, this would be worth a unit test.

wjones127 (changes requested):
Sorry it took me a while to get to this. Noticed a couple issues, but all fixable.

PR #2922, opened 2026-01-19
feat: expose table uri

rpgreen:
* Expose `table.uri` property for all tables, including remote tables
* Fix bug in path calculation on windows file systems

### DIFF
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -1076,13 +1076,13 @@ impl Database for ListingDatabase {
 
 #[cfg(test)]
 mod tests {
-    use std::path::PathBuf;
     use super::*;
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -3332,4 +3356,69 @@ mod tests {
...
+            // Server returns response without location field
+            http::Response::builder()
+                .status(200)
+                .body(r#"{"version": 1, "schema": {"fields": []}}"#)
+                .unwrap()
+        });
+
+        let result = table.uri().await;
+        assert!(result.is_err());
+        assert!(matches!(&result, Err(Error::Runtime { .. })));
```
wjones127: Maybe this should be something like `Error::NotSupported`?

PR #2928, opened 2026-01-21
test: string type conversion in pandas 3.0+

jackye1995:
Pandas 3.0+ string now converts to Arrow large_utf8. This PR mainly makes sure our test accounts for the difference across the pandas versions when constructing schema.

### DIFF
--- a/python/python/tests/conftest.py
+++ b/python/python/tests/conftest.py
@@ -3,22 +3,22 @@
 
 from datetime import timedelta
 
-import pandas as pd
-import pyarrow as pa
-
 from lancedb.db import AsyncConnection, DBConnection
 import lancedb
 import pytest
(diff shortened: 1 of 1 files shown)


python/python/tests/conftest.py
```diff
@@ -2,12 +2,27 @@
...
 import lancedb
 import pytest
 import pytest_asyncio
 
 
+# pandas 3.0+ uses large_string for string columns, pandas 2.x uses string
+PANDAS_VERSION = tuple(int(x) for x in pd.__version__.split(".")[:2])
+
+
+def pandas_string_type():
```
jackye1995: we could just ping pandas to <3.0, but I felt this might be better so we are aware of pandas behavior in tests in case anything else breaks.

PR #2939, opened 2026-01-23
feat(rust): implement `TableProvider::insert_into()` for LanceDB tables

wjones127:
Implements `InsertExec` and `RemoteInsertExec` to support running inserts in DataFusion.

## Context

In https://github.com/lancedb/lancedb/pull/2929, I've prototyped moving the insert pipeline into DataFusion. This will enable parallelism at two levels:

1. Running preprocessing, such as casting the input schema or computing embeddings
2. Writing out files

This PR is just the first part of running the actual writes. In the end, the plans might look like:

```
InsertExec
  RepartitionExec num_partitions=<write_parallelism>
    ProjectionExec vector=compute_embedding()
      RepartitionExec num_partitions=<num_cpus>
        DataSourceExec
```

where `num_cpus` is used to take advantage of all cores, while `write_parallelism` might be less than `num_cpus` if there are too few rows to want to split writes across `num_cpus` files.

Later PRs will move the preprocessing steps into DataFusion, and then hook this up to the `Table::add()` implementations.

## Relation to future SQL work

We eventually plan on having the Remote SDK go through a FlightSQL endpoint. Then for most queries we will send just the SQL string to the server, and not run any sort of DataFusion plan on the client.

However, I think writes will be a little special, especially bulk writes where we need to upload large streams of data and likely want parallelism. So we'll have different code paths for writes, and I think using DataFusion makes sense, especially as long as we are doing the pre-processing on the client side still.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5001,6 +5001,7 @@ dependencies = [
  "datafusion-common",
  "datafusion-execution",
  "datafusion-expr",
+ "datafusion-physical-expr",
  "datafusion-physical-plan",
  "futures",
  "half",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ datafusion-common = { version = "50.1", default-features = false }
 datafusion-execution = "50.1"
 datafusion-expr = "50.1"
 datafusion-physical-plan = "50.1"
+datafusion-physical-expr = "50.1"
 env_logger = "0.11"
 half = { "version" = "2.6.0", default-features = false, features = [
     "num-traits",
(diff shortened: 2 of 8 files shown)


jackye1995 (approved):
the code looks good to me! 

One question: how are we thinking about cutting the line between what is in lance and what is in lancedb? Feels if we follow what is going on in merge-insert, then ideally insert in lance should also be driven by a DataFusion plan execution , and if we do that, it would have quite some overlap with a lot of what is provided here for non-remote execution.

wjones127:
> Feels if we follow what is going on in merge-insert, then ideally insert in lance should also be driven by a DataFusion plan execution , and if we do that, it would have quite some overlap with a lot of what is provided here for non-remote execution.

I think we can try to upstream what is generic to Lance, whenever possible. Like I said in the PR description, writes are a weird case where we actually want to run a plan on the client. On the query side, I think we can put all of the execution plans into Lance.

I think one arrangement we can do it have in LanceDB a `MergeInsert` logical node, and then for local table it transforms into the Lance plan while for remote tables it converts into something custom that communicates with remote servers.

PR #2948, opened 2026-01-27
feat(rust)!: accept `RecordBatch`, `Vec<RecordBatch>` in `create_table()` and `Table.add()`

wjones127:
BREAKING CHANGE: Arbitrary `impl RecordBatchReader` is no longer accepted, it must be made into `Box<dyn RecordBatchReader>`.

This PR replaces `IntoArrow` with a new trait `Scannable` to define input row data. This provides the following advantages:

1. **We can implement `Scannable` for more types than `IntoArrow`, such as `RecordBatch` and `Vec<RecordBatch>`.** The `IntoArrow` trait was implemented for arbitrary `T: RecordBatchReader`, and the Rust compiler would prevent us from implementing it for foreign types like `RecordBatch` because (theoretically) those types might implement `RecordBatchReader` in the future. That's why we implement `Scannable` for `Box<dyn RecordBatchReader>` instead; since it's a concrete type it doesn't block implementing for other foreign types.
2. **We can potentially replay `Scannable` values**. Previously, we had to choose between buffering all data in memory and supporting retries of writes. But because `Scannable` things can optionally support re-scanning, we now have a way of supporting retries while also streaming.
3. **`Scannable` can provide hints like `num_rows`, which can be used to schedule parallel writers.** Without knowing the total number of rows, it's difficult to know whether it's worth writing multiple files in parallel.

We don't yet fully take advantage of (2) and (3) yet, but will in future PRs. For (2), in order to be ready to leverage this, we need to hook the `Scannable` implementation up to Python and NodeJS bindings. Right now they always pass down a stream, but we want to make sure they support retries when possible. And for (3), this will need to be hooked up to #2939 and to a pipeline for running pre-processing steps (like embedding generation).

## Other changes

* Moved `create_table` and `add_data` into their own modules. I've created a follow up issue to split up `table.rs` further, as it's by far the largest file: https://github.com/lancedb/lancedb/issues/2949
* Eliminated the `HAS_DATA` generic for `CreateTableBuilder`. I didn't see any public-facing places where we differentiated methods, which is why I felt this simplification was okay.
* Added an `Error::External` variant and integrated some conversions to allow certain errors to pass through transparently. This will fully work once we upgrade Lance and get to take advantage of changes in https://github.com/lance-format/lance/pull/5606
* Added LZ4 compression support for write requests to remote endpoints. I checked and this has been supported on the server for > 1 year.


python/src/connection.rs
```diff
@@ -121,7 +121,8 @@ impl Connection {
 
         let mode = Self::parse_create_mode_str(mode)?;
 
-        let batches = ArrowArrayStreamReader::from_pyarrow_bound(&data)?;
+        let batches: Box<dyn arrow::array::RecordBatchReader + Send> =
+            Box::new(ArrowArrayStreamReader::from_pyarrow_bound(&data)?);
```
westonpace, line 125: What if we had a `Scannable::from_reader` method?  Something like...

```
fn from_reader<T: RecordBatchReader>(reader: T) -> impl Scannable {
  Box::new(reader)
}
```

Then it would be...

```
let batches = Scannable::from_reader(ArrowArrayStreamReader::from_pyarrow_bound(&data)?);
```

I'm not sure if that is more readable or not.

rust/lancedb/examples/openai.rs
```diff
@@ -64,26 +62,20 @@ async fn main() -> Result<()> {
...
-        "Winter Parka",
-        "Hooded Sweatshirt",
-    ]);
-    let price = Float64Array::from(vec![10.0, 50.0, 100.0, 30.0]);
-    let schema = Arc::new(schema);
-    let rb = RecordBatch::try_new(
-        schema.clone(),
-        vec![Arc::new(id), Arc::new(text), Arc::new(price)],
+fn make_data() -> RecordBatch {
+    arrow_array::record_batch!(
```
westonpace, line 66: Thanks for switching to `record_batch!`

rust/lancedb/src/data/scannable.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// and subsequent calls return an empty stream.
+    fn scan_as_stream(&mut self) -> SendableRecordBatchStream;
+
+    /// Optional hint about the number of rows.
+    ///
+    /// When available, this allows the pipeline to estimate total data size
+    /// and choose appropriate partitioning.
+    fn num_rows(&self) -> Option<usize> {
+        None
+    }
```
westonpace, line 48: How would you then partition it?  Wouldn't you need to support something like `offset` and `limit` in `scan_as_stream`?

rust/lancedb/src/data/scannable.rs
```diff
@@ -0,0 +1,683 @@
...
+        self.inner.num_rows()
+    }
+
+    fn rescannable(&self) -> bool {
+        self.inner.rescannable()
+    }
+}
+
+/// A scannable that might have embeddings applied to it.
+pub enum MaybeEmbeddedScannable {
```
westonpace: Can you expand on this comment?  Maybe provide some motivation why this might be needed?

rust/lancedb/src/connection.rs
```diff
@@ -656,35 +351,17 @@ impl Connection {
     ///
     /// * `name` - The name of the table
     /// * `initial_data` - The initial data to write to the table
-    pub fn create_table<T: IntoArrow>(
+    pub fn create_table<T: Scannable + 'static>(
```
westonpace, line 354: This breaking change is going to be significant.  The method name stays the same but the meaning has changed.  I don't see an easy way to avoid it unless we want to introduce a `create_table_new` or something (which we probably don't).

rust/lancedb/src/connection.rs
```diff
@@ -656,35 +351,17 @@ impl Connection {
...
-        )
-    }
-
-    /// Create a new table from a stream of data
-    ///
-    /// # Parameters
-    ///
-    /// * `name` - The name of the table
-    /// * `initial_data` - The initial data to write to the table
-    pub fn create_table_streaming<T: IntoArrowStream>(
```
westonpace, line 706: Can we keep this method (and maybe emit a deprecation warning?)  We'd have to keep IntoArrowStream around but could adapt it to `Scannable` maybe?

python/src/connection.rs
wjones127: Something like that would be nice. I am finding the boxing of readers to be tedious.

rust/lancedb/src/data/scannable.rs
wjones127: Either round-robin the batches or, for merge-insert, hash partition them on the join key. `num_rows` helps us estimate the target number of partitions to use so we don't get too many fragments.

rust/lancedb/src/data/scannable.rs
```diff
@@ -0,0 +1,683 @@
...
+    fn schema(&self) -> SchemaRef;
+
+    /// Read data as a stream of record batches.
+    ///
+    /// For rescannable sources (in-memory data like RecordBatch, Vec<RecordBatch>),
+    /// this can be called multiple times and returns cloned data each time.
+    ///
+    /// For non-rescannable sources (streams, readers), the first call returns data
+    /// and subsequent calls return an empty stream.
+    fn scan_as_stream(&mut self) -> SendableRecordBatchStream;
```
wjones127, line 40: TODO: we should accept a `batch_size` parameter.

wjones127: I'll do this as a follow up. For the types we implement so far, the benefit isn't yet clear.

rust/lancedb/src/data/scannable.rs
wjones127: I think we don't actually need it. I will remove it.

python/src/connection.rs
wjones127: I think you can't implement a method on a trait like that, so instead I just made a function `box_reader()` that does the boxing. Seems to work okay.

westonpace: Could `box_reader` be a method in some `ScannerExt` trait that we give a blanket implementation for record batch readers (I'm thinking of something like `FutureExt` and `StreamExt` which have a `boxed` method)?  Or does that get us back into the same original problem?

Anyways, this is probably bike shedding at this point.  Don't feel the need to keep digging here.

rust/lancedb/src/data/scannable.rs
westonpace: Doesn't have to be done now but I think more details are needed here.  I'm not convinced there is enough here yet.  For example, let's say I see 10M rows and decide to do 10 partitions.  If I can't communicate that to the `Scannable` then how does it know what to use for round robin?  Or are you saying we create a single stream and then poll the stream from multiple readers?  A stream's poll method needs `&mut self` (more or less) and so this can't really be done.

The way we support round-robin in the file reader is the read method returns a stream of futures (read tasks).  This initial poll to get the future is a choke-point (intentionally) where only one reader can be in at a time but the resulting future can be farmed off to different threads.

Since this is returning a stream of batches I don't think round robin is possible here.

In datafusion we have the `execute` method which takes an equivalent `partition` identifier.  Do we need something like this in `Scannable`?

rust/lancedb/src/data/scannable.rs
westonpace: Would making a `ScanOptions` struct (especially if we need partition) be a good idea for future-proofing?

python/src/connection.rs
wjones127: Oh hmm, let me see.

rust/lancedb/src/data/scannable.rs
wjones127: Yeah, right now we only support a single-threaded stream as input. This stream will be provided as a single-partition input to a DataFusion plan that runs the write pipeline. This pipeline will perform preprocessing steps, such as running embedding functions and casting to the target table schema, and then finally writing to data files / to the remote server.

For example, here's what the pipeline will look like as a DataFusion plan. Say we have 32 CPUs, and we know we are writing 10 million files, so will have 10 parallel writes. So we first repartition to match the number of CPUs so we can do the compute-heavy work of computing the embeddings. Then we repartition again to match the number of files we want to write before passing into InsertExec.

```
InsertExec
  RepartitionExec num_partitions=10
    ProjectionExec vector=compute_embedding()
      RepartitionExec num_partitions=32
        ReadScannableExec (1 partition)
```

The Scannable might internally have threading that might allow it to have parallelism in the pipeline. For example, I believe both of the following will produce a single output stream but internally will use parallelism:

* `pyarrow.dataset.Dataset.scanner().to_reader()`
* `duckdb.sql("SELECT * FROM my_arrow_table").arrow(chunk_size)`

> In datafusion we have the `execute` method which takes an equivalent `partition` identifier. Do we need something like this in `Scannable`?

I think this a future direction we will want to look into. Exchanging batches through a single-threaded stream might be a bottle neck. For now, I lean toward keeping the API simple. Later we can add a partition concept to Scannable and see if that gives us a performance improvement.

Issue #2949, opened 2026-01-28
Refactor: split up table.rs


wjones127:
This file has over 5k lines.

Can put implementation and tests of individual operations in submodules, like so:

```
/rust/lancedb/src/table.rs
              .../table/add_data.rs
              .../table/delete.rs
              ...
```

* [x] add_data.rs
* [x] delete.rs
* [x] update.rs
* [x] merge_insert.rs
* [x] schema_evolution.rs (add, drop, alter column)
* [ ] optimize.rs
* [ ] create_index.rs
* [x] query.rs

We should move the code, but keep the APIs the same. So each of the submodule should be private and we can re-export the members in table.rs.

## Improving tests

As we refactor, we should move over the tests, and find opportunities to improve testing. This means both improving the existing tests (often making them more concise) and adding new test coverage.

Improvements for existing tests often include:

* Switching to `memory://` URI so you don't have to create a temporary directory
* Switch to creating literal data with `arrow_array::record_batch!()` macro
* Parameterizing test cases

New test coverage often includes:

* Covering methods on builders not yet tested
* Testing common error cases
* Testing both local and remote database

ChinmayGowda71:
Hi @wjones127 I'd like to help with this. Since add_data.rs is checked, I can start by refactoring delete.rs and others into their own submodule and migrating the relevant tests to match. Does that sound good?

wjones127:
Yeah feel free to make a PR working on one of these. I'd prefer to start with just doing one at a time, rather than a whole bunch, just to make the individual PRs easier to review.

I'm doing `add_data` in https://github.com/lancedb/lancedb/pull/2948. `delete` would be a good one to start with.

ChinmayGowda71:
I submitted a pull request for delete I ran a simple delete test it worked properly and I was able to refactor it do let me know if theres anything wrong I can fix it. 

I'll get to update now if theres no issues

ChinmayGowda71:
Fixed the two up based on your comments thank you for the pointers lemme know if anything else needs to be changed I'll get to merge insert

Abhisheklearn12:
Hi @wjones127 and @ChinmayGowda71, I'd like to work on schema_evolution.rs (add, drop, alter column) and optimize.rs, one at a time. Since @ChinmayGowda71 is handling delete.rs, update.rs, and merge_insert.rs, I thought these would be a good fit for me. let me know if that works for you both.

ChinmayGowda71:
[...]

PR #2952, opened 2026-01-29
refactor: modularize table.rs and extract delete logic

ChinmayGowda71:
References #2949 Moved DeleteResult and delete() implementation to src/table/delete.rs. No functional changes. Added a test delete which works. Will work on refactoring update next.

### DIFF
--- a/rust/lancedb/src/table/delete.rs
+++ b/rust/lancedb/src/table/delete.rs
@@ -34,7 +34,7 @@ pub(crate) async fn execute_delete(table: &NativeTable, predicate: &str) -> Resu
 #[cfg(test)]
 mod tests {
     use crate::connect;
-    use arrow_array::{Int32Array, RecordBatch, RecordBatchIterator, record_batch};
+    use arrow_array::{record_batch, Int32Array, RecordBatch, RecordBatchIterator};
     use arrow_schema::{DataType, Field, Schema};
     use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/table/delete.rs
```diff
@@ -0,0 +1,96 @@
...
+    use arrow_schema::{DataType, Field, Schema};
+    use std::sync::Arc;
+
+    use crate::query::ExecutableQuery;
+    use futures::TryStreamExt;
+    #[tokio::test]
+    async fn test_delete_simple() {
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let conn = connect(uri).execute().await.unwrap();
```
wjones127:
```suggestion
        let conn = connect("memory://").execute().await.unwrap();
```

rust/lancedb/src/table/delete.rs
```diff
@@ -0,0 +1,96 @@
...
+    dataset.delete(predicate).await?;
+
+    // Return the result with the new version
+    Ok(DeleteResult {
+        version: dataset.version().version,
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
wjones127, line 35: Could we add two other tests:

1. `table.delete("true")` -> All rows are removed, but schema remains the same.
2. `table.delete("false")` -> No change to row data, but table version is incremented.

ChinmayGowda71:
Added the two tests and changed it to run in memory as well

wjones127:
Thanks @ChinmayGowda71 !

PR #2960, opened 2026-01-30
chore!: drop pydantic 1 support

eddyxu:
drop pydantic version 1 support


.github/workflows/python.yml
```diff
@@ -195,31 +195,3 @@ jobs:
       # Make sure wheels are not included in the Rust cache
       - name: Delete wheels
         run: rm -rf target/wheels
-  pydantic1x:
```
wjones127, line 198: Could you keep a version of this job around called 'min-deps' or something? we also use this to test the minimum version of `pyarrow`, for example.

eddyxu: will do

wjones127:
Closing this for now. We can re-open this later if we want.

PR #2961, opened 2026-01-30
fix(rust): support embeddings in create_empty_table

### LINKED ISSUE #2759: Rust: Docs discrepancy for creating empy tables between Rust and Python.
### Description

tl;dr: Rust SDK's `Connection::create_empty` does not support setting embedding columns. It purely uses `TableDefinition::new_from_schema(schema)` which marks each column as physical.

The Python SDK does some funny things here. It does not use `create_empty_table` which you would assume when looking at `AsyncConnection.create_table(data=None, schema=<Something derive from e.g. pydantic>)`

https://github.com/lancedb/lancedb/blob/0b7b27481eca8742c299c4ba3b814dee82924b9e/python/python/lancedb/db.py#L1154-L1173

Instead, `sanitize_create_table` will create an empty record batch and return it as `data`.

Let's take a look at the following Rust example:
```rust
let schema = Arc::new(Schema::new(Fields::from_iter([Arc::new(Field::new(
        "name",
        DataType::Utf8,
        true,
    ))])));
let ed = EmbeddingDefinition {
        source_column: "name".to_owned(),
        dest_column: Some("name_embedding".to_owned()),
        embedding_name: "something-registered".to_owned(),
    };
let table = connection
        .create_empty_table("test", schema)
        .mode(CreateTableMode::Overwrite)
        .add_embedding(ed)
        .unwrap();
```

I would expect that `Table::add` would now create embeddings for my input batch. Instead the created table does not have a `name_embedding` column.

Having a look at Python, `schema` is generated, e.g. based on the Pydantic model. The schema passed to Rust's `create_table` is complete, including the embedding fields in the Arrow schema and metadata for the table definition. 
Trying to create a similar Schema in Rust and pass it to `create_empty_table` does not work because `TableDefinition::new_from_schema` erases the whole table definition, and I assume that is the reason Python doesn't use this and instead creates an empty batch to make use of `CreateTableBuilder::into_request` which hands the embedding function definitions correctly to WithEmbeddings.

Nothing in the Rust docs tells me that I am not able to create embedding columns using `create_empty_table`, and based on the public API design, I think the internal works seem a bit off.
`CreateTableBuilder<false>::new` and `CreateTableBuilder<false>::execute` should support adding embeddings.

### Link

_No response_

Mesut-Doner:
Fixes the Rust SDK's `create_empty_table` to properly support embedding column definitions, bringing it to parity with the Python SDK.

## Problem

The Rust SDK's `Connection::create_empty_table` did not support setting embedding columns. When using `.add_embedding()` on the builder, the embedding column definitions were lost because `TableDefinition::new_from_schema(schema)` marks all columns as physical only, without embedding metadata.

The Python SDK worked around this by creating an empty record batch with proper schema metadata rather than using `create_empty_table` directly.

## Solution
Modified `CreateTableBuilder<false>` to handle embeddings

Closes #2759

### DIFF
--- a/rust/lancedb/src/connection.rs
+++ b/rust/lancedb/src/connection.rs
@@ -251,8 +251,36 @@ impl CreateTableBuilder<false> {
     /// Execute the create table operation
     pub async fn execute(self) -> Result<Table> {
         let parent = self.parent.clone();
-        let table = parent.create_table(self.request).await?;
-        Ok(Table::new(table, parent))
+        let embedding_registry = self.embedding_registry.clone();
+        let request = self.into_request()?;
+        Ok(Table::new_with_embedding_registry(
+            parent.create_table(request).await?,
+            parent,
+            embedding_registry,
+        ))
+    }
+
+    fn into_request(self) -> Result<CreateTableRequest> {
+        if self.embeddings.is_empty() {
+            return Ok(self.request);
+        }
+
+        let CreateTableData::Empty(table_def) = self.request.data else {
+            unreachable!("CreateTableBuilder<false> should always have Empty data")
+        };
+
+        let schema = table_def.schema.clone();
+        let empty_batch = arrow_array::RecordBatch::new_empty(schema.clone());
+
+        let reader = Box::new(std::iter::once(Ok(empty_batch)).collect::<Vec<_>>());
+        let reader = arrow_array::RecordBatchIterator::new(reader.into_iter(), schema);
+        let with_embeddings = WithEmbeddings::new(reader, self.embeddings);
+        let table_definition = with_embeddings.table_definition()?;
+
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Thanks for doing this! I was noticing this gap recently.

valkum:
It seems this didn't fix all of the issues of my issue https://github.com/lancedb/lancedb/issues/2759.
`CreateTableData::schema` returns a rich schema for the non-empty variants (constructed by WithEmbeddings::schema) and a shallow schema (without "lancedb::column_definitions") for the empty variant. <s> Will provide a PR shortly.</s> Might be fixed by https://github.com/lancedb/lancedb/commit/c0230f91d2518cea3a866dc609aac5051551b5e1

PR #2962, opened 2026-01-30
feat(python): expose fast_search in synchronous API (Fixes #2612)

### LINKED ISSUE #2612: Feature: fast_search from table
### SDK

Python

### Description

Currently, when using the Python SDK (v0.24.3), enabling fast search is a bit opaque.

```python
    query = (
        embeddings_table.search(
            np.random.rand(2560),
            vector_column_name="image_embedding",
        )
        .distance_type("cosine")
        .select(["id", "_distance"])
        .limit(100)
    )
    query._fast_search = True
    results = query.to_pandas()
```

Preferable would be if the following worked:
```python
    results = (
        embeddings_table.search(
            np.random.rand(2560),
            vector_column_name="image_embedding",
        )
        .fast_search()
        .distance_type("cosine")
        .select(["id", "_distance"])
        .limit(100)
        .to_pandas()
    )
```

Ra5hidIslam:
Fixes #2612

This PR exposes the private _fast_search attribute via a public fast_search() method in the synchronous LanceVectorQueryBuilder.

Previously, enabling fast search in the sync API required accessing a private member (query._fast_search = True). This change aligns the synchronous API with the Async and Remote APIs, allowing for cleaner, more Pythonic method chaining.

Changes:
Added fast_search() method to LanceVectorQueryBuilder in python/python/lancedb/query.py.
Added a unit test verifying the flag works with high-dimensional data (2560 dims) and chaining.
Example Usage:

Before:

```
query = table.search(vector)
query._fast_search = True  # Private attribute usage
results = query.limit(10).to_pandas()
```

After:

```
results = (
    table.search(vector)
    .fast_search()
    .limit(10)
    .to_pandas()
)
```

Verification:
I have added a test case (test_fast_search_high_dimension) that replicates the scenario described in the issue (2560 dimensions, cosine distance) to ensure the pipeline constructs the query correctly without errors.

Checklist:

- [ ]  I have added tests to cover my changes.
- [ ]  All new and existing tests passed.
- [ ]  Documentation has been updated (inline docstrings).

### DIFF
--- a/python/python/tests/test_query.py
+++ b/python/python/tests/test_query.py
@@ -1499,3 +1499,30 @@ def test_search_empty_table(mem_db):
     # Search on empty table should return empty results, not crash
     results = table.search([1.0, 2.0]).limit(5).to_list()
     assert results == []
+
+
+def test_fast_search(tmp_path):
+    db = lancedb.connect(tmp_path)
+
+    # Generate data matching the async test style
+    vectors = pa.FixedShapeTensorArray.from_numpy_ndarray(
+        np.random.rand(256, 32)
+    ).storage
+
+    table = db.create_table("test", pa.table({"vector": vectors}))
+
+    # FIX: Pass arguments directly instead of using 'config=IvfPq(...)'
+    table.create_index(vector_column_name="vector", num_partitions=1, num_sub_vectors=1)
+
+    # Add data to ensure table has enough segments/rows
+    table.add(pa.table({"vector": vectors}))
+
+    q = [1.0] * 32
+
+    # 1. Normal Search -> Should include "LanceScan" (Brute Force / Scan)
+    plan = table.search(q).explain_plan(True)
+    assert "LanceScan" in plan
+
+    # 2. Fast Search -> Should NOT include "LanceScan" (Uses Index)
+    plan = table.search(q).fast_search().explain_plan(True)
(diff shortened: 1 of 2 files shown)


python/python/tests/test_query.py
```diff
@@ -1499,3 +1499,48 @@ def test_search_empty_table(mem_db):
     # Search on empty table should return empty results, not crash
     results = table.search([1.0, 2.0]).limit(5).to_list()
     assert results == []
+
+
+def test_fast_search_high_dimension(tmp_path):
```
wjones127: Could you instead write a test that imitates this one from the async tests? I think using `explain_plan` is a more direct check that `fast_search` is working.

https://github.com/lancedb/lancedb/blob/9be28448f56259315df0a9d9f8e7d7bc7dcf69d9/python/python/tests/test_query.py#L756-L772

Ra5hidIslam: Sure, will update once done.

Ra5hidIslam: Hi @wjones127 I have changed the test.

Ra5hidIslam:
Hi @wjones127 , I believe I have fixed the issues regarding the test failures. The CI checks should pass now. Ready for review!

wjones127 (approved):

PR #2964, opened 2026-01-30
refactor: extract update logic to src/table/update.rs

ChinmayGowda71:
References #2949 Part 2 of table.rs refactor. Moved UpdateResult, UpdateBuilder, and execution logic to src/table/update.rs. No functional changes API remains identical.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -77,11 +77,13 @@ use self::merge::MergeInsertBuilder;
 
 pub mod datafusion;
 pub(crate) mod dataset;
+pub mod delete;
 pub mod merge;
-
 pub mod update;
+
 use crate::index::waiter::wait_for_index;
 pub use chrono::Duration;
+pub use delete::DeleteResult;
 use futures::future::{join_all, Either};
 pub use lance::dataset::optimize::CompactionOptions;
 pub use lance::dataset::refs::{TagContents, Tags as LanceTags};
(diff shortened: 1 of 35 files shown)


rust/lancedb/src/table.rs
wjones127, line 1: Could you move the existing tests over to the new module. Here are some tests I see for update:

* test_update_with_predicate
* test_update_all_types
* test_update_via_expr

rust/lancedb/src/table/update.rs
```diff
@@ -0,0 +1,274 @@
...
+                assert_eq!(val, 1, "Row with id {} should stay 1", id);
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_update_string() {
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let conn = connect(uri).execute().await.unwrap();
```
wjones127: Unless a test needs to read files from disk, it's best to use the `memory://` URI. (These in-memory databases are distinct per call, so no need to provide a unique URI.)
```suggestion
        let conn = connect("memory://").execute().await.unwrap();
```

rust/lancedb/src/table/update.rs
```diff
@@ -0,0 +1,274 @@
...
+        let batch = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(Int32Array::from_iter_values(0..5)),
+                Arc::new(StringArray::from(vec![
+                    "alice", "bob", "charlie", "dave", "eve",
+                ])),
+            ],
+        )
+        .unwrap();
```
wjones127: It's much more succinct to use the `arrow_array::record_batch!()` macro, so I suggest you use that in tests whenever we need to create a literal batch.
```suggestion
        let batch = record_batch!(
            ("id", Int32, [0, 1, 2, 3, 4]),
            ("name", Utf8, ["alice", "bob", "charlie", "dave", "eve"])
        ).unwrap();
        let schema = batch.schema();
```

wjones127 (changes requested):
Nice work so far. I have a few pointers.

ChinmayGowda71:
Since the tests in table.rs were more all encompassing and made my tests redundant I added those, changed it to use memory:// when connecting, as for record_batch! I implemented it on test_update_with_predicate() successfully but couldn't do it for test_update_all_types() due to it not being able to implement date32 and vec32 datatypes and etc

PR #2965, opened 2026-01-31
fix: include _rowid in hash and calculated split projections

wjones127:
## Summary

- PR #2957 changed the permutation builder to only select `_rowid` from the base table, but `Splitter::project()` for hash and calculated splits replaced the selection entirely, dropping `_rowid`.
- Include `_rowid` in the column selections for hash and calculated split projections.
- Fix a Python test that queried the permutation table for base table columns no longer materialized.

Fixes the `test_split_hash`, `test_split_hash_with_discard`, `test_split_calculated`, `test_shuffle_combined_with_splits`, and `test_filter_with_splits` failures in `test_permutation.py`.

## Test plan

- [x] `cargo test -p lancedb -- permutation` (22 passed)
- [x] `pytest python/tests/test_permutation.py` (46 passed)
- [x] `npm test __test__/permutation.test.ts` (20 passed)

### DIFF
--- a/python/python/tests/test_permutation.py
+++ b/python/python/tests/test_permutation.py
@@ -441,6 +441,13 @@ def test_filter_with_splits(mem_db):
     # Verify the permutation table only contains row_id and split_id
     assert set(permutation_tbl.schema.names) == {"row_id", "split_id"}
 
+    row_ids = permutation_tbl.search(None).to_arrow().to_pydict()["row_id"]
+    data = tbl.take_row_ids(row_ids).to_arrow().to_pydict()
+    categories = data["category"]
+
+    # All categories should be A or B
+    assert all(cat in ("A", "B") for cat in categories)
+
 
 def test_filter_with_shuffle(mem_db):
     """Test filtering combined with shuffling."""


python/python/tests/test_permutation.py
```diff
@@ -438,11 +438,8 @@ def test_filter_with_splits(mem_db):
     row_count = permutation_tbl.count_rows()
     assert row_count == 67
 
-    data = permutation_tbl.search(None).to_arrow().to_pydict()
-    categories = data["category"]
-
-    # All categories should be A or B
-    assert all(cat in ["A", "B"] for cat in categories)
+    # Verify the permutation table only contains row_id and split_id
+    assert set(permutation_tbl.schema.names) == {"row_id", "split_id"}
```
westonpace, line 442: Hmm, this isn't quite testing the same thing but it's probably fine that the row_count is 67.  Otherwise we'd have to do a take on the row ids to get the category and make sure there is only A & B.

wjones127: It might be nice to do the take. I'll try that now.

PR #2973, opened 2026-02-03
refactor: extract schema evolution logic from table.rs into submodule

Abhisheklearn12:
Continues the modularization effort of schema evolution operations as outlined in #2949             

## Summary                                                                                                                           
  - Extracts schema evolution operations (add_columns, alter_columns, drop_columns) from `table.rs` into `table/schema_evolution.rs`                                                      
  - Public API remains unchanged via re-exports                                                                                        
                                                                                                                                       
  ## Test plan                                                                                                                         
  - [x] All new schema evolution tests pass                                                                                         
  - [x] All existing tests pass                                                                                                    
  - [x] `cargo clippy` passes with no warnings                                                                                         
  - [x] `cargo fmt --check` passes

### DIFF
--- a/rust/lancedb/src/table/schema_evolution.rs
+++ b/rust/lancedb/src/table/schema_evolution.rs
@@ -240,14 +240,11 @@ mod tests {
             .unwrap();
 
         let batch = &batches[0];
-        let values: Vec<i64> = batch
-            .column(0)
+        let values = batch["constant"]
             .as_any()
             .downcast_ref::<arrow_array::Int64Array>()
             .unwrap()
-            .iter()
-            .map(|v| v.unwrap())
-            .collect();
+            .values();
         assert!(values.iter().all(|&v| v == 42));
     }
 
@@ -358,6 +355,23 @@ mod tests {
             schema.field_with_name("num").unwrap().data_type(),
             &DataType::Int64
         );
+
+        // Query the data and verify the returned type is correct
+        let batches = table
+            .query()
+            .execute()
+            .await
+            .unwrap()
+            .try_collect::<Vec<_>>()
+            .await
+            .unwrap();
+        let batch = &batches[0];
+        let values = batch["num"]
+            .as_any()
+            .downcast_ref::<arrow_array::Int64Array>()
+            .unwrap()
+            .values();
+        assert_eq!(values.as_ref(), &[1i64, 2, 3]);
     }
 
     #[tokio::test]
@@ -380,7 +394,12 @@ mod tests {
         let result = table
             .alter_columns(&[ColumnAlteration::new("num".into()).cast_to(DataType::Float64)])
             .await;
-        assert!(result.is_err());
+        let err = result.unwrap_err();
+        assert!(
+            err.to_string().contains("cast"),
+            "Expected error message to contain 'cast', got: {}",
+            err
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/table/schema_evolution.rs
```diff
@@ -0,0 +1,637 @@
...
+            .unwrap();
+
+        // Cast Int32 to Int64 (a supported cast)
+        table
+            .alter_columns(&[ColumnAlteration::new("num".into()).cast_to(DataType::Int64)])
+            .await
+            .unwrap();
+
+        // Verify type changed
+        let schema = table.schema().await.unwrap();
```
wjones127, line 353: It's probably worth querying the data and asserting the returned type is correct too.

rust/lancedb/src/table/schema_evolution.rs
```diff
@@ -0,0 +1,637 @@
...
+            )
+            .execute()
+            .await
+            .unwrap();
+
+        // Casting Int32 to Float64 is not supported
+        let result = table
+            .alter_columns(&[ColumnAlteration::new("num".into()).cast_to(DataType::Float64)])
+            .await;
+        assert!(result.is_err());
```
wjones127: Could you assert the error message matching some substring? Want to makes sure we are getting a useful error.

rust/lancedb/src/table/schema_evolution.rs
```diff
@@ -0,0 +1,637 @@
...
+                "test_drop_nonexistent",
+                RecordBatchIterator::new(vec![Ok(batch)], schema),
+            )
+            .execute()
+            .await
+            .unwrap();
+
+        // Try to drop a column that doesn't exist
+        let result = table.drop_columns(&["nonexistent"]).await;
+        assert!(result.is_err());
```
wjones127: Also want to assert the error message here as well.

rust/lancedb/src/table/schema_evolution.rs
```diff
@@ -0,0 +1,637 @@
...
+            )
+            .execute()
+            .await
+            .unwrap();
+
+        // Try to alter a column that doesn't exist
+        let result = table
+            .alter_columns(&[ColumnAlteration::new("nonexistent".into()).rename("new".into())])
+            .await;
+        assert!(result.is_err());
```
wjones127: Same here: assert message

rust/lancedb/src/table/schema_evolution.rs
```diff
@@ -0,0 +1,637 @@
...
+        let batch = &batches[0];
+        let values: Vec<i64> = batch
+            .column(0)
+            .as_any()
+            .downcast_ref::<arrow_array::Int64Array>()
+            .unwrap()
+            .iter()
+            .map(|v| v.unwrap())
+            .collect();
+        assert!(values.iter().all(|&v| v == 42));
```
wjones127: FYI: you can use `batch[col_name]` in tests as a short-hand for getting a column. This panics, so not good in implementation code but good for tests.

Also, instead of collecting into a vec, I would recommend just accessing the values. That can be interpreted as a slice and thus can be iterated.
```suggestion
        let values: Vec<i64> = batch["constant"]
            .as_any()
            .downcast_ref::<arrow_array::Int64Array>()
            .unwrap()
            .values();
        assert!(values.iter().all(|&v| v == 42));
```

rust/lancedb/src/table/schema_evolution.rs
Abhisheklearn12: sure

rust/lancedb/src/table/schema_evolution.rs
Abhisheklearn12: gotcha

PR #2979, opened 2026-02-05
refactor: extract optimize logic from table.rs into submodule

Abhisheklearn12:
## Summary

  Continues the modularization effort of table operations as outlined in #2949.

  - Extracts optimization operations (`OptimizeAction`, `OptimizeStats`, `execute_optimize`, `compact_files_impl`, `cleanup_old_versions`, `optimize_indices`) from
  `table.rs` into `table/optimize.rs`
  - Public API remains unchanged via re-exports
  - Adds comprehensive tests including error cases with message assertions

  ## Test plan

  - [x] All new optimization tests pass
  - [x] All existing tests pass
  - [x] `cargo clippy` passes with no warnings
  - [x] `cargo fmt --check` passes

### DIFF
--- a/rust/lancedb/src/table/optimize.rs
+++ b/rust/lancedb/src/table/optimize.rs
@@ -214,8 +214,8 @@ pub(crate) async fn execute_optimize(
 mod tests {
     use arrow_array::{Int32Array, RecordBatch, RecordBatchIterator, StringArray};
     use arrow_schema::{DataType, Field, Schema};
+    use rstest::rstest;
     use std::sync::Arc;
-    use tempfile::tempdir;
 
     use crate::connect;
     use crate::index::{scalar::BTreeIndexBuilder, Index};
@@ -225,9 +225,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_optimize_compact_simple() {
-        let tmp_dir = tempdir().unwrap();
-        let uri = tmp_dir.path().to_str().unwrap();
-        let conn = connect(uri).execute().await.unwrap();
+        let conn = connect("memory://").execute().await.unwrap();
 
         // Create a table with initial data
         let schema = Arc::new(Schema::new(vec![Field::new("i", DataType::Int32, false)]));
@@ -313,9 +311,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_optimize_prune_versions() {
-        let tmp_dir = tempdir().unwrap();
-        let uri = tmp_dir.path().to_str().unwrap();
-        let conn = connect(uri).execute().await.unwrap();
+        let conn = connect("memory://").execute().await.unwrap();
 
         // Create a table
         let schema = Arc::new(Schema::new(vec![Field::new("i", DataType::Int32, false)]));
@@ -364,8 +360,13 @@ mod tests {
             .await
             .unwrap();
 
-        // Verify prune stats are returned
-        assert!(stats.prune.is_some());
+        // Prune-only operation should not have compaction stats
+        assert!(stats.compaction.is_none());
+
+        // Verify prune stats
+        let prune_stats = stats.prune.unwrap();
+        assert!(prune_stats.bytes_removed > 0);
+        assert_eq!(prune_stats.old_versions, 5);
 
         // Verify data is still intact
         let final_row_count = table.count_rows(None).await.unwrap();
@@ -393,9 +394,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_optimize_index() {
-        let tmp_dir = tempdir().unwrap();
-        let uri = tmp_dir.path().to_str().unwrap();
-        let conn = connect(uri).execute().await.unwrap();
+        let conn = connect("memory://").execute().await.unwrap();
 
         // Create a table with data
         let schema = Arc::new(Schema::new(vec![Field::new("i", DataType::Int32, false)]));
@@ -433,6 +432,14 @@ mod tests {
             .await
             .unwrap();
 
+        // Verify index stats before optimization
+        let indices = table.list_indices().await.unwrap();
+        assert_eq!(indices.len(), 1);
+        let index_name = indices[0].name.clone();
+        let stats_before = table.index_stats(&index_name).await.unwrap().unwrap();
+        assert_eq!(stats_before.num_indexed_rows, 100);
+        assert_eq!(stats_before.num_unindexed_rows, 100);
+
         // Run index optimization
         let stats = table
             .optimize(OptimizeAction::Index(Default::default()))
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+    use crate::index::{scalar::BTreeIndexBuilder, Index};
+    use crate::query::ExecutableQuery;
+    use crate::table::{CompactionOptions, OptimizeAction, OptimizeStats};
+    use futures::TryStreamExt;
+
+    #[tokio::test]
+    async fn test_optimize_compact_simple() {
+        let tmp_dir = tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let conn = connect(uri).execute().await.unwrap();
```
wjones127: Let's use the memory store for test when possible.
```suggestion
        let conn = connect("memory://").execute().await.unwrap();
```

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+        all_values.sort();
+        let expected: Vec<i32> = (0..600).collect();
+        assert_eq!(all_values, expected);
+    }
+
+    #[tokio::test]
+    async fn test_optimize_prune_versions() {
+        let tmp_dir = tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let conn = connect(uri).execute().await.unwrap();
```
wjones127:
```suggestion
        let conn = connect("memory://").execute().await.unwrap();
```

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+            .optimize(OptimizeAction::Prune {
+                older_than: Some(chrono::Duration::try_days(0).unwrap()),
+                delete_unverified: Some(true),
+                error_if_tagged_old_versions: None,
+            })
+            .await
+            .unwrap();
+
+        // Verify prune stats are returned
+        assert!(stats.prune.is_some());
```
wjones127: We should be able to assert that:

1. `stats.prune.bytes_removed > 0`
2. `stast.prune.old_versions == 5` (or similar value)

wjones127: Should we also assert `stats.compaction.is_none()`?

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+        all_values.sort();
+        let expected: Vec<i32> = (0..60).collect();
+        assert_eq!(all_values, expected);
+    }
+
+    #[tokio::test]
+    async fn test_optimize_index() {
+        let tmp_dir = tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let conn = connect(uri).execute().await.unwrap();
```
wjones127:
```suggestion
        let conn = connect("memory://").execute().await.unwrap();
```

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+            vec![Arc::new(Int32Array::from_iter_values(100..200))],
+        )
+        .unwrap();
+        table
+            .add(RecordBatchIterator::new(vec![Ok(batch)], schema.clone()))
+            .execute()
+            .await
+            .unwrap();
+
+        // Run index optimization
```
wjones127, line 443: We should grab the index statistics and assert things about the index stats before and after.

https://docs.rs/lancedb/latest/lancedb/index/struct.IndexStatistics.html

In particular `num_indexed_rows` should go up to all rows, and `num_unindexed_rows` should go to zero. Also should assert what happens with `num_indices`

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+        all_values.sort();
+        let expected: Vec<i32> = (0..200).collect();
+        assert_eq!(all_values, expected);
+    }
+
+    #[tokio::test]
+    async fn test_optimize_all() {
+        let tmp_dir = tempdir().unwrap();
+        let uri = tmp_dir.path().to_str().unwrap();
+        let conn = connect(uri).execute().await.unwrap();
```
wjones127:
```suggestion
        let conn = connect("memory://").execute().await.unwrap();
```

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+    #[tokio::test]
+    async fn test_optimize_stats_default() {
+        // Verify OptimizeStats default values
+        let stats: OptimizeStats = Default::default();
+        assert!(stats.compaction.is_none());
+        assert!(stats.prune.is_none());
+    }
+
+    #[tokio::test]
+    async fn test_compact_with_deferred_index_remap() {
```
wjones127, line 548: Might want to leave a comment stating this is just a smoke test for now. We don't make any real assertions that verify remap is actually happening, unfortunately. At least we can tell it doesn't error 🤷

rust/lancedb/src/table/optimize.rs
```diff
@@ -0,0 +1,934 @@
...
+        assert!(stats.prune.is_some());
+
+        // Verify table is still empty but schema is preserved
+        assert_eq!(table.count_rows(None).await.unwrap(), 0);
+        let current_schema = table.schema().await.unwrap();
+        assert_eq!(current_schema, schema);
+    }
+
+    #[tokio::test]
+    async fn test_optimize_fails_on_checked_out_table() {
```
wjones127: Could you make this test and the ones below into a parameterized test via rstest? Then it's not so verbose.

rust/lancedb/src/table/optimize.rs
Abhisheklearn12: sure done, I switched all tests to use the memory store

rust/lancedb/src/table/optimize.rs
Abhisheklearn12: goood call, added assertions for bytes_removed > 0, old_versions == 5, and also compaction.is_none(), done this

rust/lancedb/src/table/optimize.rs
Abhisheklearn12: makes sense, added a comment clarifying that

rust/lancedb/src/table/optimize.rs
Abhisheklearn12: I refactored into a single parameterized test with rstest. now much cleaner, done

rust/lancedb/src/table/optimize.rs
Abhisheklearn12: added now checking `num_indexed_rows`, `num_unindexed_rows`, and `num_indices` before and after optimization

PR #2980, opened 2026-02-05
fix: avoid holding RwLock during async I/O in DatasetConsistencyWrapper

j2h4u:
## Summary

`DatasetConsistencyWrapper` methods (`reload`, `as_latest`, `as_time_travel`) performed async I/O operations (`checkout_latest`, `checkout_version`, `latest_version_id`) while holding the `RwLock`. Under contention this blocks the tokio event loop — we observed **30+ minute stalls** in a Node.js application (via napi-rs bindings) where a `db.search()` call blocked the entire event loop after an embedding write.

The fix follows the pattern already established in `insert.rs:188` (*"Don't hold the lock over an await point"*):

1. **Read lock** — clone the `Dataset` and capture parameters (fast, no I/O)
2. **No lock** — perform async I/O on the cloned dataset
3. **Write lock** — update state with the new dataset (fast, only memory ops)

This also removes the now-unused `DatasetRef` methods (`reload`, `need_reload`, `as_latest`, `as_time_travel`) whose logic has been moved into `DatasetConsistencyWrapper`.

## Discovery Context

- Node.js application using LanceDB 0.23.0 via `@lancedb/lancedb`
- 30-minute gap in Docker logs: embedding write completed, then complete silence until next heartbeat
- Container was running (0 restarts, system healthy) — the event loop was simply blocked
- Gap occurred right after `table.add()` → next operation was `db.search()` for duplicate checking
- NAPI bindings are correct (async). Root cause is in Rust core.

## Changes

- **`DatasetConsistencyWrapper::reload()`** — clone dataset under read lock, check `latest_version_id()` and `checkout_version()` outside lock, update state under write lock
- **`DatasetConsistencyWrapper::as_latest()`** — same pattern: clone under read lock, I/O outside, update under write lock with re-check
- **`DatasetConsistencyWrapper::as_time_travel()`** — same pattern with early return if version matches
- **Removed unused `DatasetRef` methods** — `reload`, `need_reload`, `as_latest`, `as_time_travel`

## Test plan

- [ ] Existing test `test_iops_open_strong_consistency` passes
- [ ] `cargo test --features remote -p lancedb`
- [ ] `cargo clippy --features remote --tests`
- [ ] Concurrent read/write stress test (if applicable)

🤖 Calculon (feat. [Claude Opus 4.6](https://claude.com/claude-code))

### DIFF
--- a/rust/lancedb/src/table/dataset.rs
+++ b/rust/lancedb/src/table/dataset.rs
@@ -205,7 +205,7 @@ impl DatasetConsistencyWrapper {
         let need_reload = if is_latest {
             dataset.latest_version_id().await? != current_version
         } else {
-            target_version.map_or(false, |v| current_version != v)
+            target_version.is_some_and(|v| current_version != v)
         };
         if !need_reload {
             return Ok(());


wjones127:
> Under contention this blocks the tokio event loop — we observed **30+ minute stalls** in a Node.js application (via napi-rs bindings) where a `db.search()` call blocked the entire event loop after an embedding write.

Hello, thanks for reporting this. Could you describe what happened in more detail what happened?

We use an async lock, which doesn't block the event loop:

https://github.com/lancedb/lancedb/blob/6329b576042d6f6ed7d7415683ef09d8e05bbf1f/rust/lancedb/src/table/dataset.rs#L11

This is designed to be held across await points.

Some background: the lock here has two purposes:
1. If two or more concurrent calls to `ensure_up_to_date()` happen, we don't want to call `reload()` multiple times.
2. Try to prevent race conditions of different write operations calling `set_latest()` or even read operations calling `reload()`.

I think (1) is important to keep. I think we could loosen (2), especially considering we have this logic in place:

https://github.com/lancedb/lancedb/blob/6329b576042d6f6ed7d7415683ef09d8e05bbf1f/rust/lancedb/src/table/dataset.rs#L134-L136

j2h4u:
Thanks for the quick response and the clarification about the async lock — you're right that `tokio::sync::RwLock` doesn't block the thread. Let me provide more context.

## What happened

We run [OpenClaw](https://github.com/j2h4u/openclaw) (a Node.js AI assistant) on a home server. It uses `@lancedb/lancedb` 0.23.0 for memory storage (embeddings). The table is opened with strong consistency (`readConsistencyInterval: 0`). Storage is a local SSD (ext4).

On 2026-02-05 we noticed two cron jobs were skipped entirely. Docker logs showed a 30-minute gap:

```
05:44:25 - embedding write completed (table.add)
           ... nothing for 30 minutes ...
06:14:09 - next heartbeat
```

Container was running (0 restarts), system was healthy (CPU idle, RAM available, SSD healthy at 83% life). The gap occurred right after an embedding write — the next operation was `db.search()` for duplicate checking.

We don't know the exact cause of the 30-minute stall. The event loop was completely unresponsive during this period.

## Our analysis (and where we may be wrong)

We initially attributed this to the write lock being held during `checkout_latest().await` in `reload()`. But as you correctly point out, this is an async lock — it doesn't block the tokio runtime or the Node.js event loop in the traditional sense.

So the question becomes: **what could cause a 30-minute stall after `table.add()` when the next operation is `db.search()`?**

We're open to the possibility that the root cause is elsewhere. But looking at the code path — `search()` → `get()` → `ensure_up_to_date()` → `reload()` — the `reload()` method does perform I/O under the write lock, which seemed like the most likely suspect.

## What our change does

Regardless of the root cause, the change follows the pattern from `insert.rs:188` ("Don't hold the lock over an await point") and preserves both guarantees you mentioned:

1. **No duplicate reloads** — we check `need_reload` before acquiring the write lock, and re-check after acquiring it (same double-check pattern as the original)
2. **Race condition safety** — the write lock update checks versions before overwriting (only updates if newer)

The main difference: if `checkout_version().await` takes a long time, readers calling `.get()` can still proceed with the current version instead of waiting for the write lock to release.

## Open question

Any ideas on what could cause a 30-minute stall in this code path on healthy local SSD storage? We'd be happy to provide more diagnostics (strace, profiling, etc.) if helpful.

wjones127:
FYI I'm working on an alternative proposal here: https://github.com/lancedb/lancedb/pull/3021

j2h4u:
Closing in favor of #3021 — it's a much stronger solution that addresses the root cause architecturally:

- Replacing `RwLock` with `Mutex` that's never held across `.await` points
- `get()` returning `Arc<Dataset>` instead of a guard — readers never block each other
- Coalescing reloader to deduplicate concurrent refresh calls
- Proper consistency modes (Lazy / Strong / Eventual)

Our PR was a targeted fix to work around the symptom. #3021 is the right long-term solution. Happy to test it once it's ready — we have a real-world reproduction scenario (Node.js app with strong consistency on local storage).

Thanks @wjones127 for looking into this and for the thorough rework!

PR #2997, opened 2026-02-08
fix(python): implement Permutation.__getitem__ for torch DataLoader support

### LINKED ISSUE #2996: bug(python): torch dataloder doesn't work with permutation object
### LanceDB version

v0.29.0

### What happened?

Based on the docs, users should be able to create torch dataloader directly from `Permutation` objects. But creating a torch loader fails, because it always receives a batch of Nones, as it relies on __getitem__ which is not implemented - https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/permutation.py#L747

Minimal example:
Here collate fn will receive all Nones, and the assertion fails

```
import lancedb
import torch
from lancedb.permutation import Permutations, permutation_builder


def test_permutation_dataloader_returns_rows():
    mem_db = lancedb.connect("memory://")

    table = mem_db.create_table(
        "perm_table",
        pa.table(
            {
                "video": [f"video-{i}" for i in range(6)],
                "caption": [f"caption-{i}" for i in range(6)],
            }
        ),
    )

    perm_table = permutation_builder(table).execute()
    dataset = Permutations(table, perm_table)[0].select_columns(["video", "caption"])

    seen_batches = []

    def collate_fn(batch):
        assert all(row is not None for row in batch) # <--- Fails

    dataloader = torch.utils.data.DataLoader(
        dataset, batch_size=2, shuffle=False, collate_fn=collate_fn
    )

```

veeceey:
## Summary
- Implement `Permutation.__getitem__` which was stubbed out with `pass`, always returning `None`
- This fixes torch `DataLoader` usage since it relies on `__getitem__` to fetch individual rows
- The implementation uses `with_offset`/`with_limit` on the reader to fetch a single row and returns it as a Python dictionary
- Supports negative indexing and proper `IndexError` for out-of-range access

## Test plan
- Added `test_getitem` - tests basic indexing, negative indexing, and out-of-range errors
- Added `test_getitem_with_select_columns` - tests `__getitem__` works after column selection
- Added `test_getitem_with_torch_dataloader` - reproduces the exact scenario from the issue report
- All 49 permutation tests pass

Fixes #2996

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -754,7 +754,22 @@ class Permutation:
         processing use [iter](#iter) or a torch data loader to perform batched
         processing.
         """
-        pass
+        num_rows = self.num_rows
+        if index < 0:
+            index += num_rows
+        if index < 0 or index >= num_rows:
+            raise IndexError(
+                f"index {index} is out of range for permutation with {num_rows} rows"
+            )
+
+        async def do_getitem():
+            reader = await self.reader.with_offset(index)
+            reader = await reader.with_limit(1)
+            stream = await reader.read(self.selection, batch_size=1)
+            batch = await stream.__anext__()
+            return {col: batch.column(col)[0].as_py() for col in batch.schema.names}
+
+        return LOOP.run(do_getitem())
 
     @deprecated(details="Use with_skip instead")
     def skip(self, skip: int) -> "Permutation":
--- a/python/python/tests/test_permutation.py
+++ b/python/python/tests/test_permutation.py
@@ -945,3 +945,70 @@ def test_custom_transform(mem_db):
     batch = batches[0]
 
     assert batch == pa.record_batch([range(10)], ["id"])
+
+
+def test_getitem(mem_db):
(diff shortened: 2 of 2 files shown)


westonpace:
Thanks for working on this!  I ended up implementing something at the rust level (and using __getitems__ instead of __getitem__) as I think we will need to cache some information there to get this to perform well and it opens up the door for an equivalent nodejs API at some point.

Do you want to take a look at https://github.com/lancedb/lancedb/pull/3013 and see if that would satisfy your need?

veeceey:
Thanks for the heads up! I took a look at #3013 and it looks great — implementing it at the Rust level with `take_offsets` and `__getitems__` is definitely the right approach, especially for performance and the future nodejs API. That would fully satisfy my use case with torch DataLoader.

Happy to close this PR in favor of #3013. Let me know if there's anything useful I can help with on that one.

veeceey:
Closing in favor of #3013 — the Rust-level implementation is the better approach here. Thanks @westonpace!

PR #3006, opened 2026-02-09
feat: show reranker info in hybrid search explain plan

### LINKED ISSUE #3000: In hybrid search explain plan, show the reranker information
Right now, we just show the vector plan and the FTS plan:

https://github.com/lancedb/lancedb/blob/9fac56252e626cf5787f74415fde4841d1b50431/python/python/lancedb/query.py#L2121-L2133

Instead, we should show something like:

```
ReciprocalRankFusion: k=60
  <vector search plan>
  <FTS search plan>
```

dask-58:
Closes #3000

The hybrid search `explain_plan` now shows the reranker as the top-level node with
the vector and FTS sub-plans indented underneath, instead of just listing them
separately with no reranker context.

**Before:**
```
Vector Search Plan:
ProjectionExec: ...
FTS Search Plan:
ProjectionExec: ...
```

**After:**
```
RRFReranker(K=60)
  Vector Search Plan:
  ProjectionExec: ...
  FTS Search Plan:
  ProjectionExec: ...
```

Other rerankers display similarly ; e.g. `LinearCombinationReranker(weight=0.7, fill=1.0)`,
`MRRReranker(weight_vector=0.5, weight_fts=0.5)`, `CohereReranker(model_name=name)`.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -2128,11 +2128,7 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
         # Indent sub-plans under the reranker
         indented_vector = "\n".join("  " + line for line in vector_plan.splitlines())
         indented_fts = "\n".join("  " + line for line in fts_plan.splitlines())
-        return (
-            f"{reranker_label}\n"
-            f"  Vector Search Plan:\n{indented_vector}\n"
-            f"  FTS Search Plan:\n{indented_fts}"
-        )
+        return f"{reranker_label}\n  {indented_vector}\n  {indented_fts}"
 
     def analyze_plan(self):
         """Execute the query and display with runtime metrics.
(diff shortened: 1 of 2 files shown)


python/python/lancedb/query.py
```diff
@@ -3192,12 +3199,20 @@ async def explain_plan(self, verbose: Optional[bool] = False):
...
+        indented_vector = "\n".join(
+            "  " + line for line in vector_plan.splitlines()
+        )
+        indented_fts = "\n".join(
+            "  " + line for line in fts_plan.splitlines()
+        )
+        return (
+            f"{self._reranker}\n"
+            f"  Vector Search Plan:\n{indented_vector}\n"
+            f"  FTS Search Plan:\n{indented_fts}"
```
wjones127: Honestly, I think it would be fine if we removed those two annotations, then it would be more like other plans where we don't explicitly label it as vector or FTS search. What do you think?

```suggestion
            f"{self._reranker}\n"
            f"  {indented_vector}\n"
            f"  {indented_fts}"
```

PR #3015, opened 2026-02-11
perf(remote): cache schema of remote tables

### LINKED ISSUE #3014: Cache remote table schema
We should cache the schema so we avoid calling the describe endpoint frequently just to get the up-to-date schema.

## Why do we need the schema at all?

When uses call `open_table()`, we call the describe endpoint to check for existence of the table. This makes sure if the table doesn't exist the user immediately gets a "Not found" error.

Most calls after that don't strictly need the schema, as long as the user provides enough information. But some convenience APIs do require the schema. For example, `table.search("some text").to_list()` needs to look up the schema and the indexes, in order to decide to interpret this as a vector, FTS, or hybrid search query. But if the user passes `table.search("some text", query_type="FTS").to_list()`, we should be able to bypass those checks and resolve to an FTS query. Part of the reason this matters is that we return different builder classes from the `.search()` method depending on the type of query (FTS vs Vector vs Hybrid). That's why we can't defer this decision until the server receives the query.

## Handling invalidation

Schema updates are relatively rare, but they can happen. We want to minimize the chance we keep an out-of-date schema.

So we will invalidate in the following scenarios:

1. **After a 30 second TTL.** 30 seconds is short enough to feel like eventually consistency, but long enough to be better than calling describe endpoint for each query at 10 QPS.
2. **After calling methods that change the schema, like overwrite, add columns, drop columns, etc.** In this case, we know the schema has been changed.
3. **When we receive a 400 or 500 response.** This error may be due to mismatched schema expectations, so to be conservative we should invalidate the schema in these cases.

wjones127:
Caches the schema of remote tables and invalidates the cache when:

1. After 30 second TTL
2. When we do an operation that changes schema (e.g. add_columns) or checks out a different version (e.g. checkout_version)
3. When we get a 400, 404, or 500 reponse

If the schema is retrieved close to the TTL, we optimistically fetch the schema in the background. This means a continuous stream of queries will never have the schema fetch on the critical path.

Closes #3014

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -3,9 +3,11 @@
 
 pub mod insert;
 
+use crate::data::scannable::Scannable;
 use crate::index::Index;
 use crate::index::IndexStatistics;
 use crate::query::{QueryFilter, QueryRequest, Select, VectorQueryRequest};
+use crate::remote::util::stream_as_ipc;
 use crate::table::AddColumnsResult;
 use crate::table::AddResult;
 use crate::table::AlterColumnsResult;
@@ -46,10 +48,10 @@ use tokio::sync::RwLock;
 use super::client::RequestResultExt;
 use super::client::{HttpSend, RestfulLanceDbClient, Sender};
 use super::db::ServerVersion;
+use super::util::stream_as_body;
 use super::ARROW_STREAM_CONTENT_TYPE;
 use crate::index::waiter::wait_for_index;
 use crate::{
-    connection::NoData,
     error::Result,
     index::{IndexBuilder, IndexConfig},
     query::QueryExecutionOptions,
@@ -332,7 +334,10 @@ impl<S: HttpSend> RemoteTable<S> {
 
     fn reader_as_body(data: Box<dyn RecordBatchReader + Send>) -> Result<reqwest::Body> {
         // TODO: Once Phalanx supports compression, we should use it here.
-        let mut writer = arrow_ipc::writer::StreamWriter::try_new(Vec::new(), &data.schema())?;
+        let mut writer = arrow_ipc::writer::StreamWriter::try_new(
+            Vec::new(),
+            &RecordBatchReader::schema(&*data),
+        )?;
 
         //  Mutex is just here to make it sync. We shouldn't have any contention.
         let mut data = Mutex::new(data);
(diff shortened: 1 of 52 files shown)


wkalt:
@wjones127 I think this will be a big benefit! A couple questions/thoughts:
* Moka is already a workspace dependency although not used much. Would it save complexity to use that instead? What do we get by not using it?
* If we go with rolling our own, would it make sense to split the caching infrastructure from "schemas" so that the cache structure can be reused elsewhere?

wjones127:
> @wjones127 I think this will be a big benefit! A couple questions/thoughts:
> 
> * Moka is already a workspace dependency although not used much. Would it save complexity to use that instead? 

Yeah, we could use a moka cache of size 1 for this. The overhead might be a bit much, given that is a multi-threaded cache. But if would be fewer lines of our code

> What do we get by not using it?

In the Moka cache, updates always block `get()`. Here, when the trigger for update is a TTL, we let readers get a stale value while we fetch the new one. 

> * If we go with rolling our own, would it make sense to split the caching infrastructure from "schemas" so that the cache structure can be reused elsewhere?

Yeah, I think I will re-use this pattern for the `DatasetConsistencyWrapper`. So I will extract it out into a separate module.

rust/lancedb/src/remote/table.rs
```diff
@@ -61,6 +62,60 @@ use crate::{
 const REQUEST_TIMEOUT_HEADER: HeaderName = HeaderName::from_static("x-request-timeout-ms");
 const METRIC_TYPE_KEY: &str = "metric_type";
 const INDEX_TYPE_KEY: &str = "index_type";
+const SCHEMA_CACHE_TTL: Duration = Duration::from_secs(30);
+const SCHEMA_CACHE_REFRESH_WINDOW: Duration = Duration::from_secs(5);
+
+type SharedSchemaFuture =
```
LuQQiu, line 70: this is involved in the metadata cache size?

LuQQiu: or because it's too small, it does not matter

rust/lancedb/src/remote/table.rs
```diff
@@ -361,14 +429,41 @@ impl<S: HttpSend> RemoteTable<S> {
         Ok(response)
     }
 
+    /// Check if a status code should trigger schema cache invalidation
+    fn should_invalidate_cache_for_status(status: StatusCode) -> bool {
+        // Only invalidate for errors that could be schema-related
+        // Don't invalidate for auth errors (401, 403) or temporary failures (503, 502)
+        matches!(
+            status,
+            StatusCode::BAD_REQUEST // 400 - could be schema mismatch
```
LuQQiu, line 547: what happens if should update but because the status code is not include here and does not get update, then will it affect some workloads?

LuQQiu: Mainly worry if that would cause some data corruption in write?

LuQQiu (commented):
can the schema lifecycle bounded with the version lifecycle

i worry about we are at version A but shema is version B

rust/lancedb/src/remote/table.rs
wjones127: It would never cause data corruption. Data is always validated to be correct on the server.

If the status code was not included, then it would keep failing on retries until the 30 second TTL was up, and then it would reload the schema and succeed. Seems okay for such an edge case.

I think 400, 404, and 500 realistically will cover all cases of schema mismatch.

rust/lancedb/src/remote/table.rs
wjones127: For remote tables, metadata cache is on the server. This is the client.

wjones127:
> can the schema lifecycle bounded with the version lifecycle
> 
> i worry about we are at version A but shema is version B

I initially thought we might invalidate when the version changes, but later decided against it.

For some cases, when we do insert we might want the schema. If we are doing insert in a loop, that means we every time we insert, we invalidate the cached schema and need to retrieve it again. That seem sub-optimal, especially because insert doesn't change the schema.

Realistically, I think concurrent schema changes are rare. And if you change the schema while there are concurrent writes happening, there is already the possibility of creating conflicts.

LuQQiu:
Use claude to see if can find some race conditions when schema / version lifecycle is mismatch
```
1. Task A calls schema(). The cache is stale, so it needs to fetch. Before fetching, it reads the current version — say version 5.
    2. Before task A gets the lock again, task B runs checkout(6). This sets the version to 6 and clears the schema cache.
    3. Now task A grabs the lock. It sees the cache is empty (because task B just cleared it). So it starts a fetch — but it still has version 5 in hand from step 1. It
  sends
    that to the server.
    4. The fetch comes back with the schema for version 5. The counter matches (the invalidation already happened before this fetch started, so the counter was already
  bumped
    and this fetch recorded the new value). The cache now stores the version 5 schema as current.
    5. Result: the table is at version 6, but the cache holds the schema for version 5. This stale entry sticks around for up to 30 seconds until the TTL expires.
```
```
  1. Task A — schema() fast path misses. It drops the lock.
  2. Task A — calls self.current_version().await, gets Some(5). This is an async call so it can yield.
  3. Task B — checkout(6) runs: sets version to 6, calls invalidate_schema_cache() which sets state to Empty and bumps generation from 0 to 1.
  4. Task A — acquires the lock with version = Some(5). Cache state is Empty, generation is 1.
  5. Task A — determine_schema_action sees Empty, calls start_schema_fetch(cache, Some(5), None). This records generation = 1 and starts a fetch for version 5.
  6. The fetch returns the version 5 schema. The spawned task checks: cache.generation (1) == generation (1) — matches. So it writes the version 5 schema into the cache.
```

LuQQiu:
@wjones127 if the version can tell us info about whether schema changes or not, that would be really helpful, not quite sure whether that info is in version manifest or not?

wjones127:
> @wjones127 if the version can tell us info about whether schema changes or not, that would be really helpful, not quite sure whether that info is in version manifest or not?

Oh are you talking about this version? https://github.com/lancedb/lancedb/blob/14973ac9d1190cb94d1232aa3ed8d478e77fe6e8/rust/lancedb/src/remote/table.rs#L208

Or are you talking about something else? I'm confused what you want to be in sync.

wjones127:
> > * If we go with rolling our own, would it make sense to split the caching infrastructure from "schemas" so that the cache structure can be reused elsewhere?
> 
> Yeah, I think I will re-use this pattern for the `DatasetConsistencyWrapper`. So I will extract it out into a separate module.

FYI @wkalt I have a follow up PR that pulls this cache pattern into a separate module (and uses it for something else): https://github.com/lancedb/lancedb/pull/3021

PR #3016, opened 2026-02-11
feat: improve Permutation pytorch integration

westonpace:
This changes around the output format of `Permutation` in some breaking ways but I think the API is still new enough to be considered experimental.

1. In order to align with both huggingface's dataset and torch's expectations the default output format is now a list of dicts (row-major) instead of a dict of lists (column-major).  I've added a python_col option which will return the dict of lists.
2. In order to align with pytorch's expectation the `torch` format is now a list of tensors (row-major) instead of a 2D tensor (column-major).  I've added a torch_col option which will return the 2D tensor instead.

Added tests for torch integration with Permutation

~~Leaving draft until https://github.com/lancedb/lancedb/pull/3013 merges as this is built on top of that~~

### DIFF
--- a/python/python/tests/test_permutation.py
+++ b/python/python/tests/test_permutation.py
@@ -664,23 +664,20 @@ def test_iter_basic(some_permutation: Permutation):
     expected_batches = (950 + batch_size - 1) // batch_size  # ceiling division
     assert len(batches) == expected_batches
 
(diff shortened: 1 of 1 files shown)


AyushExel (approved):

westonpace:
Oops, looks like I broke all the tests by changing the default output format.  Will fix.

PR #3021, opened 2026-02-12
feat: check for dataset updates in the background

wjones127:
This updates `DatasetConsistencyWrapper` to block less:

1. `DatasetConsistencyWrapper::get()` just returns `Arc<Dataset>` now, instead of a guard that blocks writes. `DatasetConsistencyWrapper::get_mut()` is gone; now write methods just use `get()` and then later call `update()` with the new version. This means a given table handle can do concurrent reads **and** writes.
2. In weak consistency mode, will check for dataset updates in the background, instead of blocking calls to `get()`.

### DIFF
--- a/rust/lancedb/src/utils/background_cache.rs
+++ b/rust/lancedb/src/utils/background_cache.rs
@@ -157,10 +157,6 @@ where
         }
     }
 
-    /// Invalidate the cache. The next [`get()`](Self::get) will start a fresh fetch.
-    ///
-    /// Any in-flight background fetch from before this call will not update the
-    /// cache (the generation counter prevents stale writes).
     /// Pre-populate the cache with an initial value.
     ///
     /// This avoids a blocking fetch on the first [`get()`](Self::get) call.
@@ -169,6 +165,10 @@ where
         cache.state = State::Current(value, clock::now());
     }
 
+    /// Invalidate the cache. The next [`get()`](Self::get) will start a fresh fetch.
+    ///
+    /// Any in-flight background fetch from before this call will not update the
+    /// cache (the generation counter prevents stale writes).
     pub fn invalidate(&self) {
         let mut cache = self.inner.lock().unwrap();
         cache.state = State::Empty;


rust/lancedb/src/utils/background_cache.rs
```diff
@@ -0,0 +1,593 @@
...
+        }
+    }
+
+    /// Invalidate the cache. The next [`get()`](Self::get) will start a fresh fetch.
+    ///
+    /// Any in-flight background fetch from before this call will not update the
+    /// cache (the generation counter prevents stale writes).
+    /// Pre-populate the cache with an initial value.
+    ///
+    /// This avoids a blocking fetch on the first [`get()`](Self::get) call.
```
LuQQiu, line 162: the doc comments belong to invalidate instead of seed?

rust/lancedb/src/table.rs
```diff
@@ -2121,16 +2123,15 @@ impl BaseTable for NativeTable {
...
 
-        let dataset = {
-            // Limited scope for the mutable borrow of self.dataset avoids deadlock.
-            let ds = self.dataset.get_mut().await?;
-            InsertBuilder::new(Arc::new(ds.clone()))
-                .with_params(&lance_params)
-                .execute_stream(data)
-                .await?
-        };
+        self.dataset.ensure_mutable()?;
```
LuQQiu, line 2126: what if as_time_travel() happens here by another thread/task? update will fail here?

rust/lancedb/src/table.rs
```diff
@@ -2121,16 +2123,15 @@ impl BaseTable for NativeTable {
...
+        let ds = self.dataset.get().await?;
+        let dataset = InsertBuilder::new(ds)
+            .with_params(&lance_params)
+            .execute_stream(data)
+            .await?;
+
         let version = dataset.manifest().version;
-        self.dataset.set_latest(dataset).await;
+        self.dataset.update(dataset);
```
LuQQiu, line 2134: what if two writes do get() and update() together? what would happen?

rust/lancedb/src/table/optimize.rs
```diff
@@ -131,10 +129,9 @@ pub(crate) async fn cleanup_old_versions(
     delete_unverified: Option<bool>,
     error_if_tagged_old_versions: Option<bool>,
 ) -> Result<RemovalStats> {
-    Ok(table
-        .dataset
-        .get_mut()
-        .await?
+    table.dataset.ensure_mutable()?;
+    let dataset = table.dataset.get().await?;
+    Ok(dataset
```
LuQQiu, line 134: why not update here?

rust/lancedb/src/table.rs
wjones127: Update acquires a lock, and then only sets the dataset if it's version is newer than the version inside. So if `update()` is called concurrently, it will always have the newest version in the end.

rust/lancedb/src/table/optimize.rs
wjones127: Cleanup doesn't create a new version of the dataset, so it's not necessary.

rust/lancedb/src/table.rs
wjones127: That's a good question. I think I should consider making `as_time_travel` return a copy. I don't think it's all that useful to have it mutate the state of the dataset wrapper. Then it will be much safer.

rust/lancedb/src/utils/background_cache.rs
wjones127: Good point. Looks like the two functions has their comments combined.

LuQQiu (approved):
Looks good! Great changes!

PR #3023, opened 2026-02-12
fix: support dynamic projection on remote table

westonpace:
The remote server expects an object (`{"alias": "col"}`) and the client was previously sending a list of tuples `[["alias", "col"]]`


westonpace:
Example test failure: https://github.com/lancedb/lancedb/actions/runs/21967145475/job/63459792355

PR #3025, opened 2026-02-13
feat: support multi-tenant azure deployments in the remote client

wkalt:
This forwards an x-azure-tenant-id header to the remote server when supplied in storage_options. This makes it possible to have a server deployment that spans multiple Azure AD tenants.


wjones127:
Superceded by #3102

PR #3028, opened 2026-02-13
fix: handle null type in _align_field_types for nullable struct fields

### LINKED ISSUE #2654: bug(python): nullable struct field fails with AttributeError

Mr-Neutr0n:
When you have a nullable struct column and insert a row where that column is `None`, pyarrow infers the column type as `null` instead of `struct`. The `_align_field_types` function then tries to access `.fields` on the null type, which blows up with `AttributeError: 'pyarrow.lib.DataType' object has no attribute 'fields'`.

The fix checks for null type before trying to recurse into the type's structure. If the source field is null, we just use the target type directly — there's nothing to align when all values are None anyway.

Added a test that reproduces the original crash from #2654.

Fixes #2654

### DIFF
--- a/python/python/tests/test_table.py
+++ b/python/python/tests/test_table.py
@@ -327,6 +327,33 @@ def test_add_struct(mem_db: DBConnection):
     table.add(data)
 
 
+def test_add_nullable_struct_with_none(mem_db: DBConnection):
+    # https://github.com/lancedb/lancedb/issues/2654
+    schema = pa.schema(
+        [
+            pa.field("id", pa.string()),
+            pa.field(
+                "data",
+                pa.struct([pa.field("x", pa.float32())]),
+                nullable=True,
+            ),
+        ]
+    )
+
+    table = mem_db.create_table("test_nullable_struct", schema=schema)
+
+    # Adding a row with a non-null struct should work
+    table.add([{"id": "1", "data": {"x": 1.0}}])
+
+    # Adding a row with None for the nullable struct field should also work
+    table.add([{"id": "2", "data": None}])
+
+    result = table.to_arrow()
+    assert result.num_rows == 2
+    assert result.column("id").to_pylist() == ["1", "2"]
+    assert result.column("data").to_pylist() == [{"x": 1.0}, None]
+
+
 def test_add_subschema(mem_db: DBConnection):
     schema = pa.schema(
         [
--- a/python/python/tests/test_table.py
+++ b/python/python/tests/test_table.py
@@ -340,7 +340,11 @@ def test_add_nullable_struct_with_none(mem_db: DBConnection):
         ]
     )
 
-    table = mem_db.create_table("test_nullable_struct", schema=schema)
+    table = mem_db.create_table(
+        "test_nullable_struct",
+        schema=schema,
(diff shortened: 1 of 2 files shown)


python/python/tests/test_table.py
```diff
@@ -327,6 +327,33 @@ def test_add_struct(mem_db: DBConnection):
...
+            pa.field("id", pa.string()),
+            pa.field(
+                "data",
+                pa.struct([pa.field("x", pa.float32())]),
+                nullable=True,
+            ),
+        ]
+    )
+
+    table = mem_db.create_table("test_nullable_struct", schema=schema)
```
wjones127: I think Lance v2.0 might not support null structs, so you might need to specify 2.1 as the file format version to get this passing:
```suggestion
    table = mem_db.create_table(
        "test_nullable_struct",
        schema=schema,
        storage_options=dict(new_table_data_storage_version="2.1"),
    )
```

Mr-Neutr0n:
@wjones127 — done. Pushed the `2.1` file format version as `ad2d1d9` on the same branch.

Mr-Neutr0n:
Quick update: I looked into the `mergeable: false` situation here, and the actual code fix has already landed upstream — [PR #3394](https://github.com/lancedb/lancedb/pull/3394) (commit `6a431ff`, 2026-05-16) refactored `_align_field_types` to handle the same null case via nested `is_struct(field.type)` checks (and similar for list-like types), so this branch's `is_null(field.type)` check is now redundant.

Issue [#2654](https://github.com/lancedb/lancedb/issues/2654) is **still open** though — the fix landed incidentally via a feature PR and nobody added a regression test. I opened [#3483](https://github.com/lancedb/lancedb/pull/3483) with just the test from this branch's reproduction case, so once that merges we can close the issue and this PR.

I'll leave this PR (`#3028`) open until the test PR lands, since closing prematurely would block the issue from being cleaned up. But if the maintainers prefer to close this directly, the test PR is independent.

PR #3029, opened 2026-02-14
feat: hook up new writer for insert

wjones127:
This hooks up a new writer implementation for the `add()` method. The main immediate benefit is it allows streaming requests to remote tables, and at the same time allowing retries for most inputs.

In NodeJS, we always convert the data to `Vec<RecordBatch>`, so it's always retry-able.

For Python, all are retry-able, except `Iterator` and `pa.RecordBatchReader`, which can only be consumed once. Some, like `pa.datasets.Dataset` are retry-able *and* streaming.

A lot of the changes here are to make the new DataFusion write pipeline maintain the same behavior as the existing Python-based preprocessing, such as:

* casting input data to target schema
* rejecting NaN values if `on_bad_vectors="error"`
* applying embedding functions.

In future PRs, we'll enhance these by moving the embedding calls into DataFusion and making sure we parallelize them. See: https://github.com/lancedb/lancedb/issues/3048

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -1,5 +1,5 @@
 [tool.bumpversion]
-current_version = "0.27.0-beta.0"
+current_version = "0.27.0-beta.1"
 parse = """(?x)
     (?P<major>0|[1-9]\\d*)\\.
     (?P<minor>0|[1-9]\\d*)\\.
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - Cargo.toml
       - nodejs/**
+      - rust/**
       - docs/src/js/**
       - .github/workflows/nodejs.yml
       - docker-compose.yml
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - Cargo.toml
       - python/**
+      - rust/**
       - .github/workflows/python.yml
 
 concurrency:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -183,7 +183,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        msrv: ["1.88.0"] # This should match up with rust-version in Cargo.toml
+        msrv: ["1.91.0"] # This should match up with rust-version in Cargo.toml
     env:
       # Need up-to-date compilers for kernels
       CC: clang-18
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1389,9 +1389,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.10.1"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "bytes-utils"
@@ -1780,6 +1780,16 @@ version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0f58bbc28f91df819d0aa2a2c00cd19754769c2fad90579b3592b1c9ba7a3115"
 dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
+name = "crossbeam-skiplist"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df29de440c58ca2cc6e587ec3d22347551a32435fbde9d2bff64e78a9ffa151b"
+dependencies = [
+ "crossbeam-epoch",
  "crossbeam-utils",
 ]
 
@@ -3072,9 +3082,8 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "fsst"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f9e5c0b1c67a38cb92b41535d44623483beb9511592ae23a3bf42ddec758690"
+version = "3.0.0-beta.5"
+source = "git+https://github.com/lance-format/lance.git?tag=v3.0.0-beta.5#c69274bd83da9930157d5e2ceeb101af13a916a3"
 dependencies = [
  "arrow-array",
  "rand 0.9.2",
@@ -4405,9 +4414,8 @@ dependencies = [
 
 [[package]]
 name = "lance"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2b7f07b905df393a5554eba19055c620f9ea25a3e40a013bda4bd8dc4ca66f01"
+version = "3.0.0-beta.5"
+source = "git+https://github.com/lance-format/lance.git?tag=v3.0.0-beta.5#c69274bd83da9930157d5e2ceeb101af13a916a3"
 dependencies = [
  "arrow",
  "arrow-arith",
@@ -4426,6 +4434,7 @@ dependencies = [
  "byteorder",
  "bytes",
  "chrono",
+ "crossbeam-skiplist",
  "dashmap",
  "datafusion",
  "datafusion-expr",
@@ -4465,6 +4474,7 @@ dependencies = [
  "tantivy",
  "tokio",
  "tokio-stream",
+ "tokio-util",
  "tracing",
  "url",
  "uuid",
@@ -4472,9 +4482,8 @@ dependencies = [
 
 [[package]]
 name = "lance-arrow"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "100e076cb81c8f0c24cd2881c706fc53e037c7d6e81eb320e929e265d157effb"
+version = "3.0.0-beta.5"
+source = "git+https://github.com/lance-format/lance.git?tag=v3.0.0-beta.5#c69274bd83da9930157d5e2ceeb101af13a916a3"
 dependencies = [
  "arrow-array",
  "arrow-buffer",
@@ -4493,9 +4502,8 @@ dependencies = [
 
 [[package]]
 name = "lance-bitpacking"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "588318d3d1ba0f97162fab39a323a0a49866bb35b32af42572c6b6a12296fa27"
+version = "3.0.0-beta.5"
+source = "git+https://github.com/lance-format/lance.git?tag=v3.0.0-beta.5#c69274bd83da9930157d5e2ceeb101af13a916a3"
 dependencies = [
  "arrayref",
  "paste",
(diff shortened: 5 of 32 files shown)


nodejs/src/table.rs
```diff
@@ -71,6 +71,17 @@ impl Table {
     pub async fn add(&self, buf: Buffer, mode: String) -> napi::Result<AddResult> {
         let batches = ipc_file_to_batches(buf.to_vec())
             .map_err(|e| napi::Error::from_reason(format!("Failed to read IPC file: {}", e)))?;
+        let batches = batches
```
westonpace, line 74: Is there a plan to eventually add something like a `Scannable` to nodejs?  Or is the thinking that we don't have a lot of really big streaming data sources in nodejs at the moment since there isn't really a pyarrow equivalent?

python/python/lancedb/scannable.py
```diff
@@ -0,0 +1,240 @@
...
+from .pydantic import LanceModel, model_to_dict
+
+
+@dataclass
+class Scannable:
+    schema: pa.Schema
+    num_rows: Optional[int]
+    # Factory function to create a new reader each time (supports re-scanning)
+    reader: Callable[[], pa.RecordBatchReader]
+    # Whether calling reader() multiple times yields fresh data
```
westonpace: Can you expand on this comment?  It's not really clear to me what this means.

Does `rescannable = false` mean that a second scan attempt will fail?  Or does it mean that a second scan attempt will return the same data (e.g. not fresh)?

python/python/lancedb/scannable.py
```diff
@@ -0,0 +1,240 @@
...
+    return Scannable(
+        schema=data.projected_schema,
+        num_rows=None,
+        reader=data.to_reader,
+        rescannable=False,
+    )
+
+
+@singledispatch
+def to_arrow(data) -> pa.Table:
```
westonpace: These are pretty nifty.  I guess a super minor nit would be that the file is named `scannable.py` but the `to_arrow` function isn't only related to scannables.

python/python/lancedb/table.py
```diff
@@ -3727,18 +3729,31 @@ async def add(
...
+            )
+        _register_optional_converters()
+        data = to_scannable(data)
+        try:
+            return await self._inner.add(data, mode or "append")
+        except RuntimeError as e:
+            if "Cast error" in str(e):
+                raise ValueError(e)
+            elif "Vector column contains NaN" in str(e):
+                raise ValueError(e)
```
westonpace, line 3754: Minor nit: Is it not possible to change the source error to `ValueError`?  Both a cast and a "thing contains NaN" feel like they should already be `ValueError`?

python/python/tests/test_table.py
```diff
@@ -854,6 +854,57 @@ def test_add_with_nans(mem_db: DBConnection):
     assert np.allclose(v, np.array([0.0, 0.0]))
 
 
+def test_add_with_nans(mem_db: DBConnection):
```
westonpace, line 857: Do we also need a `test_merge_insert_with_nan`?  Although perhaps inserting vectors with merge_insert is not a typical operation (though maybe not too uncommon if a user is replacing an embedding)

python/src/table/scannable.rs
```diff
@@ -0,0 +1,144 @@
...
+    pyarrow::{FromPyArrow, PyArrowType},
+};
+use lancedb::{
+    arrow::{SendableRecordBatchStream, SimpleRecordBatchStream},
+    data::scannable::Scannable,
+    Error,
+};
+use pyo3::{types::PyAnyMethods, FromPyObject, Py, PyAny, Python};
+
+/// Adapter that implements SourceData for a Python reader factory callable.
```
westonpace: What is `SourceData`?  Do you mean `Scannable`?

rust/lancedb/src/data/scannable.rs
```diff
@@ -257,12 +272,29 @@ impl Scannable for WithEmbeddingsScannable {
...
+                    .map(|(i, col)| {
+                        let target_type = output_schema.field(i).data_type();
+                        if col.data_type() == target_type {
+                            Ok(col.clone())
+                        } else {
+                            arrow_cast::cast(col, target_type).map_err(Error::from)
+                        }
+                    })
+                    .collect::<Result<_>>()?;
+                let result = RecordBatch::try_new(output_schema, columns)?;
```
westonpace, line 291: We don't need to address this here but this feels kind of like something `add` should handle on its own in the rust.  In other words, if `add` is called with a schema that is close enough to cast in this way we should just do it instead of forcing the user to do it.

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -76,7 +75,15 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
         self.add_result.lock().unwrap().clone()
     }
 
-    fn stream_as_body(data: SendableRecordBatchStream) -> DataFusionResult<reqwest::Body> {
+    /// Stream the input as an IPC body, capturing any stream errors into the
+    /// provided channel. Errors from the input plan (e.g. NaN rejection) would
+    /// otherwise be swallowed inside the HTTP body upload; by stashing them in
+    /// the channel we can surface them with their original message after the
+    /// request completes.
+    fn stream_as_ipc_body(
```
westonpace: Minor nit: what does "body" in `as_ipc_body` mean?

rust/lancedb/src/table/add_data.rs
```diff
@@ -73,16 +93,152 @@ impl AddDataBuilder {
         self
     }
 
+    pub fn on_nan_vectors(mut self, behavior: NaNVectorBehavior) -> Self {
```
westonpace, line 99: We should probably comment this function since it is public

rust/lancedb/src/table/add_data.rs
```diff
@@ -73,16 +93,152 @@ impl AddDataBuilder {
...
+
+pub struct PreprocessingOutput {
+    pub plan: Arc<dyn datafusion_physical_plan::ExecutionPlan>,
+    pub overwrite: bool,
+    pub rescannable: bool,
+    pub write_options: WriteOptions,
+    pub mode: AddDataMode,
+}
+
+pub async fn local_add_table(slf: &NativeTable, add: AddDataBuilder) -> Result<AddResult> {
```
westonpace: What does `local` mean?  Does it mean "not remote"?  Why is this a free function?  Why put this here and not inside of `NativeTable` where it was before?

rust/lancedb/src/table/add_data.rs
```diff
@@ -73,16 +93,152 @@ impl AddDataBuilder {
...
+/// If a column exists in table but not in input, that is okay - it may be filled with nulls.
+///
+/// If the types are not exactly the same, we will attempt to cast later - so that is also okay at this stage.
+///
+/// If the nullability is different, that is also okay - we can relax nullability when casting.
+fn validate_schema(input: &Schema, table: &Schema) -> Result<()> {
+    validate_fields(input.fields(), table.fields())
+}
+
+fn validate_fields(input: &Fields, table: &Fields) -> Result<()> {
```
westonpace, line 179: Why are these free functions and not part of `AddDataBuilder`?

rust/lancedb/src/table/datafusion/reject_nan.rs
```diff
@@ -0,0 +1,269 @@
...
+    let projection = ProjectionExec::try_new(exprs, input).map_err(Error::from)?;
+    Ok(Arc::new(projection))
+}
+
+/// A scalar UDF that passes through FixedSizeList arrays unchanged, but errors
+/// if any float values in the list are NaN.
+#[derive(Debug, Hash, PartialEq, Eq)]
+struct RejectNanUdf {
+    signature: Signature,
+}
```
westonpace, line 78: I'm not saying it has to change but this feels like a lot of overhead compared to something like stream's `map` function.

rust/lancedb/src/table/datafusion/scannable_exec.rs
```diff
@@ -0,0 +1,118 @@
...
+use datafusion_common::{stats::Precision, DataFusionError, Result as DFResult, Statistics};
+use datafusion_execution::{SendableRecordBatchStream, TaskContext};
+use datafusion_physical_expr::{EquivalenceProperties, Partitioning};
+use datafusion_physical_plan::{
+    execution_plan::EmissionType, DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties,
+};
+
+use crate::{arrow::SendableRecordBatchStreamExt, data::scannable::Scannable};
+
+pub struct ScannableExec {
```
westonpace, line 16: I'm a little confused.  I thought the whole point of `Scannable` was to expose the row count hint and retryability to `lance` so that we could make intelligent decisions about multiple partitions and retries.  However, once it is wrapped in `ScannableExec` it seems like we've converted it back to a plain record batch stream and so those hints are lost?  Will the retries and multiple partitions happen in some higher layer above add?

rust/lancedb/src/table/add_data.rs
```diff
@@ -73,16 +93,152 @@ impl AddDataBuilder {
         self
     }
 
+    pub fn on_nan_vectors(mut self, behavior: NaNVectorBehavior) -> Self {
+        self.on_nan_vectors = behavior;
+        self
+    }
+
     pub async fn execute(self) -> Result<AddResult> {
```
westonpace, line 104: Does anyone still call this method?  Maybe for remote tables?

wjones127:
> This seems like mostly glue code so far (no actually perf improvement or new functionality in this PR yet)? Or am I missing something?

That's correct. The real performance improvement will come when we can write data files in parallel, and that will come in follow up PRs. There's a lot of glue just needed to get the same behavior in the Rust pipeline as we have in the Python pipeline, and that's part of what this PR is setting up.

nodejs/src/table.rs
wjones127: It would be nice to. I'm not sure what they have, but it would at least be nice to let them write a script that points to a bunch of Parquet or CSV files and let's them ingest them in a streaming fashion.

python/python/lancedb/scannable.py
wjones127: Means it will fail. An iterator isn't re-scannable: it can only be consumed once. But data that's materialized in memory or on disk is re-scannable.

python/python/lancedb/scannable.py
wjones127: Yeah I can move it to a separate file.

python/python/lancedb/table.py
wjones127: Let me look again. If we don't have a plan where we convert the original error to a string, then I think this is possible.

python/python/tests/test_table.py
wjones127: That's not a bad idea.

python/src/table/scannable.rs
wjones127: Sorry, should be `Scannable`. I originally called it `SourceData`.

rust/lancedb/src/remote/table/insert.rs
wjones127: Maybe I should rename `stream_as_http_body()`?

rust/lancedb/src/table/add_data.rs
wjones127: I can put it back in the `NativeTable`. Was trying to move stuff since that file has gotten large. See: https://github.com/lancedb/lancedb/issues/2949

rust/lancedb/src/table/add_data.rs
wjones127: Wouldn't a user call that? Like:

```rust
table.add(...).mode(...).execute().await?;
```

rust/lancedb/src/table/datafusion/scannable_exec.rs
wjones127: > Will the retries and multiple partitions happen in some higher layer above add?

Yeah it will dictate how the plan is constructed and run.

The multiple partitions is TODO: it will use the rowcount hint to decide how to repartition before the `InsertExec` node.

The retries are implemented already here: https://github.com/lancedb/lancedb/pull/3029/changes/BASE..b0d48a3f113134ff1d5d7875cf5f81da20bea3db#diff-230314e49b306c9418c5ba11c2c39c48ab6d4eb6cdc7972b3b50988641815325R1196-R1220

rust/lancedb/src/table/datafusion/reject_nan.rs
wjones127: Yeah I think I can make it more concise using `create_udf` or something like that.

PR #3030, opened 2026-02-15
fix(python): graceful handling of empty result sets in hybrid search

### LINKED ISSUE #2425: bug(python): Graceful Handling of Empty Result Sets After Hard Filtering

omair445:
## Problem

When applying hard filters that result in zero matches, hybrid search crashes with `IndexError: list index out of range` during reranking. This happens because empty result tables are passed through the full reranker pipeline, which expects at least one result.

Traceback from the issue:
```
lancedb/query.py: in _combine_hybrid_results
    results = reranker.rerank_hybrid(fts_query, vector_results, fts_results)
lancedb/rerankers/answerdotai.py: in rerank_hybrid
...
IndexError: list index out of range
```

## Fix

Added an early return in `_combine_hybrid_results` when both vector and FTS results are empty. Instead of passing empty tables through normalization, reranking, and score restoration (which can fail in various ways), we now build a properly-typed empty result table with the `_relevance_score` column and return it directly.

## Test

Added `test_empty_hybrid_result_reranker` that exercises `_combine_hybrid_results` directly with empty vector and FTS tables, verifying:
- Returns empty table with correct schema  
- Includes `_relevance_score` column
- Respects `with_row_ids` flag

Closes #2425

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1790,8 +1790,10 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
                 [vector_results.schema, fts_results.schema],
             )
             empty = pa.table(
-                {col: pa.array([], type=combined_schema.field(col).type)
-                 for col in combined_schema.names}
+                {
+                    col: pa.array([], type=combined_schema.field(col).type)
+                    for col in combined_schema.names
+                }
             )
             empty = empty.append_column(
                 "_relevance_score", pa.array([], type=pa.float32())


omair445:
Thanks for the approval @wjones127! Fixed the ruff formatting lint — CI should pass now.

PR #3032, opened 2026-02-16
feat(rust): add expression builder API for type-safe query filters

### LINKED ISSUE #3038: Rust - support Expr in filter

Mesut-Doner:
## Summary

Adds a Rust expression builder API as a type-safe alternative to SQL strings for query filters.

## Motivation

Filtering with raw SQL strings can be awkward when using variables and special types:

Closes  #3038

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -950,7 +950,8 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
                 Filter::Sql(sql) => sql.clone(),
                 Filter::Datafusion(expr) => expr_to_sql_string(&expr)?,
             };
-            request = request.json(&serde_json::json!({ "predicate": filter_sql, "version": version }));
+            request =
+                request.json(&serde_json::json!({ "predicate": filter_sql, "version": version }));
         } else {
             let body = serde_json::json!({ "version": version });
             request = request.json(&body);


rust/lancedb/src/remote/table.rs
wjones127, line 1: Could you add one unit test to this file, testing that we can use the datafusion expressions here?

Mesut-Doner:
> Nice work. I like that you re-used the existing APIs. Just needs a test or two.
> 
> It seems like this only handles queries, which is fine. We should do projections in a follow up. I created separate subissues and linked the one for filters to this issue.

hi, Will. Thank you so much for taking the time and attention!

rust/lancedb/src/remote/table.rs
Mesut-Doner: Thanks @wjones127 I've added UT for table.rs remote

wjones127 (approved):

wjones127:
Thanks for adding the test. I think you'll need to rebase your changes, since we merged #3035. That PR is just a refactor, so should just be a matter of moving the changes over to the new file. LMK if you want assistance with that!

Mesut-Doner:
> Thanks for adding the test. I think you'll need to rebase your changes, since we merged #3035. That PR is just a refactor, so should just be a matter of moving the changes over to the new file. LMK if you want assistance with that!

hi @wjones127 I think I need your assistance. Could you please help me on rebase process? thanks

PR #3033, opened 2026-02-16
refactor(python)!: simplify reranker API to compute_scores/needs_columns

### LINKED ISSUE #3017: Refactor Reranker API
Move reranker to just be responsible for assigning `_relevance_score`, nothing more.

## Problem

Right now, re-rankers are too involved in operations of the query. They add and remove columns within the individual reranker implementations. For example, they will do things like drop _distance column. We instead want to limit their job to deciding a new ranking and not worry about projections. This will change the public API, but make it easier to implement consistent query projection semantics across all rerankers.

## New API

Let's design a new API for this. We should make the reranker just responsible for
computing the new `_relevance_score` column. Here's an initial pass

```python
class Reranker:
    """
    Reranker takes query results from ANN and FTS queries and computes a new
    combined relevance score for each result.

    Note that it's possible for there to just be one input (e.g. ANN only or FTS only).
    It's also possible to be multiple of the same type.

    ┌─────┐ ┌───────────┐ ┌─────┐ ┌───────────┐
    │ANN  │ │ANN        │ │FTS  │ │FTS        │
    │title│ │description│ │title│ │description│
    └──┬──┘ └────────┬──┘ ├─────┘ └───┬───────┘
       │             │    │           │        
       │          ┌──▼────▼┐          │        
       └─────────►│Reranker│◄─────────┘        
                  └────────┘                   

    FTS results are provided as a PyArrow table with at most the following columns:
    - _rowid: the rowid of the matching document
    - text: the text of the matching document
    - _score: the relevance score of the matching document according to FTS

    ANN results are provided as a PyArrow table with at most the following columns:
    - _rowid: the rowid of the matching document
    - _distance: the distance of the matching document according to ANN
    - text: the text of the matching document

    Rerankers should specify which of these columns are required using `needs_columns` method.
    By providing this, you allow the system to avoid loading unnecessary fields.
    """

    def needs_columns(self) -> List[Literal["_distance", "_score", "text", "_rowid"]]: ...

    def compute_scores(
        self,
        query: str,
        results: List[RerankableResult]
    ) -> List[pa.Array]: ...

class RerankableResult:
    pass

class FtsResult(RerankableResult):
    data: pa.Table

class VectorResult(RerankableResult):
    data: pa.Table
```

## Implementation

If we change the API, we could also change how they are executed. These could be converted into a DataFusion `ExecutionPlan`, which will allow simplified projection pushdown logic and sharing the implementation across languages.

Khangdang1690:
## Summary

Resolves #3017

Refactors the reranker API so that rerankers are **only responsible for computing relevance scores**, and no longer manage column projections, merging, deduplication, or sorting.

- **New `compute_scores(query, List[RerankableResult]) -> List[pa.Array]`** — each reranker returns one float32 score array per input result set
- **New `needs_columns() -> List[Literal[...]]`** — declarative column requirements enabling projection pushdown
- **New typed wrappers** — `RerankableResult`, `FtsResult`, `VectorResult` with `data: pa.Table`
- **New `_apply_reranker()` pipeline helper** in `query.py` — centralizes wrapping, projection pushdown, scoring, merging, deduplication, sorting, and column projection
- All 6 reranker call sites (sync/async vector, FTS, hybrid) now use `_apply_reranker()` directly
- All 10 rerankers updated: RRF, LinearCombination, MRR, CrossEncoder, Answerdotai (+Colbert), Cohere, OpenAI, Jina, VoyageAI
- Removed deprecated methods: `rerank_hybrid`, `rerank_vector`, `rerank_fts`, `rerank_multivector`, `merge_results`, `_handle_empty_results`, `_keep_relevance_score`, `_merge_and_keep_scores`, `_has_new_api`

**BREAKING CHANGE:** Reranker subclasses must now implement `compute_scores()` instead of `rerank_hybrid`/`rerank_vector`/`rerank_fts`.

## Test plan

- [x] All reranker unit tests pass (13 passed, 1 skipped for torch)
- [x] All hybrid query tests pass (9 passed)
- [x] All query tests pass (53 passed, 1 skipped)
- [x] New `compute_scores` tests cover: RRF, RRF single-vector, LinearCombination, MRR, MRR multivector, empty results
- [ ] CI passes on all platforms

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -34,9 +34,104 @@ from lancedb.background_loop import LOOP
 from . import __version__
 from .arrow import AsyncRecordBatchReader
 from .dependencies import pandas as pd
-from .rerankers.base import Reranker
+from .rerankers.base import Reranker, VectorResult, FtsResult, RerankableResult
 from .rerankers.rrf import RRFReranker
 from .rerankers.util import check_reranker_result
+
+
+def _apply_reranker(
+    reranker: Reranker,
+    query: str,
+    result_sets: List[Tuple[pa.Table, str]],
+    return_score: str = "relevance",
+) -> pa.Table:
+    """Apply a reranker using the new ``compute_scores`` API.
+
+    Parameters
+    ----------
+    reranker : Reranker
+        The reranker instance.
+    query : str
+        The search query string.
+    result_sets : list of (pa.Table, str)
+        Each element is ``(table, kind)`` where *kind* is ``"vector"``
+        or ``"fts"``.
+    return_score : str
+        ``"relevance"`` to drop ``_distance``/``_score``, or ``"all"``
+        to keep them.
+
+    Returns
+    -------
+    pa.Table
+        Merged, scored, sorted, and projected result table.
+    """
+    # Wrap raw tables in typed result objects
+    wrapped: List[RerankableResult] = []
+    for table, kind in result_sets:
+        if kind == "vector":
+            wrapped.append(VectorResult(data=table))
+        else:
+            wrapped.append(FtsResult(data=table))
+
+    # Handle all-empty case
+    if all(len(r.data) == 0 for r in wrapped):
+        empty = result_sets[0][0]
+        return empty.append_column(
+            "_relevance_score", pa.array([], type=pa.float32())
+        )
+
+    # Filter out empty tables before passing to reranker
+    non_empty = [r for r in wrapped if len(r.data) > 0]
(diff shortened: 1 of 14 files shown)


Khangdang1690:
@wjones127 please review my PR

wjones127:
Thanks for looking into this.

I'm not 100% sure on the direction of how we want to refactor this, so it might take some time to review it. 

Having done the work to do this refactor, what do you think of the new API? Does it work better?

PR #3034, opened 2026-02-16
fix: non-stopping dataset version check after passing the first consistency check interval

LuQQiu:
When a table has a read consistency interval, queries within the interval skip the version check. Once the interval expires, a list call checks for new versions. If the version hasn't changed, the timer should reset so the next interval begins, but it didn't.  The timer stayed expired, so every query after that triggered a list call, even though nothing changed.

This affects all read operations (queries, schema lookups, searches) on tables with read_consistency_interval set. Each operation adds a list("_versions/") call to object storage, adding latency proportional to the store's list performance. For high-QPS workloads, this can saturate object store list throughput and significantly degrade query latency.

Bug flow:
  1. Every read operation (query, schema, search) calls ensure_up_to_date()
  2. ensure_up_to_date() calls is_up_to_date(), which compares last_consistency_check.elapsed() against
   read_consistency_interval
  3. If the interval has expired, it calls reload()
  4. reload() calls need_reload(), which calls latest_version_id() — this is the list IOP
  (list("_versions/"))
  5. If no new version, reload() returns early without resetting last_consistency_check
  6. On the next query, step 2 sees the stale timer again → step 3 → step 4 → another list IOP
  7. This repeats on every query forever

### DIFF
--- a/rust/lancedb/src/table/dataset.rs
+++ b/rust/lancedb/src/table/dataset.rs
@@ -211,11 +211,7 @@ impl DatasetConsistencyWrapper {
     }
 
     pub async fn reload(&self) -> Result<()> {
-        let mut write_guard = self.0.write().await;
-        if !write_guard.is_up_to_date() {
-            write_guard.reload().await?;
-        }
-        Ok(())
+        self.0.write().await.reload().await
     }
 
     /// Returns the version, if in time travel mode, or None otherwise
@@ -242,7 +238,12 @@ impl DatasetConsistencyWrapper {
     /// version parameters.
     async fn ensure_up_to_date(&self) -> Result<()> {
         if !self.is_up_to_date().await {
-            self.reload().await?;
+            // Re-check under write lock — another task may have reloaded
+            // while we waited for the lock.
+            let mut write_guard = self.0.write().await;
+            if !write_guard.is_up_to_date() {
+                write_guard.reload().await?;
+            }
         }
         Ok(())
     }


wjones127:
What if we just eliminated `need_reload()` and just call `reload()` straight away?

Lance's `checkout_latest()` and `checkout_version()` already have early returns if there isn't anything new.

Plus, it's technically more efficient: current code will do a list call for `dataset.latest_version_id()` and then another list call for `checkout_latest()`. So if you skip `dataset.latest_version_id()` you also reduce the total number of list calls.

wjones127:
FYI I'm rewriting most of this module later this week in https://github.com/lancedb/lancedb/pull/3021, but you can still do this now if you want a faster fix.

LuQQiu:
@wjones127 Thanks for the review! Double check the code path, the overhead of version match case is 1 list op, it's no more than having a need_reload (and need_reload actually introduce much more list ops)
Change to fully remove the need_reload API

rust/lancedb/src/table/dataset.rs
```diff
@@ -205,33 +196,7 @@ impl DatasetConsistencyWrapper {
...
-            } = &mut *write_guard
-            {
-                last_consistency_check.replace(Instant::now());
-            }
-            return Ok(());
-        }
-
-        // actually need reloading
-        write_guard.reload().await
+        self.0.write().await.reload().await
```
wjones127: We should do one more check *after* we acquire the write lock that the `last_consistency_check` was long enough ago that we need to reload. Otherwise, you could get two tasks concurrently calling this, and each would then call reload() immediately, even though the second one shouldn't need to.

LuQQiu: added the check, thanks!

wjones127 (approved):

PR #3037, opened 2026-02-17
fix: deprecate outdated files for embedding registry

prrao87:
There are old and outdated files in our embedding registry that can confuse coding agents. This PR deprecates the following files that have newer, more modern methods to generate such embeddings.

- Deprecate `embeddings/siglip.py` 
- Deprecate `embeddings/gte.py` 

## Why this change?

Per a discussion with @AyushExel, the [embedding registry directory ](https://github.com/lancedb/lancedb/tree/1840aa7edcdbf0c3f7d3a303000c8f75842a2f55/python/python/lancedb/embeddings) in the LanceDB repo has a number of outdated files that need to be deprecated.

See https://github.com/lancedb/docs/issues/85 for the docs gaps that identified this.
- Add note in `openclip` docs that it can be used for SigLip embeddings, which it now supports
- Add note in the `sentence-transformers` page that ALL text embedding models on Hugging Face can be used

### DIFF
--- a/python/python/lancedb/embeddings/gte.py
+++ b/python/python/lancedb/embeddings/gte.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
 
+import warnings
 from typing import List, Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


AyushExel:
Oh i'd prefer to deprecate them than just remove them as users depending on these modules will run into errors. 
Maybe lets add deprecation warning for a couple releases and then remove them?

prrao87:
Got it, will refactor this PR to do deprecation then.

PR #3055, opened 2026-02-20
fix: update DatasetConsistencyWrapper to accept same-version updates

wjones127:
## Summary

`DatasetConsistencyWrapper::update()` only stored datasets with a strictly newer
version. This caused `migrate_manifest_paths_v2` to silently drop its update since
the migration renames files without bumping the dataset version. The subsequent
`uses_v2_manifest_paths()` call would then return the stale cached dataset.

Changed the version check from `>` to `>=` so same-version updates are accepted.

## Test plan

- [x] Existing `test_create_table_v2_manifest_paths_async` Python test should pass
- [x] Existing `should be able to migrate tables to the V2 manifest paths` NodeJS test should pass
- [x] All dataset wrapper unit tests pass locally

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - Cargo.toml
       - nodejs/**
+      - rust/**
       - docs/src/js/**
       - .github/workflows/nodejs.yml
       - docker-compose.yml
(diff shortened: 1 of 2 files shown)


LuQQiu:
that's interesting.... 
migrate_manifest_paths_v2 <-- what if two happens together
what happen if migrate_manifest_paths_v2 happens when other write happen together
migrate_manifest_paths_v2 seems a special operation, why it does not create a new version?

LuQQiu (approved):

PR #3062, opened 2026-02-23
feat: parallel inserts for local tables

wjones127:
When input data is sufficiently large, we automatically split up into parallel writes using a round-robin exchange operator. We sample the first batch to determine data width, and target size of 1 million rows or 2GB, whichever is smaller.

### DIFF
--- a/rust/lancedb/src/data/scannable.rs
+++ b/rust/lancedb/src/data/scannable.rs
@@ -435,11 +435,8 @@ impl Scannable for PeekedScannable {
             }),
             (None, Some(rest)) => {
                 if let Some(err) = error_item {
-                    let prepend = futures::stream::once(std::future::ready(err));
-                    Box::pin(SimpleRecordBatchStream {
-                        schema,
-                        stream: prepend.chain(rest),
-                    })
+                    let stream = futures::stream::once(std::future::ready(err));
+                    Box::pin(SimpleRecordBatchStream { schema, stream })
                 } else {
                     rest
                 }
@@ -456,7 +453,8 @@ impl Scannable for PeekedScannable {
 ///
 /// `sample_bytes` and `sample_rows` come from a representative batch and are
 /// used to estimate per-row size. `total_rows_hint` is the total row count
-/// when known; otherwise `sample_rows` is used as the estimate.
+/// when known; otherwise `sample_rows` row count is used as a lower bound
+/// estimate.
 ///
 /// Targets roughly 1 million rows or 2 GB per partition, capped at
 /// `max_partitions` (typically the number of available CPU cores).
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -10,6 +10,7 @@ use datafusion_execution::TaskContext;
 use datafusion_expr::Expr;
 use datafusion_physical_plan::display::DisplayableExecutionPlan;
 use datafusion_physical_plan::ExecutionPlan;
+use futures::stream::FuturesUnordered;
 use futures::StreamExt;
 use lance::dataset::builder::DatasetBuilder;
 pub use lance::dataset::ColumnAlteration;
@@ -2125,9 +2126,7 @@ impl BaseTable for NativeTable {
         // write parallelism.
         let mut peeked = PeekedScannable::new(add.data);
         let num_partitions = if let Some(first_batch) = peeked.peek().await {
-            let max_partitions = std::thread::available_parallelism()
-                .map(|p| p.get())
-                .unwrap_or(1);
+            let max_partitions = lance_core::utils::tokio::get_num_compute_intensive_cpus();
             estimate_write_partitions(
                 first_batch.get_array_memory_size(),
                 first_batch.num_rows(),
@@ -2168,7 +2167,7 @@ impl BaseTable for NativeTable {
 
         // Execute all partitions in parallel.
         let task_ctx = Arc::new(TaskContext::default());
-        let mut handles = Vec::with_capacity(num_partitions);
+        let handles = FuturesUnordered::new();
         for partition in 0..num_partitions {
             let exec = insert_exec.clone();
             let ctx = task_ctx.clone();


wjones127:
## Performance improvement

Here's the results for writing `lineitem` TPC-H table at scale factor 10. Before: 350 MB/s vs now at ~1100-2000 MB/s

bench_plot


<details><summary>Details</summary>
<p>

Generate data:

```
tpchgen-cli -s 10 --tables lineitem --parts 5 --format=parquet
```

Benchmark:

```python
import time

import pyarrow.dataset as pa_ds
import os
import lancedb
import shutil

parquet_ds = pa_ds.dataset("lineitem")

shutil.rmtree("demo", ignore_errors=True)
db = lancedb.connect("demo")
table = db.create_table("lineitem", schema=parquet_ds.schema, mode="overwrite")

os.system("sync")
start = time.time()
table.add(parquet_ds)
print(f"Time taken: {time.time() - start:.2f} seconds")

# Get total size of all .lance files in the demo/lineitem directory
total_size = 0
for dirpath, dirnames, filenames in os.walk("demo/lineitem.lance"):
    for filename in filenames:
        if filename.endswith(".lance"):
            total_size += os.path.getsize(os.path.join(dirpath, filename))
print(f"Total size of .lance files: {total_size / (1024 * 1024):.2f} MB")
```

</p>
</details>

rust/lancedb/src/data/scannable.rs
```diff
@@ -349,6 +348,135 @@ pub fn scannable_with_embeddings(
...
+                schema,
+                stream: futures::stream::once(std::future::ready(Ok(batch))),
+            }),
+            (None, Some(rest)) => {
+                if let Some(err) = error_item {
+                    let prepend = futures::stream::once(std::future::ready(err));
+                    Box::pin(SimpleRecordBatchStream {
+                        schema,
+                        stream: prepend.chain(rest),
+                    })
```
westonpace: I guess this is technically accurate though it seems odd to chain `rest` when `prepend` is already an error.

rust/lancedb/src/data/scannable.rs
```diff
@@ -349,6 +348,135 @@ pub fn scannable_with_embeddings(
...
+            }
+        }
+    }
+}
+
+/// Compute the number of write partitions based on data size estimates.
+///
+/// `sample_bytes` and `sample_rows` come from a representative batch and are
+/// used to estimate per-row size. `total_rows_hint` is the total row count
+/// when known; otherwise `sample_rows` is used as the estimate.
```
westonpace: > otherwise `sample_rows` is used as the estimate.

This was a bit confusing so I had to look at the impl.  This means "if total rows is not provided we pretend the sample batch is all the data?"

I don't have any better idea but wanted to verify my understanding.

rust/lancedb/src/table.rs
```diff
@@ -2122,6 +2121,24 @@ impl BaseTable for NativeTable {
 
         let table_schema = Schema::from(&ds.schema().clone());
 
+        // Peek at the first batch to estimate a good partition count for
+        // write parallelism.
+        let mut peeked = PeekedScannable::new(add.data);
+        let num_partitions = if let Some(first_batch) = peeked.peek().await {
+            let max_partitions = std::thread::available_parallelism()
```
westonpace: Hmm...should we call into lance's "num compute threads"?  That way this is bounded by lance's environment variables?

rust/lancedb/src/table.rs
```diff
@@ -2135,18 +2152,41 @@ impl BaseTable for NativeTable {
...
+        let plan = if num_partitions > 1 {
+            Arc::new(
+                datafusion_physical_plan::repartition::RepartitionExec::try_new(
+                    output.plan,
+                    datafusion_physical_plan::Partitioning::RoundRobinBatch(num_partitions),
+                )?,
+            ) as Arc<dyn ExecutionPlan>
+        } else {
+            output.plan
+        };
```
westonpace, line 2164: Ah, so this answers my earlier question which you probably answered but I didn't understand.  If there is a lot of data you aren't actually partitioning the source.  You are just scanning the source as normal but shuffling the output.

That's why `Scannable` doesn't need a `partition` in the `scan_as_stream` method.

rust/lancedb/src/table.rs
```diff
@@ -2135,18 +2152,41 @@ impl BaseTable for NativeTable {
...
+                    batch.map_err(|e| -> Error { e.into() })?;
+                }
+                Ok::<_, Error>(())
+            }));
+        }
+        for handle in handles {
+            handle.await.map_err(|e| Error::Runtime {
+                message: format!("Insert task panicked: {}", e),
+            })??;
+        }
```
westonpace, line 2188: Minor nit: using `FuturesUnordered` here or some kind of "select" logic would allow you to detect errors on trailing partitions faster.  Otherwise if the last partition fails quickly you won't detect it until the earlier writes are already done.

rust/lancedb/src/table.rs
```diff
@@ -2135,18 +2152,41 @@ impl BaseTable for NativeTable {
...
+                    .execute(partition, ctx)
+                    .map_err(|e| -> Error { e.into() })?;
+                while let Some(batch) = stream.next().await {
+                    batch.map_err(|e| -> Error { e.into() })?;
+                }
+                Ok::<_, Error>(())
+            }));
+        }
+        for handle in handles {
+            handle.await.map_err(|e| Error::Runtime {
```
westonpace, line 2185: How are errors handled?  What is partition A finishes quickly and then partition N eventually errors out.  Does that mean some data is committed?  Or is it still an atomic write?

westonpace: Ah, nvm, I looked at `InsertExec` again, I see that it is an atomic op.  The last writer to finish executes the commit.

I guess a minor nit is that I'd almost expect there to be a `CommitExec` which has N different children and executes them all in parallel and then does the commit.  That way you wouldn't need this logic here.  The final plan would always have a single partition.  That being said, I see no reason this impl wouldn't work.

rust/lancedb/src/table.rs
```diff
@@ -2135,18 +2152,41 @@ impl BaseTable for NativeTable {
...
+        } else {
+            output.plan
+        };
 
-        let stream = execute_plan(plan, Default::default())?;
-        stream
-            .try_collect::<Vec<_>>()
-            .await
-            .map_err(crate::Error::from)?;
+        let insert_exec = Arc::new(InsertExec::new(ds_wrapper.clone(), ds, plan, lance_params));
```
westonpace, line 2166: How does this work when remote / using namespace?  It looks like `InsertExec` is always going to run directly against storage?

wjones127: There's a separate `RemoteInsertExec` here:

https://github.com/lancedb/lancedb/blob/0d3fc7860a5975238786cbc27c3a6047434568df/rust/lancedb/src/remote/table/insert.rs#L33

It only accepts a single input partition for now because the server doesn't yet have endpoints that support multipart writes. But a future version of the server (currently in progress) will have that.

rust/lancedb/src/table.rs
wjones127: I felt doing it this way was simpler for now. Didn't have to think about how to pass transaction data from the InsertExec to the CommitExec as Arrow data, for example.

rust/lancedb/src/table.rs
wjones127: That's a good point. I'll hook into that.

rust/lancedb/src/data/scannable.rs
wjones127: > This means "if total rows is not provided we pretend the sample batch is all the data?"

Yeah, it's a bit confusing. But basically, if we don't know the total rows, we won't be able to parallelize as much automatically. There's an edge case here where the first batch on it's own is large enough to justify parallelism.

I plan on adding a `write_parallelism` parameter that can be used to manually specify the write parallelism.

rust/lancedb/src/data/scannable.rs
wjones127: 🤣

PR #3067, opened 2026-02-24
feat: progress bar for `add()`

wjones127:
## Summary

Adds progress reporting for `table.add()` so users can track large write operations. The progress callback is available in Rust, Python (sync and async), and through the PyO3 bindings.

### Usage

Pass `progress=True` to get an automatic tqdm bar:

```python
table.add(data, progress=True)
# 100%|██████████| 1000000/1000000 [00:12<00:00, 82345 rows/s, 45.2 MB/s | 4/4 workers]
```

Or pass a tqdm bar for more control:

```python
from tqdm import tqdm

with tqdm(unit=" rows") as pbar:
    table.add(data, progress=pbar)
```

Or use a callback for custom progress handling:

```python
def on_progress(p):
    print(f"{p['output_rows']}/{p['total_rows']} rows, "
          f"{p['active_tasks']}/{p['total_tasks']} workers, "
          f"done={p['done']}")

table.add(data, progress=on_progress)
```

In Rust:

```rust
table.add(data)
    .progress(|p| println!("{}/{:?} rows", p.output_rows(), p.total_rows()))
    .execute()
    .await?;
```

### Details

- `WriteProgress` struct in Rust with getters for `elapsed`, `output_rows`, `output_bytes`, `total_rows`, `active_tasks`, `total_tasks`, and `done`. Fields are private behind getters so new fields can be added without breaking changes.
- `WriteProgressTracker` tracks progress across parallel write tasks using a mutex for row/byte counts and atomics for active task counts.
- Active task tracking uses an RAII guard pattern (`ActiveTaskGuard`) that increments on creation and decrements on drop.
- For remote writes, `output_bytes` reflects IPC wire bytes rather than in-memory Arrow size. For local writes it uses in-memory Arrow size as a proxy (see TODO below).
- tqdm postfix displays throughput (MB/s) and worker utilization (active/total).
- The `done` callback always fires, even on error (via `FinishOnDrop`), so progress bars are always finalized.

### TODO

- Track actual bytes written to disk for local tables. This requires Lance to expose a progress callback from its write path. See lance-format/lance#6247.

### DIFF
--- a/python/python/lancedb/remote/table.py
+++ b/python/python/lancedb/remote/table.py
@@ -4,7 +4,7 @@
 from datetime import timedelta
 import logging
 from functools import cached_property
-from typing import Any, Dict, Iterable, List, Optional, Union, Literal
+from typing import Any, Callable, Dict, Iterable, List, Optional, Union, Literal
 import warnings
 
 from lancedb._lancedb import (
@@ -35,6 +35,7 @@ import pyarrow as pa
 from lancedb.common import DATA, VEC, VECTOR_COLUMN_NAME
 from lancedb.merge import LanceMergeInsertBuilder
 from lancedb.embeddings import EmbeddingFunctionRegistry
+from lancedb.table import _normalize_progress
 
 from ..query import LanceVectorQueryBuilder, LanceQueryBuilder, LanceTakeQueryBuilder
 from ..table import AsyncTable, IndexStatistics, Query, Table, Tags
@@ -308,7 +309,7 @@ class RemoteTable(Table):
         mode: str = "append",
         on_bad_vectors: str = "error",
         fill_value: float = 0.0,
-        progress: Optional[Any] = None,
+        progress: Optional[Union[bool, Callable, Any]] = None,
     ) -> AddResult:
         """Add more data to the [Table](Table). It has the same API signature as
         the OSS version.
@@ -340,8 +341,6 @@ class RemoteTable(Table):
         AddResult
             An object containing the new version number of the table after adding data.
         """
-        from lancedb.table import _normalize_progress
-
         progress, owns = _normalize_progress(progress)
         try:
             return LOOP.run(
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -14,6 +14,7 @@ from functools import cached_property
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     Iterable,
     List,
@@ -989,7 +990,7 @@ class Table(ABC):
         mode: AddMode = "append",
         on_bad_vectors: OnBadVectorsType = "error",
         fill_value: float = 0.0,
-        progress: Optional[Any] = None,
+        progress: Optional[Union[bool, Callable, Any]] = None,
     ) -> AddResult:
         """Add more data to the [Table](Table).
 
@@ -2531,7 +2532,7 @@ class LanceTable(Table):
         mode: AddMode = "append",
         on_bad_vectors: OnBadVectorsType = "error",
         fill_value: float = 0.0,
-        progress: Optional[Any] = None,
+        progress: Optional[Union[bool, Callable, Any]] = None,
     ) -> AddResult:
         """Add data to the table.
         If vector columns are missing and the table
@@ -3821,7 +3822,7 @@ class AsyncTable:
         mode: Optional[Literal["append", "overwrite"]] = "append",
         on_bad_vectors: Optional[OnBadVectorsType] = None,
         fill_value: Optional[float] = None,
-        progress: Optional[Any] = None,
+        progress: Optional[Union[bool, Callable, Any]] = None,
     ) -> AddResult:
         """Add more data to the [Table](Table).
 
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -5,7 +5,7 @@ pub mod insert;
 
 use self::insert::RemoteInsertExec;
 use crate::expr::expr_to_sql_string;
-use crate::table::write_progress::WriteProgressTracker;
+use crate::table::write_progress::FinishOnDrop;
 
 use super::ARROW_STREAM_CONTENT_TYPE;
 use super::client::RequestResultExt;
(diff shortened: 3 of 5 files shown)


python/python/lancedb/remote/table.py
```diff
@@ -330,17 +331,31 @@ def add(
...
 
         Returns
         -------
         AddResult
             An object containing the new version number of the table after adding data.
         """
-        return LOOP.run(
-            self._table.add(
-                data, mode=mode, on_bad_vectors=on_bad_vectors, fill_value=fill_value
+        from lancedb.table import _normalize_progress
```
westonpace: Should we prefer top-level imports?

python/python/lancedb/table.py
```diff
@@ -556,6 +556,21 @@ def _table_uri(base: str, table_name: str) -> str:
...
 
 
+def _normalize_progress(progress):
+    """Normalize a ``progress`` parameter for :meth:`Table.add`.
+
+    Returns ``(progress_obj, owns)`` where *owns* is True when we created a
+    tqdm bar that the caller must close.
+    """
+    if progress is True:
+        from tqdm.auto import tqdm
```
westonpace, line 567: Same question about top-level imports (though maybe this one should stay local to keep tqdm an optional dependency?)

python/src/table.rs
```diff
@@ -312,6 +314,70 @@ impl Table {
         } else {
             return Err(PyValueError::new_err(format!("Invalid mode: {}", mode)));
         }
+        if let Some(progress_obj) = progress {
+            let is_callable = Python::attach(|py| progress_obj.bind(py).is_callable());
+            if is_callable {
+                // Callback: call with a dict of progress info.
+                op = op.progress(move |p| {
+                    Python::attach(|py| {
```
westonpace, line 322: This could be a follow-up (and maybe not all that critical) but `Python::attach` is a potentially blocking method (if anything else is using the GIL) so I'm guessing this will potentially block the tokio worker thread when the callback is called.

A slightly more efficient implementation might be to use something like a channel here to a dedicated thread that consumes the channel and calls the python callbacks.

Still, I assume we are not updating progress _that_ often and so probably will get away with it in general.

python/src/table.rs
```diff
@@ -312,6 +314,70 @@ impl Table {
...
+                    let prev = last_rows.swap(current, std::sync::atomic::Ordering::Relaxed);
+                    Python::attach(|py| {
+                        if let Some(total) = p.total_rows()
+                            && !total_set.load(std::sync::atomic::Ordering::Relaxed)
+                        {
+                            if let Err(e) = progress_obj.setattr(py, "total", total) {
+                                eprintln!("progress setattr error: {e}");
+                            }
+                            total_set.store(true, std::sync::atomic::Ordering::Relaxed);
+                        }
```
westonpace, line 359: It's slightly strange to see an atomic bool used in this way.  I'd normally expect to see `compare_exchange`.  The current implementation could allow multiple callers to call `progress_obj.setattr` if they arrived at roughly the same time.

However, I'm guessing you don't actually expect there to be concurrent callers in this callback and are only using atomics for borrow-checker convenience?

If you changed the callback to be `FnMut` instead of `Fn` then I think you could just use a standard `usize` and `bool` instead of atomics.

python/src/table.rs
```diff
@@ -312,6 +314,70 @@ impl Table {
...
+                                let mb_per_sec = p.output_bytes() as f64 / elapsed / 1_000_000.0;
+                                let postfix = format!(
+                                    "{:.1} MB/s | {}/{} workers",
+                                    mb_per_sec,
+                                    p.active_tasks(),
+                                    p.total_tasks()
+                                );
+                                progress_obj
+                                    .call_method1(py, "set_postfix_str", (postfix,))
+                                    .ok();
```
westonpace: Is this call to `ok` silently swallowing errors?

python/src/table.rs
```diff
@@ -312,6 +314,70 @@ impl Table {
...
+                            }
+                        }
+                        if p.done() {
+                            // Force a final refresh so the bar shows completion.
+                            progress_obj.call_method0(py, "refresh").ok();
```
westonpace: Is this call to `ok` silently swallowing errors?

python/src/table.rs
```diff
@@ -312,6 +314,70 @@ impl Table {
...
+                        {
+                            if let Err(e) = progress_obj.setattr(py, "total", total) {
+                                eprintln!("progress setattr error: {e}");
+                            }
+                            total_set.store(true, std::sync::atomic::Ordering::Relaxed);
+                        }
+                        let delta = current.saturating_sub(prev);
+                        if delta > 0 {
+                            if let Err(e) = progress_obj.call_method1(py, "update", (delta,)) {
+                                eprintln!("progress update error: {e}");
```
westonpace: Should this be a log warning instead of a println?

python/src/table.rs
```diff
@@ -312,6 +314,70 @@ impl Table {
...
+                let total_set = std::sync::atomic::AtomicBool::new(false);
+                op = op.progress(move |p| {
+                    let current = p.output_rows();
+                    let prev = last_rows.swap(current, std::sync::atomic::Ordering::Relaxed);
+                    Python::attach(|py| {
+                        if let Some(total) = p.total_rows()
+                            && !total_set.load(std::sync::atomic::Ordering::Relaxed)
+                        {
+                            if let Err(e) = progress_obj.setattr(py, "total", total) {
+                                eprintln!("progress setattr error: {e}");
```
westonpace: Should we log a warning instead of a println?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -262,12 +285,22 @@ impl<S: HttpSend + 'static> ExecutionPlan for RemoteInsertExec<S> {
...
         let input_stream = self.input.execute(partition, context)?;
+        let baseline = BaselineMetrics::new(&self.metrics, partition);
+        let input_schema = input_stream.schema();
+        let input_stream: SendableRecordBatchStream = Box::pin(RecordBatchStreamAdapter::new(
+            input_schema,
+            input_stream.map_ok(move |batch| {
+                baseline.record_output(batch.num_rows());
+                batch
+            }),
+        ));
```
westonpace: Could you use InstrumentedRecordBatchStreamAdapter?

rust/lancedb/src/table/datafusion/insert.rs
```diff
@@ -176,6 +182,19 @@ impl ExecutionPlan for InsertExec {
...
+        let input_schema = input_stream.schema();
+        let input_stream: SendableRecordBatchStream = Box::pin(RecordBatchStreamAdapter::new(
+            input_schema,
+            input_stream.map_ok(move |batch| {
+                let _timer = baseline.elapsed_compute().timer();
+                baseline.record_output(batch.num_rows());
+                output_bytes.add(batch.get_array_memory_size());
+                batch
+            }),
+        ));
```
westonpace: Could you use InstrumentedRecordBatchStreamAdapter?

python/python/lancedb/table.py
wjones127: I think tqdm is currently a hard dependency, so we can actually pull it up.

PR #3071, opened 2026-02-25
feat(rust): parallel inserts for remote tables via multipart write

### LINKED ISSUE #2861: Feature: Split large materialized or scannable inputs into multiple HTTP calls
### SDK

Rust

### Description

If we have a large data input to a remote add call, instead of doing just one HTTP call to the remote server, we should make several in parallel. This should improve throughput.

Should be done after #2602

wjones127:
Similar to https://github.com/lancedb/lancedb/pull/3062, we can write in parallel to remote tables if the input data source is large enough.

We take advantage of new endpoints coming in server version 0.4.0, which allow writing data in multiple requests, and the committing at the end in a single request.

To make testing easier, I also introduce a `write_parallelism` parameter. In the future, we can expose that in Python and NodeJS so users can manually specify the parallelism they get.

Closes #2861

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -10,7 +10,7 @@ use super::ARROW_STREAM_CONTENT_TYPE;
 use super::client::RequestResultExt;
 use super::client::{HttpSend, RestfulLanceDbClient, Sender};
 use super::db::ServerVersion;
-use crate::data::scannable::{estimate_write_partitions, PeekedScannable, Scannable};
+use crate::data::scannable::{PeekedScannable, Scannable, estimate_write_partitions};
 use crate::index::Index;
 use crate::index::IndexStatistics;
 use crate::index::waiter::wait_for_index;


rust/lancedb/src/remote/table/insert.rs
```diff
@@ -262,32 +316,43 @@ impl<S: HttpSend + 'static> ExecutionPlan for RemoteInsertExec<S> {
...
+                    })?
+                };
 
-            {
                 let mut res_lock = add_result.lock().map_err(|_| {
                     DataFusionError::Execution("Failed to acquire lock for add_result".to_string())
                 })?;
                 *res_lock = Some(parsed_result);
+            } else {
+                // We don't use the body in this case, but we should still consume it.
```
westonpace: No harm since the body is probably small but why bother?

rust/lancedb/src/table/add_data.rs
```diff
@@ -101,6 +103,15 @@ impl AddDataBuilder {
         self
     }
 
+    /// Set the number of parallel write streams.
+    ///
+    /// By default, the number of streams is estimated from the data size.
+    /// Setting this to `1` disables parallel writes.
+    pub fn write_parallelism(mut self, parallelism: usize) -> Self {
+        self.write_parallelism = Some(parallelism.max(1));
```
westonpace: Minor nit but should we throw an error on 0 just to avoid users thinking it means "default" or "unbounded"?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -38,21 +42,61 @@ pub struct RemoteInsertExec<S: HttpSend = Sender> {
...
         input: Arc<dyn ExecutionPlan>,
         overwrite: bool,
     ) -> Self {
+        Self::new_inner(table_name, identifier, client, input, overwrite, None)
+    }
+
+    /// Create a multi-partition RemoteInsertExec for use with multipart writes.
+    ///
+    /// Each partition's insert is staged under the given `upload_id` without
+    /// committing. The caller is responsible for calling the complete (or abort)
```
westonpace, line 63: What happens if the caller never does?  Is there an automatic cleanup at some point?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -25,10 +27,12 @@ use crate::Error;
...
-/// This plan:
-/// 1. Requires single partition (no parallel remote inserts yet)
-/// 2. Streams data as Arrow IPC to `/v1/table/{id}/insert/` endpoint
-/// 3. Stores AddResult for retrieval after execution
+/// Streams data as Arrow IPC to `/v1/table/{id}/insert/` endpoint.
+///
+/// When `upload_id` is set, inserts are staged as part of a multipart write
+/// session and the plan supports multiple partitions for parallel uploads.
+/// Without `upload_id`, the plan requires a single partition and commits
+/// immediately.
```
westonpace, line 35: I don't think this is your fault or neccesarily fixable but my brain, when it hears multipart, always thinks of the HTTP content type multipart, which is "multipart parts, one request" and not the S3 multipart definition which is "multiple HTTP requests"

rust/lancedb/src/remote/table/insert.rs
wjones127: I'm looking into that. Lance's cleanup should handle that automatically, but need to make sure that is configured correctly on servers.

rust/lancedb/src/table/add_data.rs
wjones127: That's fair. We can do that.

rust/lancedb/src/remote/table/insert.rs
wjones127: I guess I can remove it. Was thinking it was important to consume, but might not be since we already got a successful status.

PR #3075, opened 2026-02-26
fix: propagate cast errors in `add()`

wjones127:
When we write data with `add()`, we can input data to the table's schema. However, we were using "safe" mode, which propagates errors as nulls. For example, if you pass `u64::max` into a field that is a `u32`, it will just write null instead of giving overflow error. Now it propagates the overflow. This is the same behavior as other systems like DuckDB.

### DIFF
--- a/rust/lancedb/src/table/delete.rs
+++ b/rust/lancedb/src/table/delete.rs
@@ -1,3 +1,4 @@
+use futures::FutureExt;
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 use serde::{Deserialize, Serialize};
@@ -23,7 +24,7 @@ pub struct DeleteResult {
 pub(crate) async fn execute_delete(table: &NativeTable, predicate: &str) -> Result<DeleteResult> {
     table.dataset.ensure_mutable()?;
     let mut dataset = (*table.dataset.get().await?).clone();
-    let delete_result = dataset.delete(predicate).await?;
+    let delete_result = dataset.delete(predicate).boxed().await?;
     let num_deleted_rows = delete_result.num_deleted_rows;
     let version = dataset.version().version;
     table.dataset.update(dataset);


rust/lancedb/src/table/datafusion/cast.rs
```diff
@@ -495,4 +554,266 @@ mod tests {
...
+        let ctx = SessionContext::new();
+        let stream = casted.execute(0, ctx.task_ctx()).unwrap();
+        let result: Result<Vec<RecordBatch>, _> = stream.try_collect().await;
+        assert!(result.is_err(), "expected overflow error at execution time");
+    }
+
+    #[tokio::test]
+    async fn test_float_to_int_rejected() {
+        // Arrow's safe:false does not error on truncation (1.5 → 1 silently), so
+        // float→int must be rejected at plan time rather than relying on runtime errors.
```
wkalt: this is the behavior I'd expect on a float to int cast. Should we be rejecting it?

rust/lancedb/src/table/datafusion/cast.rs
```diff
@@ -495,4 +554,266 @@ mod tests {
...
+        assert!(is_safe_cast(&DataType::UInt32, &DataType::Int64));
+        assert!(is_safe_cast(&DataType::Int64, &DataType::Float64));
+        assert!(is_safe_cast(
+            &DataType::LargeList(Arc::new(Field::new("item", DataType::Float64, false))),
+            &DataType::FixedSizeList(Arc::new(Field::new("item", DataType::Float16, false)), 32)
+        ));
+
+        // Disallowed casts:
+        assert!(!is_safe_cast(&DataType::Int32, &DataType::Utf8));
+        assert!(!is_safe_cast(&DataType::Float64, &DataType::Int32));
```
wkalt: in postgres for instance - 
```
[local]:wyatt=# select 3.14::int;
 int4 
------
    3
(1 row)

Time: 0.184 ms
[local]:wyatt=# select 3.14::double precision::int;
 int4 
------
    3
(1 row)

Time: 0.201 ms
[local]:wyatt=#
```

wkalt: duckdb:
```
D select 3.14::int;
┌───────────────────────┐
│ CAST(3.14 AS INTEGER) │
│         int32         │
├───────────────────────┤
│           3           │
└───────────────────────┘
D
```

wjones127: This situation is a bit different. This is specifically for cast on insert, not when a user requests a cast. Like if the user does:

```python
table = db.create_table("table", schema=pa.schema({'value': pa.int32()})

table.add([{'value': 1.5}])
```

We want to error.

wkalt: I don't know if you accept the argument, but at least some other systems do not error here:

```
[local]:wyatt=# create table foo(x int);                                                                                                                                           
CREATE TABLE
Time: 16.452 ms
[local]:wyatt=# insert into foo values(3.15);                                                                                                                                      
INSERT 0 1
Time: 15.468 ms
[local]:wyatt=# select * from foo;                                                                                                                                                 
 x 
---
 3
(1 row)
```

```
D create table foo(x int);
D insert into foo values (3.14);
D select * from foo;
┌───────┐
│   x   │
│ int32 │
├───────┤
│   3   │
└───────┘
D
```

wjones127: hmmm Okay. If that's the case, then I'm fine opening up to more casts. I was worried it might be surprising to users, but I guess it exists in other systems.

wjones127: How much more should I open up? Two other examples:

* int -> string?
* decimal(20, 5) -> decimal(10, 2)?  (loss of precision in decimals?)

wkalt (approved):

PR #3096, opened 2026-03-03
fix: pre-filtering on hybrid search

marca116:
When using hybrid search with a where filter, the prefilter argument is silently inverted. Passing prefilter=True actually performs post-filtering, and prefilter=False actually performs pre-filtering.


wjones127 (changes requested):
Could you write a unit test for this? To test this, you should be able to call `.explain_plan()` on the hybrid query and assert that the filter comes before the vector query and the FTS query.

marca116:
> Could you write a unit test for this? To test this, you should be able to call `.explain_plan()` on the hybrid query and assert that the filter comes before the vector query and the FTS query.

I've added a unit test for the issue. Checking whether full_filter=id appears in the plan seems sufficient to validate it.

Would you also like me to verify that the filter comes before the vector query?

marca116:
> Looks good. Could you fix the formatting before merging? Make sure you have the matching version of `ruff`.
> 
> I would do that for you, but since the PR is off you `main` branch I can't push to it.

The formatting should be fixed now.

PR #3102, opened 2026-03-04
feat: allow passing azure client/tenant ID through remote SDK

wkalt:
Prior to this commit we supported passing the azure storage account name to the lancedb remote SDK through headers. This adds support for client ID and tenant ID as well.


wkalt:
> Just the perhaps naive question of "why aren't we just serializing _all_ options"?

Good question, not sure. Let me revisit that in a followup; I don't have a clear sense of whether there is any plumbing or tradeoffs involved with that. We don't need users sending us e.g unrelated API keys we aren't gonna use, but I don't know if there is a risk of that.

Issue #2789, opened 2025-11-18
Feature: prewarm cache for all index types


hpvd:
> There is a prewarm function for indexes. We don't yet fully support it for all index types (notably I don't think it is supported for vector indexes yet) but that's just a matter of work, not an intentional omission.

=> prewarming should be available for all index types to make also first hit fast

from https://github.com/lancedb/lancedb/issues/2777#issuecomment-3544902135

PR #3110, opened 2026-03-06
feat: support prewarm_index and prewarm_data on remote tables

esteban:
## Summary

- Implement `RemoteTable.prewarm_data(columns)` calling `POST /v1/table/{id}/page_cache/prewarm/`
- Implement `RemoteTable.prewarm_index(name)` calling `POST /v1/table/{id}/index/{name}/prewarm/` (previously returned `NotSupported`)
- Add `BaseTable::prewarm_data(columns)` trait method and `Table` public API in Rust core
- Add PyO3 bindings and Python API (`AsyncTable`, `LanceTable`, `RemoteTable`) for `prewarm_data`
- Add type stubs for `prewarm_index` and `prewarm_data` in `_lancedb.pyi`
- Upgrade Lance to 3.0.0-rc.3 with breaking change fixes

### DIFF
--- a/python/python/lancedb/remote/table.py
+++ b/python/python/lancedb/remote/table.py
@@ -643,8 +643,12 @@ class RemoteTable(Table):
     def prewarm_index(self, name: str) -> None:
         """Prewarm an index in the table.
 
-        This loads the index into memory on the server, reducing cold-start
-        latency for subsequent queries.
+        This is a hint to the database that the index will be accessed in the
+        future and should be loaded into memory if possible.  This can reduce
+        cold-start latency for subsequent queries.
+
+        This call initiates prewarming and returns once the request is accepted.
+        It is idempotent and safe to call from multiple clients concurrently.
 
         Parameters
         ----------
@@ -654,10 +658,19 @@ class RemoteTable(Table):
         return LOOP.run(self._table.prewarm_index(name))
 
     def prewarm_data(self, columns: Optional[List[str]] = None) -> None:
-        """Prewarm data (page cache) for the table.
+        """Prewarm data for the table.
 
-        This loads column data pages into the server's disk cache, reducing
-        cold-start latency for subsequent queries.
+        This is a hint to the database that the given columns will be accessed
+        in the future and the database should prefetch the data if possible.
+        Currently only supported on remote tables.
+
+        This call initiates prewarming and returns once the request is accepted.
+        It is idempotent and safe to call from multiple clients concurrently.
+
+        This operation has a large upfront cost but can speed up future queries
+        that need to fetch the given columns.  Large columns such as embeddings
+        or binary data may not be practical to prewarm.  This feature is intended
+        for workloads that issue many queries against the same columns.
 
         Parameters
         ----------
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2219,12 +2219,18 @@ class LanceTable(Table):
 
     def prewarm_index(self, name: str) -> None:
         """
-        Prewarms an index in the table
+        Prewarm an index in the table.
 
-        This loads the entire index into memory
+        This is a hint to the database that the index will be accessed in the
+        future and should be loaded into memory if possible.  This can reduce
+        cold-start latency for subsequent queries.
 
-        If the index does not fit into the available cache this call
-        may be wasteful
+        This call initiates prewarming and returns once the request is accepted.
+        It is idempotent and safe to call from multiple clients concurrently.
+
+        It is generally wasteful to call this if the index does not fit into the
+        available cache.  Not all index types support prewarming; unsupported
+        indices will silently ignore the request.
 
         Parameters
         ----------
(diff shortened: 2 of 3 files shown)


hpvd:
related to https://github.com/lancedb/lancedb/issues/2789

LuQQiu:
@jackye1995 Does it work for the namespace API path?
@esteban could you try to understand the lancedb, lance and lance-namespace API hierachy and see whether your changes work for lance-namespace API?

python/python/lancedb/remote/table.py
```diff
@@ -640,6 +640,32 @@ def drop_columns(self, columns: Iterable[str]) -> DropColumnsResult:
...
+        ----------
+        name: str
+            The name of the index to prewarm
+        """
+        return LOOP.run(self._table.prewarm_index(name))
+
+    def prewarm_data(self, columns: Optional[List[str]] = None) -> None:
+        """Prewarm data (page cache) for the table.
+
+        This loads column data pages into the server's disk cache, reducing
```
LuQQiu: server's disk cache -> server's data cache
could be in memory cache only or mix of in memory cache and disk cache

LuQQiu:
@esteban several improvements that can be included in future PRs
1. node API 
2. double check namespace API
3. optionally to make the prewarm index and prewarm data async and rerunning the same prewarm index/data can report the status (completed or not)

python/python/lancedb/remote/table.py
```diff
@@ -640,6 +640,32 @@ def drop_columns(self, columns: Iterable[str]) -> DropColumnsResult:
...
+
+        Parameters
+        ----------
+        name: str
+            The name of the index to prewarm
+        """
+        return LOOP.run(self._table.prewarm_index(name))
+
+    def prewarm_data(self, columns: Optional[List[str]] = None) -> None:
+        """Prewarm data (page cache) for the table.
```
westonpace: I don't know if "page cache" is a term we've ever used before.  Is that what we want the feature to be called?  I'm worried we might have other caching strategies in the future.  This is also only a feature implemented by remote tables at the moment so many users may not have any information about it.  Perhaps we can say something like...

```
Prewarm data for the table.

This is a hint to the database that the columns will be accessed in the future
and the database should prefetch the data if possible.  This is currently only
supported on remote tables. 
```

python/python/lancedb/remote/table.py
```diff
@@ -640,6 +640,32 @@ def drop_columns(self, columns: Iterable[str]) -> DropColumnsResult:
...
+
+        This loads column data pages into the server's disk cache, reducing
+        cold-start latency for subsequent queries.
```
westonpace: What are the limits for this?  Should I always prewarm columns before I query them?  Are there any columns that I shouldn't prewarm?  Can we give some simple advice here for users and LLMs like...

This operation has a large upfront cost but can speed up future queries that need to fetch the given columns.  Large columns such as embeddings or binary data may not be practical to prewarm.  This feature is intended for cases where you plan on issuing many queries close together against the same columns.

python/python/lancedb/remote/table.py
```diff
@@ -640,6 +640,32 @@ def drop_columns(self, columns: Iterable[str]) -> DropColumnsResult:
...
+        Parameters
+        ----------
+        name: str
+            The name of the index to prewarm
+        """
+        return LOOP.run(self._table.prewarm_index(name))
+
+    def prewarm_data(self, columns: Optional[List[str]] = None) -> None:
+        """Prewarm data (page cache) for the table.
+
```
westonpace, line 662: Does this call block until the prewarm is complete?  Or does it simply initiate prewarming?  If it is the latter then how do I check the progress?

python/python/lancedb/remote/table.py
westonpace: Is this call idempotent?  What if I have multiple query clients, can I safely call it from each of them?  Or will that repeat the prewarming process?  Is there anyway to check if a column is already prewarmed?  Or should I just call this every time my client starts?

esteban:
Updated docstrings across all layers (Rust core, Python `AsyncTable`/`LanceTable`/`RemoteTable`) to address review feedback:

- Dropped "page cache" terminology — now describes these as prefetch hints
- Clarified blocking semantics: initiates prewarming and returns once request is accepted
- Documented idempotency: safe to call from multiple clients concurrently
- Added usage guidance: upfront cost tradeoff, large columns (embeddings/binary) may be impractical, intended for many-query workloads
- Changed "disk cache" → removed cache-type specifics per @LuQQiu's note (could be memory, disk, or mixed)
- Updated `NativeTable` error message to match new terminology

@westonpace @LuQQiu Could you take another look?

esteban:
@westonpace resolving as improving docs in the API.

esteban:
Thanks @LuQQiu, in-line:

> 1. node API
Next PR
> 2. double check namespace API
It should be fine, since the namespace is defined by delimiter.
> 3. optionally to make the prewarm index and prewarm data async and rerunning the same prewarm index/data can report the status (completed or not)
Will follow up with that one later, probably same pattern returning an operation ID that can be polled.

PR #3111, opened 2026-03-07
fix: propagate managed versioning for namespace connection

jackye1995:
Without this fix, if user directly use the native table to do operations like `add_columns`, even if it is configured to use namespace db connection, it is not really propagated through.

The fix is to bring lancedb's python binding up to date and do a similar implementation as https://github.com/lance-format/lance/pull/5968, and make sure the namespace is fully propagated through all the related calls.


rust/lancedb/src/database/namespace.rs
```diff
@@ -206,83 +208,55 @@ impl Database for LanceNamespaceDatabase {
         let mut table_id = request.namespace.clone();
         table_id.push(request.name.clone());
 
-        // Try declare_table first, falling back to create_empty_table for backwards
```
jackye1995, line 209: also remove deprecated create_empty_table usage along the way

rpgreen (approved):

PR #3118, opened 2026-03-08
fix(python): preserve null structs on round-trip through to_pylist()

### LINKED ISSUE #3105: bug(python): null objects are stored/retrieved as non-null with default values
### LanceDB version

v0.29.2

### What happened?

objects inside a LanceModel can be nullable, but when stored/retrieved as a null, the value becomes a non-null object with primitives set to all default values. The expectation is that the value should respect the null and return as just a null value.

Refer to this simple test: 

```
import lancedb
from pydantic import BaseModel
import json

class InnerStruct(BaseModel):
    value: str


class NestedStruct(BaseModel):
    name: str
    inner: InnerStruct | None


class TestObject(LanceModel):
    test_id: str
    object: NestedStruct | None
    list: list[NestedStruct | None] | None
    primitive_field: int | None

LOCAL_URI = "./lancedb_local"
conn = await lancedb.connect_async(LOCAL_URI)

print(f"Creating table: test_object")
table = await conn.create_table("test_object", schema=TestObject, mode="overwrite")

print(f"\nAdding records")
records = [
    TestObject(
        test_id="All null",
        object=None,
        list=None,
        primitive_field=None
    ),
    TestObject(
        test_id="Nested null",
        object=NestedStruct(name="null inner", inner=None),
        list=[None],
        primitive_field=1
    )
]
await table.add(records)

print(f"\nPrinting records")
records = await table.query().to_list()
print(json.dumps(records, indent=2))
```

Output of records:
```
[
  {
    "test_id": "All null",
    "object": {
      "name": "",
      "inner": {
        "value": ""
      }
    },
    "list": null,
    "primitive_field": null
  },
  {
    "test_id": "Nested null",
    "object": {
      "name": "null inner",
      "inner": {
        "value": ""
      }
    },
    "list": [
      {
        "name": "",
        "inner": {
          "value": ""
        }
      }
    ],
    "primitive_field": 1
  }
]
```

Expected records
```
[
  {
    "test_id": "All null",
    "object": null,
    "list": null,
    "primitive_field": null
  },
  {
    "test_id": "Nested null",
    "object": {
      "name": "null inner",
      "inner": null
    },
    "list": [null],
    "primitive_field": 1
  }
]
```

MaxwellCalkin:
## Summary

Fixes #3105. Nullable struct fields in `LanceModel` lose their `None` values on read-back, becoming dicts populated with default values (empty strings, zeros, etc.).

### Root cause

PyArrow's `to_pylist()` has a known behavior where null struct values are converted to dicts with default field values instead of `None`. All four `to_pylist()` call sites in `query.py` (sync/async `to_list` and `to_pydantic`) are affected.

### Fix

- Add a `table_to_pylist()` helper in `arrow.py` that calls `to_pylist()` then walks the Arrow validity bitmaps to restore `None` for every null struct -- and for null struct items inside lists -- at any nesting depth.
- Includes a fast path that skips processing entirely when no struct columns are present.
- Replace all four bare `to_pylist()` call sites in `query.py` with this helper.

### Before (broken)

```python
records = await table.query().to_list()
# object=None becomes object={'name': '', 'inner': {'value': ''}}
# inner=None becomes inner={'value': ''}
# list=[None] becomes list=[{'name': '', 'inner': {'value': ''}}]
```

### After (fixed)

```python
records = await table.query().to_list()
# object=None stays None
# inner=None stays None
# list=[None] stays [None]
```

### Files changed

- `python/python/lancedb/arrow.py` -- new `table_to_pylist()` and helpers
- `python/python/lancedb/query.py` -- replace `.to_pylist()` with `table_to_pylist()`
- `python/python/tests/test_struct_null_roundtrip.py` -- unit tests covering top-level null structs, nested null structs, lists of nullable structs, and the exact scenario from #3105

---

I am an AI (Claude Opus 4.6, built by Anthropic). This contribution was made transparently and in good faith. Happy to address any review feedback.

### DIFF
--- a/python/python/lancedb/arrow.py
+++ b/python/python/lancedb/arrow.py
@@ -10,6 +10,115 @@ import pyarrow as pa
 from ._lancedb import RecordBatchStream
 
 
+
+def _has_struct_type(data_type):
+    """Check if a data type contains a struct type (directly or nested)."""
+    if pa.types.is_struct(data_type):
+        return True
+    if pa.types.is_list(data_type) or pa.types.is_large_list(data_type):
+        return _has_struct_type(data_type.value_type)
+    if pa.types.is_fixed_size_list(data_type):
+        return _has_struct_type(data_type.value_type)
+    return False
+
+
+def _fix_struct_nulls(value, data_type, array, idx):
+    """Recursively fix a single value from a struct array.
+
+    If the array element at *idx* is null, return None regardless of what
+    ``to_pylist()`` produced.  Otherwise walk the struct children and
+    recursively fix any nested struct / list-of-struct fields.
+    """
+    if not array.is_valid(idx):
+        return None
+    if value is None:
+        return None
+    for j, sub_field in enumerate(data_type):
+        if _has_struct_type(sub_field.type):
+            child_array = array.field(j)
+            value[sub_field.name] = _fix_value(
+                value[sub_field.name], sub_field.type, child_array, idx
+            )
+    return value
+
+
+def _fix_list_nulls(value, data_type, array, idx):
+    """Fix null struct values inside a list element."""
+    if not array.is_valid(idx):
+        return None
+    if value is None:
+        return None
+
+    value_type = data_type.value_type
+    offsets = array.offsets.to_pylist()
+    start = offsets[idx]
+    values_array = array.values
+
+    for k in range(len(value)):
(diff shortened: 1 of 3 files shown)


westonpace:
This change will need to pass unit tests before it can be merged.  You claim this is a known issue in `to_pylist`.  Do you have any issue that you can link to which explains the problem?

wjones127:
Upstream issue closed.

Issue #2963, opened 2026-01-30
bug(python): RuntimeError: lance error: LanceError(IO): External error: Not found (maybe NFS related)


jmhsieh:
### LanceDB version

0.25.1

### What happened?
```
RuntimeError: lance error: LanceError(IO): External error: Not found: /path/to/table.lance/data/110011001110101001000000d7bb294ad7af3a6b5a7e37b052.lance
```
```
I'll try to find a way to reproduce it consistently. I'm doing concurrent writes on NFS, which I suspect can be the culprit.
```

<img width="827" height="226" alt="Image" src="https://github.com/user-attachments/assets/2a2637f6-a7a4-4f57-8d8a-115ad41c31e5" />

<img width="670" height="314" alt="Image" src="https://github.com/user-attachments/assets/1c362ffc-6109-4ce2-a68f-97a755b79608" />

https://discord.com/channels/1030247538198061086/1030247538667827251/threads/1466389135253176404

sprytnyk:
I have the same issue and don't use NFS.
It happens when adding new vectors into tables and try to index them.
This bug longs from old releases to the latest one and difficult to reproduce because it happens intermittently.

I have such indices for my table:
```python
>>> from lancedb import connect_async
... db = await connect_async('/lancedb')
>>> table = await db.open_table('table')
>>> await table.list_indices()
[Index(BTree, columns=["pk"], name="pk_idx"),
 Index(Bitmap, columns=["visibility"], name="visibility_idx"),
 Index(IvfPq, columns=["embedding"], name="embedding_idx")]
```

During the day run such script to index newly added data:
```python
    for table_name in table_names:
        table = await db.open_table(table_name)

        # Clean up files in dirs _versions, _transactions, merge data in data
        # dir, add new vectors into scalar and vector indices.
        logger.warning(f"Optimizing {table_name} table...")
        stats = await table.optimize(cleanup_older_than=timedelta(seconds=0), delete_unverified=True)
        logger.info(stats)
        logger.info(f"Finished optimizing {table_name} table.")
```

It works for days until it's not.

Any ideas on this @wjones127, @changhiskhan?

sprytnyk:
Per docs [`optimize`](https://lancedb.github.io/lancedb/python/python/#lancedb.table.Table.optimize) method docs, optimization covers three operations:
- Compaction: Merges small files into larger ones
- Prune: Removes old versions of the dataset
- Index: Optimizes the indices, adding new data to existing indices

[...]

PR #3122, opened 2026-03-10
fix(rust): change `optimize` steps ordering

sprytnyk:
Indexing should happen before purning in order to avoid lance error: LanceError(IO): External error: Not found data.

It causes dodgy race conditions intermittently with parameters arguments like:
- cleanup_older_than=timedelta(seconds=0)
- delete_unverified=True

I ran tests with `cargo test --features remote --tests` and all them passing.

Refs: #2963

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -1495,7 +1495,6 @@ class Table(ABC):
          * Index: Optimizes the indices, adding new data to existing indices
          * Prune: Removes old versions of the dataset
 
-
         Parameters
         ----------
         cleanup_older_than: timedelta, optional default 7 days
@@ -3007,7 +3006,6 @@ class LanceTable(Table):
          * Compaction: Merges small files into larger ones
          * Index: Optimizes the indices, adding new data to existing indices
          * Prune: Removes old versions of the dataset
-
 
         Parameters
         ----------
@@ -4563,7 +4561,6 @@ class AsyncTable:
          * Compaction: Merges small files into larger ones
          * Index: Optimizes the indices, adding new data to existing indices
          * Prune: Removes old versions of the dataset
-
 
         Parameters
         ----------


westonpace:
I would want some more details into what exact failure we are fixing.  Why did the old order not work correctly?  I'm not against changing the order (as long as compaction comes before indexing) but I don't expect it to fix any bugs.

> It causes dodgy race conditions intermittently with parameters arguments like:
> 
>     cleanup_older_than=timedelta(seconds=0)
>     delete_unverified=True

This is a very dangerous combination, with or without this fix.  If this optimization is run concurrent with any other operation it can lead to the error described.  That is because, with this configuration, the cleanup job will delete files from operations that are in-progress.

sprytnyk:
> I would want some more details into what exact failure we are fixing.

Basically, I have such scenario:
1. Adding continuously new vectors into tables
2. Run `optimize` script every hour

I have such indices for my tables:
```python
>>> from lancedb import connect_async
... db = await connect_async('/lancedb')
>>> table = await db.open_table('table')
>>> await table.list_indices()
[Index(BTree, columns=["pk"], name="pk_idx"),
 Index(Bitmap, columns=["visibility"], name="visibility_idx"),
 Index(IvfPq, columns=["embedding"], name="embedding_idx")]
```

During the day run such script to index newly added data:
```python
    for table_name in table_names:
        table = await db.open_table(table_name)

        # Clean up files in dirs _versions, _transactions, merge data in data
        # dir, add new vectors into scalar and vector indices.
        logger.warning(f"Optimizing {table_name} table...")
        stats = await table.optimize(cleanup_older_than=timedelta(seconds=0), delete_unverified=True)
        logger.info(stats)
        logger.info(f"Finished optimizing {table_name} table.")
```

I don't have any kind of concurrency when run the above script, execution of the script is not overlapping, the same table isn't used twice as well.
This works well for weeks but at some point the next time I try to run the script it throws `lance error: LanceError(IO): External error: Not found`, so somehow `await table.optimize(cleanup_older_than=timedelta(seconds=0), delete_unverified=True)` corrupts data, index directories.

> This is a very dangerous combination, with or without this fix.

If so, then why it's not explicitly stated in docs?

It's up to you whether accept this PR or not. I don't know all ins and outs of this project, you have this context, but to me is more logical to create index first and then delete any kind of data that potentially may be used for indexing process.

westonpace:
> I don't have any kind of concurrency when run the above script, execution of the script is not overlapping, the same table isn't used twice as well.

Are you continuously adding new vectors while you are executing the optimize script?  Even if the add is happening in a different script?

> If so, then why it's not explicitly stated in docs?

That's a good catch.  I thought it was stated.  I've opened https://github.com/lancedb/lancedb/issues/3125 to fix that.

> It's up to you whether accept this PR or not. I don't know all ins and outs of this project, you have this context, but to me is more logical to create index first and then delete any kind of data that potentially may be used for indexing process.

Makes a lot of sense to me.  I just remember we debated this order for a while when it was first introduced and want to refresh my memory on why we chose this path.

sprytnyk:
> Are you continuously adding new vectors while you are executing the optimize script? Even if the add is happening in a different script?

Yes, I don't stop adding new data when run `optimize` from a different script.
Want to amend it with such parameters arguments, run every 2 hours, and see how it will go:
```python
        await table.optimize(
            cleanup_older_than=timedelta(hours=3),
            delete_unverified=False,
        )
```

I think it's more of problem with `delete_unverified=True` because adding new data may mark some datasets as _in-progress_, but with such flag, it schedules for deletion by `optimize`.

Here's a full script:
https://sprytnyk.dev/posts/running-lancedb-in-production/#clean-up-script

PR #3123, opened 2026-03-11
feat(nodejs): add order_by method to Query

ddupg:


ddupg:
The current CI failure appears to be unrelated to my changes in this PR.
```
cargo clippy --profile ci --all --all-features -- -D warnings
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    RUST_BACKTRACE: 1
    CC: gcc-12
    CXX: g++-12
    CARGO_INCREMENTAL: 0
    CARGO_PROFILE_DEV_DEBUG: 0
    CARGO_TERM_COLOR: always
    RUSTFLAGS: -D warnings
    CARGO_UNSTABLE_SPARSE_REGISTRY: true
    CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
    CACHE_ON_FAILURE: false
error: failed to parse lock file at: /home/runner/work/lancedb/lancedb/Cargo.lock

Caused by:
  package `snafu-derive` is specified twice in the lockfile
Error: Process completed with exit code 101.
```

Fixed it in #3124

jackye1995:
@ddupg I merged 3124, could you help rebase?

rust/lancedb/src/query.rs
```diff
@@ -530,6 +563,11 @@ impl<T: HasQuery> QueryBase for T {
         self.mut_query().norm = Some(norm);
         self
     }
+
+    fn order_by(mut self, column: impl Into<String>, direction: SortDirection) -> Self {
```
jackye1995: why not match the lance api and just expose `order_by(&mut self, ordering: Option<Vec<ColumnOrdering>>)`?

nodejs/lancedb/query.ts
```diff
@@ -416,6 +417,36 @@ export class StandardQueryBase<
...
+   *
+   * // Method chaining: filter then sort
+   * await table.query()
+   *   .where("category = 'electronics'")
+   *   .orderBy("rating", "desc")
+   *   .limit(10)
+   *   .toArray();
+   * ```
+   */
+  orderBy(column: string, direction: "asc" | "desc" = "asc"): this {
```
jackye1995: in general we need to keep parity of python and typescript implementation, this is small enough change can we just add them in the same PR?

ddupg: Thanks for the review.

Python currently supports ordering by providing the ordering_field_name field during search, like:
```
    def search(
        self,
        query: Optional[
            Union[VEC, str, "PIL.Image.Image", Tuple, FullTextQuery]
        ] = None,
        vector_column_name: Optional[str] = None,
        query_type: QueryType = "auto",
        ordering_field_name: Optional[str] = None,
        fts_columns: Optional[Union[str, List[str]]] = None,
    ) -> LanceQueryBuilder:
```

When you mentioned "keep parity of Python and TypeScript implementation", are you referring to the **ordoring direction**?

jackye1995: I think we can improve `ordering_field_name` to also support the full ordering config, and we can deprecate `ordering_field_name`

jackye1995 (approved):
looks good to me! Sorry for the delayed review. Could you rebase the change? @ddupg

jackye1995:
rebased and looks like there are still some failures, @ddupg if you could fix

rust/lancedb/src/remote/table.rs
```diff
@@ -518,6 +518,21 @@ impl<S: HttpSend> RemoteTable<S> {
             }
         }
 
+        if let Some(order_by) = &params.order_by {
+            body["order_by"] = serde_json::Value::Array(
```
ddupg, line 522: For the remote implementation, the current tests for remote table are mock-based and do not exercise a real integration environment. I’m not fully confident that the request protocol is correct for the remote API. Could you @jackye1995 please help double-check whether this is compatible with the actual remote service?

ddupg (commented):
Thanks @jackye1995  for reviewing this.

rust/lancedb/src/remote/table.rs
jackye1995: I think it is fine. Remote is at this moment basically the Rest catalog protocol, just hand-rolled instead of auto-generated. it would be a great followup task to add this to the rest catalog spec as well.

PR #3139, opened 2026-03-13
chore: fix ci

wjones127:
* Move away from buildjet, which is shutting down runners for GHA [^1]
* Add `Cargo.lock` to build jobs, so when we upgrade locked dependencies we check the builds actually pass. CI started failing because dependencies were changed in #3116 without running all build jobs.
* Add fixes for aws-lc-rs build in NodeJS.

[^1]: https://buildjet.com/for-github-actions/blog/we-are-shutting-down


westonpace:
Python failures seemed like flaky tests.  Re-running

westonpace:
The python problem is not a flaky test, but it seems unrelated to the current change.

PR #3143, opened 2026-03-14
docs: fix rendering issues with missing index types in API docs

prrao87:
## Problem

The generated Python API docs for `lancedb.table.IndexStatistics.index_type` were misleading because mkdocstrings renders that field’s type annotation directly, and the existing `Literal[...]` listed only a subset of the actual canonical SDK index type strings.

Current (missing index types):
image

## Fix

- Update the `IndexStatistics.index_type` annotation in `python/python/lancedb/table.py` to include the full supported set of canonical values, so the generated docs show all valid index_type strings inline.
- Add a small regression test in `python/python/tests/test_index.py` to ensure the docs-facing annotation does not drift silently again in case we add a new index/quantization type in the future.
- Bumps mkdocs and material theme versions to mkdocs 1.6 to allow access to more features like hooks

After fix (all index types are included and tested for in the annotations):
image

### DIFF
--- a/python/python/tests/test_index.py
+++ b/python/python/tests/test_index.py
@@ -301,4 +301,7 @@ def test_index_statistics_index_type_lists_all_supported_values():
         "LABEL_LIST",
     }
 
-    assert set(get_args(get_type_hints(IndexStatistics)["index_type"])) == expected_index_types
+    assert (
+        set(get_args(get_type_hints(IndexStatistics)["index_type"]))
+        == expected_index_types
+    )


docs/requirements.txt
```diff
@@ -1,8 +1,8 @@
-mkdocs==1.5.3
+mkdocs==1.6.1
 mkdocs-jupyter==0.24.1
-mkdocs-material==9.5.3
+mkdocs-material==9.6.23
 mkdocs-autorefs>=0.5,<=1.0
-mkdocstrings[python]==0.25.2
+mkdocstrings[python]>=0.24,<1.0
```
westonpace, line 5: Why pin this below 1.0?

prrao87: I think there's a lot of changes coming in mkdocs + mkdocstrings as they move forward with their 2.0 release of mkdocs, so I preferred to keep it constrained.

PR #3145, opened 2026-03-16
feat(nodejs): exponential backoff retry for embedding fns

### LINKED ISSUE #2343: Feature: Support rate limiting in Node.js embeddings API
### SDK

Node

### Description

Same as #614 but for Node.

Farzan-Hashmi:
[Already implemented](https://github.com/lancedb/lancedb/pull/614) in the Python SDK. Resolves https://github.com/lancedb/lancedb/issues/2343

### DIFF
--- a/nodejs/lancedb/embedding/embedding_function.ts
+++ b/nodejs/lancedb/embedding/embedding_function.ts
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
+import { setTimeout } from "node:timers/promises";
 import "reflect-metadata";
 import {
   DataType,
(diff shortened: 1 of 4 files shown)


Farzan-Hashmi:
@wjones127, just an FYI, I think you reviewed a similar PR that addressed the same issue a while back. Can you PTAL at this when you get the chance? Thanks!

Farzan-Hashmi:
@wjones127 just bumping this if you get some time to review

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -17,6 +18,86 @@ import {
...
+export interface RetryOptions {
+  /** Initial delay in seconds before the first retry. Defaults to 1. */
+  initialDelay?: number;
+  /** Base for the exponential backoff multiplier. Defaults to 2. */
+  exponentialBase?: number;
+  /** Whether to add random jitter to the delay. Defaults to true. */
+  jitter?: boolean;
+  /** Maximum number of retries. Set to 0 to disable retries. Defaults to 7. */
+  maxRetries?: number;
+}
```
wjones127, line 33: issue(blocking): there is no upper bound as written. Can you add one? Otherwise the exponential backoff can easily get into the hours, which is not desireable.

nodejs/lancedb/embedding/embedding_function.ts
```diff
@@ -17,6 +18,86 @@ import {
...
+
+function isNonRetryableError(error: unknown): boolean {
+  if (typeof error !== "object" || error === null) {
+    return false;
+  }
+  const err = error as Record<string, unknown>;
+
+  if (err.constructor?.name === "AuthenticationError") {
```
wjones127, line 41: **question (non-blocking):** This relies on `constructor.name === "AuthenticationError"` surviving minification in downstream bundles — have you considered `instanceof` checks against known SDK error classes or a duck-typed marker instead, to avoid silently retrying auth failures if the name gets mangled?

wjones127 (changes requested):
Sorry for the delay. This generally looks good but I think we want a max retry delay.

Issue #3044, opened 2026-02-17
Python - add function support to expr


wjones127:

pratik0316:
@wjones127 this issue can also be closed

wjones127:
Closed by #3150

Issue #3043, opened 2026-02-17
Python - basic support for Expr in filter


wjones127:

pratik0316:
@wjones127 can we close this issue?

wjones127:
Closed by #3150

PR #3150, opened 2026-03-17
feat(python): add type-safe expression builder API

pratik0316:
Introduces col(), lit(), func(), and Expr class as alternatives to raw SQL strings in .where() and .select(). Expressions are backed by DataFusion's Expr AST and serialized to SQL for remote table compat.

Resolves: 
- https://github.com/lancedb/lancedb/issues/3044 (python api's)
- https://github.com/lancedb/lancedb/issues/3043 (support for filter)
- https://github.com/lancedb/lancedb/issues/3045 (support for projection)

### DIFF
--- a/python/tests/test_expr.py
+++ b/python/tests/test_expr.py
@@ -373,9 +373,7 @@ class TestColNamingIntegration:
 
     def test_space_in_col_filter(self, special_col_table):
         result = (
-            special_col_table.search()
-            .where(col("first name") == lit("B"))
-            .to_arrow()
+            special_col_table.search().where(col("first name") == lit("B")).to_arrow()
         )
         assert result.num_rows == 1
 


pratik0316:
@wjones127 , @jackye1995  can you pls help review this.
Thanks

python/python/lancedb/_lancedb.pyi
```diff
@@ -27,6 +27,32 @@ from .remote import ClientConfig
...
+class PyExpr:
+    """A type-safe DataFusion expression node (Rust-side handle)."""
+
+    def eq(self, other: "PyExpr") -> "PyExpr": ...
+    def ne(self, other: "PyExpr") -> "PyExpr": ...
+    def lt(self, other: "PyExpr") -> "PyExpr": ...
+    def lte(self, other: "PyExpr") -> "PyExpr": ...
+    def gt(self, other: "PyExpr") -> "PyExpr": ...
+    def gte(self, other: "PyExpr") -> "PyExpr": ...
+    def and_(self, other: "PyExpr") -> "PyExpr": ...
```
wjones127, line 39: I honestly wonder if there would be an easier way to write these intermediate bindings. Maybe something like:

```python
class PyExpr:
    def func(self, name: str, args: List["PyExpr"]): ...
```

I would think this would at least work for `upper` and `lower`, for example.

python/python/lancedb/expr.py
```diff
@@ -0,0 +1,261 @@
...
+        """Cast values to *data_type*.
+
+        Parameters
+        ----------
+        data_type:
+            One of ``"bool"``, ``"int8"``, ``"int16"``, ``"int32"``,
+            ``"int64"``, ``"uint8"``–``"uint64"``, ``"float32"``,
+            ``"float64"``, ``"string"``, ``"date32"``, ``"date64"``.
+        """
+        return Expr(self._inner.cast(data_type))
```
wjones127: It would be nice to accept a PyArrow type. I think you could do this:


```suggestion
    def cast(self, data_type: str | pa.DataType) -> "Expr":
        """Cast values to *data_type*.

        Parameters
        ----------
        data_type:
            One of ``"bool"``, ``"int8"``, ``"int16"``, ``"int32"``,
            ``"int64"``, ``"uint8"``–``"uint64"``, ``"float32"``,
            ``"float64"``, ``"string"``, ``"date32"``, ``"date64"``.
        """
        data_type = pa.infer_type(data_type)
        return Expr(self._inner.cast(data_type))
```

And then in Rust, you want to do something like:

```rust
    fn cast(&self, PyArrowType(data_type): DataType) -> PyResult<Self> {
        Ok(Self(expr_cast(self.0.clone(), dt)))
    }
```

This is the module with `PyArrowType`: https://docs.rs/arrow-pyarrow/latest/arrow_pyarrow/

python/src/expr.rs
```diff
@@ -0,0 +1,200 @@
...
+        "large_string" | "large_utf8" => Ok(DataType::LargeUtf8),
+        "date32" | "date" => Ok(DataType::Date32),
+        "date64" => Ok(DataType::Date64),
+        other => Err(PyValueError::new_err(format!(
+            "unsupported data type: '{}'. Supported: bool, int8/16/32/64, \
+             uint8/16/32/64, float32, float64, string, date32, date64",
+            other
+        ))),
+    }
+}
```
wjones127: If you implement my other suggestion, you shouldn't need this.
```suggestion
```

python/tests/test_expr.py
```diff
@@ -0,0 +1,279 @@
...
+import pytest
+import pyarrow as pa
+import lancedb
+from lancedb.expr import Expr, col, lit, func
+
+
+# ── unit tests for Expr construction ─────────────────────────────────────────
+
+
+class TestExprConstruction:
```
wjones127, line 15: These tests just assert the data type returned. At a minimum, I think we need to check that the SQL output is what we expect for the expression.

python/tests/test_expr.py
```diff
@@ -0,0 +1,279 @@
...
+    def test_col_returns_expr(self):
+        e = col("age")
```
wjones127, line 17: Let's make sure we handle camelCase and special characters in column names correctly. That can be tricky. Might want an integration test for that as well.

python/tests/test_expr.py
```diff
@@ -0,0 +1,279 @@
...
+
+class TestExprFilter:
+    def test_simple_gt_filter(self, simple_table):
+        result = simple_table.search().where(col("age") > lit(20)).to_arrow()
+        assert result.num_rows == 3  # ages 25, 30, 22
+
+    def test_compound_and_filter(self, simple_table):
+        result = (
+            simple_table.search()
+            .where((col("age") > lit(18)) & (col("score") > lit(2.0)))
```
wjones127, line 265: Are we required to do `lit()` here? Or can we omit that?
```suggestion
            .where((col("age") > 18) & (col("score") > 2.0))
```

python/src/expr.rs
```diff
@@ -0,0 +1,200 @@
...
+#[pyfunction]
+pub fn expr_col(name: &str) -> PyExpr {
+    PyExpr(ldb_col(name))
+}
+
+/// Create a literal value expression.
+///
+/// Supported Python types: `bool`, `int`, `float`, `str`.
+#[pyfunction]
+pub fn expr_lit(value: Bound<'_, PyAny>) -> PyResult<PyExpr> {
```
wjones127, line 146: Can be done in a follow up PR, but hopefully we can add `date`, `datetime`, `bytes`, and so on.

pratik0316:
Thanks @wjones127 for the review. let me try to adress those today

python/python/lancedb/expr.py
pratik0316: This is much cleaner thanks for the suggestion’s

python/src/expr.rs
pratik0316: Thanks removed this

python/python/lancedb/expr.py
pratik0316: @wjones127 actually **pa.infer_type** infers a type from a sequence of values, not a type name string, so we need to handle if a string type is passed explicitly

python/tests/test_expr.py
pratik0316: sure added the sql checks too

python/tests/test_expr.py
pratik0316: No its not mandatory, kept a test where we have lit and dont have in the other

python/python/lancedb/expr.py
wjones127: Ah okay. It worked in one small test I did, but maybe not general enough. I wonder if there is another function.

python/tests/test_expr.py
pratik0316: Good point added camel casing col names and unicode chars too.
Sure probably we can take the integration tests as a follow up PR!?

pratik0316:
@wjones127 can you pls have a look again. Thanks!

python/tests/test_expr.py
wjones127: Created follow up issue: https://github.com/lancedb/lancedb/issues/3211

python/src/expr.rs
wjones127: Created follow up issue: https://github.com/lancedb/lancedb/issues/3212

PR #3152, opened 2026-03-18
fix: preserve valid OpenAI embeddings in mixed batches

### LINKED ISSUE #1677: Feature: openai embedding function returns None if the result is not valid

saudademjj:
## Summary
- retry OpenAI embedding batches row-by-row when a batch-level bad request occurs
- keep successful rows, log the failing row index, and leave invalid rows as null embeddings
- add regression tests for row-level fallback and nullable ingestion behavior

## Testing
- remote x86_64 validation on `141.164.59.25`
- `PYTHONPATH=/home/lixdj/lancedb-codex/python/python python -m ruff check python/python/lancedb/embeddings/openai.py python/python/tests/test_embeddings.py`
- `PYTHONPATH=/home/lixdj/lancedb-codex/python/python python -m pytest -q python/python/tests/test_embeddings.py::test_openai_bad_request_can_store_null_embeddings python/python/tests/test_embeddings.py::test_openai_bad_request_falls_back_to_row_level_requests python/python/tests/test_embeddings.py::test_openai_no_retry_on_401`

Closes #1677

### DIFF
--- a/python/python/lancedb/embeddings/openai.py
+++ b/python/python/lancedb/embeddings/openai.py
@@ -97,29 +97,85 @@ class OpenAIEmbeddings(TextEmbeddingFunction):
                 valid_texts.append(text)
                 valid_indices.append(idx)
 
+        if not valid_texts:
+            return [None] * len(texts)
+
         # TODO retry, rate limit, token limit
         try:
-            kwargs = {
-                "input": valid_texts,
-                "model": self.name,
-            }
-            if self.name != "text-embedding-ada-002":
-                kwargs["dimensions"] = self.dim
-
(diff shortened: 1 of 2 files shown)


saudademjj:
Validated on a remote x86_64 environment and happy to adjust anything if needed. Would appreciate a review when someone has a chance. Thanks!

saudademjj:
Friendly follow-up here. The PR is in a good state on my side, and I am happy to make any changes if needed.

PR #3156, opened 2026-03-19
Fix FixedSizeList sanitizing with empty vectors

### LINKED ISSUE #1670: bug(python): wrong behavior of sanitizing FixeSizeList
### LanceDB version

_No response_

### What happened?

When creating a list of vector columns, while some of them are empty list. It still throws arrow error when `on_bad_vectors="drop"` is specified. 

Additionally, the sanitizing code change the size (by averaging zero sized list and correct sized list)

```python
def test_add_table_with_empty_embeddings(tmp_path):

    class MySchema(LanceModel):
        text: str
        embedding: Vector(16)

    table = db.create_table("test", schema=MySchema)
    table.add(
        [{"text": "hello", "embedding": []}, {"text": "bar", "embedding": [0.1] * 16}],
        on_bad_vectors="drop",
    )
```

```
E   pyarrow.lib.ArrowTypeError: Size of FixedSizeList is not the same. input list: fixed_size_list<item: float>[8] output list: fixed_size_list<item: float>[16]
```

qianchongyang:
## Problem

When adding rows with mixed empty and valid vectors, `on_bad_vectors="drop"` still throws an ArrowTypeError. The sanitizing logic incorrectly modifies the FixedSizeList size by attempting to average empty vectors with valid ones.

Example:
```python
table.add(
    [{"text": "hello", "embedding": []}, {"text": "bar", "embedding": [0.1] * 16}],
    on_bad_vectors="drop",
)
# ArrowTypeError: Size mismatch between empty list and fixed size list
```

## Solution

Fixed the sanitizing logic for FixedSizeList vectors:
- Empty vectors are now correctly identified and skipped when `on_bad_vectors="drop"`
- FixedSizeList size is preserved, not averaged across empty and valid vectors
- Invalid vectors are dropped without causing type errors

## Validation

```python
def test_add_table_with_empty_embeddings(tmp_path):
    table = db.create_table("test", schema=MySchema)
    table.add(
        [{"text": "hello", "embedding": []}, {"text": "bar", "embedding": [0.1] * 16}],
        on_bad_vectors="drop",
    )
    # Passes - empty vector is dropped, valid vector is added
    assert len(table) == 1
```

Fixes #1670

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3346,7 +3346,13 @@ def _infer_target_schema(
 
 def _modal_list_size(arr: Union[pa.ListArray, pa.ChunkedArray]) -> int:
     # Use the most common length of the list as the dimensions
-    return pc.mode(pc.list_value_length(arr))[0].as_py()["mode"]
+    # Filter out empty lists first, as they should not affect dimension calculation
+    non_empty_mask = pc.greater(pc.list_value_length(arr), 0)
+    non_empty_arr = arr.filter(non_empty_mask)
+    if len(non_empty_arr) == 0:
+        # All lists are empty, fall back to mode of original (will likely error later)
+        return pc.mode(pc.list_value_length(arr))[0].as_py()["mode"]
+    return pc.mode(pc.list_value_length(non_empty_arr))[0].as_py()["mode"]
 
 
 def _validate_schema(schema: pa.Schema):


python/python/lancedb/table.py
```diff
@@ -3346,7 +3346,13 @@ def _infer_target_schema(
 
 def _modal_list_size(arr: Union[pa.ListArray, pa.ChunkedArray]) -> int:
     # Use the most common length of the list as the dimensions
-    return pc.mode(pc.list_value_length(arr))[0].as_py()["mode"]
+    # Filter out empty lists first, as they should not affect dimension calculation
+    non_empty_mask = pc.greater(pc.list_value_length(arr), 0)
+    non_empty_arr = arr.filter(non_empty_mask)
+    if len(non_empty_arr) == 0:
+        # All lists are empty, fall back to mode of original (will likely error later)
+        return pc.mode(pc.list_value_length(arr))[0].as_py()["mode"]
```
westonpace, line 3354: It seems at this point you could probably just return 0 right?  No need to spend CPU cycles calculating a fallback.
```suggestion
        return 0
```

westonpace (changes requested):
One minor nit but seems fine otherwise.  This will need at least one test case before we can merge however.

PR #3158, opened 2026-03-19
fix(python): sanitize bad vectors before Arrow cast

### LINKED ISSUE #1670: bug(python): wrong behavior of sanitizing FixeSizeList

myandpr:
## Problem

`on_bad_vectors="drop"` is supposed to remove invalid vector rows before write, but for some schema-defined vector columns it can still fail later during Arrow cast instead of dropping the bad row.

Repro:
```python
class MySchema(LanceModel):
    text: str
    embedding: Vector(16)

table = db.create_table("test", schema=MySchema)
table.add(
    [
        {"text": "hello", "embedding": []},
        {"text": "bar", "embedding": [0.1] * 16},
    ],
    on_bad_vectors="drop",
)
```
Before:
```
RuntimeError
Arrow error: C Data interface error: Invalid: ListType can only be casted to FixedSizeListType if the lists are all the expected size.
```
After:
```
rows 1
texts ['bar']
```
## Solution

Make bad-vector sanitization use schema dimensions before cast, while keeping the handling scoped to vector columns identified by schema metadata or existing vector-name heuristics.

This also preserves existing integer vector inputs and avoids applying on_bad_vectors to unrelated fixed-size float columns.

Fixes #1670

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -271,15 +271,17 @@ def _sanitize_data(
         reader,
         on_bad_vectors=on_bad_vectors,
         fill_value=fill_value,
+        target_schema=target_schema,
+        metadata=metadata,
     )
 
     if target_schema is None:
         target_schema, reader = _infer_target_schema(reader)
 
     if metadata:
-        new_metadata = target_schema.metadata or {}
-        new_metadata.update(metadata)
-        target_schema = target_schema.with_metadata(new_metadata)
+        target_schema = target_schema.with_metadata(
+            _merge_metadata(target_schema.metadata, metadata)
+        )
 
     _validate_schema(target_schema)
     reader = _cast_to_target_schema(reader, target_schema, allow_subschema)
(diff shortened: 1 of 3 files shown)


myandpr:
Hi @wjones127 @jackye1995 ,could you please review this. Thanks

myandpr:
format fix

myandpr:
Hi @wjones127 I’ve rebased on master and fixed the CI failures triggered by you. Could you take another look when you have a moment?

myandpr:
> This looks good. Thanks for the PR!

Thanks for the review and merge!

PR #3172, opened 2026-03-20
fix: respect max_batch_length for Rust vector and hybrid queries

### LINKED ISSUE #1540: bug(python): batch_size is ignored in to_batches
### LanceDB version

v0.12.0

### What happened?

The value of the batch_size (or max_batch_size for the AsyncVectorQuery) parameter of the to_batches function is ignored. A batch size of 8192 is always used.

This snippet reproduces the bug.

```python
import lancedb
from lancedb.query import LanceVectorQueryBuilder
import pandas as pd
import numpy as np

db = lancedb.connect('data/test')
try:
  df = pd.DataFrame({'vector': [np.random.rand(128) for _ in range(10000)]})
  tbl = db.create_table('rand', df)
  r = LanceVectorQueryBuilder(tbl, np.random.rand(128), 'vector').limit(10000).to_batches(100)
  for batch in r:
    print('Got batch size:', len(batch))
    assert len(batch) <= 100
finally:
  db.drop_table('rand')
  ```

This produces an assertion error.
```
Got batch size: 8192
  File "/test/sync_batch.py", line 13, in <module>
    assert len(batch) <= 100
AssertionError
```

### PR DESCRIPTION
**Pull request by myandpr**
Fixes #1540

I could not reproduce this on current `main` from Python, but I could still reproduce it from the Rust SDK.

Python no longer reproduces because the current Python vector/hybrid query paths re-chunk results into a `pyarrow.Table` before returning batches. Rust still reproduced because `max_batch_length` was passed into planning/scanning, but vector search could still emit larger `RecordBatch`es later in execution (for example after KNN / TopK), so it was not enforced on the final Rust output stream.

This PR enforces `max_batch_length` on the final Rust query output stream and adds Rust regression coverage.

Before the fix, the Rust repro produced:
`num_batches=2, max_batch=8192, min_batch=1808, all_le_100=false`

After the fix, the same repro produces batches `<= 100`.

## Runnable Rust repro

Before this fix, current `main` could still return batches like `[8192, 1808]` here even with `max_batch_length = 100`:

```rust
use std::sync::Arc;

use arrow_array::{
    types::Float32Type, FixedSizeListArray, RecordBatch, RecordBatchReader, StringArray,
};
use arrow_schema::{DataType, Field, Schema};
use futures::TryStreamExt;
use lancedb::query::{ExecutableQuery, QueryBase, QueryExecutionOptions};

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let tmp = tempfile::tempdir()?;
    let uri = tmp.path().to_str().unwrap();

    let rows = 10_000;
    let schema = Arc::new(Schema::new(vec![
        Field::new("id", DataType::Utf8, false),
        Field::new(
            "vector",
            DataType::FixedSizeList(Arc::new(Field::new("item", DataType::Float32, true)), 4),
            false,
        ),
    ]));

    let ids = StringArray::from_iter_values((0..rows).map(|i| format!("row-{i}")));
    let vectors = FixedSizeListArray::from_iter_primitive::<Float32Type, _, _>(
        (0..rows).map(|i| Some(vec![Some(i as f32), Some(1.0), Some(2.0), Some(3.0)])),
        4,
    );
    let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(ids), Arc::new(vectors)])?;
    let reader: Box<dyn RecordBatchReader + Send> = Box::new(
        arrow_array::RecordBatchIterator::new(vec![Ok(batch)].into_iter(), schema),
    );

    let db = lancedb::connect(uri).execute().await?;
    let table = db.create_table("test", reader).execute().await?;

    let mut opts = QueryExecutionOptions::default();
    opts.max_batch_length = 100;

    let mut stream = table
        .query()
        .nearest_to(vec![0.0, 1.0, 2.0, 3.0])?
        .limit(rows)
        .execute_with_options(opts)
        .await?;

    let mut sizes = Vec::new();
    while let Some(batch) = stream.try_next().await? {
        sizes.push(batch.num_rows());
    }

    println!("{sizes:?}");
    Ok(())
}
```

### DIFF
--- a/rust/lancedb/src/query.rs
+++ b/rust/lancedb/src/query.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use std::{future::Future, time::Duration};
 
 use arrow::compute::concat_batches;
-use arrow_array::{Array, Float16Array, Float32Array, Float64Array, make_array};
+use arrow_array::{Array, Float16Array, Float32Array, Float64Array, RecordBatch, make_array};
 use arrow_schema::{DataType, SchemaRef};
 use datafusion_expr::Expr;
 use datafusion_physical_plan::ExecutionPlan;
@@ -17,15 +17,17 @@ use lance_datafusion::exec::execute_plan;
 use lance_index::scalar::FullTextSearchQuery;
 use lance_index::scalar::inverted::SCORE_COL;
 use lance_index::vector::DIST_COL;
-use lance_io::stream::RecordBatchStreamAdapter;
 
 use crate::DistanceType;
 use crate::error::{Error, Result};
 use crate::rerankers::rrf::RRFReranker;
 use crate::rerankers::{NormalizeMethod, Reranker, check_reranker_result};
 use crate::table::BaseTable;
-use crate::utils::TimeoutStream;
-use crate::{arrow::SendableRecordBatchStream, table::AnyQuery};
+use crate::utils::{MaxBatchLengthStream, TimeoutStream};
+use crate::{
+    arrow::{SendableRecordBatchStream, SimpleRecordBatchStream},
+    table::AnyQuery,
+};
 
 mod hybrid;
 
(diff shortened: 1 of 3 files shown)


myandpr:
Hi @westonpace @wjones127 @jackye1995 , could you take a look when you have a chance? This fixes #1540 on the Rust side. Many thanks!

myandpr:
@wjones127 Thanks for the review — appreciate it!🙌

PR #3175, opened 2026-03-22
feat(nodejs): expr.ts added to enable Expression API. Basic support for Expr in fil…

ChinmayGowda71:
…ter and add function support. expr.test.ts for tests. Addresses Issues #3041 and #3040

### DIFF
--- a/nodejs/lancedb/expr.ts
+++ b/nodejs/lancedb/expr.ts
@@ -1,4 +1,5 @@
-import { no } from "zod/v4/locales";
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
 import { type IntoSql, toSQL as valueToSQL } from "./util"; 
 
 export type binaryOp =  "=" | "!=" | "<" | "<=" | ">=" | ">" | "AND" | "OR" | "+" | "*" | "-" | "/" | "%" | "LIKE"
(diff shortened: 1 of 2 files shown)


ChinmayGowda71:
@wjones127 hello for the expression builder I added all the common SQL keywords do let me know if i should remove things or add some or anything else as well. Thank you!

nodejs/lancedb/expr.ts
wjones127, line 1: issue(blocking): There's nothing in this implementation or tests that guarantees the SQL generated is compatible with the underlying engine.

I had in mind we would bind the NodeJs expressions to Rust's DataFusion expressions, like how we did in #3150. That would be one way of guaranteeing they work together.

I'm not entirely opposed to doing the SQL translation within Typescript. It makes it possible to use this later in a Typescript-only client, if we want to.

However, if we go the route of doing the SQL translation within TypeScript, I think we need to write integration tests against DataFusion to make sure the SQL generated is actually valid.

ChinmayGowda71: Got it, and I'm happy to go either way with it. I can write up integrations tests or go about how we did it in #3150 is there one that you'd specifically prefer?

ChinmayGowda71: With the current I do feel like SQL translation within TS is easier and I could definitely draft some tests up too and you can let me know if thats satisfactory @wjones127

wjones127: Yeah, doing it in TS is fine. Just want a good suite of integration tests then.

PR #3184, opened 2026-03-24
fix: table.add(mode='overwrite') infers vector column types

### LINKED ISSUE #3183: bug(python): table.add(..., mode="overwrite") changes vector type from fixed_size_list<float> to list<double>
### LanceDB version

v0.3.1

### What happened?

📌 Description

When using lancedb==0.30.0, calling table.add(data, mode="overwrite") unexpectedly changes the schema of the vector column:

From: fixed_size_list<float>
To: list<double>

This leads to vector search failure, since the schema no longer matches the expected vector type.

🔁 Reproduction
```python
import lancedb

def main(): 
    data = [
        {"id": 4, "text": "123", "label": "Knight", "vector": [0.1, 0.2, 0.3, 0.4]},
        {"id": 5, "text": "123", "label": "Wizard", "vector": [0.9, 0.1, 0.8, 0.7]},
        {"id": 6, "text": "123", "label": "Knight", "vector": [0.3, 0.5, 0.6, 0.2]},
    ]
    table = lancedb.connect("./tl.db").create_table("knights", data)
    print(table.schema)

    table.add(data, mode="overwrite")
    print(table.schema)

if __name__ == "__main__":
    main()
```

Actual Behavior
```
# Before overwrite
vector: fixed_size_list<item: float>[4]

# After overwrite
vector: list<item: double>
```

Expected Behavior
Schema should remain consistent after overwrite:
```
vector: fixed_size_list<item: float>[4]
```

majiayu000:
Fixes #3183

## Summary

When `table.add(mode='overwrite')` is called, PyArrow infers input data types (e.g. `list<double>`) which differ from the original table schema (e.g. `fixed_size_list<float32>`). Previously, overwrite mode bypassed `cast_to_table_schema()` entirely, so the inferred types replaced the original schema, breaking vector search.

This fix builds a merged target schema for overwrite: columns present in the existing table schema keep their original types, while columns unique to the input pass through as-is. This way `cast_to_table_schema()` is applied unconditionally, preserving vector column types without blocking schema evolution.

## Changes

- `rust/lancedb/src/table/add_data.rs`: For overwrite mode, construct a target schema by matching input columns against the existing table schema, then cast. Non-overwrite (append) path is unchanged.
- Added `test_add_overwrite_preserves_vector_type` test that creates a table with `fixed_size_list<float32>`, overwrites with `list<double>` input, and asserts the original type is preserved.

## Test Plan

- `cargo test --features remote -p lancedb -- test_add_overwrite` — all 4 overwrite tests pass
- Full suite: 454 passed, 2 failed (pre-existing `remote::retry` flakes unrelated to this change)

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -3857,26 +3857,23 @@ class AsyncTable:
 
         # _santitize_data is an old code path, but we will use it until the
         # new code path is ready.
-        if on_bad_vectors != "error" or (
+        if mode == "overwrite":
+            # For overwrite, apply the same preprocessing as create_table
+            # so vector columns are inferred as FixedSizeList.
+            data, _ = sanitize_create_table(
+                data, None, on_bad_vectors=on_bad_vectors, fill_value=fill_value
+            )
+        elif on_bad_vectors != "error" or (
             schema.metadata is not None and b"embedding_functions" in schema.metadata
         ):
-            # For overwrite, use None as target schema so _infer_target_schema
-            # is triggered, matching the create_table schema inference path.
-            target = None if mode == "overwrite" else schema
             data = _sanitize_data(
                 data,
-                target,
+                schema,
                 metadata=schema.metadata,
                 on_bad_vectors=on_bad_vectors,
                 fill_value=fill_value,
                 allow_subschema=True,
             )
-        elif mode == "overwrite":
-            # Apply schema inference for overwrite to detect vector columns
-            # and convert them to FixedSizeList, matching create_table behavior.
-            reader = _into_pyarrow_reader(data, None)
-            target_schema, reader = _infer_target_schema(reader)
-            data = _cast_to_target_schema(reader, target_schema)
         _register_optional_converters()
         data = to_scannable(data)
         progress, owns = _normalize_progress(progress)


majiayu000:
Thanks for the feedback! You're right — I've reworked the approach:

**Reverted**: Removed the Rust-side schema preservation logic in `add_data.rs`. Overwrite correctly sets a new schema.

**New fix**: Applied schema inference on the Python overwrite path to match `create_table` behavior. When `mode='overwrite'`, `_infer_target_schema()` is now called so columns named "vector" or "embedding" are properly converted from `list<double>` to `fixed_size_list<float32>`, just like `create_table` does.

Changes:
- `python/python/lancedb/table.py`: In `AsyncTable.add()`, when mode is `"overwrite"`, pass `target_schema=None` to `_sanitize_data` (or directly call `_infer_target_schema` + `_cast_to_target_schema`) so vector column inference has parity with `create_table`.
- `python/python/tests/test_table.py`: Added regression test `test_add_overwrite_infers_vector_schema`.
- `rust/lancedb/src/table/add_data.rs`: Reverted to original behavior (skip casting on overwrite).

python/python/lancedb/table.py
```diff
@@ -3860,14 +3860,23 @@ async def add(
...
                 on_bad_vectors=on_bad_vectors,
                 fill_value=fill_value,
                 allow_subschema=True,
             )
+        elif mode == "overwrite":
+            # Apply schema inference for overwrite to detect vector columns
+            # and convert them to FixedSizeList, matching create_table behavior.
+            reader = _into_pyarrow_reader(data, None)
+            target_schema, reader = _infer_target_schema(reader)
+            data = _cast_to_target_schema(reader, target_schema)
```
wjones127: issue(blocking): this is different than the preprocessing we call for `create_table`:

https://github.com/lancedb/lancedb/blob/cca6a7c989d4e951485c64766175abe30eb2ba9a/python/python/lancedb/db.py#L1499-L1502

I worry this could mean differences in behavior. Could we re-use `sanitize_create_table` here?

python/python/lancedb/table.py
```diff
@@ -3860,14 +3860,23 @@ async def add(
...
             data = _sanitize_data(
                 data,
-                schema,
+                target,
                 metadata=schema.metadata,
                 on_bad_vectors=on_bad_vectors,
                 fill_value=fill_value,
                 allow_subschema=True,
             )
+        elif mode == "overwrite":
```
wjones127: issue(non-blocking): currently, we only use `_santize_data` when the Rust-based pre-processing doesn't support it. This is when `on_bad_vectors = "error"` or there are embedding functions. Eventually, I'd like the Rust-based pre-processing to support all cases, and we will remove `_sanitize_data` and `sanitize_create_table`. The way you have written this, mode="overwrite" will **always** drop down to the less performant Python-based preprocessing. This is okay for now, but I want to note we will need to change this in a later PR.

python/python/lancedb/table.py
majiayu000: switched to `sanitize_create_table` in the latest commit — should match the create_table path now.

python/python/lancedb/table.py
majiayu000: yeah, understood — noted for the future cleanup.

majiayu000:
@wjones127 Thanks for your review, fixed. Could you take a look at this? If anything wrong plz figure out.

PR #3188, opened 2026-03-25
fix: change _client reference to _conn

dantasse:
This code previously referenced `self._client`, which does not exist. This change makes it correctly call `self._conn.close()`


wjones127:
Thanks, @dantasse!

PR #3192, opened 2026-03-26
fix: handle empty/wrong-length vectors returned by embedding functions

### LINKED ISSUE #1672: bug(python): can not handle bad embedding function output
### LanceDB version

_No response_

### What happened?

Using `ollama` embedding, when the input is not valid, the embedding functions returns a zero-length list `[]`, which crashes the pipeline even we set `table.add(.., on_bad_vector="drop")`

Install ollama and `ollama pull nomic-embed-text`

```python
import lancedb

from lancedb.pydantic import Vector, LanceModel

func = get_registry().get("ollama").create(name="nomic-embed-text")

class PdfVectorDb(LanceModel):
    doc_id: int
    chunk_id: int
    text: str = embedding_function.SourceField()
    vector: Vector(embedding_function.ndims()) = embedding_function.VectorField()

ldb = lancedb.connect(db)
tbl = ldb.create_table("pdfs_embeddings", schema=_PdfVectorDb, mode="overwrite")

tbl.add(
    [
        {
            "doc_id": 1,
            "chunk_id": 1,
            "text": "",
        }
    ],
    on_bad_vectors="drop",
)
```

error message

```
  File "/Users/lei/work/framellm/framellm/ingestion.py", line 59, in ingest_pdfs
    tbl.add(
  File "/Users/lei/work/framellm/.venv/lib/python3.12/site-packages/lancedb/table.py", line 1400, in add
    data, _ = _sanitize_data(
              ^^^^^^^^^^^^^^^
  File "/Users/lei/work/framellm/.venv/lib/python3.12/site-packages/lancedb/table.py", line 114, in _sanitize_data
    data = _append_vector_col(data, metadata, schema)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lei/work/framellm/.venv/lib/python3.12/site-packages/lancedb/table.py", line 214, in _append_vector_col
    pa.field(vector_column, type=dtype), pa.array(col_data, type=dtype)
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/array.pxi", line 368, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 42, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Length of item not correct: expected 768 but got array of size 0
```

Spectual:
## Summary

- When an embedding function returns an empty list (e.g. `[]`) for an input row — as can happen when a model produces no output for a blank string — `_append_vector_columns` crashed with `ArrowInvalid: Length of item not correct: expected N but got array of size 0` because PyArrow cannot fit a zero-length value into a fixed-size list element.
- The fix adds a validation step in `gen()`, inside `_append_vector_columns`, that replaces any vector whose length does not match the expected `ndims` (including empty lists and `None`) with `None` before `pa.array()` is called.
- `None` is a valid null in a PyArrow fixed-size list array, so the bad entry flows into `_handle_bad_vectors` and is handled according to the caller-supplied `on_bad_vectors` policy (`error` / `drop` / `fill` / `null`) instead of causing an unconditional crash.

## Test plan

- [ ] Added `test_embedding_with_empty_output_vectors` in `python/python/tests/test_embeddings.py` that uses an embedding function returning `[]` for empty-string inputs, calls `table.add(..., on_bad_vectors="drop")`, and asserts no crash and that bad rows are correctly dropped.
- [ ] Existing `test_embedding_with_bad_results` continues to pass (NaN vectors still handled correctly).
- [ ] Verified manually that `pa.array([[1.,2.,3.,4.], []], type=pa.list_(pa.float32(), 4))` raises `ArrowInvalid` without the fix, and succeeds with `None` in place of `[]`.

Fixes #1672


wjones127 (approved):

wjones127 (commented):
Hi @Spectual. It looks like your test is failing. Here's what my AI said about the failure:


> The root cause is clear: when _append_vector_columns replaces empty/wrong-length vectors with None, _handle_bad_vector_column doesn't detect them as bad because:                                                                                                            
>   - pc.list_value_length(null) → null          
>   - pc.not_equal(null, 128) → null (not True)                                                                                                                                                                                                                                  
>   - pc.any([null]) → False                                                                                                                                                                                                                                                                                                                  
>  So nulls slip through undetected and on_bad_vectors="drop" never fires.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
>  The fix is to add has_null = pc.is_null(vec_arr) detection in _handle_bad_vector_column:

wjones127 (approved):
This looks good, can merge once the formatting is fixed.

PR #3198, opened 2026-03-27
fix(python): don't assign dict.update() return value in _sanitize_data

lennylxx:
dict.update() mutates in place and returns None. Assigning its result caused with_metadata(None) to strip all schema metadata when embedding metadata was merged during create_table with embedding_functions.


wjones127:
CI failures are unrelated.

PR #2395, opened 2025-05-20
feat(python): handle strenum in pydantic

happysalada:
This PR aims at handling strenum in pydantic conversion in python.

(happy to squash and/or edit commit message), opening this as a proof of concept to get some feedback.

closes https://github.com/lancedb/lancedb/issues/2392

## Summary by CodeRabbit

- **New Features**
  - Enhanced support for StrEnum fields in Pydantic models with optimized dictionary-encoded string handling.
- **Tests**
  - Added tests validating schema conversion and data querying for models using StrEnum fields (Python 3.11+).

PR #3232, opened 2026-04-02
feat(python): support Enum types in Pydantic to Arrow schema conversion

### LINKED ISSUE #1846: Feature: Pydantic to Arrow Schema convertion support for Py Enum types
### SDK

Python

### Description

https://github.com/lancedb/lancedb/discussions/1836

Zelys-DFKH:
## Summary

Fixes #1846.

Python `Enum` fields raised `TypeError: Converting Pydantic type to Arrow Type: unsupported type <enum 'SomethingTypes'>` when converting a Pydantic model to an Arrow schema.

The fix adds Enum detection in `_pydantic_type_to_arrow_type`. When an Enum subclass is encountered, the value type of its members is inspected and mapped to the appropriate Arrow type:

- `str`-valued enums (e.g. `class Status(str, Enum)`) → `pa.utf8()`
- `int`-valued enums (e.g. `class Priority(int, Enum)`) → `pa.int64()`
- Other homogeneous value types → the Arrow type for that Python type
- Mixed-value or empty enums → `pa.utf8()` (safe fallback)

This covers the common `(str, Enum)` and `(int, Enum)` mixin patterns used in practice.

## Changes

- `python/python/lancedb/pydantic.py`: add Enum branch in `_pydantic_type_to_arrow_type`
- `python/python/tests/test_pydantic.py`: add `test_enum_types` covering `str`, `int`, and `Optional` Enum fields

## Note on #2395

PR #2395 handles `StrEnum` (Python 3.11+) specifically, using a dictionary-encoded type. This PR handles the broader `(str, Enum)` / `(int, Enum)` mixin pattern that works across all Python versions and stores values as their natural Arrow type.

AI assistance was used in developing this fix.

### DIFF
--- a/python/python/tests/test_pydantic.py
+++ b/python/python/tests/test_pydantic.py
@@ -696,7 +696,7 @@ def test_enum_types():
 
     schema = pydantic_to_schema(TestModel)
 
-    assert schema.field("status").type == pa.utf8()
+    assert schema.field("status").type == pa.dictionary(pa.int32(), pa.utf8())
     assert schema.field("priority").type == pa.int64()
-    assert schema.field("opt_status").type == pa.utf8()
+    assert schema.field("opt_status").type == pa.dictionary(pa.int32(), pa.utf8())
     assert schema.field("opt_status").nullable
--- a/python/python/lancedb/pydantic.py
+++ b/python/python/lancedb/pydantic.py
@@ -316,11 +316,17 @@ def _pydantic_type_to_arrow_type(tp: Any, field: FieldInfo) -> pa.DataType:
             # For regular Vector
             return pa.list_(tp.value_arrow_type(), tp.dim())
         if _safe_issubclass(tp, Enum):
-            # Map Enum to the Arrow type of its value.  Fall back to utf8 if the
-            # value type cannot be determined (e.g. mixed-type or empty enums).
+            # Map Enum to the Arrow type of its value.
+            # For string-valued enums, use dictionary encoding for efficiency.
+            # For integer enums, use the native type.
+            # Fall back to utf8 for mixed-type or empty enums.
             value_types = {type(m.value) for m in tp}
             if len(value_types) == 1:
-                return _py_type_to_arrow_type(value_types.pop(), field)
+                value_type = value_types.pop()
+                if value_type is str:
+                    # Use dictionary encoding for string enums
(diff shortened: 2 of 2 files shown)


python/python/tests/test_pydantic.py
```diff
@@ -673,3 +674,29 @@ class TestModel(LanceModel):
...
+    class TestModel(pydantic.BaseModel):
+        status: StrStatus
+        priority: IntPriority
+        opt_status: Optional[StrStatus] = None
+
+    schema = pydantic_to_schema(TestModel)
+
+    assert schema.field("status").type == pa.utf8()
+    assert schema.field("priority").type == pa.int64()
+    assert schema.field("opt_status").type == pa.utf8()
```
wjones127: issue(blocking): enum strings can always be better encoded as dictionary, so I think we should map to that:


```suggestion
    assert schema.field("status").type == pa.dictionary(pa.int8(), pa.utf8())
    assert schema.field("priority").type == pa.int64()
    assert schema.field("opt_status").type == pa.dictionary(pa.int8(), pa.utf8())
```

(though if there are > 128 unique values, we'd need to use pa.int16()).

wjones127 (commented):
Thanks for working on this. It looks cleanly implemented and nicely tested. However, I do think we should use dictionary encoding for string enums.

For integers, dictionary encoding doesn't make sense so we can leave those as is.

Zelys-DFKH:
## Implementation Update

Addressed maintainer feedback: string enums now use Arrow dictionary encoding for efficiency.

**Changes made:**
- String-valued enums: `pa.dictionary(pa.int32(), pa.utf8())` 
- Integer-valued enums: `pa.int64()` (unchanged)
- Mixed/empty enums: `pa.utf8()` (fallback, unchanged)

**Why dictionary encoding?**
Dictionary encoding stores each unique enum value once + uses integer indices, reducing storage overhead for repeated enum values.

**Test expectations updated:**
- `test_enum_types()` now expects dictionary-encoded types for `StrStatus` fields
- Integer and optional field tests remain consistent with the new encoding

This aligns with the pattern used in PR #2395 for `StrEnum` support and follows the principle of using the most efficient Arrow representation for each data type.

wjones127 (approved):
This looks good. Thank you!

Zelys-DFKH:
Thanks for the review and the feedback on the dictionary encoding approach. Appreciate the time!

PR #3236, opened 2026-04-04
feat(nodejs): support branches in Node API

majin1102:
## Summary

  This PR adds branch-aware table opening and branch management support for the Rust core and the Node SDK only.

  Scope and non-goals:
  - This PR does **not** add Python support. Python will follow in a separate PR.
  - This PR does **not** add `checkout_branch`. LanceDB table handles and Lance dataset references do not have identical lifecycle and refresh semantics, and allowing branch switching on an existing handle would
  introduce cross-branch state transition risk for table mutation and refresh behavior.
  - For now, branch selection is only supported at `openTable` / `open_table` time.
  - Remote support is still uncertain, so this PR treats it as unsupported and fails explicitly instead of making assumptions.

  ## What’s Added

  ### Opening a table by reference
  - Open the latest state of a branch
  - Open a specific version
  - Open a specific tag
  - Open a specific version within a branch

  ### Branch management
  - Create a branch
  - Delete a branch
  - List branches
  - Inspect the current handle reference via `current_ref`

  ### Node bridge cleanup
  - The Node SDK keeps the same public `openTable({ ref })` API
  - Internally, the bridge now passes a single reference model instead of flattening it into separate `version / tag / branch_name / branch_version_number` arguments

  ## API Details

  ### Rust
  - `open_table(...).version(...)`
  - `open_table(...).tag(...)`
  - `open_table(...).branch(...).version_number(...)`
  - `Table::create_branch(...)`
  - `Table::delete_branch(...)`
  - `Table::list_branches(...)`
  - `Table::current_ref(...)`

  ### Node
  - `openTable(name, { ref: { branchName } })`
  - `openTable(name, { ref: { branchName, versionNumber } })`
  - Existing `ref.versionNumber` / `ref.tagName` remain supported
  - Branch management and `currentRef()` are exposed for local/native tables

  ## Behavior and Boundaries

  - Branch selection happens only when opening a table handle.
  - An opened handle stays on its own branch timeline.
  - `checkout(version)` and `checkout(tag)` still operate on the current handle’s branch timeline.
  - `checkout_tag` remains branch-local.
  - A tag must belong to the same branch as the current table handle.
  - To read a tag on another branch, open a new handle on that branch first.
  - `create_branch(..., from=None)` defaults to creating from the latest head of the current branch, not from a stale cached view.
  - Deleting the currently checked out branch is rejected.

  ### Unsupported cases
  - Remote tables do not support branch management.
  - Remote tables do not support `current_ref`.
  - Remote `open_table` with references is rejected explicitly.
  - Namespace-backed tables with server-side query pushdown enabled reject opening at a specific reference.

  ## Compatibility

  - Rust: this is primarily additive; existing `open_table(...).execute()` and existing checkout flows continue to work.
  - Node public API remains compatible; the `openTable({ ref })` shape does not change.
  - The only internal compatibility change is in the native Node bridge signature, where the implementation now uses a unified reference payload instead of flattened selector fields.
  - No Python API changes are included in this PR.


rust/lancedb/src/connection.rs
```diff
@@ -268,6 +278,105 @@ impl OpenTableBuilder {
...
+    /// To read a different branch later, open a new table handle with this
+    /// selector again. Existing table handles do not support branch checkout.
+    pub fn branch(self, branch: impl Into<String>) -> OpenTableBranchBuilder {
+        OpenTableBranchBuilder {
+            builder: self.with_reference(Reference::Version(Some(branch.into()), None)),
+        }
+    }
+}
+
+impl OpenTableBranchBuilder {
```
wjones127, line 308: issue(non-blocking): It seems very verbose to have to repeat these methods. Is there a reason they have to be new methods?

rust/lancedb/src/table.rs
```diff
@@ -1205,6 +1220,38 @@ impl Table {
         self.inner.tags().await
     }
 
+    /// Create a new branch for this table.
+    pub async fn create_branch(&self, branch: &str, from: Option<Reference>) -> Result<()> {
+        let native = self.inner.as_native().ok_or_else(|| Error::NotSupported {
+            message: "Branch management is not supported for remote tables".to_string(),
+        })?;
+        native.create_branch(branch, from).await
+    }
```
wjones127, line 1229: issue(blocking): This adds a lot of top-level methods on `Table`. For tags, we have a method that gives you further options:

https://github.com/lancedb/lancedb/blob/e646e9013656f24e10a932105d2a807b3b7d5d5c/rust/lancedb/src/table.rs#L310

And that exposes:

https://github.com/lancedb/lancedb/blob/e646e9013656f24e10a932105d2a807b3b7d5d5c/rust/lancedb/src/table.rs#L217-L232

Could we do the same for branches?

rust/lancedb/src/connection.rs
```diff
@@ -268,6 +278,105 @@ impl OpenTableBuilder {
...
+    pub fn version(self, version: u64) -> Self {
+        self.with_reference(Reference::VersionNumber(version))
+    }
+
+    /// Open the table at a specific tag.
+    pub fn tag(self, tag: impl Into<String>) -> Self {
+        self.with_reference(Reference::Tag(tag.into()))
+    }
+
+    /// Open the table at the latest version of a specific branch.
```
wjones127, line 297: issue(blocking): When you open a table without a branch, it will track the latest version. Can we make it so that when you check out a branch it will track the latest version on that branch? Or is there a reason that doesn't make sense?

wjones127 (changes requested):

rust/lancedb/src/table.rs
majin1102: Sure.

Original lance dataset has independent create_branch because there's a shallow clone operation within it and that might need a storage_options. We can unify this trait at lancedb level and reuse storage_options in dataset. I will follow this point

rust/lancedb/src/connection.rs
majin1102: This matches the non-branch behavior. A branch-opened handle is still opened in latest-tracking mode; whether it observes newer commits automatically depends on the existing `read_consistency_interval` setting, just like a main-branch handle.

In this PR, `Reference::Version(branch, None)` is opened via `DatasetBuilder::with_branch(branch, None)` and then wrapped with `DatasetConsistencyWrapper::new_latest(...)`, so it is not pinned to a specific version. The existing refresh path calls `checkout_latest()` on that branch-opened Lance dataset, which resolves latest relative to that branch.

I can add a regression test to make this explicit.

rust/lancedb/src/connection.rs
majin1102: The separate `OpenTableBranchBuilder` was added to support `.branch(...).version_number(...)` while keeping branch-version selection scoped under a branch selector. That does make the implementation verbose because the normal `OpenTableBuilder` options need to be forwarded.

I agree the extra builder is probably not worth that complexity. Since the stable Node API remains `{ ref: { branchName, versionNumber } }`, we can simplify the Rust builder API and map Node internally to the simpler request shape.

majin1102:
Hi @wjones127 , thanks for pushing this so far.

I simplified the branch API here. The previous version tried to expose too much of Lance’s ref flexibility directly through table open, but I think that makes the LanceDB table-handle semantics less clear.

The current direction is:

1. `openTable` / `open_table` only accepts a branch selector, not a version selector.
2. After a table is opened on a branch, users can still use `checkout(version)` to inspect a fixed version within that branch.
3. `checkout_tag` is also scoped to the opened branch, so it does not become another implicit way to switch branch timelines outside of `open`.

Overall, this is intentionally less flexible than Lance’s dataset-level ref API. In LanceDB, a table handle can track and refresh to the latest version, so keeping branch selection as an open-time property makes the behavior safer and easier to reason about.

wjones127:
I think we can close this in favor of https://github.com/lancedb/lancedb/pull/3490

PR #3240, opened 2026-04-06
feat: add user_id field to ClientConfig for user identification

### LINKED ISSUE #3230: Feature: auto-configure user agent based on environment variable
### SDK

Python

### Description

Today configuring user agent requires:

```
custom_client_config = ClientConfig(  
  user_agent="MyAwesomeLanceDBApp/1.0 (Contact: support@example.com)" 
)

try:                                                                                                             
  db = lancedb.connect(    
     "db://my_database",                                                                                                          
      api_key="YOUR_API_KEY",                                        
      region="us-east-1", 
      client_config=custom_client_config 
  )   
```

User would like to have it more transparent.

For example, suppose a company always set env variable `MYCOMPANY_USER_ID`, and want that to be set automatically for anyone running lancedb, we can automatically pick it up.

To achieve this, I suggest 2 env variables:
1. `LANCEDB_USER_AGENT`, when set, use this value as the user agent value
2. `LANCEDB_USER_AGENT_ENV_KEY`, when set, use this value to find the environment variable to find the user agent value.

jackye1995:
## Summary

- Add a `user_id` field to `ClientConfig` that allows users to identify themselves to LanceDB Cloud/Enterprise
- The user_id is sent as the `x-lancedb-user-id` HTTP header in all requests
- Supports three configuration methods:
  - Direct assignment via `ClientConfig.user_id`
  - Environment variable `LANCEDB_USER_ID`
  - Indirect env var lookup via `LANCEDB_USER_ID_ENV_KEY`

Closes #3230

### DIFF
--- a/docs/src/js/interfaces/ClientConfig.md
+++ b/docs/src/js/interfaces/ClientConfig.md
@@ -64,7 +64,7 @@ optional userId: string;
 
 User identifier for tracking purposes.
 
-This is sent as the `x-lancedb-user-id` header in requests to LanceDB Cloud.
+This is sent as the `x-lancedb-user-id` header in requests to LanceDB Cloud/Enterprise.
 It can be set directly, or via the `LANCEDB_USER_ID` environment variable.
 Alternatively, set `LANCEDB_USER_ID_ENV_KEY` to specify another environment
 variable that contains the user ID value.
--- a/nodejs/src/remote.rs
+++ b/nodejs/src/remote.rs
@@ -94,7 +94,7 @@ pub struct ClientConfig {
     pub tls_config: Option<TlsConfig>,
     /// User identifier for tracking purposes.
     ///
-    /// This is sent as the `x-lancedb-user-id` header in requests to LanceDB Cloud.
+    /// This is sent as the `x-lancedb-user-id` header in requests to LanceDB Cloud/Enterprise.
     /// It can be set directly, or via the `LANCEDB_USER_ID` environment variable.
     /// Alternatively, set `LANCEDB_USER_ID_ENV_KEY` to specify another environment
     /// variable that contains the user ID value.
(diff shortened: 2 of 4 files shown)


jackye1995:
@wkalt you proposed originally

```python
db.connect("db://my_db", ..., storage_options={"labels": {"user": env.get("ORG_USER")}})
```

I think we should not use `storage_options`, because that is supposed to set storage configurations and this is not, but putting it in client config feels pretty natural.

I ended up using the term `user-id`, and map it to a vendor specific header `x-lancedb-user-id`. Another alternative could be `client-id`, but I feel this is for self-identification purpose only, and `client-id` is the standard term that I think we should reserve for when we do actual RBAC, and map to header `x-client-id`.

docs/src/js/interfaces/ClientConfig.md
```diff
@@ -53,3 +53,18 @@ optional tlsConfig: TlsConfig;
...
+
+```ts
+optional userId: string;
+```
+
+User identifier for tracking purposes.
+
+This is sent as the `x-lancedb-user-id` header in requests to LanceDB Cloud/Enterprise.
+It can be set directly, or via the `LANCEDB_USER_ID` environment variable.
+Alternatively, set `LANCEDB_USER_ID_ENV_KEY` to specify another environment
```
wkalt, line 69: is this ENV_KEY pattern an existing convention we use? Seems more direct to make the user specify LANCEDB_USER_ID.

wkalt:
@jackye1995 I agree regarding storage options. Just one question on this patch. Thanks!

docs/src/js/interfaces/ClientConfig.md
jackye1995: yeah, for the speciifc customer request, they have some existing environment variable that already stores the user identity, thus it would be easier to reference that environment variable, rather than repeating the user identity twice with different environment variables.

wkalt: if the user does not already have such a variable, this approach forces them to make one and then indirectly reference it. Just seems a bit strange, but not a blocker.

PR #3247, opened 2026-04-06
feat(web): serverless lancedb (WebAssembly) browser client for searching S3/CDN directly

justsml:
## Why

Today we can publish a Lance table, but we do not have a first-party way to open that published artifact directly from the browser and search it without standing up a separate service. This PR fills that gap with an intentionally narrow contract: published, read-only, HTTP-hosted tables only.

That scope is deliberate. It keeps the first version reviewable, avoids conflating this with Cloud or general remote execution, and gives us a concrete browser story for static hosting and edge-friendly deployments.

### 1. Rust-side publish artifacts for browser clients
Files:
- `rust/lancedb/src/table.rs`
- `rust/lancedb/src/table/dataset.rs`
- `rust/lancedb/src/table/web_publish.rs`

What changed:
- table commit/open paths are patched to maintain browser sidecars next to the dataset
- writes emit `_web.json`, `_snapshot.json`, `_latest.manifest`, and `_latest.version`
- publish metadata advertises only browser-compatible vector and FTS columns
- snapshots are marked `isComplete=false` when they depend on external bases or legacy data files that the browser path cannot read safely

### 2. New `lancedb-wasm` runtime
Files:
- `rust/lancedb-wasm/src/lib.rs`
- `rust/lancedb-wasm/src/browser.rs`
- `rust/lancedb-wasm/src/browser_expr.rs`
- `rust/lancedb-wasm/src/fetch_object_store.rs`

What changed:
- opens published Lance tables over HTTP
- performs ranged object-store reads
- executes schema, refresh, vector search, text search, and hybrid search against published artifacts
- evaluates filters and selections in the browser path
- rejects unsupported browser features up front (`hamming`, `fastSearch`, incomplete snapshots, unsupported base-path layouts)

### 3. New `@lancedb/lancedb-web` package
Files:
- `web/src/index.ts`
- `web/src/worker.ts`
- `web/src/worker_protocol.ts`
- `web/src/transformers.ts`
- `web/README.md`

What changed:
- exposes `openTable()` as a read-only browser API
- prefers a Worker-backed execution path by default
- adds an optional `./transformers` wrapper for client-side query embedding generation
- removes the earlier process-wide `globalThis.fetch` swapping approach and instead requires the WASM data plane to use the installed global fetch implementation consistently

### 4. Docs-site coverage for the new public API
Files:
- `docs/mkdocs.yml`
- `docs/src/js/web.md`

What changed:
- adds a docs-site entry for the web WASM client so this is not README-only documentation
- wires the page into the JavaScript/TypeScript docs navigation

## Size / review map

Net diff against `main`: `+14,485 / -15` across 32 files.

Largest areas:
- `rust/lancedb-wasm`: `+5,166`
- `rust/lancedb/src/table/web_publish.rs`: `+1,071`
- other `rust/lancedb` plumbing: `+104 / -10`
- `web/src`: `+1,958`
- `web/__tests__`: `+1,538`
- `rust/lancedb-wasm/tests/http_search.rs`: `+586`
- docs: `+90 / -1`
- package/build/generated/docs: remainder, including `web/package-lock.json` (`+4,054`)

Suggested review order:
1. `rust/lancedb/src/table/web_publish.rs`
2. `rust/lancedb-wasm/src/lib.rs`
3. `rust/lancedb-wasm/src/browser.rs`
4. `web/src/index.ts`
5. `web/src/transformers.ts`
6. docs and tests

## Why I think this shape is right

- It is intentionally read-only and published-table-only, which keeps the browser contract explicit.
- It fails closed when the published artifacts are not actually browser-safe, instead of advertising capabilities the client cannot execute.
- It keeps the publish logic in Rust, where the table metadata and snapshot semantics already live, and keeps the web layer focused on transport and runtime concerns.

## Test coverage added

- Rust end-to-end HTTP coverage in `rust/lancedb-wasm/tests/http_search.rs`
- Rust unit coverage in `rust/lancedb/src/table/web_publish.rs`
- Rust browser-path validation coverage in `rust/lancedb-wasm/src/browser.rs`
- JS API validation coverage in `web/__tests__/index.test.ts`
- transformers wrapper coverage in `web/__tests__/transformers.test.ts`

## Out of scope

- writes or mutation APIs
- Cloud / remote execution parity
- browser support for unsupported published layouts
- `optimize_indices_for_search` (split into a separate PR so this stays focused)

## Maintainer guidance I'd especially value

1. Should browser sidecar generation happen automatically on local-table commits, or would maintainers prefer an explicit publish/export step?
2. If we keep automatic sidecar generation, should a sidecar publish failure fail the table commit instead of logging a warning?
3. Should `_web.json` / `_snapshot.json` be treated as an internal format we can iterate on, or as a versioned compatibility surface now?
4. Is including the optional `./transformers` wrapper in this PR helpful, or would maintainers prefer landing only the base `openTable()` API first?


westonpace:
I really like this idea!  Although 15K LOC is a lot :smile:

To start with can you explain what actually gets published?

> * table commit/open paths are patched to maintain browser sidecars next to the dataset
> * writes emit _web.json, _snapshot.json, _latest.manifest, and _latest.version
> * publish metadata advertises only browser-compatible vector and FTS columns
> * snapshots are marked isComplete=false when they depend on external bases or legacy data files that the browser path cannot read safely

What is a browser sidecar?  Is this some modified / reduced copy of the dataset?

What is contained in _web.json, _snapshot.json?

Is the published artifact "searchable"?  Or does it just describe the table?

wjones127:
Closing due to lack of activity. If you want to push this further, I'd suggest re-opening with a smaller initial PR that's easier to review, and outlining the plan for the rest of the package.

Issue #3242, opened 2026-04-06
Feature: First-class PyTorch dataset support


AyushExel:
### SDK

Python

### Description

Currently, there are a few rough edges that users have to figure out when working with pytorch integration. 
1. Because table/permutation objects are not serializable and lance requires running mp in fork mode, with num_workers>1, creating torch dataloaders always error out
2. Subclassing torch.datasets works, but then users have to figure out how to work with permutation/table by digging into API reference. For example, here's a boiler plate user has to write to get torch's map-style dataset/loader working:
```
class LanceArrowDataset(torch.utils.data.Dataset):
    def __init__(self, uri, table_name, **connect_kwargs):
        # ... initialization details ...
        self._perm = None
        self.uri = uri # store for re-init
        self.table_name = table_name
        self.connect_kwargs = connect_kwargs

    def __getstate__(self):
        state = self.__dict__.copy()
        state["_perm"] = None   # must zero Rust state manually — not obvious
        return state

    def _ensure_open(self):
        if self._perm is None:
            import lancedb # import must be inside worker
            from lancedb.core import Permutation
            db = lancedb.connect(self.uri, **self.connect_kwargs)
            self._perm = (
                Permutation.identity(db.open_table(self.table_name))
                .select_columns(["image_bytes", "label"])
                .with_format("arrow")
            )

    def __getitems__(self, indices):
        self._ensure_open()
        return self._perm.__getitems__(indices)
```

## Proposal

While we’re working on making these objects serializable, we could just offer an abstraction for creating torch loaders that don’t require users to hit these errors, and dig the docs reference to find getstate like workarounds 

```
from lancedb.integrations.torch import LanceTorchDataset

# Pass a table
table = lancedb.connect(uri, **connect_kwargs).open_table("images_224")
dataset = LanceTorchDataset(table, columns=["image_bytes", "label"])

loader = torch.utils.data.DataLoader(dataset, batch_size=256, num_workers=8, ...)

# Or pass a permutation built from table
perm = Permutation.identity(db.open_table(self.table_name))
		     .select_columns(["image_bytes", "label"])
.with_format("arrow")
dataset = LanceTorchDataset(perm, columns=["image_bytes", "label"])
loader = torch.utils.data.DataLoader(dataset, batch_size=256, num_workers=8, ...)
```
Note: this is specifically for map-style datasets. Iterable style might not be needed for users if they already have their data in lance format, but to we could also think of supporting similar pattern for that if needed

Issue #3244, opened 2026-04-06
Feature: Bring permutation API closer to hf/torch style dataset manipulation apis


AyushExel:
### SDK

Python

###  2-step permutation might not intuitive for existing torch/hf users. 

PermutationBuilder and Permutation are two separate classes with overlapping concerns. The builder handles row selection (shuffle, filter, split); Permutation handles read-time concerns (column selection, format, transform). This distinction matters to a data engineer but is invisible and irrelevant to an ML engineer loading a training set. 
```
# Today: Permutation.shuffle() raises AttributeError
perm_tbl = permutation_builder(table).shuffle().execute()
perm = Permutation.from_tables(table, perm_tbl)

# What ML engineers might expect
perm = Permutation.from_table(table).shuffle()
```

PR #3254, opened 2026-04-08
feat: add permutation index snapshots for Python dataset integrations

Xuanwo:
This is the first step towards first-class PyTorch dataset support in Python. It teaches `Permutation` to materialize its current view as an ordered row-id snapshot, so later dataset integrations can serialize training views without depending on live table handles.

The change adds a `PermutationReader.snapshot_indices` path in Rust, exposes it through the Python bindings as an internal `_snapshot_indices()` helper, and validates identity, split ordering, and skip/take behavior in the Python test suite.

Part of #3242.


AyushExel:
I don't remember if there was a discussion if the long term solution for pickling error needs to be from the rust layer so I'll defer to @westonpace for reviews on this, but maybe meanwhile a simpler approach for better supporting torch-only workflows could also be to have a `LanceTorchDataset` class that takes care of resetting state variables so it doesn't throw pickling error:
```
class LanceTorchDataset(torch.utils.data.Dataset):
    def __init__(self, permutation, **connect_kwargs):

    def __getstate__(self):
        state = self.__dict__.copy()
        state["_perm"] = None   # must zero Rust state manually — not obvious
        return state

    def _ensure_open(self):
        if self._perm is None:
            db = lancedb.connect(self.uri, **self.connect_kwargs)
            self._perm = (...)

    def __getitems__(self, indices):
        self._ensure_open()
        return self._perm.__getitems__(indices)
 ```

Oh and lets add a torch test with multiple workers? If reopen metadata is the path to go, then basically something like this should work without errors:
```
permutation = Permutation.identity(table)
dataloader = torch.utils.data.DataLoader(permutation, num_workers=4)
```

**Comment by Xuanwo** [comment]
Yes, `LanceTorchDataset` is the next step after this PR. We need to be able to retain the metadata needed to reconstruct itself in another process.

**Review by westonpace — CHANGES_REQUESTED** [review]
The problem is that python multiprocessing uses pickle to send object across the process boundary.  Currently `LanceTable` cannot be pickled so I don't think this current solution solves the issue.

Or maybe it does and I am missing something.  Still, we should have at least one test that does some pickling of the permutation.

**Comment by Xuanwo** [comment]
You are right on this! I have updated PR and added a test case for pickling of the permutation.

File: python/python/lancedb/permutation.py [hunk 1]
```diff
@@ -1,21 +1,98 @@
...
+            "Permutation pickling only supports LanceTable-backed permutations"
+        )
+    if table._namespace_client is not None:
+        raise pickle.PicklingError(
+            "Permutation pickling does not yet support namespace-backed tables"
+        )
+    if table._conn.uri.startswith("memory://"):
+        raise pickle.PicklingError(
+            "Permutation pickling does not support in-memory databases"
+        )
```
westonpace: These are pretty significant limitations and pickling is probably necessary for normal use of this feature.  I'd rather find a more robust solution.  Also, in-memory tables (transient permutations) are the default, we need a way to support that.

I think there are some potential approaches.

Approach 1: Pickle all the things

We would need to drive the pickling support down into Rust.  This would mean we need a way to serialize and deserialize tables.  For local tables this is easy (as you have shown).  For remote tables this would mean we need to serialize the connection parameters and I'm not sure how we would handle authN/authZ in such a scenario.  For namespace tables this means we need to drive serialization into the namespace spec itself (each impl would need to provide some way to serialize itself).

Approach 2: Require user to provide connection factory

Currently we ask the user to provide a table when we create a permutation (or we silently provide an in-memory table).  Instead we could ask the user to provide a connection factory function (that can be pickled) and a table name.  Or just a pickleable table access function.  However, this doesn't work with in-memory tables either.

Approach 3: A little of both

We could support three kinds of permutations:

* The default behavior is an in-memory permutation.  It pickles by serializing the entire permutation itself (either add serialization to the in-memory object store or just dump the whole table to a record batch)
* If the user wants persistence they can call the `persist` function (exists already) and provide a connection and table name.  We can restrict this so it only works on native connections.  Then we pickle as you have provided here by just serializing the dataset.  If the user calls this method and provides a non-native connection we direct them to the next option.
* Add a new option which allows the user to call `persist` with a "connection factory" and a table name.  The connection factory is a pickleable method that takes no arguments and returns a connection object.  This places the burden of serializing authentication information onto the user.

westonpace: I think https://github.com/lancedb/lancedb/issues/3244 might be related as well.  That PR is requesting that we merge the concepts of `PermutationBuilder` and `Permutation` into one thing.

I've been looking more at how HF does this and they basically always use an in-memory permutation table (they call it `_indices`).  So pickling is just serializing the permutation as a record batch.  This also solves the "two different APIs" problem because every permutation would have an in-memory table and so methods like shuffle or filter are just manipulating that in-memory table.

So maybe a faster short-term goal is to just always have the permutation table be an in-memory table, add pickle support for that, and then decide how to add permutation persistence later.  My original concern is that users would want to create permutations with billions or tens of billions of rows.  The resulting permutation table would be massive and so persistence would be important (so we don't have to load the whole thing in RAM).

However, more and more, I am seeing that even though the base table is in the billions the actual model is never trained on that much data.  In other words, the filtered permutation is typically much smaller (millions) and so the permutation table would be smaller.

This could be a python-only change or we could push it down into the Rust code as well.

westonpace: If we want to add permutation persistence later I think I would propose doing it differently anyways.  Instead of requiring it to be "connection based" the `persist` method should take in a callback or ABC for a "persistence manager".  This lets the user decide how to persist it (and how to serialize this persistence manager)

Xuanwo: Let me know if this change aligns what you think: https://github.com/lancedb/lancedb/pull/3254/commits/533c81ce03c65c0396c4467f7ddc202eae021390

Xuanwo:
Closing this PR because the implementation direction has been superseded. The next PR will target remote table and remote-backed permutation support for PyTorch multiprocessing as a complete feature.

PR #3264, opened 2026-04-11
feat(python): add map() method to Permutation for dict-based transforms

Its-Tanay:
### Summary :
Adds `Permutation.map(fn, *, batched=False)` — a convenience wrapper over `with_transform` that handles Arrow-to-Python conversion so users can write `fn(dict) -> dict` without touching`pyarrow.RecordBatch`.
- `batched=False` (default): fn receives one row dict, returns one row dict.                                                                                                                
- `batched=True`: fn receives `dict[str, list]`, returns `dict[str, list]`.
- `with_transform` remains available for direct Arrow access.

Closes [#3246](https://github.com/lancedb/lancedb/issues/3246)

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -762,6 +762,36 @@ class Permutation:
         assert transform is not None, "transform is required"
         return Permutation(self.reader, self.selection, self.batch_size, transform)
 
+    def map(self, fn: Callable, *, batched: bool = False) -> "Permutation":
+        """
+        Apply a transformation function to each row or batch.
+
+        When batched=False (default), fn receives a single row as a dict
+        and should return a dict. This is applied to every row.
+
+        When batched=True, fn receives a column-oriented dict (keys are
+        column names, values are lists) and should return a dict of the
+        same shape.
+
+        This is a lazy transform -> fn is not called until data is
+        iterated or accessed. This replaces any previously set format
+        or transform.
+        """
+        if not callable(fn):
+            raise TypeError("fn must be callable")
+
+        if batched:
+
+            def _wrapper(batch: pa.RecordBatch):
+                return fn(Transforms.arrow2pythoncol(batch))
+
+        else:
+
+            def _wrapper(batch: pa.RecordBatch):
+                return [fn(row) for row in Transforms.arrow2python(batch)]
+
+        return self.with_transform(_wrapper)
+
     def __getitem__(self, index: int) -> Any:
         """
         Returns a single row from the permutation by offset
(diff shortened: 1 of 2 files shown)


Its-Tanay:
@AyushExel , @wjones127 can you folks please help review this.

Its-Tanay:
Resurfacing this folks, @wjones127 @AyushExel (really sorry to bother)

python/python/lancedb/permutation.py
```diff
@@ -758,10 +758,42 @@ def with_transform(self, transform: Callable[pa.RecordBatch, Any]) -> "Permutati
...
+
+            def _wrapper(batch: pa.RecordBatch):
+                return fn(Transforms.arrow2pythoncol(batch))
+
+        else:
+
+            def _wrapper(batch: pa.RecordBatch):
+                return [fn(row) for row in Transforms.arrow2python(batch)]
+
+        return self.with_transform(_wrapper)
```
westonpace, line 795: It looks like this would replace any existing transform that has been set?  It will also replace any existing call to map that has been set?

What happens in huggingface if there is already a transform that has been set?

Does it also replace previous transformations or can it chain transformations?

westonpace (commented):
Thanks for the contribution!  Quick question before proceeding.

python/python/lancedb/permutation.py
Its-Tanay: Thanks @westonpace  — I looked into HF’s implementation a bit more. Their `.map()` is eager, so each call materializes the dataset and carries formatting forward through a decorator that reapplies the previous format.

Our `.map()` is lazy, so that approach doesn’t translate directly. But that’s more of a design choice on my side than a hard limitation. We could split `transform_fn` into a map chain plus a final formatter and do a dict ↔ arrow conversion at the boundary to preserve formats across `.map()` calls. Probably around ~80 lines of work.

Since we’re still defining the semantics here, I think there are three reasonable options for v1:

1. Replace (current PR)
   Simple, but `.map().map()` silently drops the first transform, which will probably surprise anyone familiar with HF or most `.map()` APIs in Python.

2. Chain `.map()` with `.map()`, and treat `.with_format()` / `.with_transform()` as resets in either direction
   Small change, fixes the surprising behavior, and avoids getting into the more ambiguous edge cases.

3. Literal HF behavior
   Chain everything and preserve formats across `.map()` calls. Most complete, but we’d also need to define semantics for cases HF themselves still have TODOs around, like whether `fn` should receive tensors or dicts after `.with_format("torch").map(...)`.

My preference is (2). It gives users the behavior they’d naturally expect from `.map()` without us having to lock in semantics for the weird cases yet. Happy to do (3) instead if you’d rather land the full behavior in one go.

Its-Tanay: Hey @westonpace, would love to know your thoughts on this 🙇

PR #3270, opened 2026-04-12
fix(node): prevent reranker from keeping process alive

### LINKED ISSUE #3269: bug(node): using reranker keeps Node.js process from exiting
### LanceDB version

v0.27.2

### What happened?

Minimal reproducible example, `test.js`:

```js
import lancedb from '@lancedb/lancedb'

const data = [
  { text: 'hello', vector: [1, 2, 3] }
]

const db = await lancedb.connect('./db')
const table = await db.createTable('test', data, { mode: 'overwrite' })

await table.createIndex('text', { config: lancedb.Index.fts() })
await table.waitForIndex(['text_idx'], 30)

const queryText = 'hello'
const queryVector = [1, 2, 3]
const reranker = await lancedb.rerankers.RRFReranker.create()

const results = await table
  .query()
  .nearestTo(queryVector)
  .fullTextSearch(queryText)
  .rerank(reranker)
  .toArray()

console.log(results)

table.close()
db.close()
```

```sh
node test.js
```

Process is unable to exit naturally unless I comment out `//.rerank(reranker)`.

Like it lacks `reranker.close()` method or something like that.

snigenigmatic:
Fixes #3269.

## What I observed
Using a reranker in a hybrid query could keep the Node.js process alive even after `table.close()` and `db.close()`.

## Root cause
The reranker callback bridge used a `ThreadsafeFunction` in referenced mode, which can keep the event loop alive longer than intended.

## Minimal fix
- In `nodejs/src/rerankers.rs`, create the reranker callback TSFN in weak mode (`.weak::<true>()`).
- Add a regression test in `nodejs/__test__/rerankers.test.ts` that spawns a child process, runs a rerank query, and asserts the process exits naturally.

## Validation
- Built Node bindings successfully.
- Ran targeted tests: `rerankers.test.ts` passes (including new regression test).
- Pre-commit checks for changed files were run and clean.

### DIFF
--- a/nodejs/__test__/rerankers.test.ts
+++ b/nodejs/__test__/rerankers.test.ts
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
+import { spawn } from "node:child_process";
+import * as path from "node:path";
 import { RecordBatch } from "apache-arrow";
 import * as tmp from "tmp";
 import { Connection, Index, Table, connect, makeArrowTable } from "../lancedb";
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This looks good! Thanks for submitting the PR. I appreciate the test.

snigenigmatic:
Thanks for the approval @wjones127. Let me know if anything else is needed from my side before merge.

PR #3271, opened 2026-04-12
feat(nodejs): add Scannable primitive for streaming ingestion

### LINKED ISSUE #3223: feat(nodejs): Add streaming data support to Rust bindings
NodeJS currently serializes all input data into a single in-memory IPC buffer before
passing it to Rust. This limits the size of data that can be added and prevents
streaming.

Add a streaming/Scannable interface to the napi-rs bindings (similar to Python's
\`PyScannable\`) so that data can be sent incrementally to Rust without materializing
the entire dataset in memory. The exact API design is TBD.

### Blocks
- NodeJS add() migration
- NodeJS create_table() migration
- NodeJS merge_insert() migration

Its-Tanay:
## **Summary**

This PR adds a **Scannable primitive** to the Node.js bindings, bringing parity with Python's `PyScannable`.

A `Scannable` wraps a schema, an optional row count hint, a rescannable flag, and a batch producing callback. On the Rust side it implements `lancedb::data::scannable::Scannable`. The goal is to give consumers such as `Table.add`, `createTable`, and `mergeInsert` a way to stream data without materializing the full dataset in JS memory.

This PR introduces only the primitive. Migrating existing consumers to use it will come in follow up work.

---

### **Transport**

The transport uses the **Arrow IPC Stream format, one batch at a time**.

The JS side encodes each `RecordBatch` into a self contained IPC Stream message containing schema, batch, and end of stream. The message is returned as a `Buffer` through a napi `ThreadsafeFunction`. The Rust side decodes it using `arrow_ipc::reader::StreamReader`.

Only one batch is active at a time, so JS memory stays bounded by the batch size. The Node `Buffer` size limit of about 4 GiB therefore does not constrain the stream as a whole.

I initially evaluated the Arrow C Data Interface, which is the approach used in Python. I dropped that path after confirming that the `apache-arrow` npm package does not expose a C Data Interface export in any supported version from 15 to 18. JavaScript is not listed in Arrow's C Data Interface implementation table, and the upstream tracking issue remains open with no scheduled work.

Third party FFI shims would introduce additional dependency risk without solving the core maintenance problem. Using IPC adds one encode and decode step per batch, but the cost is predictable and typically dominated by Lance's write path.

---

### **API**

```ts
class Scannable {
  readonly schema: Schema
  readonly numRows: number | null
  readonly rescannable: boolean

  static fromFactory(schema, factory, opts?)
  static fromTable(table, opts?)
  static fromIterable(schema, iter, opts?)
  static fromRecordBatchReader(reader, opts?)
}
```

The FFI boundary consists of a single callback:

`getNextBatch(isStart: boolean): Promise<Buffer | null>`

`isStart` is `true` on the first call of each new scan and `false` for every call after it. The JS side uses it to drop any cached iterator and re-invoke the factory at scan boundaries. This is what makes a rescannable source restart at batch 0 on every `scan_as_stream` call, even when a previous scan ended mid stream, for example a retried write after a network error. Without this signal a retry would resume a stale iterator and silently skip already emitted batches.

In addition, a schema only IPC buffer is transferred once during construction.

---

## **Changes**

* `nodejs/src/scannable.rs`
  Adds `NapiScannable` and the `LanceScannable` implementation. Implements `schema()`, `num_rows()`, `rescannable()`, and `scan_as_stream()`. Includes per batch schema validation against the declared schema, one shot enforcement for non rescannable sources, and a scan boundary reset signal (`isStart`) so rescannable sources restart from batch 0 on every `scan_as_stream` call rather than resuming a stale iterator.

* `nodejs/src/lib.rs`
  Module registration.

* `nodejs/lancedb/scannable.ts`
  Defines the `Scannable` class and the four constructors listed above. Each constructor rejects option combinations it cannot honor, for example a `rescannable: true` request on a one shot iterable or reader, and a `numRows` that disagrees with an in memory table's row count.

* `nodejs/lancedb/index.ts`
  Exports the new primitive.

* `nodejs/__test__/scannable.test.ts`
  Test suite for the primitive.

---

## **Validation**

Before implementing the bridge, I ran an end to end harness with a JS producer feeding a standalone Rust consumer built against the same `arrow-ipc` version used in the bridge.

The harness covered the following scenarios:

* happy path
* empty stream
* 1,000 small batches
* 10 large batches
* mixed primitive types with nullables
* nested `List<Struct<>>`
* truncated stream error handling
* declared schema mismatch validation
* a 6 GB stress test through the pipe

All scenarios completed with bounded memory usage. The goal of this harness was to confirm that the IPC Stream transport works correctly end to end and that Node's `Buffer` size limit does not constrain the overall stream.

Separately, the rescannable restart contract was verified with a focused harness. A rescannable source is consumed partially and the scan is dropped mid stream, then re-scanned. The re-scan replays from batch 0 rather than resuming the stale iterator. The same harness was run with the `isStart` reset path disabled and the mid stream restart case failed as expected, confirming the test exercises the real regression.

These harnesses are not meant to replace the full test suite, which is described below.

---

## **Tests**

`__test__/scannable.test.ts` covers construction, metadata reflection, per constructor defaults and overrides, construction time validation, the native handle surface, and schema variety across empty tables, nested types, `FixedSizeList`, and wide schemas.

Runtime scan behavior including `scan_as_stream`, one shot enforcement on non rescannable sources, schema mismatch detection, IPC decode failures, and rescannable restart semantics is not exercised here. There is no in tree JS consumer of `NapiScannable` yet. This mirrors Python's `PyScannable`, which has no dedicated test file and is covered transitively through the consumers that accept a Scannable.

Runtime coverage will follow in the consumer migration work.

---

## **Status**

Ready for review.

Closes #3223

---


Its-Tanay:
@wjones127 do share your feedback and review on the overall design of the primitive. Thanks!

Its-Tanay:
Resurfacing this, @wjones127 (really sorry to bother)

Its-Tanay:
@wjones127 requesting your review on this, I have fixed the JSDoc issues from the last CI run.

Its-Tanay:
@wjones127 could you please approve the workflow, I have fixed the issues from the last run. Also need your review on this PR.

Its-Tanay:
@wjones127 @Xuanwo Dropping a gentle reminder to review this MR, so we can close the issue.

nodejs/lancedb/scannable.ts
```diff
@@ -0,0 +1,195 @@
...
+    return new Scannable(native, schema, numRows, rescannable);
+  }
+
+  /**
+   * Build a Scannable from an in-memory Arrow `Table`. Rescannable by
+   * default — the table's batches are replayed on each scan.
+   */
+  static fromTable(table: ArrowTable, opts: ScannableOptions = {}): Scannable {
+    return Scannable.fromFactory(table.schema, () => table.batches, {
+      numRows: opts.numRows ?? table.numRows,
```
westonpace: Should it be an error if `opts.numRows` and `table.numRows` are both set but don't match?

nodejs/lancedb/scannable.ts
```diff
@@ -0,0 +1,195 @@
...
+      rescannable: opts.rescannable ?? true,
```
westonpace: Ditto, should we even look at `opts.rescannable` in this case?

nodejs/lancedb/scannable.ts
```diff
@@ -0,0 +1,195 @@
...
+   * schema so the consumer can validate before any batch is pulled.
+   */
+  static fromIterable(
+    schema: Schema,
+    iter: AsyncIterable<RecordBatch> | Iterable<RecordBatch>,
+    opts: ScannableOptions = {},
+  ): Scannable {
+    return Scannable.fromFactory(schema, () => iter, {
+      numRows: opts.numRows,
+      rescannable: opts.rescannable ?? false,
```
westonpace, line 185: I don't think it's valid to allow `rescannable` to be set to true.  The iterator would be exhausted.  We should error if `opts.rescannable` is true.

nodejs/lancedb/scannable.ts
```diff
@@ -0,0 +1,195 @@
...
+   * The reader must already be opened (via `.open()`) so its `.schema` is
+   * populated — `RecordBatchReader.from(...)` returns an unopened reader.
+   */
+  static fromRecordBatchReader(
+    reader: RecordBatchReader,
+    opts: ScannableOptions = {},
+  ): Scannable {
+    return Scannable.fromFactory(reader.schema, () => reader, {
+      numRows: opts.numRows,
+      rescannable: opts.rescannable ?? false,
```
westonpace: Same comment as above.  A default to false is not correct, it should be error if true.

nodejs/lancedb/scannable.ts
```diff
@@ -0,0 +1,195 @@
...
+
+// Uses IPC File format (not Stream) because the Rust side parses it with
+// `ipc_file_to_schema`, which expects File format.
+function encodeSchema(schema: Schema): Buffer {
+  const writer = new RecordBatchFileWriter();
+  writer.reset(undefined, schema);
+  writer.finish();
+  writer.close();
+  return Buffer.from(writer.toUint8Array(true));
+}
```
westonpace: Do we do something like this (serializing as IPC) when accepting arrow tables / record batch readers on the normal path?

If so, don't these functions exist somewhere already?

If not, why are we doing it here?

I think serializing it as IPC is probably cheap enough I just want to avoid having too many different paths with different behaviors if we can avoid it.

westonpace (changes requested):
A few questions and suggestions but overall I really like this, thank you!

nodejs/lancedb/scannable.ts
Its-Tanay: Fair, silently overriding `table.numRows` was problematic. Throws now if `opts.numRows` is set and doesn't match.

nodejs/lancedb/scannable.ts
Its-Tanay: Agreed, accepting `rescannable: false` would actually let the user under claim a feat the table already has. Throws now if it's set, with a message pointing them to either omit it or pass true.

nodejs/lancedb/scannable.ts
Its-Tanay: You're right for generators, async generators, and already-iterators. Those return themselves from `[Symbol.iterator]()` so they exhaust after one pass. But the parameter type also admits arrays, sets, and custom iterables whose `[Symbol.iterator]()` mints a fresh iterator each call, where `rescannable: true` is honest.
                                                                                                                                                                  
I decided to move with a partial fix: at construction, detect whether the input is structurally one-shot by calling `[Symbol.iterator]()` once and checking identity. Throw with a pointer to `fromFactory(schema, () => createIter(), { rescannable: true })` if it is, pass through otherwise. Closes the issue you flagged for the cases that deserve it without breaking the array/set case.                                                                                                            
                             
Also added a sub-describe covering the full matrix (array, set, custom replayable, sync/async generators, empty cases, IterableIterator from .values(), raw iterators, null/undefined) so each detection branch is exercised.

nodejs/lancedb/scannable.ts
Its-Tanay: Agreed, unlike fromIterable, this one's unconditional. `RecordBatchReader[Symbol.iterator]()` always returns this and we never call `reader.reset()` between scans, so no input makes `rescannable: true` honest here. Throws now, pointing to `fromFactory(schema, () => openReader(), { rescannable: true })` for genuinely replayable readers.

nodejs/lancedb/scannable.ts
Its-Tanay: Oh yeah, originally I kept these local because I wanted the constructors to stay sync (`encodeSchema` was using `toUint8Array(true)` to avoid the `await`), and there wasn't an exact stream-format-per-batch helper in `arrow.ts` to point at. but yeah you're right, two paths producing the same bytes can drift later, better to just converge now.

Updated:                   

- `encodeBatch` → moved to `arrow.ts` as `fromRecordBatchToStreamBuffer`, similar to `fromRecordBatchToBuffer` (matches the `fromTableToBuffer` ↔ `fromTableToStreamBuffer` pattern that's already there).
- `encodeSchema` → removed, scannable now calls `fromTableToBuffer(makeEmptyTable(schema))` directly, which is the same for `connection.ts:499-500` already uses same bytes, same exec path on both sides .                                                                                                                      
   
only side effect is that `Scannable.fromXYZ` factories are async now since `fromTableToBuffer` is.

Its-Tanay:
@westonpace Thanks for taking the time to review my work. I’ve worked through your comments and would love your feedback!

nodejs/src/scannable.rs
```diff
@@ -0,0 +1,229 @@
...
+                Err(Error::InvalidInput {
+                    message: "Scannable has already been consumed (non-rescannable source)"
+                        .to_string(),
+                })
+            });
+            return Box::pin(SimpleRecordBatchStream::new(err_stream, schema));
+        }
+        self.scanned = true;
+
+        let tsfn = Arc::clone(&self.get_next_batch);
```
wjones127, line 128: issue(blocking): there doesn't seem to be a way to reset the scannable when it is `rescannable`, is there? For example, consider this situation: there is a Scannable with 10 batches, we read 5 batches and error, then we retry the write by calling `scan_as_stream` again. How does the `get_next_batch` function know to start over from the beginning?

nodejs/src/scannable.rs
```diff
@@ -0,0 +1,229 @@
...
+            "[scannable/js-iterator] napi error status={}, reason={}",
+            e.status, e.reason
+        ),
+    })
+}
+
+/// Decode one IPC Stream buffer (schema + batch + EOS) into a `RecordBatch`.
+/// Each buffer is a standalone IPC stream, so every decoded stream schema must
+/// match the one declared at construction.
+fn decode_one_batch(buf: &[u8], declared: &SchemaRef) -> LanceResult<RecordBatch> {
```
wjones127, line 228: note: in the future, we could make this deserialization zero-copy. We have some utilities for this in https://github.com/lance-format/lance/blob/bcf5e7e01d89fe8d834423f1c11e71f97a7c2cfb/rust/lance-arrow/src/ipc.rs#L301, but I think these aren't low level enough to use here. They expect the full stream to be in one buffer, including the schema. We could eventually create similar functions that parse the schema and each batch separately. That can be a future PR though.

nodejs/lancedb/scannable.ts
```diff
@@ -0,0 +1,269 @@
...
+    const getNextBatch = async (): Promise<Buffer | null> => {
+      if (iter === null) {
+        iter = normalizeIterator(factory());
+      }
+      const result = await iter.next();
+      if (result.done) {
+        iter = null;
+        return null;
+      }
+      return fromRecordBatchToStreamBuffer(result.value);
```
wjones127, line 110: issue(blocking): related to what I was saying in the other comment, I think this assumes that we only rescan after fully consuming the original iterator. This may not always be the case. The purpose of rescanning is to be able to restart in case of any failures, like networking errors. These could happen mid stream.

wjones127 (changes requested):
I'm excited for this change!

I'm noticing one key issue with the way rescanning is working here. Otherwise, it looks like a good change.

nodejs/src/scannable.rs
Its-Tanay: I’ve fixed it by passing an `isStart` flag through the FFI callback so Rust can signal JS to recreate the iterator at the start of each `scan_as_stream` call, instead of reusing the cached one. That should also bring it in line with how `PyScannable` behaves, there the factory gets invoked for every scan.

nodejs/lancedb/scannable.ts
Its-Tanay: Definitely missed that. I’m updating it so every `scan_as_stream` call recreates the JS iterator regardless of how the previous scan ended.

nodejs/src/scannable.rs
Its-Tanay: Thanks, that’s a helpful reference. Yeah, I don’t think those helpers quite fit this layout since each buffer here is its own stream with a schema header attached. Could plan a future PR for split helpers (schema parse once + batch parse against existing schema) and we could switch `decode_one_batch` over to that path later.

wjones127 (approved):
Looks good now, thank you!

PR #3277, opened 2026-04-14
fix(python): allow _rowid in Permutation.select_columns

### LINKED ISSUE #3132: bug(python): Permutation `select_columns` rejects `_rowid`, breaking Blob V2 workflows
### LanceDB version

0.30.0-beta.5

### What happened?

`Permutation.select_columns(["_rowid", ...])` raises `ValueError: Cannot select column '_rowid' because it does not exist`. This prevents using `take_blobs()` with permutations, since `take_blobs` requires `_rowid` to resolve blob references.

The Rust layer already fully supports `_rowid` selection and even has explicit logic to detect whether `_rowid` was requested — [`has_row_id()`](https://github.com/lancedb/lancedb/blob/216c1b5f7743b991dceca31d26579e8b80736153/rust/lancedb/src/dataloader/permutation/reader.rs#L318-L343) checks all three selection variants (`Select::All`, `Select::Columns`, `Select::Dynamic`) and uses the result to decide whether to [keep or drop `_rowid` from the output batch](https://github.com/lancedb/lancedb/blob/216c1b5f7743b991dceca31d26579e8b80736153/rust/lancedb/src/dataloader/permutation/reader.rs#L279-L284). The only gap is the Python validation layer rejecting it before it reaches Rust.

```python
from lancedb import connect
from lancedb.permutation import permutation_builder, Permutation

db = connect("memory:///")
tbl = db.create_table("t", {"id": range(10), "value": range(10)})
perm_tbl = permutation_builder(tbl).shuffle(seed=42).execute()
perm = Permutation.from_tables(tbl, perm_tbl)

# This fails:
perm.select_columns(["_rowid", "id"])
# ValueError: Cannot select column `_rowid` because it does not exist
```

Zelys-DFKH:
## Summary

`Permutation.select_columns(["_rowid", ...])` raised a `ValueError` because the Python validation layer built its allowed set from Arrow schema names, and `_rowid` is a virtual column not present in the schema.

The Rust `PermutationReader` already handles `_rowid` selection via `has_row_id()`. The fix treats `_rowid` as a special case in `select_columns` and passes it through directly.

## Changes

- `python/python/lancedb/permutation.py`: skip the schema lookup for `_rowid` in `select_columns`
- `python/python/tests/test_permutation.py`: add `test_select_columns_rowid` covering `["_rowid", "id"]` and `["_rowid"]` alone

## Related

Fixes #3132

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -648,6 +648,9 @@ class Permutation:
 
         new_selection = {}
         for name in columns:
+            if name == "_rowid":
+                new_selection[name] = name
+                continue
             value = self.selection.get(name, None)
             if value is None:
                 raise ValueError(
--- a/python/python/tests/test_permutation.py
+++ b/python/python/tests/test_permutation.py
@@ -699,6 +699,18 @@ def test_select_columns(some_permutation: Permutation):
         some_permutation.select_columns([])
 
 
+def test_select_columns_rowid(some_permutation: Permutation):
+    # _rowid is a virtual column produced by the Rust reader; it is not part of
+    # the Arrow schema, so it was previously rejected by select_columns().
+    result = some_permutation.select_columns(["_rowid", "id"])
+    assert "_rowid" in result.schema.names
+    assert "id" in result.schema.names
+
+    # Selecting _rowid alone should also work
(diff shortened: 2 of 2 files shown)


wjones127:
Closing in favor of https://github.com/lancedb/lancedb/pull/3133, which has more thorough testing.

PR #3278, opened 2026-04-15
docs: fix broken documentation links

Gezi-lzq:
## Summary

Replace old `lancedb.com/docs/*` URLs with `docs.lancedb.com/*` across the codebase.

## Problem

URLs with sub-paths like `lancedb.com/docs/storage/`, `lancedb.com/docs/tables/`, and `lancedb.com/docs/quickstart/` all return **Page Not Found
**. Only `lancedb.com/docs/` (without sub-path) correctly redirects to `docs.lancedb.com`.

image

## Changes

Updated 19 files (Rust, Python, TypeScript, docs, README, issue template) to use the correct `docs.lancedb.com` domain.

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation.yml
+++ b/.github/ISSUE_TEMPLATE/documentation.yml
@@ -18,6 +18,6 @@ body:
       label: Link
       description: >
         Provide a link to the existing documentation, if applicable.
-      placeholder: ex. https://lancedb.com/docs/tables/...
+      placeholder: ex. https://docs.lancedb.com/tables/...
     validations:
       required: false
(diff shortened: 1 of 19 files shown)


Gezi-lzq:
As a note, this issue could also be addressed by adding a wildcard redirect rule on the server side (e.g. `lancedb.com/docs/*` → `docs.lancedb.com/*`). Currently only `lancedb.com/docs/` redirects correctly, while sub-paths like `/docs/storage/` and `/docs/tables/` return 404.

A server-side redirect would serve as a safety net for any links not covered by this PR (e.g. external blog posts, bookmarks, etc.).

Xuanwo (approved):
Thank you!

PR #3285, opened 2026-04-17
fix(node): remove redundant `postbuild:release` script to fix build failure

### LINKED ISSUE #3284: bug(node): `npm run build:release` fails due to missing file in `postbuild:release`
### LanceDB version

v0.27.2

#### 📋 Description
When attempting to run `npm run build:release` in the `nodejs` directory, the build process fails during the `postbuild:release` hook. The error indicates that the source file for the copy command cannot be found.

#### ℹ️ Environment
- **OS:** Linux
- **Node.js Version:** v22.22.2

#### 📝 Steps to Reproduce
1. Navigate to the `nodejs` directory.
2. Run the command: `npm run build:release`

#### ❌ Error Log
```bash
> @lancedb/lancedb@0.27.2 postbuild:release
> shx mkdir -p dist && shx cp lancedb/*.node dist/

cp: no such file or directory: lancedb/*.node
```

#### 🔍 Analysis
I have analyzed the scripts defined in `package.json`:

- **`build:release`**: `"napi build --platform --release --dts ../lancedb/native.d.ts --js ../lancedb/native.js --output-dir dist"`
- **`postbuild:release`**: `"shx mkdir -p dist && shx cp lancedb/*.node dist/"`

**Root Cause:**
The `build:release` command uses the flag `--output-dir dist`, which correctly generates the `lancedb.*.node` files directly into the `dist/` directory. Consequently, the `lancedb/` source directory does not contain any `*.node` files.

When the `postbuild:release` script attempts to execute `cp lancedb/*.node dist/`, it fails because the source files do not exist in the `lancedb/` folder.

#### 💡 Proposed Solution
I suggest fixing this issue using one of the following methods:

1. **Remove the redundant script:**
   Simply delete the `postbuild:release` script from `package.json`, as the files are already in the correct destination (`dist/`) after the build.

   **OR**

2. **Adjust the output directory:**
   Modify the `build:release` script to point the `--output-dir` parameter to `lancedb` instead of `dist` (if the copy step is strictly required for a specific structure).

If there are no objections, I would be happy to open a PR to fix this issue (likely via solution 1).

See the description above.

qingfeng-occ:
The `build:release` command already outputs the `*.node` files directly to the `dist/` directory via the `--output-dir dist` flag.

Therefore, the `postbuild:release` script, which attempts to copy `*.node` files from the `lancedb/` source directory, fails with a "no such file or directory" error because the source files do not exist there.

This commit removes the redundant `postbuild:release` script to resolve the build failure.

fix #3284

### DIFF
--- a/nodejs/package.json
+++ b/nodejs/package.json
@@ -75,7 +75,6 @@
     "build:debug": "napi build --platform --dts ../lancedb/native.d.ts --js ../lancedb/native.js --output-dir lancedb",
     "postbuild:debug": "shx mkdir -p dist && shx cp lancedb/*.node dist/",
     "build:release": "napi build --platform --release --dts ../lancedb/native.d.ts --js ../lancedb/native.js --output-dir dist",
-    "postbuild:release": "shx mkdir -p dist && shx cp lancedb/*.node dist/",
     "build": "npm run build:debug && npm run tsc",
     "build-release": "npm run build:release && npm run tsc",
     "tsc": "tsc -b",


qingfeng-occ:
Hi @wjones127  @jackye1995  @Xuanwo  ,
Could anyone available please take a look at this PR?  I'd really appreciate your feedback.
Thanks!

westonpace (approved):
Thanks for cleaning this up.  I guess it raises a larger question of why we aren't using this in our publish workflows.  Those use...

```
npx napi build --platform --release \
   --features ${{ matrix.settings.features }} \
   --target ${{ matrix.settings.target }} \
   --dts ../lancedb/native.d.ts \
   --js ../lancedb/native.js \
   --strip \
   $EXTRA_ARGS \
   --output-dir dist/
```

So we've already deviated a little on `--strip`.

PR #3288, opened 2026-04-19
fix(rust): materialize declared namespace tables on create

jackye1995:
## Summary
- handle `declare_table` already-exists conflicts in the Rust namespace database create path
- reuse declared-but-not-materialized table metadata instead of failing create mode
- preserve overwrite behavior while allowing declared Geneva system tables to be materialized

### DIFF
--- a/rust/lancedb/src/database/namespace.rs
+++ b/rust/lancedb/src/database/namespace.rs
@@ -239,35 +239,62 @@ impl Database for LanceNamespaceDatabase {
                     .filter(|o| !o.is_empty());
                 (loc, opts, response.managed_versioning)
             } else {
-                let response = self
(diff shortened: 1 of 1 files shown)


LuQQiu (approved):
In any case that we will declare but not create the table? 
When declare, are you using some system table metadata or some kind of metadata on object storage to reserve the table creation?

jackye1995:
This happened in a retry, basically the table was declared during create and the create itself failed. Then the table ended up in a state that is declared but not created. The second attempt sees the table is available (declared) and ended up not able to create it. 

Currently declare writes an empty .lance-reserved file inside the table root to indicate it is reserved, and record the table information in the __manifest system table.

PR #3289, opened 2026-04-20
feat(python): support model-backed native FTS tokenizers

### LINKED ISSUE #2168: Feature: Support Chinese Language in FTS

Xuanwo:
This wires Lance's existing `jieba/*` and `lindera/*` native FTS tokenizers through the Python SDK instead of leaving them behind disabled features and narrow public typing. It also documents the `LANCE_LANGUAGE_MODEL_HOME` model layout and adds Python coverage for successful CJK indexing plus missing-model error guidance.

Closes #2168.


python/python/lancedb/table.py
```diff
@@ -958,7 +1009,8 @@ def create_fts_index(
         tokenizer_name: str, default "default"
             A compatibility alias for native tokenizer configs. Can be "raw",
             "default" or the 2 letter language code followed by "_stem". So
-            for english it would be "en_stem".
+            for english it would be "en_stem". Prefer ``base_tokenizer`` for
```
westonpace: Why prefer `base_tokenizer`?

westonpace (approved):
Looks good assuming we update the docs to describe how to use this feature soon.  How big are these language models, out of curiosity?  How are they normally distributed?  I wonder if there is some way we can help users get them installed, for example, download them by default to ~/.cache/lancedb/language-models.

python/python/lancedb/table.py
Xuanwo: `tokenizer_name` is the legacy name used by tantivy, I update the docs to make it more clear.

Xuanwo:
> How big are these language models, out of curiosity? How are they normally distributed? I wonder if there is some way we can help users get them installed, for example, download them by default to ~/.cache/lancedb/language-models.

It's about 10MiB to 100MiB levels. I personally think it's a good idea but not sure about how we expose this.

PR #3290, opened 2026-04-20
feat(nodejs): add col() and lit() SQL helpers for camelCase column names

### LINKED ISSUE #3211: Make sure `col()` handlings camelCase column names
Make sure `col()` can be used on camelCase and other column names. May need to require wrapping in backticks or some other quoting character.

We should add an integration test for this.

This is most important for NodeJS, where camelCase fields are common.

Follow up from https://github.com/lancedb/lancedb/pull/3150

octo-patch:
Fixes #3211

### DIFF
--- a/nodejs/lancedb/expr.ts
+++ b/nodejs/lancedb/expr.ts
@@ -1,0 +1,146 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+/**
+ * SQL expression helpers for type-safe query construction.
+ *
+ * These utilities make it easy to reference column names — including names
+ * with camelCase, spaces, or other special characters — without writing raw
+ * SQL strings by hand.
+ *
+ * @example
+ * ```ts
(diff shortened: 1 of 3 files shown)


nodejs/lancedb/expr.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Use `col()` inside template literals when building `.where()` predicates:
+ *
+ * ```ts
+ * await table.query()
+ *   .where(`${col("firstName")} = ${lit("Alice")} AND ${col("age")} > ${lit(18)}`)
+ *   .toArray();
+ * ```
+ */
+export function col(name: string): string {
```
wjones127, line 119: issue(blocking): We should not return strings from this function. It should return an expression object instead, which can be converted to SQL in the end. Using strings as the representation of expressions will make it very hard to build up expressions and will make us reply on a lot of string parsing.

wjones127 (changes requested):

pranshu314:
Would you like me to address the comments and raise a PR for the same? Or is this already patched.

PR #3301, opened 2026-04-21
feat(rust): batch_knn via DataFusion extension node

wjones127:
## Summary

Implements batch brute-force KNN as a DataFusion extension node, replacing the
serial per-query `batch_knn_flat` in `analyze_index` with a single batched
operation.

- `DataFrameKnnExt` trait: public API to attach KNN to any DataFusion DataFrame
- `KnnNode` logical plan + `PartialKnnExec`/`MergeKnnExec` physical plans
- `KnnProjectVectorRule` optimizer: drops vector column from `RowConverter` when
  not projected downstream
- Streaming top-k heap per query with `OwnedRow`, threshold-guarded `.owned()`
- `MergeKnnExec` uses `interleave_record_batch` and single-pass grouping
- `BaseTableAdapter.with_row_id()` for `_rowid` support through DataFusion path
- `batch_knn` is `pub(crate)` — only `DataFrameKnnExt` is public

**Depends on #3303** (`task/index-analyze`) — should be rebased onto that branch
before merging.


rust/lancedb/src/table/datafusion/knn.rs
```diff
@@ -0,0 +1,629 @@
...
+use crate::DistanceType;
+
+// ---------------------------------------------------------------------------
+// Schema helper
+// ---------------------------------------------------------------------------
+
+/// Extend an input Arrow schema with `query_index: Int32` and `_distance: Float32`.
+pub(crate) fn knn_output_schema(input_schema: &SchemaRef) -> SchemaRef {
+    let mut fields = input_schema.fields().to_vec();
+    fields.push(Arc::new(Field::new("query_index", DataType::Int32, false)));
```
wjones127: suggestion: Let's make any hardcoded fields LazyStatics, so they are only allocated once.

wjones127: suggestion: I think we can have two fields:

* `_query_index`: Int32 that references the position of original query vector
* `_query_vector`: `Dict(Int32, FixedSizeList(<value-type>))` that has the original query vectors. The dictionary will just be the original input vectors and the indices will be the same buffer as _query_index, so it will be efficient to represent.

rust/lancedb/src/table/datafusion/knn.rs
```diff
@@ -0,0 +1,629 @@
...
+/// all query vectors at once.
+///
+/// The output schema is the input schema augmented with:
+/// - `query_index: Int32` — 0-based index into `query_vectors`
+/// - `_distance: Float32` — distance from that query to this row
+///
+/// At most `k` rows are emitted per distinct `query_index`.
+pub(crate) struct KnnNode {
+    pub input: LogicalPlan,
+    pub column: String,
```
wjones127: suggestion: I wonder if this should be `Option<TableReference>, Arc<Field>` as returned by `LogicalPlan::qualified_field()`. Then we could validate that it exists in the input on construction.

rust/lancedb/src/table/datafusion/knn.rs
```diff
@@ -0,0 +1,629 @@
...
+        let k = self.k;
+        let lance_dist: LanceDistanceType = self.distance_type.into();
+        let schema = self.schema();
+
+        let fut = async move {
+            // Buffer all input batches for this partition.
+            let mut batches: Vec<RecordBatch> = Vec::new();
+            while let Some(batch) = input_stream.next().await {
+                batches.push(batch?);
+            }
```
wjones127: issue(blocking): We absolutely do not want to collect all these batches into memory. Imagine that this can be 100GB of data. Instead, you want to store these values in a heap. Use `OwnedRow` from DataFusion to keep the row values around that are relevant. Not that for example we likely don't want to keep the vector column, so we may need to note in an option in the node whether the original vector should be projected or not.

rust/lancedb/src/table/datafusion/knn.rs
```diff
@@ -0,0 +1,629 @@
...
+
+/// Scans one partition and returns the top-k nearest rows for **every** query
+/// vector in `query_vectors`.
+///
+/// Output schema: input columns + `query_index: Int32` + `_distance: Float32`.
+/// Each query vector contributes at most `k` rows, differentiated by
+/// `query_index` (0-based position in `query_vectors`).
+pub(crate) struct PartialKnnExec {
+    input: Arc<dyn ExecutionPlan>,
+    column: String,
```
wjones127: issue: once we get to a physical plan, we should use field indices rather than column names. That way we don't have to pay the lookup cost once per batch.

rust/lancedb/src/table/datafusion/knn.rs
```diff
@@ -0,0 +1,629 @@
...
+                .column(dist_col_idx)
+                .as_primitive::<arrow_array::types::Float32Type>();
+
+            let max_qi = qi_arr.values().iter().copied().max().unwrap_or(-1);
+            if max_qi < 0 {
+                return Ok(RecordBatch::new_empty(schema));
+            }
+
+            // For each query group, pick top-k rows by distance.
+            let mut kept_rows: Vec<u32> = Vec::new();
```
wjones127: suggestion: you should be able to presize this array with capacity for K.

rust/lancedb/tests/knn_test.rs
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+use std::sync::Arc;
+
+use arrow_array::{Array, FixedSizeListArray, Float32Array, Int32Array, RecordBatch, UInt8Array};
+use arrow_schema::{DataType, Field, Schema};
+use lancedb::{DistanceType, Result, connect, table::knn::batch_knn};
+
+fn make_float_batch(num_rows: usize, dim: usize) -> RecordBatch {
```
wjones127: suggestion: I think you should use `lance_datagen` for generating test data instead. There may be some examples in lancedb, but many more examples in lance-format/lance repo.

wjones127:
Superseded by https://github.com/wkalt/lancedb/pull/1 (targeting task/index-analyze branch directly)

PR #3303, opened 2026-04-21
feat: index analyze support

### LINKED ISSUE #3359: Feature: Analyze Index command for IVF_PQ indexes
### SDK

Python

### Description

Have a command to measure recall on a built IVF_PQ index to evaluate it.

wkalt:
Add support for analyze_index for IVF indexes. This will sample N vectors from the dataset, compute exact KNN, and then do a parameter sweep over nprobes and refine_factor to calculate recall + latencies on a range of parameter tunings.

This will allow users to cheaply understand how search parameters affect search quality on their particular dataset.

Closes #3359


rust/lancedb/src/remote/table.rs
```diff
@@ -1925,6 +1925,61 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
...
+                "/v1/table/{}/index/{}/analyze",
+                self.identifier, index_name
+            ))
+            .json(&body);
+
+        let (request_id, response) = self.send(request, true).await?;
+        let response = self.check_table_response(&request_id, response).await?;
+        let bytes = response.bytes().await.err_to_http(request_id.clone())?;
+
+        let mut reader = FileReader::try_new(Cursor::new(bytes), None)?;
```
wjones127, line 1967: suggestion: could we make this an IPC stream and not an IPC file? Then we could stream results to client, which might be nice if it's long-running. Should be done as part of this PR since it would be hard to change later.

rust/lancedb/src/index/analyze.rs
```diff
@@ -0,0 +1,803 @@
...
+                samples.len(),
+                sample_size
+            ),
+        });
+    }
+    Ok(samples)
+}
+
+// `nearest_to` casts to f32 internally, so we do the same for sampled vectors.
+fn vector_values_to_f32(inner: &ArrayRef, col: &str) -> Result<Vec<f32>> {
```
wjones127, line 551: issue(non-blocking): this doesn't seem to be right. We should support u8 (binary) vectors with hamming distance. Can be done as a follow up though.

https://github.com/lancedb/lancedb/blob/6745d438fd8a4d21623865123fb28ed20fd0933c/rust/lancedb/src/table/query.rs#L156-L165

rust/lancedb/src/index/analyze.rs
```diff
@@ -0,0 +1,803 @@
...
+                    out_p90.push(l90);
+                    out_p99.push(l99);
+                    out_max.push(lmax);
+                }
+            }
+        }
+    }
+
+    let schema = analyze_index_schema();
+    RecordBatch::try_new(
```
wjones127, line 323: suggestion(non-blocking): it would be cool to stream out results as they come in. Can be done in the future though.

wjones127 (commented):
This is awesome! Noted

wjones127:
Closing this for now. We'll re-open something for this soon.

PR #3304, opened 2026-04-21
fix: enable hostname verification by default

jackye1995:
## Summary

- make `TlsConfig::default()` enable hostname verification by default
- align the Rust default with the documented Python and Node behavior
- update the Rust unit test to lock in the safe default


wjones127 (approved):
🙌

PR #3309, opened 2026-04-22
test: verify LanceHybridQueryBuilder.where() default is prefilter mode

### LINKED ISSUE #3095: bug(python): LanceHybridQueryBuilder: prefilter logic is inverted when a where filter is applied
### LanceDB version

v0.25.3

### What happened?

When using hybrid search with a where filter, the prefilter argument is silently inverted. Passing prefilter=True actually performs post-filtering, and prefilter=False actually performs pre-filtering.

import lancedb
import pyarrow as pa
import numpy as np

DIM = 8
N = 20

db = lancedb.connect("data/sample-lancedb")

vectors = np.zeros((N, DIM), dtype=np.float32)
vectors[0] = np.full(DIM, 100.0, dtype=np.float32)  # id=1 is far from query

texts = [f"document {i}" for i in range(1, N + 1)]

table = db.create_table(
    "hybrid_prefilter_test",
    pa.table({
        "id": list(range(1, N + 1)),
        "text": texts,
        "vector": pa.array(vectors.tolist(), type=pa.list_(pa.float32(), DIM)),
    }),
    mode="overwrite",
)
table.create_fts_index("text", replace=True)

QUERY_VEC = np.zeros(DIM, dtype=np.float32)
QUERY_TEXT = "test"
FILTER = "id = 1"
LIMIT = 3

"""
Test 1: prefilter=True 
Expected correct behaviour: prefilter applies BEFORE search → id=1 is the only result
"""
results_prefilter_true = (
    .vector(QUERY_VEC)
    .text(QUERY_TEXT)
    .where(FILTER, prefilter=True)
    .limit(LIMIT)
    .to_list()
)

""" expected ≥1 — bug returns 0 """
print(f"prefilter=True: {len(results_prefilter_true)} result(s)")

"""
Test 2: prefilter=False 
Expected correct behaviour: filter applied AFTER search (postfilter) → id=1 not in top-3 → returns 0 results.
"""
results_prefilter_false = (
    .vector(QUERY_VEC)
    .text(QUERY_TEXT)
    .where(FILTER, prefilter=False)
    .limit(LIMIT)
    .to_list()
)

""" expected 0 — bug returns ≥1 """
print(f"prefilter=False: {len(results_prefilter_false)} result(s)")

## Expected output:
prefilter=True:  1 result(s)
prefilter=False: 0 result(s)

## Actual output (buggy):
prefilter=True: 0 result(s)
prefilter=False: 1 result(s)

xodn348:
## Summary

Closes #3095

The root cause of #3095 was that `_create_query_builders()` in `LanceHybridQueryBuilder` passed `self._postfilter` directly to the sub-query `.where()` calls instead of `not self._postfilter`, inverting the prefilter/postfilter semantics. This was already fixed in #3096.

This PR adds a regression test that was missing: it verifies that calling `.where()` with no explicit `prefilter=` argument (the default) correctly pushes the filter into the scan (prefilter mode), rather than applying it as a postfilter.

## What changed

- `python/python/tests/test_hybrid_query.py`: added `test_hybrid_prefilter_default` which asserts that `.where("id = 1")` on a hybrid query produces an explain plan containing `full_filter=id = Int64(1)` (the scan-level pushdown marker), confirming prefilter-by-default behaviour.

## Why this test fills a gap

The existing `test_hybrid_prefilter_explain_plan` only tested explicit `prefilter=True` and explicit `prefilter=False`. It did not test the no-argument default, which is the common user-facing API and the exact scenario described in #3095.

### DIFF
--- a/python/python/tests/test_hybrid_query.py
+++ b/python/python/tests/test_hybrid_query.py
@@ -231,6 +231,27 @@ def test_hybrid_prefilter_explain_plan(table_with_id: Table):
     )
 
 
+def test_hybrid_prefilter_default(table_with_id: Table):
+    plan_default = (
+        table_with_id.search(query_type="hybrid")
+        .vector([0.0, 0.0])
+        .text("dog")
+        .where("id = 1")
+        .limit(2)
+        .explain_plan(verbose=True)
+    )
+
+    assert "full_filter=id = Int64(1)" in plan_default, (
+        f"Default .where() should act as prefilter=True and push filter into the scan."
+        f"\nPlan:\n{plan_default}"
+    )
+
+    assert "full_filter=id = Int64(1)" in plan_default, (
+        f"Default .where() should act as prefilter=True and push filter into the scan."
+        f"\nPlan:\n{plan_default}"
+    )
+
+
 def test_normalize_scores():
     cases = [
         (pa.array([0.1, 0.4]), pa.array([0.0, 1.0])),


xodn348:
Closing duplicate — same change covered in #3308

xodn348:
Closing duplicate — same change covered in #3308

PR #3313, opened 2026-04-25
ci: avoid passing GPG passphrase on command line in Java publish workflow

### LINKED ISSUE #3299: security: avoid passing GPG passphrase on command line in Java publish workflow
`.github/workflows/java-publish.yml:61` passes the GPG passphrase via the command line, which makes it visible in process listings on the runner and risks being echoed into debug logs.

## Proposed

Switch to a passphrase mechanism that doesn't expose the secret as a CLI arg:
- `gpg --pinentry-mode loopback --passphrase-fd 0` with the passphrase piped on stdin, OR
- Maven's `gpg.passphrase` property via environment variable, OR
- `gpg-agent` preload before running Maven

Also worth checking: ensure `-X`/`--debug` is not enabled on Maven in the publish job (would echo secrets).

## Related

Longer-term, consider replacing GPG-based signing with Sigstore via `sigstore-maven-plugin` — removes the need to manage a long-lived GPG key and passphrase at all. That's a bigger change; tracking separately if we pursue it.

Part of security review follow-up.

octo-patch:
Fixes #3299

## Problem

Two security issues exist in `.github/workflows/java-publish.yml`:

1. **`gpg-passphrase` input is misused**: `actions/setup-java`'s `gpg-passphrase` input expects the **name** of an environment variable (default: `GPG_PASSPHRASE`), not the secret value itself. The previous value `${{ secrets.GPG_PASSPHRASE }}` was setting the env var name to the actual secret, which is incorrect.

2. **Passphrase visible on the command line**: `-Dgpg.passphrase=${{ secrets.GPG_PASSPHRASE }}` passes the GPG passphrase as a Maven system property argument, making it visible in process listings and potentially echoed in debug logs — a supply-chain security risk for release workflows.

## Solution

- Fix `gpg-passphrase: MAVEN_GPG_PASSPHRASE` — use the correct env var name so `actions/setup-java` generates a proper Maven `settings.xml` entry that reads from `MAVEN_GPG_PASSPHRASE`.
- Remove `-Dgpg.passphrase=...` from the Maven CLI invocation.
- Add `MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}` to the `env:` block of the Publish step, so the passphrase is available as an environment variable rather than a CLI argument.

## Testing

The Java publish workflow only runs on tag pushes, so this cannot be exercised in a PR build. The logic change is straightforward: `actions/setup-java` is documented to write a `settings.xml` that reads `<gpg.passphrase>` from the named env var, and `maven-gpg-plugin` picks it up from there without any CLI argument.

### DIFF
--- a/.github/workflows/java-publish.yml
+++ b/.github/workflows/java-publish.yml
@@ -43,7 +43,7 @@ jobs:
           server-username: SONATYPE_USER
           server-password: SONATYPE_TOKEN
           gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
-          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          gpg-passphrase: MAVEN_GPG_PASSPHRASE
       - name: Set git config
         run: |
           git config --global user.email "dev+gha@lancedb.com"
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Looks right to me.

Issue #3324, opened 2026-04-26
bug(python): import lancedb crashes with SIGILL on pre-Haswell x86_64 (Sandy Bridge / Ivy Bridge / FX-7500)


tobocop2:
### LanceDB version

0.31.0b9 (every published wheel since the `target-cpu=haswell` baseline was set is affected)

### What happened?

On x86_64 CPUs without AVX2 — Intel Sandy Bridge / Ivy Bridge / Westmere, and AMD Bulldozer / Piledriver / Steamroller (e.g. FX-7500) — `import lancedb` crashes immediately:

```
$ python -c "import lancedb"
[1]  Illegal instruction (core dumped)
```

The Rust extension is compiled with `target-cpu=haswell` + `target-feature=+avx2,+fma,+f16c` (this repo's `.cargo/config.toml`), so AVX2 and FMA instructions are baked into every compiled function. The binary traps on the first AVX2 instruction at load time, before any Python-level lancedb code runs.

Numpy and pyarrow run cleanly on the same hardware via runtime CPU dispatch. On my Sandy Bridge Xeon E5-2609:

| Library | What it ships | Behavior |
|---|---|---|
| pyarrow | `runtime_info.simd_level == 'avx'` | Imports cleanly |
| numpy | `baseline=X86_V2` | Imports cleanly |
| lancedb | no runtime guard, heavy AVX2 | SIGILL |

The root cause lives in lance (which lancedb embeds) — tracked at [lance-format/lance#6618](https://github.com/lance-format/lance/issues/6618). This issue is the lancedb-side companion: even after lance lands runtime dispatch, the lancedb wheel-build config (`.cargo/config.toml`) needs its baseline lowered to actually ship a wheel that runs on the affected hardware.

### Are there known steps to reproduce?

On any pre-Haswell x86_64 Linux box, or in a `qemu-x86_64 -cpu Nehalem` container:

```
pip install lancedb
python -c "import lancedb"
```

Reliably reproduces the SIGILL.

The candidate lancedb-side fix lives at https://github.com/tobocop2/lancedb/pull/2 — small and config-only: bump the lance dep to a runtime-dispatch-aware version, lower this repo's `.cargo/config.toml` baseline to `x86-64-v2` (matches numpy), and add a wheel-import smoke test in CI.

[...]

PR #3327, opened 2026-04-28
feat(wheels): publish lancedb-compat for pre-Haswell x86_64 hosts

tobocop2:
Tracks #3324. On x86_64 CPUs without AVX2 (Sandy Bridge / Ivy Bridge / Westmere on Intel; Bulldozer / Piledriver / Steamroller on AMD), `import lancedb` SIGILLs because the wheel bakes AVX2 + FMA into every compiled function. Per [westonpace's review](https://github.com/lancedb/lancedb/issues/3324#issuecomment-4328944354), the default `lancedb` wheel stays fast; pre-Haswell users get a separately-published `lancedb-compat` wheel.

## Summary

- Adds a `lancedb-compat` matrix entry to `pypi-publish.yml` that builds with `RUSTFLAGS="-C target-cpu=x86-64-v2"` (Nehalem-class baseline). Same Python API (`import lancedb` works) — files install to the same namespace, so the two wheels conflict at install time and users pick one. Same pattern as `psycopg2` / `psycopg2-binary` and `tensorflow` / `tensorflow-cpu`.
- Generalizes `build_linux_wheel` and `upload_wheel` composites with optional `package-name` and `rustflags` inputs (defaults preserve the existing 4 `lancedb` matrix entries verbatim).
- Documents the choice in `python/README.md`: `pip install lancedb-compat` for pre-Haswell hosts.

The default `.cargo/config.toml` baseline is unchanged.

## Sequencing

1. ~~lance-format/lance#6630 merges → runtime SIMD dispatch lands in lance.~~ **Done — merged.**
2. lancedb's lance dep is bumped to a release that includes it (separate PR / normal cadence).
3. This PR's `lancedb-compat` wheel build path starts producing a wheel that runs on pre-Haswell hardware. **Maintainer setup**: register `lancedb-compat` on PyPI and configure trusted publishing.

## Verified end-to-end on Sandy Bridge Xeon E5-2609

Verification was done locally against a fork-pinned lance dep that includes the runtime dispatch implementation, using the same `RUSTFLAGS="-C target-cpu=x86-64-v2"` flags this PR uses in CI:

```
$ RUSTFLAGS="-C target-cpu=x86-64-v2" maturin build --release
$ pip install ./target/wheels/lancedb-*.whl
$ python verify.py
PASS: import + simd dispatch + table create + vector search all work.
```

Pre-fix on the same CPU (default `pip install lancedb`): `Illegal instruction (core dumped)`. Full reproducer (deps + clone + build + verification): https://gist.github.com/tobocop2/2e341358b55c143527416edfdb1e37df. Fork-internal verification PR with the dep bump and full logs: [`tobocop2/lancedb#2`](https://github.com/tobocop2/lancedb/pull/2).

## Benchmarks — no regressions on modern CPUs from the lance-side change

These are the numbers I ran for the lance PR, confirming the runtime dispatch doesn't slow down the default (`target-cpu=haswell`) wheel that existing users install. Criterion, one machine, one session, base → PR, no `RUSTFLAGS` override. Full methodology, null experiments, and logs: [lance-format/lance#6630 benchmark comment](https://github.com/lance-format/lance/pull/6630#issuecomment-4933063394) and the [logs gist](https://gist.github.com/tobocop2/3c6d0f449cbd736aa2501f89a7fe56a2).

| benchmark | EPYC 7B13 (`avx2`, `fma`, no `avx512f`) | Xeon Cascade Lake (`avx512f`) |
|---|---|---|
| `Cosine(f32, scalar)` *(control)* | +0.04% | +0.09% |
| `Cosine(f64, scalar)` | −0.34% | −1.94% |
| `Cosine(u8, SIMD)` | +2.30% | +3.63% |
| `Dot(f16, SIMD)` | −0.58% | +0.61% |
| `Dot(f32, SIMD)` | +0.34% | **−6.08%** |
| `Dot(f32, arrow_arity)` | +0.02% | −0.00% |
| `L2(f32, scalar)` | −0.10% | −0.02% |
| `L2(f32, simd)` (dim 1024) | +2.63% | −0.53% |
| **`L2(simd,f32x8)` (dim 8)** | **−45.9%** | **−25.1%** |
| `L2(u8, SIMD)` | +0.42% | −3.11% |
| `NormL2(f32, SIMD)` | −1.02% | −4.17% |
| `NormL2(f64, SIMD)` | +3.51% | −0.58% |

Nothing regresses beyond the noise floor. Dim 8 — the PQ sub-vector width — improves 25–46%.

---

To be transparent: this isn't my domain of expertise and the lance-side implementation is AI-generated. I verified it works end-to-end on the failing hardware. Happy to roll in feedback.


tobocop2:
friendly bump @wjones127 @westonpace — lance-format/lance#6630 has been merged so this is now ready for review and I'm taking it out of draft.

I updated the PR description to now have the benchmarks from the lance work as well

.github/workflows/pypi-publish.yml
```diff
@@ -31,11 +31,28 @@ jobs:
...
+            rustflags: ""
+          # `lancedb-compat`: pre-Haswell-friendly variant for x86_64 hosts
+          # without AVX2 (Sandy Bridge / Ivy Bridge / Westmere on Intel,
+          # Bulldozer / Piledriver / Steamroller on AMD). Compiled at the
+          # `x86-64-v2` baseline; runtime SIMD dispatch in lance-linalg
+          # picks the appropriate tier (scalar / AVX / AVX+FMA / AVX2+FMA
+          # / AVX-512) at load time. Same import as `lancedb` -- conflicts
+          # at install time, so users pick one.
+          - platform: x86_64
+            manylinux: "2_17"
```
wjones127: suggestion: we dropped support for 2_17, as have most projects. Can we use 2_28 instead? (Might need to change the `name` of the job too to prevent collisions.) The purposes of this is to support outdated hardware, not software, right?

wjones127 (changes requested):

.github/workflows/pypi-publish.yml
tobocop2: that's right, I'll update it now.

tobocop2: Updated in 2b7be76c.

tobocop2: Also, this exposed a bug. 
  
  I updated the compat wheel artifact name. it was wheels-linux-{platform}-{manylinux}, which collides with the default. 
  
  
  It now includes the package name, same as the job name. 
  
  Also found an issue where beta tags would have built the compat wheel but skipped publishing it.

tobocop2:
One thing here needs maintainer setup: `lancedb-compat` needs a PyPI trusted publisher before the first release tag after this lands:  a [pending publisher](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/)

Without it the tag publishes the `lancedb` wheels normally and then 403s on the compat wheel: a red release job, nothing broken for users. Fury needs nothing extra.

I also fixed a bug of my own in 7277917: the empty `rustflags` default was reaching the container as `-e RUSTFLAGS=`, and cargo treats `RUSTFLAGS` as an override even when it is empty, so it was quietly dropping `-Ctarget-cpu=haswell` from the defaul wheel. It is now only set when non-empty, so the stock wheels build exactly as they do on main.

wjones127 (approved):

PR #3335, opened 2026-04-29
fix(python): make Permutation picklable for PyTorch multiprocessing

westonpace:
## Summary

When pytorch is used with multiprocessing and the mp mode is spawn then the Permutation needs to be pickled.  It could not be pickled because `Table` and `Connection` are not serializable.  This PR adds pickle support to Permutation without adding general pickle support to `Table` or `Connection`.  To add general support we probably need to start by adding serialization in the namespace client.

In the meantime this PR enable pickling by adding special cases for:

 * In-memory tables (just serialize as Arrow IPC)
 * Native tables (serialize the URI)

If a user is not using one of the above cases (e.g. using a remote connection) then they will need to provide a connection factory that can be pickled.

## Breaking change

`PermutationBuilder.persist(...)` is removed from the Python bindings; the permutation table is now always in-memory. The underlying Rust `PermutationBuilder::persist` API is untouched and can be re-exposed later if needed.  It probably won't make sense to do that until we have a way to serialize `Table` and `Connection`.


python/python/lancedb/permutation.py
```diff
@@ -51,15 +47,6 @@ def __init__(self, table: LanceTable):
...
 
-    def persist(
-        self, database: Union[DBConnection, AsyncConnection], table_name: str
-    ) -> "PermutationBuilder":
-        """
-        Persist the permutation to the given database.
-        """
-        self._async.persist(database, table_name)
-        return self
-
```
westonpace, line 62: This is a breaking change.  We're just removing this capability for now.  We could add it back in the future but we'd have to think about it some more to figure out how we can pickle this.

python/python/lancedb/permutation.py
```diff
@@ -365,6 +354,20 @@ def impl(batch: pa.RecordBatch) -> pl.DataFrame:
...
+    sink = pa.BufferOutputStream()
+    with pa.ipc.new_stream(sink, table.schema) as writer:
+        writer.write_table(table)
+    return sink.getvalue().to_pybytes()
+
+
+def _deserialize_arrow_table(data: bytes) -> pa.Table:
+    """Read an Arrow IPC stream back into a pyarrow Table."""
+    with pa.ipc.open_stream(pa.BufferReader(data)) as reader:
+        return reader.read_all()
```
wjones127: note: I had thought `pa.Table` was already pickle-able.

wjones127 (approved):
Looks good. I think the approach works. I think we can later make Connection and Table pickle-able. I think instead of a `connection_factory`, for those we'll just need an optional `credential_factory`. Otherwise we can just reconstruct stuff from the pickled data.

python/python/lancedb/permutation.py
westonpace: Good catch.  Removed these helpers.

PR #3336, opened 2026-04-30
feat(python): add table serialization

jackye1995:
Adds Table.serialize() and lancedb.deserialize_table() so Python table handles can be reopened from serialized connection and table identity, including namespace paths and checked-out versions.


jackye1995:
Closing this for now. Geneva is staying on its own TableReference serialization path and no longer needs LanceDB table serialization for this namespace unification work.

PR #3337, opened 2026-04-30
docs: note that `namespace_path` and `namespace_client` used to be called `namespace`

dantasse:
Hey all! Kind of in a weird spot in our docs; right now the docs say to use `namespace_path` and `namespace_client` for these params, but that doesn't work; latest released non-beta lancedb (0.30.2) still has `namespace` for both of them, so code trying to use this doesn't work. This PR would update all our docstrings so you can at least figure out to use `namespace` if you're on an old version of lancedb.

Adding Jack author of https://github.com/lancedb/lancedb/pull/3205, plus Prashanth as arbiter of taste in docs. I'm not super tied to this; it fixes the issue but is kind of noisy and will be sort of moot as soon as we release 0.31. So I'm happy to merge this, or just accept the inconsistency for now, whatever you all think.

### DIFF
--- a/python/python/lancedb/db.py
+++ b/python/python/lancedb/db.py
@@ -79,6 +79,7 @@ class DBConnection(EnforceOverrides):
         namespace_path: List[str], default []
             The parent namespace to list namespaces in.
             Empty list represents root namespace.
+            Previously called ``namespace`` in 0.30.2 and earlier.
         page_token: str, optional
             Token for pagination. Use the token from a previous response
             to get the next page of results.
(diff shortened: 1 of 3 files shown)


docs/src/python/python.md
```diff
@@ -92,14 +92,6 @@ of raw SQL strings with [where][lancedb.query.LanceQueryBuilder.where] and
 
 ::: lancedb.context.Contextualizer
 
-## Full text search
```
dantasse, line 95: ok, actually, we should merge this change whether we merge the whole PR or not, because while this code is still here referencing a nonexistent file, the docs don't build.

dantasse: oh this is just the same fix as #3328

prrao87:
@dantasse have you tested this with agents? Do the new docstrings help avoid the error you described? Overall LGTM

docs/src/python/python.md
dantasse: and this has since been fixed, reverting this

prrao87 (approved):

dantasse:
Closing this now that 0.32.0 is out and `namespace_path` and `namespace_client` are the correct param names.

PR #3340, opened 2026-05-01
fix(python): add missing space in hybrid query error message

n1teshy:
Hi, the hybrid query error message looks like it can use a space, just added it.

```python
def _validate_query(self, query, vector=None, text=None):
    if query is not None and (vector is not None or text is not None):
        raise ValueError(
            "You can either provide a string query in search() method"
            "or set `vector()` and `text()` explicitly for hybrid search."
            "But not both."
        )
```


westonpace (approved):

westonpace:
Thanks!

PR #3346, opened 2026-05-04
perf: skip Dataset::index_statistics() for all index types

wjones127:
`Dataset::index_statistics()` loads index files and does meaningful CPU work to serialize low-level info. Most fields `NativeTable::index_stats()` needs are available from manifest metadata via `Dataset::describe_indices()`, which is much cheaper.

`NativeTable::index_stats()` now:

- Calls `describe_indices()` filtered by name; returns `Ok(None)` if no match.
- Parses `distance_type` from `description.details()` JSON (the `VectorIndexDetails` proto stored in the manifest by recent Lance versions).
- Falls back to `index_statistics()` only for vector indices where `details()` returns no `distance_type` — this handles older Lance datasets that didn't write `VectorIndexDetails`.
- `Unknown` index types (e.g. Lance's internal `FragReuseIndex`) are explicitly filtered out of `list_indices` rather than erroring.

## Test plan
- [x] `test_create_scalar_index` — asserts `index_type`, `distance_type`, and `num_unindexed_rows > 0` after adding rows post-index
- [x] `test_create_fm_index`, `test_create_bitmap_index`, `test_create_label_list_index` — added `index_stats` assertions
- [x] IvfPq, IvfHnswPq, IvfHnswSq, IvfHnswFlat tests assert `distance_type == Some(L2)`
- [x] `test_list_indices_skip_frag_reuse` — FragReuseIndex is filtered by the Unknown guard in `list_indices`

### DIFF
--- a/rust/lancedb/src/index.rs
+++ b/rust/lancedb/src/index.rs
@@ -317,6 +317,8 @@ pub enum IndexType {
     // FTS
     #[serde(alias = "INVERTED", alias = "Inverted")]
     FTS,
+    /// Catch-all for index types not recognized by this version of LanceDB.
+    Unknown,
 }
 
 impl std::fmt::Display for IndexType {
@@ -334,6 +336,7 @@ impl std::fmt::Display for IndexType {
             Self::LabelList => write!(f, "LABEL_LIST"),
             Self::Fm => write!(f, "FM"),
             Self::FTS => write!(f, "FTS"),
+            Self::Unknown => write!(f, "UNKNOWN"),
         }
     }
 }
@@ -355,9 +358,7 @@ impl std::str::FromStr for IndexType {
             "IVF_HNSW_PQ" => Ok(Self::IvfHnswPq),
             "IVF_HNSW_SQ" => Ok(Self::IvfHnswSq),
             "IVF_HNSW_FLAT" => Ok(Self::IvfHnswFlat),
-            _ => Err(Error::InvalidInput {
-                message: format!("the input value {} is not a valid IndexType", value),
-            }),
+            _ => Ok(Self::Unknown),
         }
     }
 }
@@ -427,8 +428,7 @@ pub(crate) struct IndexMetadata {
     pub metric_type: Option<DistanceType>,
 }
 
-// This struct is used to deserialize the JSON data returned from the Lance API
-// Dataset::index_statistics().
+// Deserializes the JSON returned by Dataset::index_statistics().
 #[skip_serializing_none]
 #[derive(Debug, Deserialize)]
 pub(crate) struct IndexStatisticsImpl {
--- a/rust/lancedb/src/index.rs
+++ b/rust/lancedb/src/index.rs
@@ -425,8 +425,6 @@ pub struct IndexConfig {
 #[derive(Debug, Deserialize)]
 pub(crate) struct IndexMetadata {
     pub metric_type: Option<DistanceType>,
-    // Sometimes the index type is provided at this level.
-    pub index_type: Option<IndexType>,
 }
 
 // This struct is used to deserialize the JSON data returned from the Lance API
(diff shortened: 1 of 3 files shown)


rust/lancedb/src/index.rs
```diff
@@ -425,20 +426,15 @@ pub struct IndexConfig {
 #[derive(Debug, Deserialize)]
 pub(crate) struct IndexMetadata {
     pub metric_type: Option<DistanceType>,
-    // Sometimes the index type is provided at this level.
-    pub index_type: Option<IndexType>,
```
westonpace, line 429: Nice!  I'm hoping we can get to a world where that enum can just go away.

rust/lancedb/src/table.rs
```diff
@@ -2930,17 +2932,21 @@ impl BaseTable for NativeTable {
...
+                    // Internal or future index types that this version doesn't recognize
+                    // (e.g. Lance's internal FragReuseIndex) are silently excluded from
+                    // the user-visible index listing.
+                    log::debug!(
+                        "Skipping unrecognized index '{}' (type '{}') in list_indices",
+                        idx_desc.name(),
+                        idx_desc.index_type(),
+                    );
+                    return None;
+                }
```
westonpace, line 2949: This is fine for now (it doesn't make things worse) but we should not rely on `IndexType`.  It will not be present once we move to plugins.  So at some point we need to get rid of it in this path.

rust/lancedb/src/table.rs
```diff
@@ -3007,40 +3013,83 @@ impl BaseTable for NativeTable {
...
+        // VectorIndexDetails in the manifest carries distance_type for indices written
+        // by recent Lance versions. For older datasets that didn't write those details
+        // we fall back to index_statistics() for vector index types.
+        let dataset = self.dataset.get().await?;
+
+        let mut descriptions = dataset
+            .describe_indices(Some(IndexCriteria::default().with_name(index_name)))
+            .await?;
+        let Some(description) = descriptions.pop() else {
+            return Ok(None);
```
westonpace, line 3026: We already have a lot of methods on dataset but it does seem like a `describe_index` helper might be nice.

rust/lancedb/src/table.rs
```diff
@@ -3007,40 +3013,83 @@ impl BaseTable for NativeTable {
...
         };
 
-        let mut stats: IndexStatisticsImpl =
-            serde_json::from_str(&stats).map_err(|e| Error::InvalidInput {
-                message: format!("error deserializing index statistics: {}", e),
-            })?;
+        let index_type: crate::index::IndexType = description
+            .index_type()
+            .parse()
+            .unwrap_or(crate::index::IndexType::Unknown);
```
westonpace, line 3032: How do you imagine this working once we move to a world of plugins?  We can't rely on an enum of all possible index types.  Is the thought that `index_type` will just always be unknown in those cases and users should use `describe_indices` to get the true type (the type url in this case)?

westonpace (approved):
Just one question on where this is going.  `index_stats` currently reports `index_type`.  However, once we move to a plugin model, there is no similar concept (we don't have a single enum of all possible indexes anymore).  How do you imagine that evolving?

Does `index_type` just become "unknown" in the `index_stats` method and users know they can get the true type from the `describe_indices` method?

Xuanwo (approved):
Nice change! Sharing the same concern with @westonpace, others LGTM.

rust/lancedb/src/table.rs
wjones127: Understood. IIRC we had an idea that index type could be a string, and be derived from the protobuf definition. Is that still an idea floating around? I get that the enum will go away.

wjones127: Happy to make a follow up issue for this.

wjones127: https://github.com/lancedb/lancedb/issues/3664

PR #3354, opened 2026-05-07
feat(table): route merge_insert through the MemWAL LSM write path

touch-of-grey:
## Summary

When an `LsmWriteSpec` is installed on a table (#3396), `merge_insert` upsert
calls are dispatched through Lance's MemWAL `ShardWriter` (LSM-style append)
instead of the standard merge path.

- **`use_lsm_write`** — a `merge_insert` builder option, default `true`; set it
  `false` to use the standard path for a call even when a spec is set.
- **`assume_pre_sharded`** — a `merge_insert` builder option, default `false`;
  skips the per-row shard check and routes by the first row only.
- **`close_lsm_writers`** — drains and closes the table's cached MemWAL shard
  writers.
- The `merge_insert` **`on`** columns default to, and are validated against,
  the table's unenforced primary key.
- Shard writers are cached alongside the dataset (in
  `DatasetConsistencyWrapper`) and reused for the session.
- `MergeResult` gains **`num_rows`** — on the LSM path the insert/update
  breakdown is unknown until compaction, so only the total is reported.

Routing covers all three sharding strategies — bucket (murmur3,
Iceberg-compatible), identity, and unsharded. Each `merge_insert` call targets
a single shard; the whole input is collected and validated before a single
atomic `ShardWriter::put`, so a validation failure leaves the MemWAL untouched.

Bindings: Python (`merge_insert(...).use_lsm_write(...)` /
`.assume_pre_sharded(...)`, `Table.close_lsm_writers`) and TypeScript
(`mergeInsert(...).useLsmWrite(...)` / `.assumePreSharded(...)`,
`Table.closeLsmWriters`).

## Context

Reconstructed from the original #3354 branch onto current `main`: the branch
predated the #3394 (unenforced primary key) / #3396 (`LsmWriteSpec`) split and
has been rebuilt on that merged foundation. Depends on Lance `v7.0.0-beta.13`.

The MemWAL read path (reading un-flushed shard data back into queries) and
remote (LanceDB Cloud) LSM support are follow-ups.


touch-of-grey:
@jackye1995 PTAL — this is the LanceDB-side companion to lance-format/lance#6706 (now released as v7.0.0-beta.6).

nodejs/src/table.rs
```diff
@@ -790,6 +790,7 @@ pub struct MergeResult {
     pub num_updated_rows: i64,
     pub num_deleted_rows: i64,
     pub num_attempts: i64,
+    pub num_inserted_or_updated_rows: i64,
```
jackye1995: probably better to just do num_rows, becuase we will eventually also allow deletion

python/python/lancedb/remote/table.py
```diff
@@ -651,6 +651,18 @@ def alter_columns(
     def drop_columns(self, columns: Iterable[str]) -> DropColumnsResult:
         return LOOP.run(self._table.drop_columns(columns))
 
+    def set_unenforced_primary_key(self, column: str) -> None:
```
jackye1995: we should support multiple columns for this API. it is okay to check in merge_insert and say that for initial implementation of lsm write, we only support single column primary key table.

python/src/table.rs
```diff
@@ -167,6 +169,56 @@ impl From<lancedb::table::MergeResult> for MergeResult {
...
+
+    #[getter]
+    pub fn num_buckets(&self) -> u32 {
+        self.num_buckets
+    }
+}
+
+impl From<MergeInsertSpec> for lancedb::table::MergeInsertSpec {
+    fn from(spec: MergeInsertSpec) -> Self {
+        Self::Bucket {
```
jackye1995: I think we still want to also surface this to python, so this is of type `bucket`. We should not just omit this in python layer just because it is the only supported one right now.

rust/lancedb/src/table/merge/lsm.rs
```diff
@@ -0,0 +1,824 @@
...
+/// the writer, so the entry wraps it in [`RwLock<Option<ShardWriter>>`] —
+/// `put` takes a read lock on the option, while `close` takes a write
+/// lock and replaces the option with `None`.
+#[derive(Default, Debug)]
+#[allow(clippy::redundant_pub_crate)]
+pub(crate) struct ShardWriterCache {
+    /// `Some((bucket, entry))` once a writer has been opened. Only one
+    /// bucket is cached at a time. Switching buckets requires draining
+    /// this slot via `drain_and_close()` first.
+    slot: RwLock<Option<(u32, Arc<ShardWriterEntry>)>>,
```
jackye1995: I think we should only support a single ShardWriter to cache. we should not auto split data, instead we should just ask user to split record batches upstream and then use multiple lancedb processes to write each shard. This probably fits the actual application better anyway right? Because typically the upstream engine e.g. Spark/Flink already does this for you.

rust/lancedb/src/table/merge/lsm.rs
```diff
@@ -0,0 +1,824 @@
...
+// Single-bucket validation
+// =============================================================================
+
+/// Confirm that every row in `batch` hashes to the same bucket under
+/// Murmur3-x86-32 (seed 0, Iceberg-compatible) modulo `num_buckets`, and
+/// return that bucket index. Errors if rows split across multiple buckets,
+/// the primary key column is missing, or any primary key value is null.
+///
+/// Empty batches return `None`; the caller should treat them as a no-op.
+fn validate_single_bucket(
```
jackye1995: yeah, looks like you already are doing what I was saying above, so there is further no need to keep a map of shard writers.

rust/lancedb/src/table/merge/lsm.rs
jackye1995, line 1: should probably call this module lsm.rs

rust/lancedb/src/table.rs
```diff
@@ -224,6 +224,20 @@ pub trait Tags: Send + Sync {
...
+/// Today the only supported variant is [`MergeInsertSpec::Bucket`], which
+/// hash-partitions input by the unenforced primary key into a fixed number of
+/// buckets and routes each bucket through an independent MemWAL `ShardWriter`.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum MergeInsertSpec {
+    /// Hash-bucket partitioning by the unenforced primary key column.
+    ///
+    /// `column` must equal the table's currently-set unenforced primary key.
+    /// `num_buckets` must be in `[1, 1024]`.
+    Bucket { column: String, num_buckets: u32 },
```
jackye1995: looks like we are missing the idea of no partitioning in https://lance.org/format/table/mem_wal/#shard-expression. Or I guess we could use an expression that is basically a constant like `1`. I would suggest we add this to the MemWAL spec as `unsharded`.

jackye1995 (commented):
thanks for the work! Added some comment for the initial pass. Would be great to also get some performance numbers from this.

nodejs/src/table.rs
touch-of-grey: Renamed to `num_rows` so the same field can carry delete counts in the future.

python/python/lancedb/remote/table.py
touch-of-grey: `set_unenforced_primary_key` now accepts an ordered list of columns; the bucket `merge_insert` spec still rejects multi-column PKs but the table-level API does not.

python/src/table.rs
touch-of-grey: Added `spec_type` discriminator getter ("bucket" or "unsharded"); `column` and `num_buckets` are now `Optional` and only populated for bucket.

rust/lancedb/src/table/merge/lsm.rs
touch-of-grey: Confirmed: single ShardWriter cache, no auto-split, caller pre-shards. Already implemented this way.

rust/lancedb/src/table/merge/lsm.rs
touch-of-grey: Right — there's no shard-writer map, just a single-slot cache.

rust/lancedb/src/table/merge/lsm.rs
touch-of-grey: Renamed module `merge/shard.rs` → `merge/lsm.rs` (preserved git history via rename).

rust/lancedb/src/table.rs
touch-of-grey: Added `MergeInsertSpec::Unsharded` variant. Persists a ShardSpec with transform="unsharded" and num_shards=1; merge_insert dispatches all rows to bucket 0 and skips per-row hashing. Worth proposing the named transform on the Lance side too.

touch-of-grey:
## Benchmark — c7i.12xlarge / S3 us-east-1

Bench script: [`bench/bench_merge_insert.py`](https://github.com/touch-of-grey/lancedb/blob/MergeInsertShardWriter/bench/bench_merge_insert.py) (in this branch's analysis folder; not part of the merge). 100k-row seed table, single-process measurement.

### large (batch_rows=10000, calls=20)
| mode | rows/s | p50 ms | p90 ms | p99 ms |
|---|---:|---:|---:|---:|
| baseline | 15,165 | 593.75 | 1224.86 | 1305.83 |
| lsm_unsharded | **60,980** (4.02×) | 139.49 | 198.99 | 472.16 |
| lsm_bucket(N=4) | **34,526** (2.28×) | 170.40 | 247.40 | 336.65 |

### medium (batch_rows=1000, calls=50)
| mode | rows/s | p50 ms | p90 ms | p99 ms |
|---|---:|---:|---:|---:|
| baseline | 1,065 | 876.08 | 1577.91 | 1738.61 |
| lsm_unsharded | **7,334** (6.88×) | 128.45 | 181.44 | 368.13 |
| lsm_bucket(N=4) | **7,194** (6.75×) | 115.51 | 167.90 | 274.04 |

### small (batch_rows=100, calls=50)
| mode | rows/s | p50 ms | p90 ms | p99 ms |
|---|---:|---:|---:|---:|
| baseline | 107 | 882.70 | 1462.01 | 2125.26 |
| lsm_unsharded | **769** (7.16×) | 122.38 | 170.10 | 264.45 |
| lsm_bucket(N=4) | **714** (6.66×) | 128.08 | 195.87 | 246.67 |

**Headline:** the LSM path is **2.3× to 7.2× faster** than baseline merge_insert across batch sizes, with much tighter per-call latency (~120-170 ms p50 vs. 600-880 ms baseline). The win comes from skipping the hash-join scan against the target table — the LSM path is just a WAL append and one durable-write round-trip per call.

bucket(N=4) trails unsharded slightly (per-row Murmur3 hash + single-bucket validation overhead), but enables the multi-process scale-out story by letting independent processes own different buckets.

Caveats: single-process measurement, modest seed size — the gap should widen with larger seeds because baseline scales with target size while LSM append cost is independent of it.

Full report + raw JSON: `~/ai/analysis/lancedb/MergeInsertShardWriter/expose-mem-wal-shard-writer-in-merge-insert/bench/`.

touch-of-grey:
## Bench addendum — 1024-dim vector schema + HNSW maintained

Re-ran the same workload against an `id: utf8`, `v: int64`, `vec: FixedSizeList<float32, 1024>` schema, in two variants. Same EC2 host / region / build as the prior round (c7i.12xlarge, us-east-1, `--release`).

### (1) Vector schema, no maintained index

| sweep | baseline | lsm_unsharded | lsm_bucket(N=4) |
|-------|---------:|--------------:|----------------:|
| large (10k rows/call) | 6,598 rows/s | **20,357** (3.09×) | 11,593 (1.76×) |
| medium (1k rows/call) | 1,024 rows/s | **4,781** (4.67×) | 4,004 (3.91×) |
| small (100 rows/call) | 108 rows/s | 599 (5.52×) | **677** (6.25×) |

The LSM advantage is ~25-30% smaller than on the string-only schema (was 4-7×, now 3-6×) — both paths now spend more time on vector payload I/O so the relative win shrinks.

### (2) Vector schema + IVF_HNSW_PQ maintained on `vec`

`MergeInsertSpec.bucket(...).with_maintained_indexes(["vec_idx"])` (and the same for `unsharded`). MemWAL keeps an HNSW-style in-memory copy synced to the writes.

| sweep | lsm_unsharded+hnsw | lsm_bucket(N=4)+hnsw |
|-------|-------------------:|---------------------:|
| medium (1k rows/call) | 646 rows/s | 659 rows/s |
| small (100 rows/call) | 628 rows/s | 608 rows/s |

Throughput plateaus at ~600-650 rows/s once HNSW maintenance kicks in — consistent with ~1.5 ms/row HNSW insertion cost on 1024-dim with `M=20` and a 100k-vertex graph (~M·log₂N ≈ 240 distance comps per insertion).

### Cost of maintaining HNSW vs. not (lsm_unsharded only)

| sweep | no maintained index | +hnsw | overhead |
|-------|--------------------:|------:|---------:|
| small | 599 rows/s | 628 rows/s | ~0 (WAL flush dominates) |
| medium | 4,781 rows/s | 646 rows/s | **86% slower** |

For small batches the HNSW work fits within the WAL-flush window. For larger batches HNSW becomes the bottleneck. The trade-off is intentional: in exchange for a 7× write slowdown at medium batches, the index stays queryable immediately — no separate `optimize` / index rebuild needed for the freshly-written rows.

Full numbers + raw JSON in this branch's analysis folder (`bench/REPORT-vec.md`, `bench/results-vec/`). EC2 host (`lsm-bench-vec`, `i-0f906316...`) terminated.

touch-of-grey:
## Bench addendum #2 — small batches + checked-vs-unchecked + HNSW

Sweep at `batch_rows ∈ {1, 10, 20, 50, 100}` on the same 1024-dim vector schema. Adds two new factors:
- `lsm_bucket_unchecked` — `MergeInsertSpec.bucket(...).assume_pre_sharded()`. Skips the per-row Murmur3 check; only the first row picks the writer.
- HNSW maintained — `IVF_HNSW_PQ` (num_partitions=1) listed in `with_maintained_indexes(...)`.

### Throughput (rows/s) — no maintained index

| batch | baseline | lsm_unsharded | lsm_bucket | lsm_bucket/unchecked |
|------:|---------:|--------------:|-----------:|---------------------:|
|   1 |   0.4 |   9.3 |   9.6 |   **9.6** |
|  10 |   3.6 |  82.8 |  83.2 |  **87.4** |
|  20 |   6.9 | 165.5 | 163.9 | 162.0 |
|  50 |  32.3 | 386.0 | 413.7 | **434.3** |
| 100 |  64.6 | 713.7 | 686.5 | **751.6** |

### Throughput (rows/s) — HNSW maintained

| batch | unsharded+hnsw | bucket+hnsw | bucket/unchecked+hnsw |
|------:|---------------:|------------:|----------------------:|
|   1 |   9.8 |   9.1 |   9.0 |
|  10 |  89.9 |  82.6 |  91.0 |
|  20 | 154.8 | 181.6 | 156.4 |
|  50 | 430.9 | 439.6 | 377.1 |
| 100 | 731.0 | 683.4 | 715.9 |

### Three things to call out

1. **Baseline collapses at small batches** (~0.4-65 rows/s) — every call does a full hash-join scan against the seed table. LSM is **20-100× faster** at small batches. p50 baseline is 1.5-3 s; p50 LSM stays at ~100-140 ms across all batch sizes — one durable S3 PUT.

2. **You're right that bucket=4 ≈ unsharded for single-bucket targeting.** The three LSM variants land within ~5% of each other across the sweep. `assume_pre_sharded()` shows a 5-10% gain at batch≥50 without HNSW (where per-row Murmur3 finally accumulates), and **no measurable difference under HNSW** because the per-row HNSW update dominates over the few hundred ns of hashing.

3. **HNSW maintenance is essentially free at small batches**, contradicting the earlier `medium=1000 rows/call` result:

    | batch | no-idx unsharded | +hnsw | overhead |
    |------:|-----------------:|------:|---------:|
    |  1 |   9.3 |   9.8 | — |
    | 10 |  82.8 |  89.9 | — |
    | 20 | 165.5 | 154.8 | ~6% |
    | 50 | 386.0 | 430.9 | — |
    |100 | 713.7 | 731.0 | — |

    The HNSW per-row cost only shows up when batches get into the thousands of rows. **In the small-batch streaming regime LSM is built for, you get a continuously-updated HNSW index at no measurable write-throughput cost.**

Full report + raw JSON: `bench/REPORT-sm.md`, `bench/results-sm/`. EC2 host (`lsm-bench-sm`, `i-02c9e016...`) terminated.

jackye1995:
To reduce complexity, let's split the unenforced primary key part and sharding spec part so that they are separated PRs.

rust/lancedb/src/table/merge.rs
```diff
@@ -150,6 +164,35 @@ impl MergeInsertBuilder {
         self
     }
 
+    /// Controls whether `merge_insert` uses the MemWAL LSM write path when an
+    /// LSM write spec is installed on the table.
+    ///
+    /// When `true` (the default), a `merge_insert` on a table with an
```
jackye1995: this does not really default to true, it should default to None, which means the table that has the LsmWriteSpec set will use lsm write, ones without will not use it.

setting it to true for a table without LsmWriteSpec is not meaningful, and we should return an error saying to set it first.

rust/lancedb/src/table/merge.rs
```diff
@@ -150,6 +164,35 @@ impl MergeInsertBuilder {
...
+    pub fn use_lsm_write(&mut self, use_lsm_write: bool) -> &mut Self {
+        self.use_lsm_write = use_lsm_write;
+        self
+    }
+
+    /// Declares that the input of an LSM `merge_insert` is already sharded.
+    ///
+    /// When a table has an LSM write spec, every row in a single `merge_insert`
+    /// call must route to the same shard. By default every row is inspected to
+    /// verify this. When set to `true`, only the first row of each batch is
```
jackye1995: we should not even inspect the first row of a batch, we should just inspect the first row to know the shard, and then just keep using it.

I think the term "validate_single_shard" is better than "assume_pre_sharded". the default should be validate_single_shard=true.

In addition, we should allow
1. `for_shard`: setting the shard field values to avoid the initial row probe
2. `using_shard_writer`: setting the actual shard writer cache if caller want to fully reuse a shard writer.

2 is not necessary to be exposed in python and typescript.

rust/lancedb/src/table/merge/lsm.rs
```diff
@@ -92,10 +141,856 @@ pub(crate) async fn unset_lsm_write_spec(table: &NativeTable) -> Result<()> {
...
+///
+/// Held by [`DatasetConsistencyWrapper`](crate::table::dataset::DatasetConsistencyWrapper)
+/// so writers live where the dataset lives — cached for the session and shared
+/// across `merge_insert` calls. `ShardWriter::put` takes `&self`, so concurrent
+/// puts on a cached writer are safe; `close` consumes the writer, so each entry
+/// wraps it in `RwLock<Option<ShardWriter>>`.
+#[derive(Default)]
+#[allow(clippy::redundant_pub_crate)]
+pub(crate) struct ShardWriterCache {
+    writers: RwLock<HashMap<Uuid, Arc<ShardWriterEntry>>>,
```
jackye1995: we are only handling a single shard, it should not need a map.

rust/lancedb/src/table/dataset.rs
```diff
@@ -18,6 +19,10 @@ use crate::{Error, error::Result, utils::background_cache::BackgroundCache};
 pub struct DatasetConsistencyWrapper {
     state: Arc<Mutex<DatasetState>>,
     consistency: ConsistencyMode,
+    /// MemWAL `ShardWriter` cache, co-located with the dataset so open writers
+    /// are cached for the session and share the dataset's lifecycle. Shared by
+    /// `Arc` across clones of the wrapper.
+    shard_writers: Arc<ShardWriterCache>,
```
jackye1995: As said in the comment above, there should only be a single shard writer for this dataset writing to a single shard. It should not allow multiple writers to multiple shards.

rust/lancedb/src/table/merge.rs
touch-of-grey: Renaming `assume_pre_sharded` → `validate_single_shard` (default `true`) in this PR. `for_shard` and `using_shard_writer` will be added as separate follow-up PRs.

rust/lancedb/src/table/merge.rs
touch-of-grey: Done — `use_lsm_write` is now `Option<bool>` defaulting to `None`: unset routes through the LSM path only when a spec is set, `Some(false)` forces the standard path, and `Some(true)` errors when no LSM write spec is installed.

rust/lancedb/src/table/merge/lsm.rs
touch-of-grey: Done — `ShardWriterCache` now holds a single writer slot instead of a map. A dataset writes to one shard at a time; routing a `merge_insert` to a different shard errors until the current writer is closed via `close_lsm_writers`.

rust/lancedb/src/table/dataset.rs
touch-of-grey: Done — the wrapper holds a single `shard_writer` (renamed from `shard_writers`), backed by the single-slot cache. Multiple writers to multiple shards are not allowed.

jackye1995 (approved):
I think we want to actually move shard routing to lance, but that can be done separately. thanks for the work @touch-of-grey !

PR #3369, opened 2026-05-12
feat(python)!: align `Permutation.with_format("torch")` with HuggingFace `set_format("torch")`

omkar-334:
Closes #3245.

> **BREAKING CHANGE:** `with_format("torch")` no longer returns a list of stacked row tensors. It now returns per-row dicts so PyTorch's default `DataLoader` collate stacks them into `{col: tensor(B,)}`. Switch to `with_format("torch_row")` to keep the old shape.

### What changed

`"torch"` now returns a list of per-row dicts (`[{col: tensor}, ...]`) at every indexed access path. The default `DataLoader` collate stacks them into a column-keyed batched dict, no custom `collate_fn` needed. The old shape is preserved under a new `"torch_row"` literal. `"torch_col"` is unchanged.

The unbatching lives inside the transform (`batch_to_tensor_dict`), not `__getitems__`, so the shape survives pickling and works under `DataLoader(num_workers>0, multiprocessing_context="spawn")`.

### Format comparison

| Format | `iter(batch_size=N)` | `__getitems__([0,1,2])` | `DataLoader` default collate |
|---|---|---|---|
| `"torch"` (new) | `list[{col: tensor}]` length N | `list[{col: tensor}]` length 3 | `{col: tensor(B,)}` |
| `"torch_row"` (old `"torch"` behavior) | `list[tensor(n_cols,)]` length N | `list[tensor(n_cols,)]` length 3 | `tensor(B, n_cols)` |
| `"torch_col"` (unchanged) | `tensor(n_cols, N)` | `tensor(n_cols, 3)` | needs `collate_fn=lambda x: x` |

Output matches HuggingFace `Dataset.set_format("torch")` on container shape, keys, and values at every access path. The only divergence: HuggingFace downcasts `float64` to `torch.float32` by default, LanceDB preserves dtype. Verified by `scripts/verify_torch_format.py`.

### Migration

```python
# Old default — column names lost, shape was tensor(B, n_cols)
DataLoader(Permutation.identity(table).with_format("torch"))

# New default — column names preserved
DataLoader(Permutation.identity(table).with_format("torch"))     # {col: tensor(B,)}

# Keep old behavior
DataLoader(Permutation.identity(table).with_format("torch_row")) # tensor(B, n_cols)
```


python/python/lancedb/permutation.py
```diff
@@ -916,15 +917,20 @@ def with_format(
         - "python_col" - the batch will be a dict of lists (one entry per column)
         - "pandas" - the batch will be a pandas DataFrame
         - "arrow" - the batch will be a pyarrow RecordBatch
-        - "torch" - the batch will be a list of tensors, one per row
+        - "torch" - the batch will be a dict of torch tensors keyed by column name
+          (one 1-D tensor per column). This matches HuggingFace's
```
geruh: nit: I don't think we need to mention hugging face in the description here. The first line should be enough

python/python/lancedb/permutation.py
```diff
@@ -916,15 +917,20 @@ def with_format(
...
+          ``dataset.set_format("torch")`` and works with the default
+          ``torch.utils.data.DataLoader`` collate.
+        - "torch_row" - the batch will be a list of torch tensors, one per row.
+          This was the behavior of "torch" prior to the HuggingFace alignment.
```
geruh: Nit: we could just move the old torch comment here.

```
- "torch_row" - the batch will be a list of tensors, one per row
```

python/python/tests/test_permutation.py
```diff
@@ -935,14 +935,43 @@ def test_transform_fn(mem_db):
     try:
         import torch
 
-        torch_result = list(
-            permutation.with_format("torch").iter(10, skip_last_batch=False)
+        # "torch" returns a dict of column-keyed tensors via iter()
+        # (HuggingFace alignment, see issue #3245).
```
geruh: nit: like above we don't need to mention the history. git blame can do this for us

python/python/lancedb/permutation.py
```diff
@@ -983,7 +994,15 @@ async def do_getitems():
...
 
         batch = LOOP.run(do_getitems())
-        return self.transform_fn(batch)
+        result = self.transform_fn(batch)
+        # For with_format("torch"), the transform produces a dict of batched
+        # tensors. Unbatch into a list of per-row dicts so PyTorch's default
+        # DataLoader collate (which expects a list of samples) can stack them
+        # back into a batched dict. iter() still yields the batched dict
+        # directly.
+        if getattr(self, "_torch_dict_format", False):
```
geruh: I'm a bit on the fence with this need for a torch dict flag. If I'm understanding correctly this exists because the new `batch_to_tensor_dict` returns as a batched dict and the dunder `get_items` needs to unbatch it to row dicts for the Dataloader's default `collate_fn` to work. So the flag will steer the data with this unbatch logic. 

Another concern at the moment is that the flag wasn't added to the `__get_state__` function which pickles the attributes. So workers receive a `Permutation` without your added flag, and would fall back to the batched dict. You can test this locally with a dataloader like: `DataLoader(num_workers=2, multiprocessing_context="spawn")`

However, why don't we just move this unbatching logic into your util function:

```python
    torch = attempt_import_or_raise("torch", "torch")
    tensors = {
        name: torch.tensor(col.to_numpy(zero_copy_only=False))
        for name, col in zip(batch.schema.names, batch.columns)
    }
    return [{col: tensor[i] for col, tensor in tensors.items()} for i in range(batch.num_rows)]
```

That way we avoid tracking the flag entirely, __getitems__ stays format agnostic, and the pickle issue goes away by construction. I tried this locally and get something like:

```
__getitems__([0, 1, 2]) -> [{"x": tensor(0.), "label": tensor(0)}, ...]
DataLoader(batch_size=2) ->  {"x": tensor([0., 1.]), "label": tensor([0, 1])}
```

WDYT?

geruh (commented):
Thanks for opening this @omkar-334! I did a pass on the logic, and left a few comments. 

Also, this is technically a breaking change so we should use the conventional commit to mark it as such. i.e. `feat!`

omkar-334:
hey @geruh thanks for the detailed review. i agree with your last point, - i hadnt observed the pickle issue. I think one test i had written locally was failing and this was the cause, and then you basically explained it. the code also looks cleaner now.

I've made the requested changes and updated the branch. what do you think now

omkar-334:
cc @geruh

geruh (approved):
Thanks for the update @omkar-334, I think this heads in the right direction now!

I just have a comment here, this looks like a breaking change for `torch` format version so we should label it as such in the description and pr title `feat!(python)`

omkar-334:
@geruh thanks! I've updated the PR name and descrioption to align better with the PR

omkar-334:
hi @geruh , just a nudge on this

PR #3373, opened 2026-05-12
ci(nodejs): switch from npm to pnpm 11

wjones127:
## Summary

Switch the nodejs bindings and examples package from npm to pnpm 11 to pick up its stronger supply-chain defaults:

- `minimumReleaseAge` defaults to 1 day, so newly-published (potentially compromised) versions aren't resolved into installs for at least 24h.
- Install lifecycle scripts (`preinstall`/`install`/`postinstall`) are no longer run for arbitrary transitive deps; only an explicit allowlist may run them, and unapproved scripts cause install to fail (`strictDepBuilds: true`).
- Audit uses GHSA IDs and `--fix=update` to add patched versions to `minimumReleaseAgeExclude`.

This is the same class of protection that would have blunted the recent TanStack/`@uipath`/etc. compromise discussed in the [Aikido write-up](https://www.aikido.dev/blog/mini-shai-hulud-is-back-tanstack-compromised).

## Changes

- Replace `nodejs/package-lock.json` and `nodejs/examples/package-lock.json` with `pnpm-lock.yaml`.
- Pin pnpm via `packageManager: pnpm@11.1.1` in both `package.json`s.
- Add `pnpm-workspace.yaml` with the four build-script packages we actually need: `@biomejs/biome`, `onnxruntime-node`, `protobufjs`, `sharp`. Everything else is blocked from running install scripts.
- Update package.json scripts (`npm run X` → `pnpm X`).
- Update workflows: `.github/workflows/nodejs.yml`, `.github/workflows/npm-publish.yml`, and `.github/workflows/codex-fix-ci.yml` — install pnpm via `pnpm/action-setup@v4` and switch `setup-node` caches to `pnpm-lock.yaml`.
- Refresh `nodejs/AGENTS.md`, `nodejs/CLAUDE.md`, and `nodejs/CONTRIBUTING.md`.

`docs/package-lock.json` is **not** touched — out of scope for this PR.

## Test plan

- [ ] `Lint` job (lint Rust/TS + examples lint) passes on CI.
- [ ] `Linux (NodeJS 18/20)` build+test passes, including the examples test step.
- [ ] `macos` build+test passes.
- [ ] `NPM Publish` workflow's PR dry-run completes (build matrix + test matrix + dry `npm publish`).
- [ ] No new install-script approvals are required at install time.

## Follow-ups

- `update_package_lock_run_nodejs.yml` references a composite action path that doesn't exist (`./.github/workflows/update_package_lock_nodejs`); it was already broken pre-PR. We may want to either delete this workflow or rewrite it for pnpm in a follow-up.
- Consider migrating `docs/` to pnpm in a separate PR.


Xuanwo (approved):

.github/workflows/codex-fix-ci.yml
```diff
@@ -45,7 +45,8 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          # pnpm 11 (used by the nodejs install step below) requires Node >= 22.13.
+          node-version: 22
```
dantasse: might as well do 24, while you're at it; 22 is only good until october: https://nodejs.org/en/about/previous-releases

.github/workflows/nodejs.yml
```diff
@@ -42,11 +42,16 @@ jobs:
...
+      with:
+        version: 11.1.1
     - uses: actions/setup-node@v4
       with:
-        node-version: 20
-        cache: 'npm'
-        cache-dependency-path: nodejs/package-lock.json
+        # pnpm 11 requires Node >= 22.13. The library itself still
+        # supports Node >= 18 (see test matrix below).
+        node-version: 22
```
dantasse: 24 here too

nodejs/CONTRIBUTING.md
```diff
@@ -12,20 +12,22 @@ Typescript.
...
+* `examples/`: A pnpm package with the examples shown in the documentation
 
 ## Development environment
 
 To set up your development environment, you will need to install the following:
 
-1. Node.js 14 or later
-2. Rust's package manager, Cargo. Use [rustup](https://rustup.rs/) to install.
-3. [protoc](https://grpc.io/docs/protoc-installation/) (Protocol Buffers compiler)
+1. Node.js 18 or later
```
dantasse: I guess they'll need 22 or later, to use pnpm 11?

.github/workflows/nodejs.yml
dantasse: ah I guess I won't point out all the places it says 22, claude can do that :P

.github/workflows/nodejs.yml
```diff
@@ -82,60 +89,69 @@ jobs:
...
     - name: Test
       env:
         S3_TEST: "1"
-      run: npm run test
-    - name: Setup examples
-      working-directory: nodejs/examples
-      run: npm ci
+        # Newer @smithy/core uses dynamic ESM imports.
+        NODE_OPTIONS: "--experimental-vm-modules"
+      # Invoke jest directly because pnpm 11 itself requires Node 22+.
```
dantasse: I don't really get this; you'll have node 22+ anyway, right?

docs/src/js/_media/CONTRIBUTING.md
```diff
@@ -12,20 +12,22 @@ Typescript.
...
+* `examples/`: A pnpm package with the examples shown in the documentation
 
 ## Development environment
 
 To set up your development environment, you will need to install the following:
 
-1. Node.js 14 or later
-2. Rust's package manager, Cargo. Use [rustup](https://rustup.rs/) to install.
-3. [protoc](https://grpc.io/docs/protoc-installation/) (Protocol Buffers compiler)
+1. Node.js 18 or later
```
dantasse: 22 or later, for pnpm 11 or later, right?

dantasse (approved):
lgtm! 22 works too but 24 just saves us a little future work.

wjones127:
Thanks for the review!

- Bumped install-side Node 22 → 24 across `nodejs.yml`, `npm-publish.yml`, `codex-fix-ci.yml` (578b097e).
- `nodejs/CONTRIBUTING.md` (and the typedoc-imported copy in `docs/`) now say Node 22+ since pnpm 11 needs it. Kept `engines.node: ">= 18"` in `package.json` because that's the library *runtime* minimum, not the contributor toolchain.
- Re: the `npx jest` comment on `nodejs.yml:143` — that step runs *after* the second `setup-node` action drops the runtime back to the matrix version (18 or 20) to actually exercise the library on those Node versions. Pnpm itself can't run on 18/20, so we invoke jest's binary directly. Happy to drop the 18/20 matrix entries if we're OK losing that coverage, but as-is the comment is accurate.

dantasse:
ahh, yeah, ok! I missed the test-matrix version and engines.node, that sounds good to keep as is then. Thanks!

PR #3374, opened 2026-05-12
feat(nodejs): add prewarmData method on Table

### LINKED ISSUE #3362: Add prewarmData API to Node bindings
Prewarm was added to Rust and Python in https://github.com/lancedb/lancedb/pull/3110

We should add it to NodeJS as well.

You can follow the pattern from https://github.com/lancedb/lancedb/pull/2342

brendanclement:
### Summary
- Closes #3362 
- Adds `prewarmData(columns?: string[])` to the Node bindings, mirroring the Rust and Python implementations

### Testing
- [x] `npm run build` (regenerates the napi `.node` module + TS declarations)
- [x] `npm run lint`
- [x] `npm test
- [ ] live test against remote table - just waiting for my dev stack to get created

### Documentation
- updated docs

### DIFF
--- a/docs/src/js/classes/Table.md
+++ b/docs/src/js/classes/Table.md
@@ -501,6 +501,34 @@ Modeled after ``VACUUM`` in PostgreSQL.
 
 ***
 
+### prewarmData()
+
+```ts
+abstract prewarmData(columns?): Promise<void>
+```
+
+Prewarm one or more columns of data in the table.
+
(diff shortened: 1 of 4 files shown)


wjones127 (approved):
Looks good. This is admittedly hard to test in CI, but I think we can rely on previous testing for this, given it's just a wrapper with straightforward signature.

PR #3375, opened 2026-05-13
feat(python): add public take_offsets method on Permutation

### LINKED ISSUE #3243: Feature: public API for batch access on Permutation
### SDK

Python

### Description

The only way to batch-fetch by index from a custom Dataset is to call a dunder directly:
```
def __getitems__(self, indices):
    self._ensure_open()
    return self._perm.__getitems__(indices)   # calling a dunder explicitly
```
The internal implementation already uses reader.take_offsets(indices)  but it is only surfaced through a dunder. The deprecated .take(limit: int) exists but does SQL-LIMIT row capping, not index access. A user who discovers .take() and passes a list of indices gets a type error and has to dig into the source to understand why.

```
perm.take(3) # DeprecationWarning, returns Permutation with 3 rows (first 3)                                                                                    
perm.take([0, 5, 10]) # TypeError: 'list' object cannot be interpreted as an integer   
```

## Proposal
A public method to fetch permutation indices:
```
def __getitems__(self, indices):
    self._ensure_open()
    return self._perm.fetch(indices) # or something like this or simply ([0, 5, 10]) 
```

geruh:
Closes #3243.

This PR exposes a new public api `Permutation.take_offsets(offsets: list[int])`, since users initially had to call __getitems__ directly to batch-fetch rows by position. 

Currently, the name matches the existing `Table.take_offsets` pattern, and now the dunder `__getitem__` and `__getitems__` now delegate to it.

Also, fixes a parse error when `PermutationReader::take_offsets` gets an empty list. Now returns an empty `RecordBatch` with the correct schema instead. Bundled this because without the fix the new public API blows up on a perfectly reasonable input.

`__getitems__` is preserved since PyTorch's batched DataLoader requires it.

### Testing

- Added 3 new Rust tests for empty offsets including permutation table with Select::All, Select::Columns, and identity path
- Added 3 new Python tests for the public API including a happy case, and empty input on both identity and permutation 

clippy, format, check all clean!

cc: @westonpace

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -968,22 +968,32 @@ class Permutation:
         new.transform_fn = transform
         return new
 
+    def take_offsets(self, offsets: list[int]) -> Any:
+        """
+        Take rows from the permutation by offset
+
+        The returned value is passed through the permutation's current transform,
+        so `with_format` and `with_transform` affect this method in the same way
+        they affect iteration.
+        """
+
+        async def do_take_offsets():
+            return await self.reader.take_offsets(offsets, selection=self.selection)
+
+        batch = LOOP.run(do_take_offsets())
+        return self.transform_fn(batch)
+
     def __getitem__(self, index: int) -> Any:
         """
         Returns a single row from the permutation by offset
         """
-        return self.__getitems__([index])
+        return self.take_offsets([index])
 
     def __getitems__(self, indices: list[int]) -> Any:
         """
         Returns rows from the permutation by offset
         """
-
-        async def do_getitems():
-            return await self.reader.take_offsets(indices, selection=self.selection)
-
-        batch = LOOP.run(do_getitems())
-        return self.transform_fn(batch)
+        return self.take_offsets(indices)
 
     @deprecated(details="Use with_skip instead")
     def skip(self, skip: int) -> "Permutation":
--- a/python/python/tests/test_permutation.py
+++ b/python/python/tests/test_permutation.py
@@ -1080,3 +1080,29 @@ def test_getitems_invalid_offset(some_permutation: Permutation):
     """Test __getitems__ with an out-of-range offset raises an error."""
     with pytest.raises(Exception):
         some_permutation.__getitems__([999999])
+
+
+def test_take_offsets(some_permutation: Permutation):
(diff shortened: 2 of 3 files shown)


python/python/lancedb/permutation.py
```diff
@@ -968,22 +968,32 @@ def with_transform(self, transform: Callable[pa.RecordBatch, Any]) -> "Permutati
         new.transform_fn = transform
         return new
 
+    def take_offsets(self, offsets: list[int]) -> Any:
```
geruh, line 971: I know in the description i mentioned that I went with `take_offset` to match the lance table logic but, I'm still on the fence about the public facing API. I did some digging into what the ecosystem does and found:

- NumPy: `arr.take(indices)`
- Pandas: `df.take(indices) / df.iloc[indices]`
- Polars: `df.take(indices)`
- Arrow: `pa.compute.take(table, indices)`
- Lance: `Dataset.take(indices)` <-- collision with the api we already expose for rows mentioned in issue.
- HuggingFace Datasets: `dataset[indices]` / `dataset.select(indices)`

Plain `take(indices)` is pretty dominate. But as mentioned, we can't reuse it here because `Permutation.take(int)` already exists and it's `@deprecated` in favor of `with_take`, but still public.

I also considered:

- `fetch([..])` was originally coined, but seems like it can be confused with materialize and not really select by index. 
- `select(indices)` (HuggingFace) this is  a name collision with different semantics and HF returns a new Dataset, and we'd return rows.

Happy to hear what others think and if there's a stronger preference.

westonpace: CC @AyushExel as I don't have a strong opinion here.  I agree `take` is ambiguous because HuggingFace and Lance have different definitions for that term and we are trying to mirror HuggingFace.  HuggingFace has no equivalent concept.  We should probably also support indexing `permutation[57]` and sliced indexing `permutation[57:80]` if we want to mirror hugging face but that can be done in a follow-up.

AyushExel: Yup agreed. Let's go with `take_offsets` Matching with existing lance semantics is better. We don't necessarily need to match with HF 1:1, just a nice public function is good enough for dataloader implementations:)

AyushExel (approved):

PR #3378, opened 2026-05-13
ci: bump version after 0.29 release

westonpace:
The 0.29 release happened on a branch because the main line had already moved past the 6.0.0 stable lance release.  As a result the version bump commits ended up on the branch.  This merges those commits back into main.


wkalt (approved):
:+1: pending CI

PR #3386, opened 2026-05-13
feat(nodejs): add renameTable on Connection

### LINKED ISSUE #3381: feat(nodejs): add renameTable on Connection
The NodeJS `Connection` class has no `renameTable` method, even though<br>the Rust core and Python SDK both expose one. Follow-up to lancedb/lancedb#3363.

## Requirements

* A user can rename a table on a `Connection`.
* A user can rename a table that lives in a non-root namespace.
* A user can move a table to a different namespace as part of the<br>rename (cross-namespace rename).
* The default behavior when only one of the namespace paths is<br>ambiguous should not silently move tables across namespaces.<br>(Python's current behavior — independently defaulting both<br>`cur_namespace_path` and `new_namespace_path` to `[]` — is a footgun<br>worth avoiding here; see `python/python/lancedb/db.py:1098` for the<br>precedent we'd improve on.)
* Errors on closed connections, like the other `Connection` methods.
* Has tests covering same-namespace rename, cross-namespace rename,<br>and the closed-connection error.
* Documentation is regenerated.

brendanclement:
Adds `Connection.renameTable` to the Node SDK. Closes #3381.

### DIFF
--- a/docs/src/js/classes/Connection.md
+++ b/docs/src/js/classes/Connection.md
@@ -441,18 +441,28 @@ Open a table in the database.
 
 ```ts
 abstract renameTable(
-   oldName,
+   currentName,
    newName,
-   namespacePath?): Promise<void>
+   options?): Promise<void>
 ```
+
+Rename a table.
+
+Currently only supported by LanceDB Cloud. Local OSS connections and
+namespace-backed connections (via [connectNamespace](../functions/connectNamespace.md)) reject with
+a "not supported" error.
 
 #### Parameters
 
-* **oldName**: `string`
+* **currentName**: `string`
+    The current name of the table.
 
 * **newName**: `string`
+    The new name for the table.
 
-* **namespacePath?**: `string`[]
+* **options?**: [`RenameTableOptions`](../interfaces/RenameTableOptions.md)
+    Optional namespace paths. When
+    `newNamespacePath` is omitted the table stays in `namespacePath`.
 
 #### Returns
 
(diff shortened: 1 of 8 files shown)


brendanclement:
@claude review once

docs/src/js/classes/Connection.md
```diff
@@ -437,6 +437,39 @@ Open a table in the database.
...
+   currentName,
+   newName,
+   options?): Promise<void>
+```
+
+Rename a table.
+
+Currently only supported by LanceDB Cloud. Local OSS connections and
+namespace-backed connections (via [connectNamespace](../functions/connectNamespace.md)) reject with
+a "not supported" error.
```
wjones127, line 453: question: is this true? I would expect at least some namespaces to support this.

wjones127: It seems it isn't implement:

https://github.com/lancedb/lancedb/blob/20556e23a943274be567e93eab5402e56216444b/rust/lancedb/src/database/namespace.rs#L489-L499

It should be supported by the namespace spec, though. (https://lance.org/format/namespace/operations/#operation-list). I've created a ticket: https://github.com/lancedb/lancedb/issues/3412

wjones127 (approved):
Looks good! I can merge once the conflicts are resolved.

brendanclement:
Note: while this was in review, https://github.com/lancedb/lancedb/commit/13c6dae9a359398d726a9f5a500178170810bbc0 landed an initial renameTable with a positional (oldName, newName, namespacePath?) shape that only supports same-namespace renames, internally it passes the single namespace as both cur and new to the Rust core, so cross-namespace moves aren't reachable from TS. 

This PR supersedes that surface with an options bag that exposes both paths, satisfying the cross-namespace bullet in #3381 The signature is incompatible with the prior version, but nothing has shipped to npm yet

wjones127 (approved):

PR #3389, opened 2026-05-14
feat(nodejs): surface skip_auto_cleanup on add and merge insert

brendanclement:
### Summary

Surfaces lance-core's `WriteParams.skip_auto_cleanup` (and the matching `MergeInsertBuilder` option) through the LanceDB Rust SDK and the Node.js/TypeScript SDK. When set, the auto-cleanup hook that normally runs during commit is suppressed.

Use case: high-frequency writers that prefer to manage version cleanup themselves (e.g. via a periodic optimize job), or writers without delete permissions on the underlying storage that would otherwise log a cleanup failure on every commit.

### Changes
- TS: `skipAutoCleanup?: boolean` on `AddDataOptions`; `skipAutoCleanup()` builder method on `MergeInsertBuilder`
- NAPI + Rust SDK: plumbing through `WriteOptions` and `MergeInsertBuilder`
- Drive-by fix in `InsertExec`: the commit was discarding `WriteParams.skip_auto_cleanup`, so even existing Rust/Python users of the `lance_write_params` escape hatch weren't getting cleanup suppressed.

### Testing
- updated tests / pnpm passed

### DIFF
--- a/nodejs/__test__/table.test.ts
+++ b/nodejs/__test__/table.test.ts
@@ -115,6 +115,12 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
       await expect(table.countRows()).resolves.toBe(1);
     });
 
+    it("should accept skipAutoCleanup on add()", async () => {
+      await table.add([{ id: 1 }], { skipAutoCleanup: true });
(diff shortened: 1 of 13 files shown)


brendanclement:
todo: need to generate docs

PR #3394, opened 2026-05-16
feat: support setting unenforced primary key

touch-of-grey:
## Summary

Adds `Table::set_unenforced_primary_key` — records a single column as the
table's unenforced primary key in Lance schema field metadata. "Unenforced"
means LanceDB does not check uniqueness on write; the key is metadata that
`merge_insert` consumes.

- Single-column only; the column must exist and have a supported dtype
  (Int32, Int64, Utf8, LargeUtf8, Binary, LargeBinary, FixedSizeBinary). The
  API accepts an iterable for binding ergonomics but requires exactly one
  column — compound keys are rejected.
- The primary key is immutable: calling this on a table that already has an
  unenforced primary key is rejected. Concurrent writers racing to set the key
  fail at commit time rather than silently overriding it.
- `RemoteTable` returns `NotSupported`.
- Bindings: Python (`AsyncTable`, `LanceTable`, `RemoteTable`) and TypeScript
  (`Table.setUnenforcedPrimaryKey`).

## Context

Split out from #3354 per review feedback, so the unenforced primary key and the
`merge_insert` sharding spec land as separate reviewable PRs.

No Lance dependency bump — `main` is already on v7.0.0-beta.10, which includes
the field-metadata round-trip fix the API relies on. Enforcing primary-key
immutability at the Lance commit layer (so the cross-column concurrent race is
also rejected) is a companion Lance change: lance-format/lance#6810.


rust/lancedb/src/table/primary_key.rs
```diff
@@ -0,0 +1,127 @@
...
+                        "set_unenforced_primary_key: column '{}' has dtype {:?} which is not supported as a primary key. Supported: Int32, Int64, Utf8, LargeUtf8, Binary, LargeBinary, FixedSizeBinary",
+                        name,
+                        field.data_type()
+                    ),
+                });
+            }
+            new_pk_field_ids.insert(field.id);
+
+            let mut metadata = field.metadata.clone();
+            metadata.remove(LANCE_UNENFORCED_PRIMARY_KEY);
```
jackye1995: I don't think we should support overriding primary key. If the keys is already set, it should fail. We currently don't support evolving primary key. We also don't really support writing with compound primary key, so probably should also fail that.

jackye1995 (commented):
Can you also add typescript support accordingly?

jackye1995 (approved):
looks good to me

PR #3395, opened 2026-05-17
fix(python): check all table pages for db membership

### LINKED ISSUE #2727: bug(python): Table cannot be found
### LanceDB version

0.25.2

### What happened?

All tables should exist. But there are some tables that were not found.

```python
import lancedb

db = lancedb.connect('test_db')

for n in range(20):
    db.create_table(f'table_{n}', data=[{'id': n, 'name': 'name_' + str(n)}], exist_ok=True)

for n in range(20):
    print(f'table_{n} exists', f'table_{n}' in db)

```

Output
```
table_0 exists True
table_1 exists True
table_2 exists False
table_3 exists False
table_4 exists False
table_5 exists False
table_6 exists False
table_7 exists False
table_8 exists False
table_9 exists False
table_10 exists True
table_11 exists True
table_12 exists True
table_13 exists True
table_14 exists True
table_15 exists True
table_16 exists True
table_17 exists True
table_18 exists False
table_19 exists False

```

pragnyanramtha:
## Summary

- Fix `name in db` and `len(db)` for local Python connections with more than one page of tables.
- Use `list_tables()` pagination instead of deprecated `table_names()` with its default 10-item page.
- Add regression coverage with 20 tables so later pages are included.

Fixes #2727.

## Validation

- `python3 -m py_compile python/python/lancedb/db.py python/python/tests/test_db.py`
- No-build Python harness that extracts and executes the edited `LanceDBConnection` pagination methods: passed
- `uvx ruff check python/python/lancedb/db.py python/python/tests/test_db.py`
- `uvx ruff format --check python/python/lancedb/db.py python/python/tests/test_db.py`

Note: `uv run pytest python/tests/test_db.py::test_db_contains_and_len_include_all_table_name_pages -q` was attempted first, but it stayed in the broad Rust/PyO3 native extension build and was stopped before pytest started.


python/python/lancedb/db.py
```diff
@@ -847,11 +847,21 @@ def table_names(
...
 
+    def _all_table_names(self) -> List[str]:
+        names = []
+        page_token = None
+        while True:
+            response = self.list_tables(page_token=page_token)
+            names.extend(response.tables)
+            page_token = response.page_token
+            if not page_token:
+                return names
```
wjones127: suggestion: for large databases, it might be nice just to return an iterator. That way you can manage memory use and return early for things like `__contains__`:


```suggestion
    def _all_table_names(self) -> Generator[str, None, None]:
        page_token = None
        while True:
            response = self.list_tables(page_token=page_token)
            for table in response.tables:
                yield table
            page_token = response.page_token
            if not page_token:
                return
```

python/python/lancedb/db.py
```diff
@@ -847,11 +847,21 @@ def table_names(
...
+
     def __len__(self) -> int:
-        return len(self.table_names())
+        return len(self._all_table_names())
```
wjones127: suggestion: if you did use a generator, I think this would have to be count instead:


```suggestion
        return itertools.count(self._all_table_names())
```

wjones127 (approved):
Thanks for working on this. I've identified some optional improvements you can make if you'd like. Otherwise, I'll merge in a day or two, if CI is passing.

PR #3396, opened 2026-05-17
feat: support setting LSM write spec for a table

touch-of-grey:
## Summary

Adds `LsmWriteSpec` and `Table::set_lsm_write_spec` / `unset_lsm_write_spec` to
install and clear the spec that selects Lance's MemWAL LSM-style write path for
`merge_insert`.

`LsmWriteSpec` offers three sharding strategies, all built on Lance's
`InitializeMemWalBuilder`:

- `LsmWriteSpec::bucket(column, num_buckets)` — hash-bucket sharding by the
  single-column unenforced primary key.
- `LsmWriteSpec::identity(column)` — identity sharding by the raw value of a
  scalar column.
- `LsmWriteSpec::unsharded()` — a single MemWAL shard.

Each can be refined with `with_maintained_indexes(...)` (indexes the MemWAL
keeps up to date as rows are appended) and `with_writer_config_defaults(...)`
(default `ShardWriter` configuration recorded in the MemWAL index, so every
writer starts from the same defaults). All variants require the table to have
an unenforced primary key.

- `set_lsm_write_spec` installs the spec by initializing the MemWAL index;
  `unset_lsm_write_spec` removes it (dropping the MemWAL index), reverting to
  the standard `merge_insert` path. `unset` is idempotent.
- Bindings: Python (`LsmWriteSpec.bucket` / `.identity` / `.unsharded`,
  `set_lsm_write_spec` / `unset_lsm_write_spec`) and TypeScript
  (`setLsmWriteSpec` with `specType` `"bucket"` / `"identity"` /
  `"unsharded"`). `RemoteTable` returns `NotSupported`.

The actual `merge_insert` LSM dispatch and `ShardWriter` write path are a
follow-up — this PR only installs and clears the spec.

## Context

Split out from #3354; the unenforced primary key half landed in #3394.

The bucket / identity / unsharded `ShardingSpec` construction lives in Lance's
`InitializeMemWalBuilder` (lance-format/lance#6815); this PR calls the builder
rather than hand-building specs. Depends on Lance `v7.0.0-beta.12`.


python/src/table.rs
```diff
@@ -167,10 +169,179 @@ impl From<lancedb::table::MergeResult> for MergeResult {
...
+/// optionally chain `with_maintained_indexes(...)` to have the MemWAL
+/// keep listed indexes up to date as rows are appended.
+#[pyclass(from_py_object)]
+#[derive(Clone, Debug)]
+pub struct MergeInsertSpec {
+    inner: lancedb::table::MergeInsertSpec,
+}
+
+#[pymethods]
+impl MergeInsertSpec {
```
jackye1995: we should call this LsmWriteSpec, since we should be able to reuse it also for inserts for tables without primary key

python/src/table.rs
```diff
@@ -818,6 +1003,24 @@ impl Table {
         })
     }
 
+    pub fn set_merge_insert_spec<'a>(
```
jackye1995: we should also add an unset function so that people can revert to the old behavior if necessary

python/src/table.rs
```diff
@@ -167,10 +169,179 @@ impl From<lancedb::table::MergeResult> for MergeResult {
...
+            inner: self.inner.clone().with_maintained_indexes(indexes),
+        }
+    }
+
+    /// For the bucket variant: skip the per-row Murmur3 check that
+    /// asserts every row hashes to the same bucket. Only the first
+    /// row of each input batch is hashed (to pick the target writer).
+    /// The caller becomes responsible for pre-sharding the input.
+    /// No-op for the unsharded variant.
+    pub fn assume_pre_sharded(&self) -> Self {
```
jackye1995: the information of this API is not really stored, it should be supplied at runtime when running merge insert, not a part of the persisted sharding spec.

rust/lancedb/src/table/merge/lsm.rs
```diff
@@ -0,0 +1,972 @@
...
+        }
+        MergeInsertSpec::Unsharded {
+            maintained_indexes,
+            tuning,
+        } => (build_unsharded_shard_spec(&tuning), 1, maintained_indexes),
+    };
+
+    let mut dataset = (*table.dataset.get().await?).clone();
+    dataset
+        .initialize_mem_wal_with_shards(
```
jackye1995: examine what information exactly are persisted by the MemWAL index, we should not record other unnecessary information in the spec. Things like tunings and assume pre-sharded are runtime configs that should not be a part of the shard spec, but should be added into the specific function call.

jackye1995 (commented):
similar to the other PR, should add typescript support

rust/lancedb/src/table/merge.rs
```diff
@@ -39,6 +41,15 @@ pub struct MergeResult {
...
     pub num_attempts: u32,
+    /// Total number of rows that were either inserted or updated.
+    ///
+    /// On the standard merge_insert path this equals `num_inserted_rows + num_updated_rows`.
+    /// On the MemWAL ShardWriter path the breakdown between insert and update is
+    /// not known until LSM compaction; in that mode `num_inserted_rows`,
+    /// `num_updated_rows`, and `num_deleted_rows` are all zero and this field
+    /// holds the total row count written.
+    #[serde(default)]
+    pub num_rows: u64,
```
jackye1995: the merge result and merge writer related changes should be added separatedly, this PR should only concern setting and clearing the spec.

python/src/table.rs
touch-of-grey: Is it really true? The default need to be chosen very carefully to maximize performance, I propose that we should store these in the MemWAL index details so that all the writers can agree on the same config to use, same memtable flush size, etc.

python/python/lancedb/table.py
```diff
@@ -3838,6 +3849,35 @@ async def set_unenforced_primary_key(
             columns = list(columns)
         await self._inner.set_unenforced_primary_key(columns)
 
+    async def set_lsm_write_spec(self, spec: "LsmWriteSpec") -> None:
+        """Install an LsmWriteSpec on this table.
+
+        The spec selects Lance's MemWAL LSM-style write path for future
+        `merge_insert` calls. For a bucket spec, the table must already
```
jackye1995: This is not just for bucket spec, currently all writes require unenforced primary key. Also it's not "bucket spec" but bucket sharding.

python/src/table.rs
```diff
@@ -171,6 +171,104 @@ impl From<lancedb::table::MergeResult> for MergeResult {
...
+pub struct LsmWriteSpec {
+    inner: lancedb::table::LsmWriteSpec,
+}
+
+#[pymethods]
+impl LsmWriteSpec {
+    /// Hash-bucket partitioning by the unenforced primary key column.
+    #[staticmethod]
+    pub fn bucket(column: String, num_buckets: u32) -> Self {
+        Self {
```
jackye1995, line 191: Also add the identity sharding to the bindings.

rust/lancedb/src/table.rs
```diff
@@ -273,6 +273,91 @@ pub trait Tags: Send + Sync {
...
+    /// No partitioning — every `merge_insert` call writes to a single
+    /// MemWAL shard. No primary key is required.
+    Unsharded {
+        /// Names of indexes (already created on the table) that the
+        /// MemWAL should maintain in-memory as rows are appended.
+        maintained_indexes: Vec<String>,
+    },
+}
+
+impl LsmWriteSpec {
```
jackye1995, line 329: As you suggested we can now expose writer config defaults to the spec settings.

docs/src/js/classes/Table.md
```diff
@@ -690,6 +690,30 @@ of the given query
...
+### setLsmWriteSpec()
+
+```ts
+abstract setLsmWriteSpec(spec): Promise<void>
+```
+
+Install an [LsmWriteSpec](../interfaces/LsmWriteSpec.md) on this table.
+
+The spec selects Lance's MemWAL LSM-style write path for future
+`mergeInsert` calls. For a `"bucket"` spec, the table must already
```
jackye1995: we should provide a more comprehensive introduction of what the Lsm write spec does. We should not just mention buket spec, we should equally introduce bucket, identity and unsharded sharding spec. Also give some code examples across python, typescript and rust for the usage.

docs/src/js/classes/Table.md
```diff
@@ -818,6 +842,22 @@ Return the table as an arrow table
...
 
+### unsetLsmWriteSpec()
+
+```ts
+abstract unsetLsmWriteSpec(): Promise<void>
+```
+
+Remove the [LsmWriteSpec](../interfaces/LsmWriteSpec.md) from this table.
+
+This is a no-op if no spec is currently set.
```
jackye1995: this should error if no spec is currently set

jackye1995 (approved):
thanks for all the iterations, looks good to me

PR #3398, opened 2026-05-18
feat(nodejs): add progress to Table.add

brendanclement:
### Summary

- Add an optional `progress` callback to `Table.add(data, { progress })`.  Callback fires once per batch written and once more with `done: true` when the write completes.
- Errors thrown from the user's callback are logged with `console.warn` and swallowed

### Testing
- npm test 
- ran smoke test script to verify functionality

### DIFF
--- a/docs/src/js/interfaces/AddDataOptions.md
+++ b/docs/src/js/interfaces/AddDataOptions.md
@@ -31,9 +31,9 @@ progress: (progress) => void;
 Optional callback invoked periodically with write progress.
 
 The callback is fired once per batch written and once more with
-`done: true` when the write completes. Calls are non-blocking — if the
-callback is slow, intermediate updates may be dropped to avoid stalling
-the write.
+`done: true` when the write completes. Calls are dispatched
+asynchronously to the JS event loop and never block the write — a slow
+callback will queue events rather than back-pressure the writer.
 
 Errors thrown from the callback are logged with `console.warn` and
 swallowed — they do not abort the write.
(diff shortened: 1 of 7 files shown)


docs/src/js/interfaces/AddDataOptions.md
```diff
@@ -19,3 +19,39 @@ mode: "append" | "overwrite";
...
 
 If "overwrite" then the new data will replace the existing data in the table.
+
+***
+
+### progress()
+
+```ts
+progress: (progress) => void;
+```
```
wjones127, line 29: suggestion(non-blocking): In Python, we allowed passing `true` to `progress`, which provided a nice default progress bar. it showed progress in terms of rows, the byte throughput and number of workers active. Does it make sense to add something like that for Typescript? Can make a follow up issue for that if you want.

```python
table.add(data, progress=True)
# 100%|██████████| 1000000/1000000 [00:12<00:00, 82345 rows/s, 45.2 MB/s | 4/4 workers]
```

https://github.com/lancedb/lancedb/pull/3067

wjones127 (approved):
This all looks very reasonable. Could be nice to provide a default progress bar like we do for Python.

docs/src/js/interfaces/AddDataOptions.md
brendanclement: Good feedback, we can look into it as a followup. I'm not sure how often the TS SDK is used in a console-like setting. Should be a quick addition if we want to add it

Issue #3402, opened 2026-05-19
Support nested field paths in native index creation


Xuanwo:
Parent: lancedb/lancedb#3401  <br>Related: [#929](<https://github.com/lancedb/lancedb/issues/929>)

## Summary

Native LanceDB index creation should accept Lance nested field paths and pass the canonical full path to Lance, instead of resolving columns with a top-level Arrow lookup or falling back to a leaf field name.

## Scope

* Update the native Rust create-index path so `metadata.user_id`, `payload.text`, and `image.embedding` resolve through Lance field path semantics.
* Preserve escaped paths such as `` `a.b` `` and `` `meta-data`.`user-id` ``.
* Ensure scalar, vector, and FTS index creation share the same field-path contract.
* Keep validation errors explicit when the field path is invalid or ambiguous.

## Acceptance criteria

* `create_scalar_index("metadata.user_id")` creates an index on the nested leaf.
* `create_index(... vector_column_name="image.embedding" ...)` creates a vector index on the nested leaf.
* FTS index creation can target `payload.text` once the Python wrapper guard is removed.
* Top-level literal-dot fields and nested paths are distinguishable.

PR #3408, opened 2026-05-19
fix: support nested field paths in native index creation

Xuanwo:
Native index creation was resolving requested columns through top-level Arrow schema lookup before handing the request to Lance, which rejected nested paths and could collapse a nested field to its leaf name. This PR resolves index targets with Lance field-path semantics, passes the canonical path through to Lance, and reports indexed columns from field ids as canonical full paths.

This also removes the Python native FTS guard that rejected dotted paths so scalar, vector, and FTS index creation share the same nested-field contract. Related to #3402.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -2171,6 +2171,33 @@ impl NativeTable {
         }
     }
 
+    fn resolve_index_field(
+        schema: &lance_core::datatypes::Schema,
+        column: &str,
+    ) -> Result<(String, Field)> {
+        lance_core::datatypes::parse_field_path(column).map_err(|e| Error::InvalidInput {
+            message: format!("Invalid field path `{}`: {}", column, e),
+        })?;
+
+        let field_path = schema
+            .resolve_case_insensitive(column)
+            .ok_or_else(|| Error::Schema {
+                message: format!(
+                    "Field path `{}` not found in schema. Available field paths: {}",
+                    column,
+                    schema.field_paths().join(", ")
+                ),
+            })?;
+        let field = field_path.last().expect("field path should be non-empty");
+        let path_segments = field_path
+            .iter()
+            .map(|field| field.name.as_str())
+            .collect::<Vec<_>>();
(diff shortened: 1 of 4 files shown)


LuQQiu:
does it count as breaking changes? 
list_indices() previous return leaf field name and now it returns full path name? and if people are doing some column name comparison in their script, it could break them

LuQQiu (approved):

westonpace:
> does it count as breaking changes?
list_indices() previous return leaf field name and now it returns full path name? and if people are doing some column name comparison in their script, it could break them

I think it is a grey area.  If a user was only using lancedb they could not have created the index in the first place (so list_indices can't be a breaking change).

However, some users may have created the index in pylance, in which case they were getting back the old name.

Let's label it a breaking change but not worry about it (e.g. don't try and introduce some backwards compatible path).

rust/lancedb/src/table.rs
```diff
@@ -3650,6 +3688,203 @@ mod tests {
...
+                &["`meta-data`.`user-id`"],
+                Index::BTree(BTreeIndexBuilder::default()),
+            )
+            .name("escaped_names_idx".to_string())
+            .execute()
+            .await
+            .unwrap();
+        table
+            .create_index(
+                &["literal.`a.b`"],
```
wjones127, line 3785: issue(non-blocking): I think this technically works, but I will note we do have some weird rules around `.` in column names. We don't allow them at the top-level, but do at nested levels.

https://github.com/lance-format/lance/blob/e5105f7b249e834c9a1d089416676f7f8a97e7d7/rust/lance-core/src/datatypes/schema.rs#L312-L317

wjones127 (approved):

rust/lancedb/src/table.rs
Xuanwo: Yep, this is intentional. Top-level dotted fields are rejected, but nested field names may contain `.`, so this case verifies that escaped nested segments round-trip as `literal.`a.b`` instead of being split into `literal.a.b`.

PR #3411, opened 2026-05-19
fix(python): drop version from Table __repr__

mackrorysd:
There have been a couple of reports of this function freezing debuggers because it triggers a network round-trip but is assumed to be extremely light-weight: https://github.com/lancedb/lancedb/discussions/2853. We'll just cache the last version we see.

I considered digging into see if we could assume or get the version at create time or after other operations, but that could be a bit of a rabbit hole as I'm a bit unfamiliar with this. Claude was having a hard time of it too 😅  I propose we see how the currently implementation goes and improve it if people find "unknown" or stale values coming up disruptively often before improving this further.

### DIFF
--- a/python/python/tests/test_table.py
+++ b/python/python/tests/test_table.py
@@ -33,7 +33,7 @@ def test_basic(mem_db: DBConnection):
     table = mem_db.create_table("test", data=data)
 
     assert table.name == "test"
-    assert "LanceTable(name='test', version=1, _conn=LanceDBConnection(" in repr(table)
+    assert "LanceTable(name='test', _conn=LanceDBConnection(" in repr(table)
     expected_schema = pa.schema(
         {
             "vector": pa.list_(pa.float32(), 2),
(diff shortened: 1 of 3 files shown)


wjones127 (commented):
Realistically, most people don't directly call `version` that often and our code doesn't either. So I think this will be often just `unknown`. I don't see another easy solution right now.

Given that, I think the best thing for now is just to remove the `version` field from the dataset.

The one exception we might make to that is if there is an explicitly pinned version, like here:

https://github.com/lancedb/lancedb/blob/ad1634a0a5466b9959f1f38ef8f932f8de55ffad/rust/lancedb/src/table/dataset.rs#L29

But that's a more complex change that should be left to a follow up.

Can you change this PR to just drop the `version` field for now?

python/python/tests/test_table.py
```diff
@@ -33,6 +33,12 @@ def test_basic(mem_db: DBConnection):
     table = mem_db.create_table("test", data=data)
 
     assert table.name == "test"
+    # __repr__ shows "unknown" until the version property is read at least
```
Xuanwo: unknown is not useful in this case, maybe we can just hide this value.

wjones127 (approved):

PR #3415, opened 2026-05-20
feat(rust): support DataFusion Expr for table row deletions

### LINKED ISSUE #3204: Feature: support datafusion expression for table row deletions
### SDK

Rust

### Description

currently the row deletion api support only SQL predicate, which is eventually translated into a datafusion Expr.
would suggest allowing for a datafusion Expr to be used directly, so, no conversion is needed.
this would require a change to the lance code. see: https://github.com/lance-format/lance/issues/6339

yuvalif:
Modified the parameter of delete to a Predicate that could be constructed from either datafusion Expr, from str (to support SQL predicate), or from String to support python and javascript bindings.
When a datafusion Expr is used, it avoids the overhead of serializing to SQL and re-parsing when callers already have an Expr (e.g. from query planning).

The native implementation uses lance's `DeleteBuilder::from_expr`. The remote implementation converts the Expr to SQL via `expr_to_sql_string` before sending to the server, consistent with the existing query and count_rows paths.

Closes #3204

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -253,6 +253,36 @@ pub enum Filter {
     Datafusion(Expr),
 }
 
+/// A predicate for filtering rows in delete operations.
+///
+/// Accepts either a SQL string or a DataFusion [`Expr`]. Use the [`From`]
+/// implementations to convert from `&str` or `&Expr` automatically.
+/// See [`Table::delete`] for usage examples.
+pub enum Predicate<'a> {
+    /// A SQL predicate string
+    String(&'a str),
+    /// A DataFusion logical expression
+    Expr(&'a Expr),
+}
+
+impl<'a> From<&'a str> for Predicate<'a> {
+    fn from(s: &'a str) -> Self {
+        Predicate::String(s)
+    }
+}
+
+impl<'a> From<&'a String> for Predicate<'a> {
+    fn from(s: &'a String) -> Self {
+        Predicate::String(s.as_str())
+    }
+}
+
+impl<'a> From<&'a Expr> for Predicate<'a> {
+    fn from(e: &'a Expr) -> Self {
+        Predicate::Expr(e)
+    }
+}
+
 #[async_trait]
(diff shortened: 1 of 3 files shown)


rust/lancedb/src/table.rs
```diff
@@ -493,6 +493,8 @@ pub trait BaseTable: std::fmt::Display + std::fmt::Debug + Send + Sync {
     async fn add(&self, add: AddDataBuilder) -> Result<AddResult>;
     /// Delete rows from the table.
     async fn delete(&self, predicate: &str) -> Result<DeleteResult>;
+    /// Delete rows from the table using a DataFusion [`Expr`] predicate.
+    async fn delete_expr(&self, predicate: Expr) -> Result<DeleteResult>;
```
wjones127: suggestion: It might be nice to not have to add a new method here. For example, could do something like:

```rust
enum Predicate<'a> {
    String(&'a str),
    Expr(&'a Expr),
}

impl From<&str> for Predicate { ... }
impl From<&Expr> for Predicate { ... }

async fn delete(&self, predicate: impl Into<Predicate<'_>>) -> Result<DeleteResult> {
    let predicate = Predicate::from(predicate);
    self.inner.delete(predicate).await
}
```

What do you think of that?

wjones127 (commented):
This seems well implemented. I have one idea about the API I'd like your thoughts on before we go further.

rust/lancedb/src/table.rs
yuvalif: agree that this is better API. but if we change the existing function, won't it force changes on all clients using it?

wjones127: > won't it force changes on all clients using it?

Why would it do that? The API currently takes `&str` and the new API also takes `&str` (because `&str` implements `Into<Predicate<'_>>`). You can validate this by changing the API to this and seeing the unit tests shouldn't need to change anything to compile.

yuvalif: made the change

wjones127 (approved):
Nice work!

PR #3417, opened 2026-05-20
docs: document Python uv agent workflow

Xuanwo:
Document the uv-managed Python development workflow in the root and Python-specific agent instructions so validation does not stop at system Python or global pytest failures.

This clarifies that agents should bootstrap or enter the uv environment, rebuild the PyO3 extension when `lancedb._lancedb` is missing or stale, and run targeted Python tests through `uv run`.


westonpace (approved):
Thank you!  I've run into the same problem I think where the agents get easily confused by uv.

PR #3422, opened 2026-05-21
fix(rust): report local write progress bytes from Lance

### LINKED ISSUE #3360: Update write progress for local tables to use bytes to disk
We temporarily measured size as in-memory size:

https://github.com/lancedb/lancedb/blob/9b21c136c6b90dd147cf1f82b543155008d8a58c/rust/lancedb/src/table/datafusion/scannable_exec.rs#L112-L114

But an upstream PR in Lance should enable tracking bytes written to disk, which is more representative of the bandwidth constraints we care about.

https://github.com/lance-format/lance/pull/6318

GHX5T-SOL:
Fixes #3360.

This updates native table writes so local write progress uses Lance writer byte stats instead of Arrow in-memory batch size once write bytes are available. The change wires the existing `WriteProgressTracker` into `InsertExec` for native `add` writes, installs a Lance `WriteProgressFn` only when no lower-level callback is already configured, and keeps the existing public `InsertExec::new` signature unchanged.

Validation:
- `cargo test -p lancedb --features remote table::write_progress::tests::test_progress_uses_lance_write_bytes_for_local_tables -- --nocapture` passed: 1 passed, 0 failed.
- `cargo test -p lancedb --features remote table::write_progress::tests -- --nocapture` passed: 7 passed, 0 failed.
- `cargo check --quiet --features remote --tests --examples` passed.
- `cargo fmt --all --check` passed.
- `git diff --check` passed.
- `git diff | gitleaks stdin --no-banner --redact --timeout 30` passed: no leaks found.

I did not run the full `cargo test --quiet --features remote --tests` suite.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -2646,7 +2646,13 @@ impl BaseTable for NativeTable {
             output.plan
         };
 
-        let insert_exec = Arc::new(InsertExec::new(ds_wrapper.clone(), ds, plan, lance_params));
+        let insert_exec = Arc::new(InsertExec::new_with_tracker(
+            ds_wrapper.clone(),
+            ds,
+            plan,
+            lance_params,
+            output.tracker.clone(),
+        ));
 
         let tracker_for_tasks = output.tracker.clone();
         if let Some(ref t) = tracker_for_tasks {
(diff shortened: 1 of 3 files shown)


wjones127 (approved):
Thanks for working on this!

PR #3425, opened 2026-05-21
fix(python): route sync BaseQueryBuilder.to_batches through async path

### LINKED ISSUE #3424: take_row_ids(...).to_batches() raises AttributeError: 'AsyncTakeQuery' object has no attribute 'execute'
## Summary

`LanceTakeQueryBuilder.to_batches()` is broken: it always raises `AttributeError: 'AsyncTakeQuery' object has no attribute 'execute'`. `to_arrow()` on the same builder works fine, and `to_batches()` on other sync query types (`search()`, vector, FTS, hybrid) works fine — the bug is specific to the take-query sync path.

Reproduced on `lancedb==0.31.0b9`; the offending code is unchanged on `main` and through `0.32.1b0`.

## Minimal repro

```python
import lancedb, pyarrow as pa, tempfile

tbl = db.create_table("t", data=pa.table({"a": list(range(100))}))

tbl.take_row_ids([0, 1, 2]).to_arrow()       # works
tbl.search().to_batches()                    # works
list(tbl.take_row_ids([0, 1, 2]).to_batches())  # AttributeError
```

Output:
```
AttributeError: 'AsyncTakeQuery' object has no attribute 'execute'
```

## Root cause

[`BaseQueryBuilder.to_batches`](https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/query.py#L3351) calls:

```python
async_iter = LOOP.run(self._inner.execute(max_batch_length, timeout))
```

It assumes `self._inner` is the native Rust query object (which has `execute()`), but for `LanceTakeQueryBuilder` `self._inner` is an `AsyncTakeQuery` (a Python `AsyncQueryBase` wrapper) — only its `_inner` (the native `TakeQuery`) has `execute()`.

Other sync builders (`LanceQueryBuilder`, `LanceVectorQueryBuilder`, etc.) all override `to_batches` and never go through this path; `LanceTakeQueryBuilder` is the only subclass that inherits `BaseQueryBuilder.to_batches` unchanged, so this bug surfaces only on take queries.

The async `AsyncQueryBase.to_batches` (`query.py:2318`) works correctly because it does `await self._inner.execute(...)` from an awaiting context.

## Proposed fix

Route the sync `to_batches` through the async `to_batches` on the background loop, so the native `execute()` is called from inside an awaiting context:

```python
def to_batches(self, *, max_batch_length=None, timeout=None):
    async def _start():
        return await self._inner.to_batches(
            max_batch_length=max_batch_length, timeout=timeout
        )
    async_reader = LOOP.run(_start())

    async def _next(it):
        return await it.__anext__()

    def iter_sync():
        try:
            while True:
                yield LOOP.run(_next(async_reader))
        except StopAsyncIteration:
            return

    return pa.RecordBatchReader.from_batches(async_reader.schema, iter_sync())
```

Verified locally with monkey-patching against 0.31.0b9: take/select/regular `to_batches()` all work after the change. Happy to send a PR if helpful.

## Impact

Any downstream that chunks `take_row_ids()` calls and streams via `to_batches()` is broken. We hit this in [geneva](https://github.com/lancedb/geneva) where a UDTF index-partition source streams chunked takes, causing every partition to be silently skipped (lossy on-error handling). The bug has been masked for weeks because callers that use `to_arrow()` instead are unaffected.

justinrmiller:
## Summary

Fixes #3424.

`LanceTakeQueryBuilder.to_batches()` raised `AttributeError: 'AsyncTakeQuery' object has no attribute 'execute'`. The inherited `BaseQueryBuilder.to_batches` called `self._inner.execute(...)`, but `self._inner` is an `AsyncQueryBase` (Python wrapper) — only its native inner exposes `execute`. Every other sync builder overrides `to_batches`, so the bug only surfaced on take-query builders, which inherit the base unchanged. `take_offsets(...).to_batches()` is broken for the same reason.

Route the sync wrapper through the async `to_batches` on the background event loop, so the native `execute` is invoked from inside an awaiting context (matching how the async path works correctly).

## Repro

```python
import lancedb, pyarrow as pa, tempfile
tbl = db.create_table("t", data=pa.table({"a": list(range(100))}))

tbl.take_row_ids([0, 1, 2]).to_arrow()        # works
tbl.search().to_batches()                     # works
list(tbl.take_row_ids([0, 1, 2]).to_batches())  # AttributeError (before)
```

## Test plan

- [x] New regression test `test_take_queries_to_batches` covers `take_offsets(...).to_batches()`, `take_row_ids(...).to_batches()`, and the `select(...)` projection — all fail on `main` with the patch reverted, all pass with the fix.
- [x] `test_take_queries`, `test_query_builder_batches`, and `test_query_schema` still pass.
- [x] `ruff format --check` and `ruff check` clean on changed files.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -3348,16 +3348,30 @@ class BaseQueryBuilder(object):
             If not specified, no timeout is applied. If the query does not
             complete within the specified time, an error will be raised.
         """
-        async_iter = LOOP.run(self._inner.execute(max_batch_length, timeout))
+
+        # ``self._inner`` is an ``AsyncQueryBase`` wrapping the native query
+        # object.  Calling ``execute`` on the native object directly from this
+        # thread fails (the pyo3 future needs a running loop), and ``execute``
+        # is not defined on ``AsyncQueryBase`` itself — so go through the
+        # async ``to_batches`` on the background loop.
+        async def _open():
+            return await self._inner.to_batches(
+                max_batch_length=max_batch_length, timeout=timeout
+            )
+
+        async_reader = LOOP.run(_open())
+
+        async def _anext(it):
+            return await it.__anext__()
 
         def iter_sync():
             try:
                 while True:
-                    yield LOOP.run(async_iter.__anext__())
+                    yield LOOP.run(_anext(async_reader))
             except StopAsyncIteration:
                 return
 
-        return pa.RecordBatchReader.from_batches(async_iter.schema, iter_sync())
+        return pa.RecordBatchReader.from_batches(async_reader.schema, iter_sync())
 
     def to_arrow(self, timeout: Optional[timedelta] = None) -> pa.Table:
         """
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -3,12 +3,14 @@
 
 from __future__ import annotations
 
+import asyncio
 from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
-from enum import Enum
 from datetime import timedelta
+from enum import Enum
 from typing import (
     TYPE_CHECKING,
+    Any,
     Dict,
     List,
     Literal,
@@ -17,41 +19,40 @@ from typing import (
     Type,
     TypeVar,
     Union,
-    Any,
 )
 
-import asyncio
 import deprecation
 import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc
 import pydantic
+from typing_extensions import Annotated
 
-from lancedb.pydantic import PYDANTIC_VERSION
+from lancedb._lancedb import fts_query_to_json
 from lancedb.background_loop import LOOP
+from lancedb.pydantic import PYDANTIC_VERSION
 
 from . import __version__
 from .arrow import AsyncRecordBatchReader
 from .dependencies import pandas as pd
+from .expr import Expr
 from .rerankers.base import Reranker
 from .rerankers.rrf import RRFReranker
 from .rerankers.util import check_reranker_result
 from .util import flatten_columns
-from .expr import Expr
-from lancedb._lancedb import fts_query_to_json
-from typing_extensions import Annotated
 
 if TYPE_CHECKING:
     import sys
+
     import PIL
     import polars as pl
 
-    from ._lancedb import Query as LanceQuery
     from ._lancedb import FTSQuery as LanceFTSQuery
     from ._lancedb import HybridQuery as LanceHybridQuery
-    from ._lancedb import VectorQuery as LanceVectorQuery
-    from ._lancedb import TakeQuery as LanceTakeQuery
     from ._lancedb import PyQueryRequest
+    from ._lancedb import Query as LanceQuery
+    from ._lancedb import TakeQuery as LanceTakeQuery
+    from ._lancedb import VectorQuery as LanceVectorQuery
     from .common import VEC
     from .pydantic import LanceModel
     from .table import Table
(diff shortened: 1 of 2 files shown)


python/python/lancedb/query.py
```diff
@@ -3348,16 +3348,30 @@ def to_batches(
             If not specified, no timeout is applied. If the query does not
             complete within the specified time, an error will be raised.
         """
-        async_iter = LOOP.run(self._inner.execute(max_batch_length, timeout))
+
+        # ``self._inner`` is an ``AsyncQueryBase`` wrapping the native query
+        # object.  Calling ``execute`` on the native object directly from this
+        # thread fails (the pyo3 future needs a running loop), and ``execute``
+        # is not defined on ``AsyncQueryBase`` itself — so go through the
+        # async ``to_batches`` on the background loop.
```
westonpace: Is this comment adding value?  It seems to be more describing the bug.

westonpace (approved):
Calling to_batches instead of execute seems correct to me

python/python/lancedb/query.py
```diff
@@ -3348,16 +3348,30 @@ def to_batches(
...
+        async_reader = LOOP.run(_open())
+
+        async def _anext(it):
+            return await it.__anext__()
 
         def iter_sync():
             try:
                 while True:
-                    yield LOOP.run(async_iter.__anext__())
+                    yield LOOP.run(_anext(async_reader))
```
wjones127: nitpick: it's not clear to me why this is necessary. Shouldn't `async_iter.__anext__()` return a future that is equivalent to `_anext(async_reader)`?

python/python/lancedb/query.py
justinrmiller: Yeah I can remove it. Thanks.

python/python/lancedb/query.py
```diff
@@ -3348,16 +3349,25 @@ def to_batches(
...
             complete within the specified time, an error will be raised.
         """
-        async_iter = LOOP.run(self._inner.execute(max_batch_length, timeout))
+
+        async def _open():
+            return await self._inner.to_batches(
+                max_batch_length=max_batch_length, timeout=timeout
+            )
+
+        async_reader = LOOP.run(_open())
```
wjones127: nitpick: also here, why not just pass the future directly into `LOOP.run()`? Why create a closure?

python/python/lancedb/query.py
```diff
@@ -3348,16 +3349,25 @@ def to_batches(
             If not specified, no timeout is applied. If the query does not
             complete within the specified time, an error will be raised.
         """
-        async_iter = LOOP.run(self._inner.execute(max_batch_length, timeout))
```
wjones127, line 3351: question: Should we be worried about the call to `execute()` here too?

https://github.com/lancedb/lancedb/blob/d861debc81aa6e8981e5de5574ac83bc7fd50904/python/python/lancedb/query.py#L2304-L2308

If not, why not?

python/python/lancedb/query.py
justinrmiller: You're right — `AsyncRecordBatchReader.__anext__` is a regular Python coroutine method, so the returned coroutine can be passed straight to `LOOP.run`. Dropped in 30d640fb.

python/python/lancedb/query.py
justinrmiller: Same answer — `AsyncQueryBase.to_batches` is a plain `async def`, no loop needed at call time. Closure dropped in 30d640fb.

python/python/lancedb/query.py
justinrmiller: No — that `execute` is the working path. In `AsyncQueryBase.__init__`, `self._inner` is typed `Union[LanceQuery, LanceVectorQuery, LanceTakeQuery]` (the native pyo3 query objects), which do expose `execute`. The sync-wrapper bug was that `BaseQueryBuilder._inner` is an `AsyncQueryBase` (Python wrapper one layer up), not the native object — same attribute name at different layers.

PR #3426, opened 2026-05-21
test(rust): fix flaky env-var-dependent client tests

wjones127:
The `test_resolve_user_id_*` tests in `remote/client.rs` mutate the process-global `LANCEDB_USER_ID` and `LANCEDB_USER_ID_ENV_KEY` environment variables. cargo runs tests in a binary across multiple threads, so one test's `remove_var` can race another's `set_var` between when it's set and when `resolve_user_id()` reads it.

This surfaced as an intermittent failure of `test_resolve_user_id_from_env_key` on Windows CI:

```
assertion `left == right` failed
  left: None
 right: Some("custom-env-user-id")
```

Annotates the five env-mutating tests with `serial_test`'s `#[serial(user_id_env)]` so they run serially with respect to each other.

Should be backported to `release/v0.28` (CI for #3421 hit this same flake).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5084,6 +5084,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "serial_test",
  "snafu 0.8.9",
  "tempfile",
  "test-log",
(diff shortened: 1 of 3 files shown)


westonpace (commented):
I prefer the `serial_test` crate but this works too.

westonpace (approved):

PR #3427, opened 2026-05-21
fix: use releases API in check_lance_release.py

wjones127:
Previously `check_lance_release.py` used `git/refs/tags` with `--paginate --jq`, which drops the last page in some `gh` versions. The 7.x Lance tags all landed on the final (partial) page, causing the script to report `v6.0.1` as the latest and never triggering an update.

Switch to the releases API with `per_page=20`, which returns the 20 most recent releases sorted newest-first — one API call, no pagination needed.

### DIFF
--- a/ci/check_lance_release.py
+++ b/ci/check_lance_release.py
@@ -110,25 +110,25 @@ def fetch_remote_tags() -> List[TagInfo]:
             "api",
             "-X",
             "GET",
-            f"repos/{LANCE_REPO}/git/refs/tags",
-            "--paginate",
+            f"repos/{LANCE_REPO}/releases",
             "--jq",
-            ".[].ref",
+            ".[].tag_name",
+            "-F",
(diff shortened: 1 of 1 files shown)


ci/check_lance_release.py
```diff
@@ -112,25 +110,25 @@ def fetch_remote_tags() -> List[TagInfo]:
...
+            "-F",
+            "per_page=20",
         ]
     )
     tags: List[TagInfo] = []
     for line in output.splitlines():
-        ref = line.strip()
-        if not ref.startswith("refs/tags/v"):
+        tag = line.strip()
+        if not tag.startswith("v"):
```
wkalt, line 123: preexisting, but this rule seems too fragile. It would be best to at least assert this is vX.Y.Z or whatever convention we use.

wkalt (approved):

PR #3291, opened 2026-04-20
fix: allow appending arrow.json data into lance.json tables

yangmeilly:
Bug Description
fixes: https://github.com/lancedb/lancedb/issues/3144

  When a table is created with pa.json_() (PyArrow's JSON extension type), it is internally stored as lance.json (backed by LargeBinary with ARROW:extension:name: lance.json field metadata). However, calling
  table.add() with pa.json_() data fails with:

   1 RuntimeError: lance error: Append with different schema:
   2   `data` should have type json but type was large_binary

  Root Cause

  In rust/lancedb/src/table/datafusion/cast.rs, the build_field_exprs function sees that the input field type (Utf8 with arrow.json metadata) differs from the table field type (LargeBinary with lance.json metadata).
  Since can_cast_types(Utf8, LargeBinary) == true, it inserts a DataFusion CastExpr(Utf8 → LargeBinary).

  The problem is that this cast preserves the input field's extension metadata (arrow.json) rather than adopting the table field's metadata (lance.json). When lance-core receives a LargeBinary field with arrow.json
  metadata, it detects a schema mismatch and rejects the append.

  Fix

  Add a special case in build_field_exprs to detect when:
   - Input field is arrow.json (PyArrow pa.json_(), backed by Utf8/LargeUtf8)
   - Table field is lance.json (backed by LargeBinary)

  In this case, pass the expression through unchanged without casting. Lance-core's write path already handles the arrow.json → lance.json conversion (including JSONB encoding), so no DataFusion cast is needed.

  Changes

   - rust/lancedb/src/table/datafusion/cast.rs:
     - Added import for lance_arrow::json::{is_arrow_json_field, is_json_field}
     - Added special case in build_field_exprs to bypass casting for arrow.json → lance.json
     - Added unit test test_arrow_json_passthrough_to_lance_json

   - rust/lancedb/src/table/add_data.rs:
     - Added integration test test_add_arrow_json_into_lance_json_table

  Testing

   - Unit test verifies that cast_to_table_schema preserves arrow.json metadata when the table expects lance.json
   - Integration test verifies that table.add() with pa.json_() data succeeds on a table created with pa.json_()

PR #3429, opened 2026-05-21
fix: allow appending arrow.json data into lance.json tables

### LINKED ISSUE #3144: bug(rust): `table.add()` fails with schema mismatch when appending PyArrow JSON extension type data
### Bug description

`create_table()` accepts `pa.json_()` (PyArrow's `arrow.json` extension type) but internally lance-core converts it to `lance.json` (backed by `large_binary` with `ARROW:extension:name: lance.json` field metadata). However, `table.add()` does not account for this: the cast layer in lancedb tries to cast `Utf8 → LargeBinary`, and the resulting field carries the *input's* `arrow.json` metadata instead of the table's `lance.json` metadata. Lance-core then sees a schema mismatch (`json` vs `large_binary`) and rejects the append.

### Minimal reproduction

```python
import pyarrow as pa
import lancedb
import tempfile

with tempfile.TemporaryDirectory() as tmp:
    db = lancedb.connect(tmp)

    # Create table with pa.json_() — internally stored as lance.json
    schema = pa.schema([pa.field("data", pa.json_(), nullable=True)])
    tbl = db.create_table("t", schema=schema)

    # Verify internal conversion happened
    f = tbl.schema.field("data")
    assert f.type == pa.large_binary()
    assert f.metadata[b"ARROW:extension:name"] == b"lance.json"

    # Append pa.json_() data — this fails
    arr = pa.array([None, '{"a": 1}'], type=pa.json_())
    tbl.add(pa.table({"data": arr}))
```

**Error:**
```
RuntimeError: lance error: Append with different schema:
  `data` should have type json but type was large_binary
```

### Expected behavior

`table.add()` should accept `pa.json_()` data on a table that was created with `pa.json_()`.

### Fix

In `build_field_exprs`, detect when the input field is an `arrow.json` extension and the table field is `lance.json`, and pass through without casting. Lance-core's write path already handles the `arrow.json → lance.json` conversion (including JSONB encoding):

```rust
if is_arrow_json_field(input_field) && is_json_field(table_field) {
    result.push((input_expr, input_field.clone()));
    continue;
}
```

### Workaround

Manually produce `large_binary` arrays with `lance.json` field metadata before calling `add()`:

```python
field = pa.field("data", pa.large_binary(), nullable=True, metadata={
    b"ARROW:extension:name": b"lance.json",
    b"ARROW:extension:metadata": b"",
})
arr = pa.array([None, b'{"a": 1}'], type=pa.large_binary())
tbl.add(pa.table({"data": arr}, schema=pa.schema([field])))
```

### Versions

- `lancedb==0.30.0b5`, `pyarrow==23.0.1`

wjones127:
When a table is created with `pa.json_()` (PyArrow's JSON extension type),
it is stored internally as `lance.json` (LargeBinary with `lance.json`
extension metadata). Calling `table.add()` with `pa.json_()` data failed
with:

```
RuntimeError: lance error: Append with different schema:
  `data` should have type json but type was large_binary
```

`build_field_exprs` in `rust/lancedb/src/table/datafusion/cast.rs` saw that
the input field (`Utf8` with `arrow.json` metadata) differed from the table
field (`LargeBinary` with `lance.json` metadata). Since
`can_cast_types(Utf8, LargeBinary)` is true, it inserted a DataFusion
`Utf8 → LargeBinary` cast. That cast preserved the input field's `arrow.json`
extension metadata instead of adopting the table's `lance.json` metadata, so
lance-core detected a schema mismatch and rejected the append.

This adds a special case in `build_field_exprs`: when the input is
`arrow.json` and the table field is `lance.json`, the expression is passed
through unchanged. Lance-core's write path already handles the
`arrow.json → lance.json` conversion (including JSONB encoding), so no
DataFusion cast is needed.

Fixes #3144

Continues #3291 from a fork (the original author's branch could not be
pushed to). The original commits are preserved; an additional commit fixes
the CI failures on that PR — formatting, a missing trait import, and
read-back assertions that assumed binary storage when a lance.json column
is read back as `Utf8`.

### DIFF
--- a/rust/lancedb/src/table/add_data.rs
+++ b/rust/lancedb/src/table/add_data.rs
@@ -993,6 +993,7 @@ mod tests {
     /// `"json vs large_binary" schema mismatch`.
     #[tokio::test]
     async fn test_add_arrow_json_into_lance_json_table() {
+        use arrow_array::{Array, cast::AsArray};
         use lance_arrow::json::{ARROW_JSON_EXT_NAME, JSON_EXT_NAME};
 
         // Build a table whose "data" column is lance.json (LargeBinary +
@@ -1013,7 +1014,10 @@ mod tests {
         let data_field = stored_field.field_with_name("data").unwrap();
         assert_eq!(data_field.data_type(), &DataType::LargeBinary);
         assert_eq!(
-            data_field.metadata().get("ARROW:extension:name").map(|s| s.as_str()),
+            data_field
+                .metadata()
+                .get("ARROW:extension:name")
+                .map(|s| s.as_str()),
             Some(JSON_EXT_NAME),
         );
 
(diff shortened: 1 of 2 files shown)


westonpace:
Feedback from Claude.  None of this seems required:

## Suggestions / Minor Issues

1. **Test duplication** (`add_data.rs`): `test_add_arrow_json_into_lance_json_table` and `test_add_arrow_json_large_utf8_into_lance_json_table` are ~95% identical. Consider extracting a helper parameterized on `DataType::Utf8` vs `LargeUtf8` and the array type. Not a blocker — explicitness is fine here.

2. **Metadata construction**: The tests build the metadata `HashMap` imperatively:
   ```rust
   let mut arrow_json_metadata = std::collections::HashMap::new();
   arrow_json_metadata.insert("ARROW:extension:name".to_string(), ARROW_JSON_EXT_NAME.to_string());
   ```
   `HashMap::from([(k, v)])` would be tidier, and the unit test in `cast.rs` already uses the `ARROW_EXT_NAME_KEY` constant — the two `add_data.rs` tests still hard-code the string `"ARROW:extension:name"`. Worth unifying.

3. **Asymmetric direction not addressed**: The fix only handles `arrow.json → lance.json`. The reverse (input is `lance.json` typed, table is `arrow.json` typed) isn't covered. In practice that's unlikely to occur for `add()` (you'd need an Arrow producer emitting `LargeBinary + lance.json` directly), so leaving it alone is reasonable — but worth a brief comment near the special case noting the asymmetry, or a follow-up issue.

4. **`is_arrow_json_field` semantics**: The fix relies on `is_arrow_json_field` accepting both `Utf8` and `LargeUtf8` backing types. The `LargeUtf8` test exercises this, but the unit test in `cast.rs` only covers `Utf8`. Consider adding a `LargeUtf8` case to the unit test as well, so a future change to `is_arrow_json_field`'s scope is caught at the projection layer too.

5. **Doc comment for the second test** says "LargeUtf8 ... is the default for PyArrow's `pa.large_string()` or when creating `pa.json_()` with certain versions." The PyArrow-version remark is a bit vague — if it's documenting real behavior, a version hint or link would help future maintainers; otherwise consider trimming.

wjones127:
Thanks @westonpace, I've addressed all but (3).

brendanclement (approved):
Rev 2 looks good! Thanks for bringing this over the line

PR #3430, opened 2026-05-22
fix: canonicalize remote nested field paths

### LINKED ISSUE #3407: Lock remote nested field index and search contract
Parent: lancedb/lancedb#3401

## Summary

Remote table APIs need the same nested field-path behavior as local tables for create index, search, list, and stats.

## Scope

* Define remote request/response semantics for nested field paths and escaped paths.
* Fix `Index::Auto` or make its nested-field limitations explicit.
* Ensure explicit remote create-index requests preserve canonical field paths.
* Align remote `list_indices()`, `index_stats()`, vector search, and FTS query behavior with local tables.

## Acceptance criteria

* Remote scalar/vector/FTS index creation accepts nested field paths where local does.
* Remote list/stats responses return canonical full paths.
* Remote explicit vector and FTS search can target nested fields.
* Local and remote tests share the same field-name matrix where practical.

Xuanwo:
Fixes #3407.

Remote tables now resolve create-index field paths against the table schema before sending requests, so nested, escaped, and case-insensitive inputs use the same canonical path contract as local tables. Remote `list_indices()` also canonicalizes returned columns against the current schema, and the remote query tests lock explicit nested vector and FTS request payloads.

### DIFF
--- a/python/python/tests/test_remote_db.py
+++ b/python/python/tests/test_remote_db.py
@@ -362,6 +362,22 @@ def test_table_create_indices():
                     schema=dict(
                         fields=[
                             dict(name="id", type={"type": "int64"}, nullable=False),
+                            dict(name="text", type={"type": "string"}, nullable=False),
+                            dict(
+                                name="vector",
+                                type={
+                                    "type": "fixed_size_list",
+                                    "fields": [
+                                        dict(
+                                            name="item",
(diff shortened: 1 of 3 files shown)


BubbleCal (approved):

Xuanwo:
The rust failure is not related to this PR.

PR #3432, opened 2026-05-22
feat(python): support remote tables in PyTorch dataloaders

Xuanwo:
This PR makes remote LanceDB tables usable from PyTorch multiprocessing workers. Remote tables now carry enough safe JSON connection state to reopen themselves after pickle/spawn or fork, and permutations lazily rebuild their reader from restored tables instead of trying to reuse process-local handles.

This addresses the remote-table gap in the PyTorch dataset path while preserving the explicit connection factory escape hatch for custom worker-side credential loading or non-serializable header providers.

Validated with targeted remote table, permutation, and PyTorch DataLoader tests.


python/python/lancedb/permutation.py
```diff
@@ -594,19 +635,24 @@ def __getstate__(self) -> dict[str, Any]:
...
+        Arrow IPC natively); otherwise, the permutation table uses its own
+        reopen state too. The reader is dropped from the wire format and
+        rebuilt lazily on first use.
         """
         permutation_data: Optional[pa.Table] = None
+        permutation_table_state: Optional[dict[str, Any]] = None
         if self.permutation_table is not None:
-            permutation_data = self.permutation_table.to_arrow()
+            try:
+                permutation_data = self.permutation_table.to_arrow()
```
westonpace: When would this not be implemented?  We stripped out the ability to persist permutation tables I think so this should always be an in-memory table.  And even if it wasn't, to_arrow is defined on all table types.

python/python/lancedb/remote/db.py
```diff
@@ -109,6 +173,20 @@ def __init__(
...
+            {
+                "connection_type": "remote",
+                "db_url": self.db_url,
+                "api_key": self.api_key,
+                "region": self.region,
+                "host_override": self.host_override,
+                "client_config": _client_config_to_dict(self.client_config),
+                "storage_options": self.storage_options,
+            }
+        )
```
westonpace, line 190: I know the `serialize` method already existed but why isn't this implemented at the rust level?

python/python/lancedb/remote/table.py
```diff
@@ -49,14 +50,78 @@ def __init__(
...
+    @_table.setter
+    def _table(self, table: AsyncTable) -> None:
+        self._table_handle = table
+        self._name = table.name
+        self._pid = os.getpid()
+
+    def _ensure_open(self) -> None:
+        pid = os.getpid()
+        if self._table_handle is not None and self._pid == pid:
+            return
```
westonpace, line 104: Why do we need to key off the pid?  Is it not sufficient that `_table_handle` is None?  Or are you trying to add support for both pickling and forking here?

python/python/lancedb/permutation.py
```diff
@@ -696,10 +718,26 @@ def __setstate__(self, state: dict[str, Any]) -> None:
...
+        if self.permutation_table is not None and hasattr(
+            self.permutation_table, "_ensure_open"
+        ):
+            self.permutation_table._ensure_open()
         self.reader = LOOP.run(self._build_reader())
+        self._pid = pid
 
     @property
     def schema(self) -> pa.Schema:
+        self._ensure_open()
```
westonpace, line 731: I'm not really a fan of needing to do this everywhere.  Either:

* Unpickling a table should open the table (e.g. don't be lazy)
* Lazy initialization of the table should be the table's concern (e.g. `_ensure_open` should be spread throughout `RemoteTable`, not in `Permutation`).

Maybe the unpickling could return some kind of auto-proxy object that opens the connection on first method call.

The problem with the current approach is that `RemoteTable` now has `getstate/setstate` but if you used those methods outside the context of a `Permutation` it would fail.

python/python/lancedb/permutation.py
```diff
@@ -354,6 +355,49 @@ def impl(batch: pa.RecordBatch) -> pl.DataFrame:
...
+        }
+
+    return {
+        "kind": "local",
+        "name": table.name,
+        "uri": base_uri,
+        "namespace": table._namespace_path,
+        "storage_options": table._conn.storage_options,
+    }
+
```
westonpace, line 385: This is a bit asymmetric now.  The logic for serializing and deserializing a `LanceTable` is here (in `Permutation`) but the logic for serializing and deserializing a `RemoteTable` is over there (in the getstate/setstate methods).  We should probably just add getstate/setstate to `LanceTable` at this point.

However, this could be done in a follow-up.

westonpace (approved):
I have some opinions.  I think my main opinion is that it would be nice for the serialization / deserialization to move to Rust and for `Table`, `LanceTable`, and `RemoteTable` to all just use that for getstate / setstate.

python/python/lancedb/permutation.py
Xuanwo: Good point. Removed the fallback and now always use `to_arrow()` for the permutation table.

python/python/lancedb/remote/db.py
Xuanwo: Agreed, that would be cleaner. I kept this PR on the existing Python `serialize()` path to avoid expanding scope. Filed #3472 to track moving this into shared Rust-backed table state.

python/python/lancedb/remote/table.py
Xuanwo: Yes, this is for both pickle and fork. Pickle clears `_table_handle`, but fork inherits a non-None handle from the parent process, so PID is needed to force reopen in the child. Added a short comment.

python/python/lancedb/permutation.py
Xuanwo: RemoteTable already handles its own lazy reopen. This `_ensure_open()` is for rebuilding `PermutationReader`, which owns Rust-side handles and can be stale after unpickle/fork. Added a comment to clarify.

python/python/lancedb/permutation.py
Xuanwo: Agreed. I kept it here to avoid expanding this PR, and filed #3473 to move local table pickle state onto `LanceTable`.

PR #3437, opened 2026-05-22
fix(rerankers): inverted scores and incorrect missing-FTS penalty in LinearCombinationReranker

### LINKED ISSUE #3154: bug(python): LinearCombinationReranker produces incorrect relevance ranking in hybrid search
### LanceDB version

main branch

### What happened?

`LinearCombinationReranker` produces incorrect hybrid search rankings:
1. The relevance ranking is inverted, the least relevant doc rank first.
2. Documents missing an FTS score are rewarded instead of being penalized, while docs missing a vector score are correctly penalized.

```
import shutil
import tempfile

import lancedb, pyarrow as pa
from lancedb.rerankers import LinearCombinationReranker

DB_PATH = tempfile.mkdtemp(prefix="testdb")

try:
    db = lancedb.connect(DB_PATH)

    data = pa.table({
        "text": ["apple orange", "banana grape", "cherry melon", "date fig"],
        "vector": [[0.1, 0.2], [0.9, 0.8], [0.5, 0.5], [0.3, 0.4]],
    })
    table = db.create_table("text", data=data, mode="overwrite")
    table.create_fts_index("text", replace=True)

    reranker = LinearCombinationReranker(weight=0.7, return_score="all")
    results = (
        .vector([0.1, 0.2])
        .text("apple orange")
        .rerank(reranker)
        .to_list()
    )
    for r in results:
        print(r["text"], r["_relevance_score"])

finally:
    shutil.rmtree(DB_PATH, ignore_errors=True)
```

**output:**
```
banana grape None 0.9999999403953552 0.699999988079071
apple orange 2.4079456329345703 0.0 0.30000001192092896
cherry melon None 0.25 0.17500002682209015
date fig None 0.08000001311302185 0.056000012904405594
```

**Expected:** `apple orange` should rank first, it's the closest vector match and the only FTS match.

devteamaegis:
## Problem

`LinearCombinationReranker.merge_results` has two related bugs that make it return **inverted relevance rankings** — the least relevant document ranks first (closes #3154).

### Bug 1 — `_combine_score` subtracts from 1, inverting the final ranking

```python
def _combine_score(self, vector_score, fts_score):
    return 1 - (self.weight * vector_score + (1 - self.weight) * fts_score)
```

Both `vector_score` (already converted via `_invert_score`) and `fts_score` (BM25 relevance) are in **higher-is-better** space. Wrapping the weighted average in `1 - (...)` flips the direction: a perfectly matching document (`vector_score=1, fts_score=1`) gets `_relevance_score = 0.0`, while a non-matching document gets a high score.

### Bug 2 — Documents missing an FTS score are rewarded, not penalised

```python
fts_score = result.get("_score", fill)  # fill=1.0 by default
```

When a document has no FTS match, `fts_score = fill = 1.0`. In `_combine_score` (with the bug-1 formula), this large value becomes a **negative penalty** via `1 - (... + 0.3 * 1.0)`, counterintuitively *boosting* the document's score. By contrast, missing vector results correctly receive `_invert_score(fill) = 0.0` (penalised).

## Fix

**Bug 1** — remove the `1 -` inversion from `_combine_score`:

```python
def _combine_score(self, vector_score, fts_score):
    return self.weight * vector_score + (1 - self.weight) * fts_score
```

**Bug 2** — use `1 - fill` for missing FTS scores so both penalties are symmetric (mirror of what `_invert_score(fill)` already does for missing vector scores):

```python
fts_score = result.get("_score", 1 - fill)  # was: fill
```

With `fill=1.0` (default): `1 - 1.0 = 0.0` — missing-FTS entries contribute `0` to the FTS term, identical to how missing-vector entries contribute `0` to the vector term.

## Verification

Concrete example from the issue. With `weight=0.7`, `fill=1.0`:

| Document | `_distance` | `_score` | Old `_relevance_score` | New `_relevance_score` |
|----------|-------------|----------|------------------------|------------------------|
| `apple orange` | 0.0 (best) | 2.41 (only FTS) | 0.30 (**wrong: ranked 2nd**) | 1.42 (**correct: ranked 1st**) |
| `banana grape` | 0.9999 (worst) | — | 0.70 (**wrong: ranked 1st**) | 0.00 (**correct: ranked last**) |

## Tests

Two regression tests added to `python/python/tests/test_rerankers.py`:

- `test_linear_combination_best_match_ranks_first` — the document with the smallest distance **and** an FTS match must have the highest `_relevance_score`.
- `test_linear_combination_missing_fts_is_penalised` — a document with any FTS score must beat an otherwise-equal document with no FTS match.

### DIFF
--- a/python/python/lancedb/rerankers/linear_combination.py
+++ b/python/python/lancedb/rerankers/linear_combination.py
@@ -102,8 +102,15 @@ class LinearCombinationReranker(Reranker):
 
         combined_list = []
         for row_id, result in results.items():
+            # Convert vector distance to a relevance score in [0, 1] where
+            # higher is better.  Missing vector entries are penalised with
+            # `_invert_score(fill)` = 1 - fill (= 0.0 for the default fill=1).
             vector_score = self._invert_score(result.get("_distance", fill))
-            fts_score = result.get("_score", fill)
+            # FTS scores (BM25) are already in a "higher = more relevant" space.
+            # Missing FTS entries are penalised symmetrically: we use
+            # `1 - fill` so that the same `fill` value drives both missing-vector
+            # and missing-FTS penalties in the same direction.
+            fts_score = result.get("_score", 1 - fill)
             result["_relevance_score"] = self._combine_score(vector_score, fts_score)
             combined_list.append(result)
 
@@ -123,8 +130,12 @@ class LinearCombinationReranker(Reranker):
         return tbl
 
     def _combine_score(self, vector_score, fts_score):
-        # these scores represent distance
-        return 1 - (self.weight * vector_score + (1 - self.weight) * fts_score)
+        # Both vector_score (inverted distance) and fts_score are in a
+        # "higher = more relevant" space.  A straight weighted average gives
+        # higher _relevance_score to better matches, as expected.
+        # Previously this returned `1 - (...)` which inverted the final
+        # ranking so that the *least* relevant document ranked first.
+        return self.weight * vector_score + (1 - self.weight) * fts_score
 
     def _invert_score(self, dist: float):
         # Invert the score between relevance and distance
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This seems good. I appreciate the clear tests; they show very obvious cases.

Issue #3370, opened 2026-05-12
Make `read_consistency_interval` behavior aligned between OSS and Remote


wjones127:
There is inconsistency in the handling of the consistency interval and related functions between OSS and Remote:

| Behavior | OSS | Remote |
|--------|--------|--------|
| `checkout_latest()` | Loads the latest version of the table | Just drops the version pin, server might still provide older table version |
| `checkout_version()` | Fixes the table version to given version | Same |
| `read_consistency_interval=None` | Never checks for updates, except when explicitly asked | Ignored |
| `read_consistency_interval=0` | Check for update on every request | Ignored |
| `read_consistency_interval=5s` | Check for update at least every 5s | Ignored | 

We'd like to bring this into alignment.

## Proposal

Add server support for two headers:

1. `LanceDB-Min-Version`: minimum table version to provide. This is used to guarantee we are reading data newer than what we've written.
2. `LanceDB-Min-Timestamp`: minimum table version created at to use. This is for situations like `checkout_latest()` where we just want to specify at a given time, but don't know what the particular version is.

Then we can make `Remote` match the OSS behavior by:

* In `checkout_latest()` save the current timestamp and pass it in `LanceDB-Min-Timestamp` in subsequent calls.
* For `read_consistency_interval=0`, always pass `LanceDB-Min-Timestamp` with the current timestamp
* For `read_consistency_interval=5s`, keep passing an increasing timestamp in `LanceDB-Min-Timestamp`

PR #3439, opened 2026-05-23
feat(remote): send read freshness headers for remote table consistency

brendanclement:
Closes client side work of #3370 

### Summary
- Plumbs `read_consistency_interval` from `ConnectBuilder` through `RestfulLanceDbClient` so remote reads attach an `x-lancedb-min-timestamp` freshness header. None = no header (default), zero = "now", positive = `now - interval`.
- Adds per-table `FreshnessState` on `RemoteTable`: write responses (`update`, `delete`, `merge_insert`, `add_columns`, `alter_columns`, `drop_columns`) track the committed version, and the next read sends `x-lancedb-min-version` so the server's cache honors read-your-write.
- `checkout(v)` / `checkout_tag(t)` / `checkout_latest()` / `restore()` reset the freshness state appropriately; the validating `/describe/` and tag-resolve requests are sent without freshness headers so they don't carry stale state.
- Updates Rust, Python, and Node docstrings and calls out that stronger consistency raises per-read latency and cost.

### Testing
- Unit tests cover default behavior, interval=0, positive interval, checkout_latest baseline, min_version-after-write, checkout clears state, and the two no-stale-header invariants on `checkout(v)` and `checkout_tag(t)`.
- Ran smoke tests against local remote table to verify functionality

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -75,14 +75,23 @@ const SCHEMA_CACHE_REFRESH_WINDOW: Duration = Duration::from_secs(5);
 
 /// Per-table state driving the freshness headers (`x-lancedb-min-version` and
 /// `x-lancedb-min-timestamp`) sent on read requests.
-///
-/// `min_version` provides read-your-write within a single handle: writes that
-/// return a version update it, and reads send it. `checkout_baseline` is the
-/// wall-clock time captured at the last [`BaseTable::checkout_latest`] call;
-/// reads send `max(baseline, now - read_consistency_interval)`.
 #[derive(Debug, Default, Clone, Copy)]
 struct FreshnessState {
+    /// Provides read-your-write within a single handle: writes that return a
+    /// version update this, and reads send it as `x-lancedb-min-version`.
     min_version: Option<u64>,
+    /// Wall-clock time captured at the last [`BaseTable::checkout_latest`]
+    /// call. Subsequent reads send
+    /// `max(baseline, now - read_consistency_interval)` as
+    /// `x-lancedb-min-timestamp`.
+    ///
+    /// Without this, `checkout_latest()` would have no effect on subsequent
+    /// reads when `read_consistency_interval` is unset (the default): a
+    /// server-side cache could still serve a snapshot older than the moment
+    /// the user explicitly asked for "latest". The baseline forces the
+    /// server to skip any cache entry older than the checkout time, so the
+    /// `checkout_latest()` signal is preserved across reads on the same
+    /// handle regardless of the configured consistency interval.
     checkout_baseline: Option<SystemTime>,
 }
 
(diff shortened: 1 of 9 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -62,15 +62,67 @@ use std::collections::HashMap;
...
+///
+/// `min_version` provides read-your-write within a single handle: writes that
+/// return a version update it, and reads send it. `checkout_baseline` is the
+/// wall-clock time captured at the last [`BaseTable::checkout_latest`] call;
+/// reads send `max(baseline, now - read_consistency_interval)`.
+#[derive(Debug, Default, Clone, Copy)]
+struct FreshnessState {
+    min_version: Option<u64>,
+    checkout_baseline: Option<SystemTime>,
+}
```
westonpace: Minor doc nit:

```suggestion
#[derive(Debug, Default, Clone, Copy)]
struct FreshnessState {
    /// Provides read-your-write within a single handle, writes that
    /// return a version update this, and reads send it.
    min_version: Option<u64>,
    /// The wall-clock time captured at the last [`BaseTable::checkout_latest`]
    /// call.  Reads send `max(baseline, now - read_consistency_interval)`.
    checkout_baseline: Option<SystemTime>,
}
```

rust/lancedb/src/remote/table.rs
```diff
@@ -62,15 +62,67 @@ use std::collections::HashMap;
...
 const SCHEMA_CACHE_TTL: Duration = Duration::from_secs(30);
 const SCHEMA_CACHE_REFRESH_WINDOW: Duration = Duration::from_secs(5);
 
+/// Per-table state driving the freshness headers (`x-lancedb-min-version` and
+/// `x-lancedb-min-timestamp`) sent on read requests.
+///
+/// `min_version` provides read-your-write within a single handle: writes that
+/// return a version update it, and reads send it. `checkout_baseline` is the
+/// wall-clock time captured at the last [`BaseTable::checkout_latest`] call;
+/// reads send `max(baseline, now - read_consistency_interval)`.
```
westonpace: This doesn't describe the "why", currently it is only explaining the "what".  Why do we need checkout_baseline?  Is there some concern that the server will regress the version after we've done a checkout?  Or is this maybe for the case where the server is doing load balancing and we might hit the checkout on node A and then if a future request hits node B we want to make sure node B is at least as up-to-date as node A?

westonpace (approved):
Does this need to be added to typescript as well?

Looks great though, just nits/questions.

rust/lancedb/src/remote/table.rs
brendanclement: Added to the docs but yes this is for the second case and we want to ensure that any future request that hits node B is at least as up to date as node A

brendanclement:
> Does this need to be added to typescript as well?
> 
> Looks great though, just nits/questions.

read_consistency_interval is already part of the typescript API surface and it's plumbed through to the rust core. I ran E2E smoke tests with the TS sdk against a remote table and it functioned correctly.  We just needed to update the docs

PR #3443, opened 2026-05-24
fix(rerankers): make MRRReranker.rerank_multivector divide by total system count, not appearances

devteamaegis:
## Summary

`MRRReranker.rerank_multivector` computed the Mean Reciprocal Rank (MRR) score using `np.mean(reciprocal_ranks)` for each document, where `reciprocal_ranks` only contained entries for the systems in which the document appeared.

MRR is defined as the average reciprocal rank across **all** ranking systems, with a reciprocal rank of 0 for systems where a document is absent. `np.mean` divides by the count of appearances rather than the total system count, inflating scores for documents with partial coverage and inverting rankings.

### Concrete counter-example (3 systems)

| Document | Systems appeared | Buggy score | Correct score |
|----------|-----------------|-------------|---------------|
| A (rank 1 in system 1 only) | 1 of 3 | `mean([1.0])` = **1.0** | `1.0 / 3` = **0.333** |
| B (rank 2 in system 1, rank 1 in systems 2 & 3) | 3 of 3 | `mean([0.5, 1.0, 1.0])` = **0.833** | `2.5 / 3` = **0.833** |

With the bug: A=1.0 > B=0.833 → A wins (wrong).  
With the fix: B=0.833 > A=0.333 → B wins (correct).

Note that `rerank_hybrid` already handles this correctly — it initialises `vector_rr = 0.0` and `fts_rr = 0.0` before the loop, so missing results contribute zero.

## Fix

Replace:
```python
mean_rr = np.mean(reciprocal_ranks)
```
with:
```python
mean_rr = sum(reciprocal_ranks) / n_systems
```
where `n_systems = len(vector_results)` is captured once before the loop.

## Test

`test_mrr_reranker_multivector_partial_coverage` in `tests/test_rerankers.py` encodes the counter-example above using pure PyArrow tables (no database required) and asserts both the ordering and the absolute score values.

### DIFF
--- a/python/python/lancedb/rerankers/mrr.py
+++ b/python/python/lancedb/rerankers/mrr.py
@@ -145,6 +145,7 @@ class MRRReranker(Reranker):
                     `search().with_row_id(True)`"
             )
 
+        n_systems = len(vector_results)
         mrr_score_map = defaultdict(list)
 
         for result_table in vector_results:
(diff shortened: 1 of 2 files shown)


wjones127 (commented):
Thank you. This is well written. Happy to approve and merge once CI is passing, looks like it's your unit test that is failing.

wjones127 (changes requested):
Just waiting on resolving conflicts.

wjones127:
Fixed by #3599.

PR #3455, opened 2026-05-27
ci: drop manylinux2_17 wheel builds

### LINKED ISSUE #3452: Drop manylinux2_17 builds
We already build for manylinux2_28. manylinux2_17 reached EOL in 2024. pyarrow stopped publishing new manylinux 2_17 many versions ago, for example.

wjones127:
manylinux2_17 reached EOL in 2024 and pyarrow stopped publishing 2_17 wheels long ago. We already build manylinux2_28 wheels, so drop the 2_17 matrix entries.

Fixes #3452

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -21,18 +21,10 @@ jobs:
       matrix:
         config:
           - platform: x86_64
-            manylinux: "2_17"
-            extra_args: ""
-            runner: ubuntu-22.04
-          - platform: x86_64
             manylinux: "2_28"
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Should we add `manylinux_2_34`?  Though RHEL8 should get security fixes into 2029 so maybe no rush...

brendanclement (approved):
All red PRs are too rare and my favourite type of PR.

Thanks for getting this out

wjones127:
> Should we add `manylinux_2_34`?

Do you think we should care much about the glibc version? It seems like both versions of the crate use gcc 14, so that's the same at least.

PR #3458, opened 2026-05-28
refactor(remote): drop x-lancedb-min-version; use timestamp baseline for strong consistency

brendanclement:
### Summary
Drops the x-lancedb-min-version freshness header on the client side. Reads still send x-lancedb-min-timestamp. min_version gave read-your-write by passing back the version a write returned and asking the server to serve at least that version. 

That doesn't hold up under the WAL: a WAL write returns a WAL entry id, not a Lance dataset version, so the value we'd send back no longer matches what the server compares against. Instead, writes now bump a per-table timestamp baseline (renamed from the old checkout_latest baseline), and the next read sends that as min_timestamp.

### Testing
Unit tests cover the timestamp-baseline-after-write path, the
read_consistency_interval cases, and that checkout(v)/checkout_tag don't carry a stale baseline.

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -66,53 +66,50 @@ use std::time::{Duration, SystemTime};
 use tokio::sync::RwLock;
 
 const REQUEST_TIMEOUT_HEADER: HeaderName = HeaderName::from_static("x-request-timeout-ms");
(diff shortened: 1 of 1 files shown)


brendanclement:
@claude review once

brendanclement:
@claude review once

wjones127:
@brendanclement is this still relevant or should we close?

PR #3459, opened 2026-05-29
fix(python): run AsyncTable.search embeddings on a dedicated executor

### LINKED ISSUE #3310: AsyncTable.search() dispatches embedding to the default executor, starving other async I/O under load
## Problem

Complementary to #3268. That issue points out `AsyncTable.search()` at least wraps the embedding call in `run_in_executor`, while `AsyncTable.add()` doesn't. However, the `search()` side still has a separate problem under concurrent load: it uses the default executor.

## Evidence

`lancedb/table.py` — `AsyncTable.search()`:

```python
async def make_embedding(embedding, query):
    if embedding is not None:
        loop = asyncio.get_running_loop()
        # This function is likely to block, since it either calls an expensive
        # function or makes an HTTP request to an embeddings REST API.
        return (
            await loop.run_in_executor(
                None,                                       # <-- default executor
                embedding.function.compute_query_embeddings_with_retry,
                query,
            )
        )[0]
```

Passing `None` uses `asyncio`'s default `ThreadPoolExecutor`, which is:

1. **Shared with every other `run_in_executor(None, …)` / `asyncio.to_thread(…)` caller in the process** — rerankers, pyarrow, fsspec, `boto3` elsewhere, user code.
2. Sized at `min(32, cpu_count + 4)` by default — as small as ~10 on low-vCPU containers.

Under concurrent load (e.g. 100 users against a FastAPI service), a blocking embedding provider (Bedrock, OpenAI, Cohere over sync HTTP) will saturate that shared pool. We measured query throughput drop from ~60 RPS (fake embeddings) to ~10 RPS (real Bedrock `invoke_model` via sync `boto3`), with the default pool being the bottleneck.

## Suggested improvements

1. **Short-term** — expose an override. Let `AsyncTable.search()` (and the future async `add()` path from #3268) accept an explicit `executor` argument, falling back to `None` for today's behavior. Also expose a module-level setter (e.g. `lancedb.set_embedding_executor(pool)`) so an app can isolate embedding I/O on a dedicated, sized pool without threading the arg through every call site.
2. **Long-term** — the async ABC from #3268 eliminates the executor entirely for providers with real async clients. Pair it with (1) so the degraded sync fallback is still isolated from the shared default pool.

## Environment

Same as #3268 — lancedb 0.30.2, Python 3.13, FastAPI.

Ar-maan05:
## Summary
  
  `AsyncTable.search()` computes the query embedding with `loop.run_in_executor(None, ...)`, which uses asyncio's **default** `ThreadPoolExecutor`. That pool is shared with all other `run_in_executor(None, ...)` work, so a slow embedding call — a heavy local model or an HTTP request to an embeddings API — ties up those threads and starves unrelated async I/O under concurrent load.
  
  This moves the (potentially blocking) embedding call onto a **dedicated executor**, isolating it from the default pool.
  
  Closes #3310.
  
  ## Problem

  `python/lancedb/table.py`, `AsyncTable.search()`:

  ```python
  return (
      await loop.run_in_executor(
          None,  # asyncio's default executor, shared with other blocking I/O
          embedding.function.compute_query_embeddings_with_retry,
          query,
      )   
  )[0]
  ```
  
  Under load, concurrent searches whose embeddings block (or any other code using the default executor) contend for the same small thread pool.
  
  ## Change

  - Add a dedicated `ThreadPoolExecutor(thread_name_prefix="lancedb-embedding")` in `background_loop.py`, exposed via `embedding_executor()`.
  - Use it in `AsyncTable.search()`'s `make_embedding` instead of the default executor.
  - Reset the executor in the existing `_reset_after_fork` hook — its worker threads don't survive `fork()`, same as the background event loop. It's recreated lazily, so this is cheap.

  ## Design notes
  
  The issue asked whether maintainers preferred a configurable executor, a dedicated internal one, or another approach (no response in the thread). I went with a **dedicated internal executor**: it fixes the starvation with no public API change and stays consistent with the existing `LOOP` singleton. Making the pool size configurable would be an easy follow-up if preferred.
  
  Scope is limited to `search()`. The broader "embedding functions need real async support" (including `add()`) is tracked separately in #3268.
  
  ## Testing
  
  - Added `test_async_search_runs_embedding_on_dedicated_executor`: patches the embedding function to record the executing thread during an async search and asserts it runs on a `lancedb-embedding` thread. Verified it **fails** against the previous `run_in_executor(None, ...)` and passes with the fix.
  - `ruff format`, `ruff check`, and `pyright` pass on the changed files.

### DIFF
--- a/python/python/lancedb/background_loop.py
+++ b/python/python/lancedb/background_loop.py
@@ -2,6 +2,7 @@
 # SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
 import asyncio
+import concurrent.futures
 import os
 import threading
 import warnings
@@ -37,6 +38,24 @@ class BackgroundEventLoop:
 
 LOOP = BackgroundEventLoop()
 
+
+def _new_embedding_executor() -> concurrent.futures.ThreadPoolExecutor:
+    return concurrent.futures.ThreadPoolExecutor(thread_name_prefix="lancedb-embedding")
+
+
+# Embedding functions can block for a long time -- a heavy local model or an
+# HTTP request to a remote embeddings API. Running them on asyncio's default
+# executor lets them starve the unrelated blocking I/O that shares that pool,
+# so they get a dedicated one. See
+# https://github.com/lancedb/lancedb/issues/3310.
+_EMBEDDING_EXECUTOR = _new_embedding_executor()
+
+
+def embedding_executor() -> concurrent.futures.ThreadPoolExecutor:
+    """Return the executor dedicated to running blocking embedding calls."""
+    return _EMBEDDING_EXECUTOR
+
+
 _FORK_WARNED = False
 
 
@@ -47,6 +66,12 @@ def _reset_after_fork():
     # the new state. The Rust-side tokio runtime is reset analogously by a
     # pthread_atfork hook installed in the _lancedb extension.
     LOOP._start()
+    # The embedding executor's worker threads are dead in the child as well.
+    # Replace it with a fresh pool (threads are spawned lazily, so this is
+    # cheap); we don't shut down the old one, since joining its dead workers
+    # could hang.
+    global _EMBEDDING_EXECUTOR
+    _EMBEDDING_EXECUTOR = _new_embedding_executor()
     global _FORK_WARNED
     if not _FORK_WARNED:
         _FORK_WARNED = True
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -30,7 +30,7 @@ from lancedb.scannable import _register_optional_converters, to_scannable
 
 from . import __version__
 from lancedb.arrow import peek_reader
-from lancedb.background_loop import LOOP
+from lancedb.background_loop import LOOP, embedding_executor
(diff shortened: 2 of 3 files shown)


Ar-maan05:
Friendly ping! Let me know if any changes are needed.

wjones127 (approved):
This looks great, assuming all CI is passing :)

Ar-maan05:
@wjones127 thank you for the approval! The only CI failure was `test_remote_db.py::test_remote_create_index_new_api` (`Schema Error: Field path 'vector' not found`), which is unrelated to this change. This branch was based on #2882, which updated the create-index API but left that test's mock-server schema fixture with only an id field; the fixture was fixed the next day in #3462. The branch just predated that fix.
  
I've rebased onto current main (no conflicts), which pulls in #3462, the embedding-executor change itself is unchanged. CI should be green now.

Ar-maan05:
Friendly ping! CI is passing and the branch is ready to be merged!

Ar-maan05:
Thank you! Let me know if there are other open issues I can help with :)

PR #3460, opened 2026-05-29
feat(python): add namespace/table exist support

### LINKED ISSUE #3419: How to verify whether namespace and table exist?
### Discussed in https://github.com/lancedb/lancedb/discussions/3414

<div type='discussions-op-text'>

<sup>Originally posted by **FarmerChillax** May 20, 2026</sup>
It seems there's no API in LanceDB to check if tables and namespaces exist. How should I determine if it exists?

i was visted the docs:
- https://lancedb.github.io/lancedb/python/python
- https://docs.lancedb.com/api-reference/rest/namespace/check-if-a-namespace-exists

if use lance namespace, we can impl like this:
```python3
def table_exists(ns: ln.LanceNamespace, table_id: list[str]) -> bool:
    """Check if a table with the given table_id exists in the namespace.

    Args:
        ns (ln.LanceNamespace): The LanceNamespace instance.
        table_id (list[str]): The list of table IDs to check existence.

    Returns:
        bool: True if the table exists, False otherwise.
    """
    try:
        ns.table_exists(
            ln.TableExistsRequest(
                id=table_id,
            )
        )
        return True
    except ln.TableNotFoundError as table_not_found_error:
        _logger.debug(
            "Table with ID: %s does not exist. Error: %s",
            table_id,
            table_not_found_error,
        )
        return False
    except ln.NamespaceNotFoundError as namespace_not_found_error:
        _logger.debug(
            "Namespace for table ID: %s does not exist. Error: %s",
            table_id,
            namespace_not_found_error,
        )
        return False
    except Exception as e:
        _logger.debug("Table with ID: %s does not exist. Error: %s", table_id, e)
        return False


def namespace_exists(ns: ln.LanceNamespace, namespace_id: list[str]) -> bool:
    """Check if a namespace with the given namespace_id exists.

    Args:
        ns (ln.LanceNamespace): The LanceNamespace instance.
        namespace_id (list[str]): The list of namespace IDs to check existence.

    Returns:
        bool: True if the namespace exists, False otherwise.
    """
    try:
        ns.namespace_exists(
            ln.NamespaceExistsRequest(
                id=namespace_id,
            )
        )
        return True
    except ln.NamespaceNotFoundError as namespace_not_found_error:
        _logger.debug(
            "Namespace with ID: %s does not exist. Error: %s",
            namespace_id,
            namespace_not_found_error,
        )
        return False
    except Exception as e:
        _logger.debug(
            "Namespace with ID: %s does not exist. Error: %s", namespace_id, e
        )
        return False
```

</div>

FarmerChillax:
In the current LanceDB usage implementation, there is no way to check whether a table or namespace already exists. This PR introduces the namespace_exists and table_exists methods to determine the existence of tables and namespaces.

useage like this:
```
# check table exists
db.table_exists(table_id=['xxx'])

# check namespace exists
db.namespace_exists(namespace_id=['xxx'])
```

fixes: #3419


prrao87:
Hey @FarmerChillax just checking if you're still working on this one?

FarmerChillax:
> Hey @FarmerChillax just checking if you're still working on this one?

Yeah, I’m still working on this PR. I’ve been held up by other tasks lately, but I’ll get it wrapped up within this week.

python/python/lancedb/namespace.py
```diff
@@ -1233,6 +1272,41 @@ async def describe_namespace(
         """
         return await self._inner.describe_namespace(namespace_path)
 
+    async def namespace_exists(self, namespace_id: List[str]) -> None:
```
wjones127: issue(blocking): I think this is a surprising API. I would expect `namespace_exists` to return a `bool` rather than raise an exception. Is this because it's implemented this way upstream?

wjones127 (changes requested):

python/python/lancedb/namespace.py
FarmerChillax: I also think this API design is rather counterintuitive. However, as you mentioned, this is how the upstream interface is structured. 

Should we catch the relevant exceptions here and return the corresponding bool? my take is any changes in here might confuse users: it throws exceptions under the lance namespace, while returning a boolean value in lancedb.

ref:
- table: https://github.com/lance-format/lance-namespace/blob/5743fcaba76b4aa337a6cb329a8f3d2e6547404f/python/lance_namespace/lance_namespace/__init__.py#L479
- namespace: https://github.com/lance-format/lance-namespace/blob/5743fcaba76b4aa337a6cb329a8f3d2e6547404f/python/lance_namespace/lance_namespace/__init__.py#L424

wjones127: My understanding is that the upstream interface is only structured like that because it is auto-generated from the openapi.yml spec, and translates the 404 response to an exception. It's not handwritten by a human, like this API.

I think it's acceptable to deviate from that auto-generated API.

wjones127: @jackye1995  do you agree?

FarmerChillax: > I think it's acceptable to deviate from that auto-generated API.

It’d be great if we didn’t have to match the upstream. I like this method signature better:

```python3
def namespace_exists(self, namespace_id: List[str]) -> bool:
```

wjones127: Consensus from my colleagues is to return a bool and allow deviating from the upstream. Could you implement that?

FarmerChillax: I’m very happy to do this. Thanks for confirme.

wjones127 (approved):
This looks good. Will merge once CI is passing.

PR #3469, opened 2026-05-31
fix(rerankers/mrr): raise ValueError on empty vector_results list

### LINKED ISSUE #3468: BUG: MRRReranker.rerank_multivector raises IndexError on empty input list
## What breaks

Calling `MRRReranker.rerank_multivector([])` raises an `IndexError: list index out of range` instead of a descriptive error.

## How to trigger

```python
from lancedb.rerankers.mrr import MRRReranker
reranker = MRRReranker()
reranker.rerank_multivector([])  # IndexError: list index out of range
```

## Traceback

```
File "lancedb/rerankers/mrr.py", line 128, in rerank_multivector
    if not all(isinstance(v, type(vector_results[0])) for v in vector_results):
                                    ~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

## Root cause

On line 128 the type-homogeneity check uses `all()` which passes vacuously on an empty iterable. Execution then reaches line 134 which accesses `vector_results[0]` unconditionally, crashing with `IndexError`. There is no guard against an empty input list.

devteamaegis:
## What's broken

`MRRReranker.rerank_multivector([])` raises `IndexError: list index out of range`. The crash happens on line 128 (the `all()` type-homogeneity check passes vacuously on an empty iterable) and on line 134 which accesses `vector_results[0]` unconditionally, with no prior guard for an empty list.

## Why it happens

`all()` over an empty iterable returns `True`, so the type check silently passes and execution falls through to `vector_results[0]` which crashes.

## Fix

Added a two-line guard at the top of `rerank_multivector` that raises a clear `ValueError("vector_results must not be empty")` before any indexing occurs.

## Test

Added `test_mrr_reranker_empty_input` in `test_rerankers.py` which calls `rerank_multivector([])` and asserts that a `ValueError` with the message "must not be empty" is raised.

Fixes #3468

### DIFF
--- a/python/python/lancedb/rerankers/mrr.py
+++ b/python/python/lancedb/rerankers/mrr.py
@@ -125,6 +125,9 @@ class MRRReranker(Reranker):
         This cannot reuse rerank_hybrid because MRR semantics require treating
         each vector result as a separate ranking system.
         """
+        if not vector_results:
+            raise ValueError("vector_results must not be empty")
+
         if not all(isinstance(v, type(vector_results[0])) for v in vector_results):
             raise ValueError(
                 "All elements in vector_results should be of the same type"
--- a/python/python/tests/test_rerankers.py
+++ b/python/python/tests/test_rerankers.py
@@ -344,6 +344,12 @@ def test_mrr_reranker(tmp_path):
     assert len(result_deduped) == len(result)
 
 
+def test_mrr_reranker_empty_input():
+    reranker = MRRReranker()
+    with pytest.raises(ValueError, match="must not be empty"):
+        reranker.rerank_multivector([])
+
+
 def test_rrf_reranker_distance():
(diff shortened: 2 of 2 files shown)


python/python/lancedb/rerankers/mrr.py
```diff
@@ -125,6 +125,9 @@ def rerank_multivector(
         This cannot reuse rerank_hybrid because MRR semantics require treating
         each vector result as a separate ranking system.
         """
+        if not vector_results:
+            raise ValueError("vector_results must not be empty")
+
```
wjones127, line 130: note: I think in an ideal world this function would just return an empty result if the results are empty. But I'm guessing in that case we don't have a schema, and thus can't construct the output table.

wjones127 (approved):

PR #3474, opened 2026-06-01
ci: stop release builds on pull requests

Xuanwo:
Release packaging workflows currently run on pull_request for dependency or workflow changes, which adds release build noise to normal PRs. This keeps the NPM release compatibility check for PRs that update Cargo dependency files, where dependency bumps have previously broken Node release builds. PyPI and Java publish workflows now run from main or release tag pushes, with publish steps still gated to release tags where applicable.

### DIFF
--- a/.github/workflows/java-publish.yml
+++ b/.github/workflows/java-publish.yml
@@ -13,11 +13,10 @@
 name: Build and publish Java packages
 on:
   push:
+    branches:
+      - main
     tags:
       - "v*"
-  pull_request:
-    paths:
-      - .github/workflows/java-publish.yml
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


wjones127:
Can we run them on PRs that update the cargo lock files? I agree just doing it for code changes can be overkill, but we’ve had previous PRs that bump dependencies not notice they broke the node releases build.

wjones127 (changes requested):
Sorry, I think I misread the existing state when I first read this PR. I thought it was vein triggered for any code change.

I actually agree with the current state. When dependencies change, we need to verify the builds still work. When we change the workflows, we should be rerunning them to make sure they still work.

Maybe it adds noise, but it definitely beats realizing the build is broken on release.

At the very least, I think the workflow files should trigger on pull request. When I am working on changes to those, I want real time feedback to validate my changes didn’t break anything.

I’m somewhat okay with making dependency changes not affect the builds. It does mean we have to watch main and make sure we are vigilant to revert any PRs that cause build issue. Of course, the problem is it makes it harder for the PR author to then write a new PR and validate that it doesn’t break the build. The PR trigger makes that much easier.

Does that make sense? I’m open to ideas. Maybe we just make sure we can easily manually trigger via workflow dispatch and that addresses the problem.

Xuanwo:
> Maybe it adds noise, but it definitely beats realizing the build is broken on release.

Yep, it makes sense. Let's close.

PR #3482, opened 2026-06-02
feat: add update_field_metadata to edit per-field metadata

brendanclement:
### Summary
Adds update_field_metadata to the client SDK (Rust core, Python, and TypeScript) so clients can edit per-field (column) Arrow metadata (schema.fields[].metadata)

### Testing
- added unit tests
- ran E2E against a local server on both local and remote tables (set → merge → delete), across Python sync/async and TypeScript

### Next steps
- deprecate replace_field_metadata in the python lancedb favor of this (typescript didn't have replace_field_metadata method). This matches Lance's API direction (Lance already deprecated replace_field_metadata for update_field_metadata)


jackye1995 (approved):
looks good to me!

PR #3483, opened 2026-06-02
test(python): add regression test for nullable struct with None (#2654)

Mr-Neutr0n:
## Summary

Regression test for [issue #2654](https://github.com/lancedb/lancedb/issues/2654) — a nullable struct column whose first batch contains only `None` values crashed in `_align_field_types` with `AttributeError: 'pyarrow.lib.DataType' object has no attribute 'fields'`.

The actual fix landed in #3394, but no test was added. This PR adds the reproducer from the issue as a test.

## Test plan

- `test_add_nullable_struct_with_none`: creates a table with a nullable struct column, adds a row with a non-null struct value, then a row with `None` for the struct field. Verifies both rows land correctly.
- Uses Lance file format v2.1 (`new_table_data_storage_version="2.1"`) because nullable structs aren't supported on v2.0.

## Related

- #3028 (the original fix attempt, now superseded)

### DIFF
--- a/python/python/tests/test_table.py
+++ b/python/python/tests/test_table.py
@@ -1264,6 +1264,45 @@ def test_add_with_empty_fixed_size_list_drops_bad_rows(mem_db: DBConnection):
     assert np.allclose(data["embedding"].to_pylist()[0], np.array([0.1] * 16))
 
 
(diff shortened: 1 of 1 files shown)


wjones127 (approved):
Looks good. Thanks for adding the test!

PR #3484, opened 2026-06-02
feat: deprecate replace_field_metadata for update_field_metadata

brendanclement:
### Summary
Deprecates the Python replace_field_metadata (on Table and AsyncTable) in favor of update_field_metadata. Mirrors Lance, which already deprecated Dataset.replace_field_metadata for update_field_metadata. 

Stacked on top of #3482 as this was a follow-up task after adding update_field_metadata

### DIFF
--- a/python/src/table.rs
+++ b/python/src/table.rs
@@ -1118,6 +1118,7 @@ impl Table {
         })
     }
 
+    #[allow(deprecated)]
     pub fn replace_field_metadata<'a>(
         self_: PyRef<'a, Self>,
         field_name: String,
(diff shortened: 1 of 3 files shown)


rust/lancedb/src/table.rs
```diff
@@ -4482,6 +4483,7 @@ mod tests {
 
         let mut new_field_metadata = HashMap::<String, String>::new();
         new_field_metadata.insert("test_field_key1".into(), "test_field_val1".into());
+        #[allow(deprecated)]
```
westonpace: Should we update this test or is it specifically verifying the method?

python/src/table.rs
```diff
@@ -1118,6 +1118,7 @@ impl Table {
         })
     }
 
+    #[allow(deprecated)]
```
westonpace: We should use the updated method ourselves and not need allow(deprecated)

westonpace (approved):
A few nits but seems fine.  Ping me when ready to merge.

PR #3488, opened 2026-06-02
feat(python): add observability metrics to torch datasets

### LINKED ISSUE #3480: Feature: Add observability to torch datasets
### SDK

Python

### Description

At a minimum we should be able to easily determine how many bytes per second we are running through I/O and how many bytes per second we are running through transformations and how many bytes per second are being consumed.  This will help distinguish between an I/O, CPU, or GPU bottleneck.

DhruvGarg111:
Resolves #3480. Adds PermutationMetrics to torch datasets so users can easily distinguish between I/O, CPU, and GPU bottlenecks by calling dataset.with_observability().metrics.summary().

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -4,6 +4,7 @@
 import copy
 import json
 import os
+import time
 
 from deprecation import deprecated
 import pyarrow as pa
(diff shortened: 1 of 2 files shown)


wjones127:
Fixed by #3509

PR #3489, opened 2026-06-02
feat(query): add use_lsm to read MemWAL LSM data

touch-of-grey:
## What

MemWAL LSM **read** support. When a table has an LSM write spec (`set_lsm_write_spec`), `merge_insert` upserts live in the MemWAL active/frozen memtables and flushed SSTables until an external compaction merges them into the base table, so a normal scan returns **stale** data. This routes reads through Lance's `LsmScanner` so queries also surface that in-flight data, deduplicated by primary key (newest generation wins).

## How

- Adds a **`use_lsm: Option<bool>`** query flag, symmetric with the `merge_insert` flag:
    - **unset** — auto-route through the LSM scanner when the table carries a write spec
    - **`use_lsm(true)`** — force the LSM path; error if there is no spec
    - **`use_lsm(false)`** — read the base table only (the escape hatch)
- Plain scan, single-column full-text search, and single-vector ANN all run through one `LsmScanner` (assembled from on-disk shard manifests plus the cached writer's in-memory memtables), so a `where` predicate is honored as a **prefilter** uniformly — including for vector search.
- **Compaction-aware snapshots:** an SSTable generation is dropped only once it is both compacted into the base table and covered by the arm's base-index catch-up (`index_catchup`); plain scans use the compaction watermark alone.
- Query shapes the scanner cannot honor hard-error with guidance to set `use_lsm(false)`: hybrid, multi/binary vectors, `with_row_id`, reranking, `order_by`, dynamic/Substrait projection or filters, `distance_range`, `use_index(false)`, postfilter, take-by-row-id/offset, reads from a time-traveled version, and an unmaintained or ambiguous (multiple) FTS/vector index. Namespace-pushdown queries fall back to local execution when a spec is present; WAL-only writers are handled.
- Exposed across the Rust core and the Python (`use_lsm`) and TypeScript (`useLsm`) bindings, including `TakeQuery`.

Rebased from Lance `7.2.0-beta.3` to `10.0.0-beta.3`.


touch-of-grey:
@jackye1995 PTAL — this adds the read side of the MemWAL LSM feature: an opt-in `use_lsm_read` flag so queries surface in-flight `merge_insert` data (active/frozen memtables + flushed generations) via Lance's `LsmScanner`, covering plain/FTS/vector reads across Rust + Python + TS.

touch-of-grey:
The NodeJS CI failures (`Linux (NodeJS 18/20)`, `macos`) are unrelated flakes: the **Test examples** step fails downloading `Xenova/all-MiniLM-L6-v2` from HuggingFace with HTTP 429 (rate limit). The `useLsmRead` jest tests and the docs check both pass. I don't have rerun permissions here, so I pushed an empty commit to retrigger; a maintainer rerun of the failed jobs would also clear it.

hamersaw:
I think it would be useful to be consistent on the flags for LSM interfacing across reads / writes. Right now there is a `use_lsm_write` flag on the `merge_insert` that is an `Option<bool> (added in https://github.com/lancedb/lancedb/pull/3354)`, it behaves by:
- `None`: route to WAL if dataset has an `LsmWriteSpec` index and write directory to dataset if not
- `Some(false)`: route to dataset directly (bypass WAL if exists)
- `Some(true)`: route to WAL (fail if `LsmWriteSpec` does not exist).

The proposed behavior of this `use_lsm_read` flag varies a bit in that `None` defaults to reading from the dataset directly (bypassing WAL if exists). So users are required to specify `use_lsm_read` to use the WAL, whereas `use_lsm_write` can be omitted.

TBH I don't like having an `Option<bool>` and using it as a WAL inclusion gate. I think that if the `LsmWriteSpec` exists, we should do writes / reads including the WAL by default. So my proposal is to make this an exclusion flag instead on both writes / reads by changing it to `disable_lsm` as a simple `bool`. This would default to `false` and then reads / writes would check if the `LsmWriteSpec` exists to determine WAL inclusion unless the user explicitly says that we want to operate without WAL.

hamersaw:
@touch-of-grey I just pushed the updates to change this flag (both `merge_insert` + query) to `use_lsm` and made the behavior the same for both writes and reads. `None` means we use LSM if the LsmWriteSpec exists, `Some(true)` fails if the `LsmWriteSpec` doesn't exist, and `Some(false)` disables the  LSM path. I think this is uniform now.

jackye1995:
Rebased this onto the latest `main` and pushed — thanks for the original work, @touch-of-grey 🙏. The PR was based on Lance `7.2.0-beta.3` and `main` is now on `10.0.0-beta.3`, so this was mostly an API port plus one fix.

**Rebase / port**
- Squashed to a single commit on top of `main` (your authorship preserved). The only git conflict was `query.py` (both sides added builder fields).
- Ported the read path to the current Lance MemWAL API: `flushed_generations` → `sstables` / `with_sstable` (the flushed→SSTable rename), `open_vector_index(&Uuid)`, and the new `LsmScanner::full_text_search(query)` signature.

**Vector prefilter fix**
- The vector arm drove `LsmVectorSearchPlanner` directly and never passed the query filter, so a `.where()` on a vector search was silently ignored. All three arms (plain / FTS / vector) now route through a single `LsmScanner`, which threads the filter into the vector planner as a true prefilter. Added regression tests (`test_lsm_read_vector_prefilter`, etc.) over base ∪ memtable.

**Hardening from review**
- `distance_range` and `use_index(false)` were the same class of silent-drop on the vector arm — they now hard-error with the `use_lsm(false)` guidance, like the other unsupported shapes.
- `read_snapshot` now captures memtables before the manifest, so a flush interleaving a read yields dedup-safe duplicates rather than missed rows.

**Heads-up (Lance behavior):** the LSM vector base arm uses `fast_search` (indexed data only), so un-indexed base rows are invisible to a `use_lsm` vector search — they surface via the memtable or `use_lsm(false)`. Documented in the module docs + test helper.

Validation: `cargo clippy`/`fmt` clean, `cargo test table::query`, 31 Python LSM tests, and the nodejs crate compiles. Possible follow-up: auto-resolve the FTS column from a single FTS index so `fts_columns` isn't required on the LSM path.

jackye1995:
@claude review once

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,415 @@
...
+    // stores). The base path matches `list_mem_wal_latest_shard_ids`.
+    let store = dataset.object_store(None).await?;
+    let base_path = dataset.branch_location().path;
+    let scan_batch_size = ShardWriterConfig::default().manifest_scan_batch_size;
+
+    let mut snapshots: Vec<ShardSnapshot> = Vec::new();
+    for shard_id in shard_ids {
+        let manifest_store =
+            ShardManifestStore::new(store.clone(), &base_path, shard_id, scan_batch_size);
+        if let Some(manifest) = manifest_store.read_latest().await? {
```
jackye1995, line 290: Could we use `MemWalIndexDetails.compacted_sstables` here, plus `index_catchup` for indexed searches, when building the shard snapshots? The shard manifest keeps flushed SSTables append-only, while compaction records its watermark in the MemWAL system index. As written, every historical SSTable is added as a generation newer than the base, so after compaction an old value can shadow a later base-only update or delete, and a retired SSTable path may also be reopened.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,415 @@
...
+///
+/// Snapshots for all shards come from their on-disk manifests; for the shard
+/// with a live cached `ShardWriter` (this session's in-flight writes) the
+/// writer's authoritative in-memory manifest and memtables override the
+/// on-disk view so a read sees data not yet flushed.
+async fn build_read_context(
+    table: &NativeTable,
+    dataset: &Dataset,
+) -> Result<(Vec<ShardSnapshot>, HashMap<Uuid, InMemoryMemTables>)> {
+    let shard_ids = dataset.list_mem_wal_latest_shard_ids().await?;
```
jackye1995, line 278: Could we coordinate the WAL snapshot with the dataset version here, or reject LSM reads from a checked-out historical handle? `dataset` may refer to an older version, but `list_mem_wal_latest_shard_ids`, `read_latest`, and the cached writer all expose current live state. A query after checkout can therefore surface WAL rows written after the requested dataset version.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,415 @@
...
+            message: "the MemWAL LSM scanner full-text search supports a single column; set use_lsm(false) to read the base table only".to_string(),
+        });
+    }
+    if columns.is_empty() {
+        return Err(Error::NotSupported {
+            message: "the MemWAL LSM scanner full-text search requires an explicit FTS column"
+                .to_string(),
+        });
+    }
+    let scanner = base_scanner(dataset, query, pk_columns, snapshots, in_memory)?
```
jackye1995, line 456: Could we validate that the requested FTS index is maintained by the MemWAL spec before planning this arm? In the pinned Lance implementation, `active_source_can_execute_fts` returns false and the active arm becomes an empty plan when no in-memory FTS index exists. Since `LsmWriteSpec::unsharded()` permits an empty maintained-index list, this query can succeed while silently omitting un-compacted documents. A fallback scan or a guided `NotSupported` error seems safer.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,415 @@
...
+    query: &VectorQueryRequest,
+    pk_columns: Vec<String>,
+    snapshots: Vec<ShardSnapshot>,
+    in_memory: HashMap<Uuid, InMemoryMemTables>,
+) -> Result<LsmScanner> {
+    let mut scanner = LsmScanner::new(Arc::new(dataset.clone()), snapshots, pk_columns);
+    for (shard_id, memtables) in in_memory {
+        scanner = scanner.with_in_memory_memtables(shard_id, memtables);
+    }
+    if let Some(columns) = selected_columns(query) {
```
jackye1995, line 377: Could we add a final projection that restores the exact user-requested schema after the LSM plan? Lance appends primary-key columns internally for deduplication and retains them in its canonical output schema, so a request such as `select(["value"])` returns `value` plus the primary key. The internal dedup key should not leak through the public LanceDB projection.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,415 @@
...
+        .nprobes(query.minimum_nprobes)
+        .distance_metric(distance_type.into());
+    if let Some(refine_factor) = query.refine_factor {
+        scanner = scanner.refine(refine_factor);
+    }
+    scanner = scanner.limit(as_i64(limit), as_i64(offset))?;
+    Ok(scanner.create_plan().await?)
+}
+
+/// Whether `column` stores binary (uint8) vectors, which the LSM vector planner
```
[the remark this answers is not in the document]
jackye1995: Verified this is handled internally, not a bug: Lance's `plan_vector` over-fetches per source — `per_source_k = nearest.k + self.offset` (`scanner/builder.rs:596`, with a Lance test covering it) — then `apply_limit_offset` skips `offset`. So `k = limit` yields the correct `[offset, offset+limit)` page. Added `test_lsm_read_vector_limit_offset` in d48c0e74 (a full, distinct second page — which would be empty if `k` ignored `offset`) to lock it in.

jackye1995:
Thanks for the thorough review, @jackye1995 (and @claude). All eight are addressed in `d48c0e74`; each fix has a regression test unless noted.

**🔴 `vector_plan` `k` ignores offset — not a bug (verified).** Lance's `plan_vector` over-fetches per source: `per_source_k = nearest.k + self.offset` (`scanner/builder.rs:596`), then `apply_limit_offset` skips `offset`. So `k = limit` is correct end-to-end. Locked in with `test_lsm_read_vector_limit_offset` (replied inline).

**🟡 TakeQuery `_rowid`/`_rowoffset`.** `reject_unsupported` now detects a Datafusion filter referencing `_rowid`/`_rowoffset` and returns the guided `NotSupported` error instead of a deep column-not-found. Added `use_lsm` to the Python `TakeQuery` (pyo3 + `AsyncTakeQuery` + `BaseQueryBuilder`) for parity with NodeJS. Test: `test_lsm_read_take_row_ids_errors`.

**🟡 postfilter silently prefiltered.** `LsmScanner` has no postfilter mode, so `!query.base.prefilter` on a vector search now hard-errors alongside `distance_range`/`use_index`. Test: `test_lsm_read_vector_postfilter_errors`.

**🟡 `use_lsm` ignored on namespace pushdown.** `requires_local_namespace_execution` now forces local execution when `use_lsm.is_some()`, mirroring the `approx_mode` guard. Test: `test_execute_query_use_lsm_with_namespace_pushdown_runs_locally`.

**`compacted_sstables` watermark.** `build_read_context` now excludes SSTable generations at or below the per-shard compaction watermark (first-per-shard, mirroring `compacted_generation_for_shard`), so a stale flushed value no longer shadows a newer base update/delete and retired paths aren't reopened. I used `generation <= watermark` (per `is_index_caught_up`'s `>= compacted_gen`) — shout if the boundary should be strict. The `index_catchup` per-index refinement for the maintained arms is a good follow-up.

**WAL snapshot vs dataset version (time-travel).** A checked-out handle (`time_travel_version().is_some()`) now rejects LSM reads — cheap, no extra version fetch — with `use_lsm(false)` reading the base at the pinned version. Test: `test_lsm_read_time_travel_errors`.

**FTS index maintained.** `fts_plan` now verifies the column's FTS index is in `maintained_indexes` (otherwise the active arm returns an empty plan and silently omits un-compacted docs) and rejects with guidance otherwise. Test: `test_lsm_read_fts_unmaintained_index_errors`.

**PK leak in projection.** A final `restore_projection` drops the internal primary-key columns Lance appends for dedup when the user didn't request them, so `select(["value"])` no longer leaks the key. Test: `test_lsm_read_projection_excludes_pk`.

Validation: `cargo clippy`/`fmt` clean, `cargo test table::query` (incl. the new namespace test) green, 27 Python LSM tests + the broader query suite green.

rust/lancedb/src/table/query.rs
```diff
@@ -71,8 +74,13 @@ fn can_execute_namespace_query(table: &NativeTable, query: &AnyQuery) -> bool {
 }
 
 fn requires_local_namespace_execution(query: &AnyQuery) -> bool {
-    // The namespace QueryTable request has no approx_mode field yet, so
-    // pushing this query down would silently ignore the user's setting.
+    // The namespace QueryTable request has no approx_mode or use_lsm field yet, so
+    // pushing these down would silently ignore the user's setting. For use_lsm that
+    // is worse than a tuning miss: MemWAL read routing lives only in `create_plan`,
+    // so a pushed-down query would return stale base-only data with no error.
+    if query.base().use_lsm.is_some() {
```
jackye1995, line 93: Could we also handle the default use_lsm == None case here? That is the documented auto-routing mode: when the table has a MemWAL spec, create_plan selects the LSM scanner. This guard only forces local execution for explicit Some values, so QueryTable pushdown still bypasses create_plan for the default query and can omit un-compacted rows. The new regression test covers Some(false) on a table without a spec, not None on a table with one.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,548 @@
...
+    maintained: &[String],
+) -> Result<bool> {
+    use lance::index::DatasetIndexExt;
+    let Some(field) = dataset.schema().field(column) else {
+        return Ok(false);
+    };
+    let indices = dataset.load_indices().await?;
+    Ok(indices
+        .iter()
+        .any(|idx| idx.fields.contains(&field.id) && maintained.iter().any(|m| m == &idx.name)))
```
jackye1995: Could we verify that the matched maintained index is actually the FTS/inverted index? This predicate accepts any maintained index whose fields include this column. If the text field has both FTS and BTree indexes and only BTree is maintained, this returns true, but Lance has no active FTS arm and silently omits un-compacted documents.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,548 @@
...
+    if !query.query_vector.is_empty() {
+        if query.lower_bound.is_some() || query.upper_bound.is_some() {
+            return unsupported("distance_range on vector search");
+        }
+        if !query.use_index {
+            return unsupported(
+                "use_index(false) / brute-force vector search (the LSM base arm is index-only)",
+            );
+        }
+        if !query.base.prefilter {
```
jackye1995: Could we extend the postfilter rejection to FTS? The new guard is inside the vector-only branch, but FTS builders also expose postfilter and the standard scanner honors it. base_scanner always installs the predicate before LSM BM25 search, so an FTS postfilter request still silently changes result semantics. Please reject prefilter == false when full_text_search is set, or implement it.

rust/lancedb/src/table/query/lsm.rs
```diff
@@ -0,0 +1,548 @@
...
+        });
+    }
+
+    let distance_type = resolve_distance_type(dataset, query, &column).await?;
+
+    // `nearest` takes a flat query vector and builds the fixed-size list itself.
+    // Guard `k` to at least 1 so a degenerate `limit(0)` is trimmed by `limit`
+    // below rather than rejected by `nearest`.
+    let k = limit.unwrap_or(DEFAULT_TOP_K).max(1);
+    let mut scanner = base_scanner(dataset, query, pk_columns, snapshots, in_memory)?
```
jackye1995, line 665: Could we configure or expose an overfetch factor for LSM searches? Neither FTS nor vector sets with_overfetch_factor, so Lance defaults to 1.0. Its scanner explicitly documents that a blocked source then fetches exactly k and can return fewer than k live rows after stale-PK filtering. Since upserts normally create blocked candidates, result pages can be silently underfilled.

rust/lancedb/src/table/merge/lsm.rs
```diff
@@ -345,6 +378,36 @@ impl ShardWriterCache {
...
+            guard.as_ref().map(|(id, entry)| (*id, entry.clone()))
+        };
+        let Some((shard_id, entry)) = cached else {
+            return Ok(None);
+        };
+        // Capture memtables before the manifest. If a flush interleaves, dedup
+        // tolerates the same rows appearing in both a memtable and a freshly
+        // flushed generation, but would drop rows present in neither. Manifest
+        // last guarantees any generation flushed mid-capture is still covered.
+        let memtables = entry.in_memory_memtable_refs().await?;
```
jackye1995, line 406: Could we handle enable_memtable=false before calling this accessor? writer_config_defaults exposes WAL-only mode, but Lance in_memory_memtable_refs explicitly errors in that mode. Once a WAL-only writer is cached, every LSM-routed read reaches this call and fails. We should either support a WAL-only snapshot/visibility policy here or reject that configuration at the LanceDB boundary.

nodejs/src/query.rs
```diff
@@ -479,6 +489,11 @@ impl TakeQuery {
         self.inner = self.inner.clone().with_row_id();
     }
 
+    #[napi]
+    pub fn use_lsm(&mut self, enable: bool) {
```
jackye1995, line 493: The native escape hatch is added here, but it is not reachable from the public TypeScript TakeQuery. nodejs/lancedb/query.ts has TakeQuery extend QueryBase directly, while useLsm is defined only on StandardQueryBase. Since take-by-row-id/offset now auto-routes to LSM and is rejected, Node users have no way to call useLsm(false). Could we expose it on TakeQuery (or hoist it to QueryBase) and add a binding test?

rust/lancedb/src/table/query/lsm.rs
jackye1995: To clarify, this is not primarily a Lance LsmScanner bug. The LSM snapshot here is the read-time source set LanceDB assembles for the scanner: the base dataset, flushed SSTable generations, and active/frozen memtables. LsmScanner trusts that source list and cannot search generations the caller omitted. This code removes SSTables as soon as the dataset compaction watermark covers them, which is correct for plain scans, but an indexed search can still miss those rows while its base index lags compaction. We should retain the affected SSTables until the relevant per-index index_catchup watermark confirms that the base index covers them.

jackye1995: Good catch — fixed in aa2d8fb3. `build_read_context` now computes a per-shard *exclusion* watermark: for an indexed arm it caps the compaction watermark at the base index's `index_catchup` generation (`IndexCatchupProgress::caught_up_generation_for_shard`), so SSTables the base index has not yet caught up to are retained and stay searchable. Plain scans still use the compaction watermark alone. `create_lsm_plan` resolves the arm's maintained index (`arm_maintained_index_name`) and threads it in.

rust/lancedb/src/table/query.rs
jackye1995: Fixed in aa2d8fb3. `can_execute_namespace_query` is now async and forces local execution whenever the table carries a MemWAL spec (checks `mem_wal_index_details`), so the default `use_lsm == None` auto-route no longer pushes down and silently omits un-compacted rows. The explicit `use_lsm.is_some()` guard in `requires_local_namespace_execution` remains for spec-less tables.

rust/lancedb/src/table/query/lsm.rs
jackye1995: Fixed in aa2d8fb3. `fts_index_maintained` now additionally requires the matched maintained index to be the inverted/FTS index (`index_details.type_url` ends with `InvertedIndexDetails`), so a maintained BTree on the same column no longer passes as an FTS index.

rust/lancedb/src/table/query/lsm.rs
jackye1995: Fixed in aa2d8fb3 — the postfilter rejection moved out of the vector-only branch and now also rejects `prefilter == false` when `full_text_search` is set. Added `test_lsm_read_fts_postfilter_errors`.

rust/lancedb/src/table/query/lsm.rs
jackye1995: Fixed in aa2d8fb3 — the vector and FTS arms now set `with_overfetch_factor(2.0)` (new `LSM_OVERFETCH_FACTOR`), so upsert-blocked candidates don't underfill result pages.

rust/lancedb/src/table/merge/lsm.rs
jackye1995: Fixed in aa2d8fb3 — `build_read_context` detects `enable_memtable=false` from `writer_config_defaults` and skips the live-writer memtable snapshot (which errors in WAL-only mode); the on-disk manifests already cover flushed SSTables. Since Lance forbids maintained indexes in WAL-only mode, only plain scans reach this path.

nodejs/src/query.rs
jackye1995: Fixed in aa2d8fb3 — `useLsm` is now exposed on the TypeScript `TakeQuery`, the generated `TakeQuery.md` docs are regenerated, and a jest binding test was added (`should expose useLsm on takeRowIds as the base-only escape hatch`).

rust/lancedb/src/table/query/lsm.rs
jackye1995: Follow-up in 90f37172, addressing both points:

1. **Index selection by type.** `arm_maintained_index_name` now matches the index the arm actually uses — `VectorIndexDetails` for ANN, `InvertedIndexDetails` for FTS — so a maintained BTree on the same column can no longer gate exclusion with its (possibly further-along) catch-up and drop SSTables the vector/FTS index still needs.

2. **Regression test.** Extracted the gating into a pure `exclusion_watermarks(details, index_name)` and added `exclusion_watermark_gates_on_lagging_index_catchup`: with compaction at generation 5 but the FTS index caught up only to generation 2, the FTS arm retains generations 3–5 (`Some(&2)`), while a plain scan drops through 5 (`Some(&5)`), and a caught-up/untracked index falls back to the compaction watermark.

jackye1995: Follow-up in the latest push: `arm_maintained_index_name` now **fails** rather than guessing when the searched field carries multiple same-type indexes. It filters the field's indexes to the arm's type (`VectorIndexDetails` / `InvertedIndexDetails`); with more than one match the base planner's choice is ambiguous and the catch-up watermarks can diverge, so it returns `NotSupported` ("...has multiple {vector|full-text} indexes; set use_lsm(false)...") instead of gating exclusion on a possibly-wrong watermark. A single match is used only when it's maintained. Added `test_lsm_read_fts_multiple_same_type_indexes_errors` (two FTS indexes on one column → rejected).

jackye1995: Follow-up in 3900008c: the vector arm now **rejects an unmaintained vector index**, mirroring FTS. Generalized the maintained-index check to `index_maintained(dataset, column, maintained, type_url_suffix)` and call it from `vector_plan` (`VectorIndexDetails`) as well as `fts_plan` (`InvertedIndexDetails`). An unmaintained base vector index has untracked catch-up, so exclusion would fall back to the plain compaction watermark and drop compacted rows the (lagging) base index has not re-indexed — it now errors (`"requires the vector index on '…' to be maintained …; set use_lsm(false)…"`) instead of silently omitting them. Added `test_lsm_read_vector_unmaintained_index_errors`.

jackye1995: Fixed in the latest push. `load_indices` returns one entry per physical index segment, so a single logical index could appear multiple times and be miscounted as "multiple indexes". Extracted `resolve_single_index(names, maintained, arm, column)` which **dedupes by index name** before counting: multiple segments of one logical index resolve to it, while genuinely distinct indexes still error. Covered by a Rust unit test (`resolve_single_index_dedupes_segments`: `["fts_idx","fts_idx"]` → `Some`, `["fts_a","fts_b"]` → `Err`, `["other"]` → `None`) plus an end-to-end `test_lsm_read_fts_optimized_index_not_rejected` (create → add → `optimize` → FTS search not rejected).

jackye1995 (approved):
thanks for the work @touch-of-grey, I updated this to the extent that I am personally satisfied, looking forward to you being back to finish the remaining parts of this work!

PR #3490, opened 2026-06-03
feat: add table branch support

brendanclement:
### Description

Adds first-class support for table branches across the Rust core and the Python and TypeScript SDKs.

Rust

```rust
use lance::dataset::refs::Ref;

// Create a branch from main and write to it — main is untouched.
let exp = table.create_branch("exp", Ref::Version(None, None)).await?;
exp.add(batches).await?;

// Reopen the branch later: check out from a table, or open it directly.
let exp = table.checkout_branch("exp").await?;
let exp = db.open_table("items").branch("exp").execute().await?;

let branches = table.list_branches().await?;
table.delete_branch("exp").await?;
```

Python

```python
# Create a branch from main and write to it
branch = await table.branches.create("exp", from_ref="main")
await branch.add(data)

# Reopen the branch later: check out from a table, or open it directly.
branch = await table.branches.checkout("exp")
branch = await db.open_table("items", branch="exp")

await table.branches.list()
await table.branches.delete("exp")
```

TypeScript

```typescript
const branches = await table.branches();

// Create a branch from main and write to it
const branch = await branches.create("exp");
await branch.add(data);

// Reopen the branch later: check out from a table, or open it directly.
const checkedOut = await branches.checkout("exp");
const opened = await db.openTable("items", undefined, { branch: "exp" });

await branches.list();
await branches.delete("exp");
```

### Testing
- Added unit tests
- ran smoke tests against python and typescript sdks on local machine

### Next steps
- Add RemoteTable support
- Add Branch Comparison support
- Merge Branching support


brendanclement:
@claude review once

brendanclement:
@claude review once

brendanclement:
@claude review once

brendanclement:
@claude review once

python/python/lancedb/namespace.py
```diff
@@ -562,14 +563,17 @@ def open_table(
...
             namespace_client=self._namespace_client,
             pushdown_operations=self._namespace_client_pushdown_operations,
             _async=async_table,
         )
+        if branch is not None:
+            return tbl.branches.checkout(branch)
+        return tbl
 
     @override
     def drop_table(self, name: str, namespace_path: Optional[List[str]] = None):
```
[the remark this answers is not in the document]
brendanclement: good catch claude!

brendanclement:
@claude review once

rust/lancedb/src/table/dataset.rs
```diff
@@ -737,4 +759,31 @@ mod tests {
         let result = wrapper.reload().await;
         assert!(result.is_err());
     }
+
+    #[tokio::test]
+    async fn test_as_branch_is_writable_and_tracked() {
```
wjones127, line 764: issue(blocking): one test I would like to see is what happens to a branch handle if it's concurrently updated? Does the branch handle track new writes to the branch? Either way we go, I think the behavior should be asserted in a test.

python/python/tests/test_table.py
```diff
@@ -903,6 +903,144 @@ async def test_async_tags(mem_db_async: AsyncConnection):
     )
 
 
+def test_branches(tmp_path):
+    db = lancedb.connect(tmp_path)
```
wjones127: issue(blocking): the default for `read_consistency_interval` is `None`, which actually makes this test less useful. I would recommend setting this to zero for these kinds of tests:


```suggestion
    db = lancedb.connect(tmp_path, read_consistency_interval=timedelta(0))
```

wjones127 (changes requested):
The API looks really well designed. I like this.

There are a few behavior / testing issues I think we need to address before we merge this.

python/python/tests/test_table.py
brendanclement: good catch, changed to 0

rust/lancedb/src/table/dataset.rs
brendanclement: I've added a test here to document the desired behaviour. A branch handle will track new writes to its branch at its read_consistency_interval.

wjones127 (approved):
Excellent. Thank you!

Issue #3495, opened 2026-06-03
feat(bindings): expose new IndexConfig fields in Python and Node.js


wjones127:
Follow-up to lancedb/lancedb#3492. Surface the new `IndexConfig` fields in language bindings.

## Python

* Extend `pyclass IndexConfig` in `python/src/index.rs` with the new fields.
* Update `__getitem__` to support the new keys.
* Update `python/python/lancedb/_lancedb.pyi` with the new type hints.
* Rebuild: `uv run --extra tests --extra dev maturin develop --extras tests,dev`.
* Tests in `python/tests/test_table.py`.

## Node.js

* Extend `IndexConfig` in `nodejs/src/table.rs`.
* Regenerate types: `npm run build`.
* Tests in `nodejs/__test__/table.test.ts`.
* Regenerate docs: `npm run docs`.

## Depends on

* lancedb/lancedb#3492 (expand `IndexConfig`)

Issue #3494, opened 2026-06-03
perf: drop N+1 in RemoteTable::list_indices via server-version gate


wjones127:
`RemoteTable::list_indices` (`rust/lancedb/src/remote/table.rs`) makes one `/index/list` call plus one `/index/{name}/stats` call per index just to recover `index_type`. The server is being updated to return all needed info in `/index/list` (cloud-side change, gated by phalanx version `0.5.0`).

## Changes

* Add `ServerVersion::support_enriched_index_list()` in `rust/lancedb/src/remote/db.rs` (`>= 0.5.0`).
* In `RemoteTable::list_indices`:
  * If supported → deserialize the enriched response, populate full `IndexConfig` (fields added in lancedb/lancedb#3492), skip the per-index fan-out.
  * Else → keep the existing list + per-index `index_stats` fallback.

## Tests

Mocked-endpoint tests in `rust/lancedb/src/remote/table.rs` for both new and legacy response shapes (including the version-header branching).

## Depends on

* lancedb/lancedb#3492 (expand `IndexConfig`)
* Server-side support (Enterprise tracker [ENT-1590](https://linear.app/lancedb/issue/ENT-1590/enrich-v1tableidindexlist-response-with-describe-indices-data)) deployed

PR #3497, opened 2026-06-03
feat: expand IndexConfig with rich per-index metadata

### LINKED ISSUE #3492: feat: expand IndexConfig with rich per-index metadata
`IndexConfig` (returned by `Table::list_indices`) currently exposes only `name`, `index_type`, `columns`. The Lance core `describe_indices` API now provides much richer info cheaply (reads manifest metadata, often cached). Surface it.

Add `Option<T>` fields to `lancedb::index::IndexConfig`:

- `index_uuid: Option<String>` (first segment uuid)
- `created_at: Option<DateTime<Utc>>` (min across segments)
- `num_indexed_rows: Option<u64>`
- `num_unindexed_rows: Option<u64>` (`dataset.count_rows() - rows_indexed`)
- `size_bytes: Option<u64>` (from `IndexDescription::total_size_bytes`)
- `index_details: Option<String>` (JSON from `IndexDescription::details`)
- `num_segments: Option<u32>`

Populate from `dataset.describe_indices(None)` in `NativeTable::list_indices` (already used there). `RemoteTable` leaves the new fields `None` until the follow-up that wires them through the server response.

Tests: extend existing `list_indices` tests in `rust/lancedb/src/table.rs` to assert the new fields.

Bindings exposure tracked separately.

wjones127:
`IndexConfig` (returned by `Table::list_indices`) previously exposed only `name`, `index_type`, and `columns`. Lance's `describe_indices` provides richer per-index info cheaply (reads manifest metadata, often cached), so this surfaces it.

Adds these `Option<T>` fields to `lancedb::index::IndexConfig`, populated in `NativeTable::list_indices` from the `IndexDescription`:

- `index_uuid`: uuid of the first segment
- `type_url`: protobuf type URL (`IndexDescription::type_url`)
- `created_at`: minimum creation time across segments
- `num_indexed_rows`: approximate rows indexed across segments
- `num_unindexed_rows`: table row count minus `num_indexed_rows`
- `size_bytes`: total size of index files across segments
- `num_segments`: number of segments making up the index
- `index_version`: on-disk index format version (first segment)
- `index_details`: index-type-specific details as JSON

This field set mirrors the lance-namespace `IndexContent` contract (lance-format/lance-namespace#348) so client and server agree on the same shape. Note these are populated **locally** via `describe_indices` — `NativeTable::list_indices` reads the dataset directly and does not depend on the namespace spec change.

`RemoteTable` leaves the new fields `None` until a follow-up wires them through the server response (#3494). Bindings exposure will also be a follow up: #3495

Existing `list_indices` tests in `rust/lancedb/src/table.rs` are extended to assert the new fields.

Fixes #3492

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -2151,6 +2151,17 @@ impl<S: HttpSend> BaseTable for RemoteTable<S> {
                         name: index.index_name,
                         index_type: stats.index_type,
                         columns,
+                        // These are left None until the server response wires
+                        // them through. See https://github.com/lancedb/lancedb/issues/3494
+                        index_uuid: None,
+                        type_url: None,
+                        created_at: None,
+                        num_indexed_rows: None,
+                        num_unindexed_rows: None,
+                        size_bytes: None,
+                        num_segments: None,
+                        index_version: None,
+                        index_details: None,
                     })),
                     Ok(None) => Ok(None), // The index must have been deleted since we listed it.
                     Err(e) => Err(e),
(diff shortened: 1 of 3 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -4233,53 +4262,43 @@ mod tests {
...
+                IndexType::BTree,
+                "MetaData.userId",
+            ),
+            ("metadata_user_id_idx", IndexType::BTree, "metadata.user_id"),
+            ("image_embedding_idx", IndexType::IvfPq, "image.embedding"),
+            ("payload_text_idx", IndexType::FTS, "payload.text"),
+            (
+                "meta_data_user_id_idx",
+                IndexType::BTree,
+                "`meta-data`.`user-id`",
```
justinrmiller, line 4282: should this be "metadata.user_id"? I'm confused what the difference is with line 4276?

wjones127: In this example, there is a column called `meta-data`, with a subfield called `user-id`. In Lance, to reference such a field, you need to escape it by wrapping in the tick character.

This is a pre-existing test case from above.

LuQQiu (approved):

PR #3503, opened 2026-06-04
fix(pydantic): bare Tuple/List crashes to_arrow_schema with AttributeError

### LINKED ISSUE #3502: Bare typing.Tuple/List field crashes to_arrow_schema with AttributeError: __args__
In `_py_type_to_arrow_type` the branch `elif getattr(py_type, "__origin__", None) in (list, tuple)` is taken for a bare, unparameterised `typing.Tuple`/`typing.List` (their `__origin__` is `tuple`/`list`). The next line reads `py_type.__args__[0]`, but a bare generic has no `__args__`, so it raises `AttributeError: __args__`.

```python
from typing import Tuple
class Doc(LanceModel):
    items: Tuple
Doc.to_arrow_schema()
```

Other unsupported types (e.g. `Dict[str, int]`) correctly raise a clear `TypeError`; the bare-generic case crashes with an opaque `AttributeError` instead.

devteamaegis:
## What's broken

A bare `typing.Tuple`/`typing.List` field on a `LanceModel` crashes `to_arrow_schema()`:

```python
class Doc(LanceModel):
    items: Tuple
Doc.to_arrow_schema()
# AttributeError: __args__
```

In `_py_type_to_arrow_type`, the list/tuple branch matches on `__origin__` (which a bare generic has) and then reads `py_type.__args__[0]` (which a bare generic does not have). Other unsupported types raise a clean `TypeError`, so this path is inconsistent.

## Why it happens

A bare `typing.List`/`typing.Tuple` has `__origin__` set but no `__args__`.

## Fix

Also require `__args__` on the branch condition. Bare generics then fall through to the existing `raise TypeError("... unsupported type ...")`, matching the contract used for every other unsupported type. Parameterised `List[int]`/`Tuple[int, ...]` still convert as before.

## Test

Added a case asserting a bare `Tuple` field makes `to_arrow_schema()` raise `TypeError` (it raised `AttributeError` before the fix).

Fixes #3502

### DIFF
--- a/python/python/lancedb/pydantic.py
+++ b/python/python/lancedb/pydantic.py
@@ -274,7 +274,9 @@ def _py_type_to_arrow_type(py_type: Type[Any], field: FieldInfo) -> pa.DataType:
     elif py_type is datetime:
         tz = get_extras(field, "tz")
         return pa.timestamp("us", tz=tz)
-    elif getattr(py_type, "__origin__", None) in (list, tuple):
+    elif getattr(py_type, "__origin__", None) in (list, tuple) and getattr(
+        py_type, "__args__", None
+    ):
         child = py_type.__args__[0]
         return _pydantic_list_child_to_arrow(child, field)
     raise TypeError(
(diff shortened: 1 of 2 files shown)


wjones127:
Closing in favor of #3511

PR #3509, opened 2026-06-05
feat: add an elastic dataloader as an iterable dataset

westonpace:
# Elastic Streaming Dataloader

## Motivation

Training large models on LanceDB tables today requires loading the entire dataset
into memory or writing bespoke batching logic. This PR introduces
`StreamingDataset`, a PyTorch `IterableDataset` that streams directly from a
LanceDB table with two hard guarantees that are difficult to achieve together:
**elastic determinism** and **resumability**.

### Elastic determinism

The dataset partitions the table into a fixed number of *splits* (controlled by
`num_splits`, `shuffle_seed`, and `epoch`). Samples are yielded by round-robining
over splits one sample per split per cycle. Because the split structure is fixed,
the set of samples that makes up each global training step is identical regardless
of `world_size` or `num_workers`. You can scale your cluster up or down between
runs and the model sees the same data in the same order — no re-sharding, no
gradient variance from topology changes.

### Resumability

`state_dict()` / `load_state_dict()` capture how many samples each split has
consumed. Because all splits are the same size and the round-robin design keeps
them in lockstep, the state reduces to a single scalar (`samples_consumed_per_split`)
that is topology-independent. A checkpoint saved with 8 GPUs can resume correctly
on 4 GPUs or 16 GPUs without any adjustment.

### PyTorch `IterableDataset` / streaming

`StreamingDataset` implements the standard PyTorch `IterableDataset` interface, so
it drops into any existing `DataLoader` pipeline without modification. Data is
fetched lazily from Lance in chunks — only the rows needed for the current batch are
ever in memory.

Compared to the map dataset this takes more work from pytorch and puts it into the dataset itself (e.g. shuffling, filtering, etc.).  We do this because we cannot achieve things like elastic determinism or prefiltering otherwise.

### Multi-worker support

DataLoader workers are automatically assigned contiguous sub-blocks of splits (the
rank's splits are divided evenly across workers). Each worker is independent:
no shared state, no inter-process coordination. The only constraint is that
`num_splits` must be divisible by `world_size * num_workers`.

That being said, multi-worker is highly discouraged as it relies on multiprocessing which is inefficient.  Still, we want to support it.

### Filters as prefilters

Filters are applied at *permutation-build time* via `PermutationBuilder.filter()`,
not re-evaluated on every fetch. The filtered row IDs are stored in the permutation
table so that subsequent reads see only the matching rows. This allows us to avoid loading rows that don't match the filter (which is the default pytorch behavior)

### Prefetching

Two parameters control the I/O pipeline:

- `read_batch_size` (default 64) — number of rows fetched per `take_offsets` call.
  Larger values amortise per-request overhead, which is critical on object storage
  where a single round-trip can cost ~100 ms.
- `prefetch_batches` (default 4) — number of batches prefetched in parallel per
  split via a `ThreadPoolExecutor`. While the model processes the current batch,
  the next several batches are already in flight, hiding storage latency behind
  compute.

If set correctly then you can get good performance even with num_workers=0 (unless you are bottlenecked on transform).

### Transform parallelism

The underlying `Permutation` API supports a `with_transform()` callback for
decoding, augmentation, and format conversion.  Unfortunately, this is not parallelized.  Pytorch typically parallelizes this with num_workers which is multiprocessing which is highly inefficient.  For simple transforms we should be able to utilize multithreading and Rust based UDFs.  For complex python UDFs we could have a dedicated multiprocessing pipeline for just the transform.  Or we could just utilize multithreading.  In both cases we would exclude the I/O stage from the multiprocessing because that ends up being very memory hungry and inefficient.

### DIFF
--- a/python/python/lancedb/streaming.py
+++ b/python/python/lancedb/streaming.py
@@ -120,9 +120,10 @@ class StreamingDataset(IterableDataset):
         self._transform = transform
         self._worker_info_override = worker_info_override
 
-        # Reference to the live per-split pending-futures deques while __iter__
-        # is running.  None when not iterating.  Used by prefetch_queue_depth.
+        # References to the live per-split deques while __iter__ is running.
+        # None when not iterating.  Used by prefetch_queue_depth.
         self._pending_ref: Optional[list[deque]] = None
+        self._ready_ref: Optional[list[deque]] = None
 
         # Cumulative bytes of Arrow buffer data fetched across all iterations.
         self._bytes_loaded: int = 0
@@ -263,6 +264,7 @@ class StreamingDataset(IterableDataset):
 
         with ThreadPoolExecutor(max_workers=n * max_prefetch) as executor:
             self._pending_ref = pending
+            self._ready_ref = ready
             try:
                 # Prime the pipeline: submit up to max_prefetch batches per split.
                 for i in range(n):
@@ -291,6 +293,7 @@ class StreamingDataset(IterableDataset):
                         yield row
             finally:
                 self._pending_ref = None
+                self._ready_ref = None
 
     @property
     def bytes_loaded(self) -> int:
@@ -325,16 +328,16 @@ class StreamingDataset(IterableDataset):
 
     @property
     def prefetch_queue_depth(self) -> int:
-        """Number of batches currently in-flight across all splits.
+        """Number of rows fetched and ready to yield across all splits.
 
-        Each in-flight batch is a ``take_offsets`` call submitted to the
-        background thread pool that has not yet been consumed.  The maximum
-        value is ``num_assigned_splits * prefetch_batches``.  Returns 0 when
-        not iterating.
+        Counts only rows whose ``take_offsets`` future has already completed
+        and whose data is sitting in memory waiting for the main thread —
+        i.e. rows that can be handed off with no storage wait.  Does not
+        count futures that are still in-flight.  Returns 0 when not iterating.
         """
-        if self._pending_ref is None:
+        if self._ready_ref is None:
             return 0
-        return sum(len(q) for q in self._pending_ref)
+        return sum(len(q) for q in self._ready_ref)
 
     def state_dict(self) -> dict:
         """Snapshot the dataset's consumption state.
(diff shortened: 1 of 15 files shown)


westonpace:
I'm going to declare this ready for review.  I've been verifying it with tests here: github.com/westonpace/lancedb-model-training-example

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -196,17 +197,10 @@ impl PermutationReader {
...
-                .iter()
-                .map(|o| o.to_string())
-                .collect::<Vec<_>>()
-                .join(",")
-        );
+        let in_list: Vec<Expr> = row_ids.iter().map(|id| lit(*id)).collect();
 
         let base_query = QueryRequest {
-            filter: Some(QueryFilter::Sql(filter)),
+            filter: Some(QueryFilter::Datafusion(col(ROW_ID).in_list(in_list, false))),
```
wjones127, line 203: praise: this is a nice win!

python/python/lancedb/streaming.py
wjones127, line 1: question: one thing Claude picked up from this PR: the dataset will ignore a random set of rows `num_rows % num_splits`, because this does `split.rs:528 floors num_rows / num_splits,` to decide splits per row, and that must be exactly equal across each worker. It did claim that this set of rows would be random per epoch, and so maybe they don't care? Or also you could always choose `num_splits` so `num_rows % num_splits == 0`. We could also emit a warning if the remainder isn't zero. What are your thoughts on that?

wjones127 (approved):
Looks awesome. I think you've designed for two really useful properties. I have a question on one detail.

python/python/lancedb/streaming.py
westonpace: This is probably an existing limitation built into the permutation builder.   However, I think you're right that we should raise the limitation into the docs at least.  We could always fix it later if we wanted.  The pytorch dataloader already has a `drop_last` which avoids irregular sized batches so we should probably include the extra rows instead of dropping them.

I'll create a follow-up.

PR #3514, opened 2026-06-07
fix(python): reject repeated where clauses

### LINKED ISSUE #2649: bug(python): using .where after a FTS search results in spurious results
### LanceDB version

_No response_

### What happened?

Running:

```
tb.search(f'"{q}"', fts_columns='_entityname', query_type='fts')
                    .select(['_entitiid', '_score'])
                    .limit(100)
```

Works well. It returns up to 100 results that match the query string `q`

However...

```
tb.search(f'"{q}"', fts_columns='_entityname', query_type='fts')
                    .select(['_entitiid', '_score'])
                    .limit(100)
                    .where('entityId != 0')
```

Results in exactly 100 rows, with no apparent reference to the fts search itself. 

setting `prefilter` to True/False does not change things. 

TBD - working on it

liuzemei:
## Summary
- reject a second `where()` call for sync and async Python query builders so filters are not silently replaced
- add regression coverage confirming the first filter remains in the serialized query object

Fixes #2649

## Tests
- `python3.13 -m py_compile python/python/lancedb/query.py python/python/tests/test_query.py`
- `/Users/neo/.cursor/extensions/charliermarsh.ruff-2026.48.0-darwin-arm64/bundled/libs/bin/ruff check python/python/lancedb/query.py python/python/tests/test_query.py`
- `git diff --check`

Not run locally: targeted pytest. `uv sync --python 3.13 --extra tests --extra dev -v` stalled during local dependency/native package setup after resolving packages, and direct pytest cannot import the uninstalled local package metadata.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -43,6 +43,11 @@ from .util import flatten_columns
 
 BlobMode = Literal["lazy", "bytes", "descriptions"]
 
+MULTIPLE_WHERE_ERROR = (
+    "where() has already been called on this query; combine predicates with AND "
+    "in a single where() call"
+)
+
 _BLOB_MODE_TO_HANDLING = {
     "lazy": "blobs_descriptions",
     "bytes": "all_binary",
@@ -888,6 +893,10 @@ class LanceQueryBuilder(ABC):
         self._bypass_vector_index = None
         self._order_by = None
 
+    def _check_no_existing_where(self):
+        if self._where is not None:
+            raise ValueError(MULTIPLE_WHERE_ERROR)
+
     @deprecation.deprecated(
         deprecated_in="0.3.1",
         removed_in="0.4.0",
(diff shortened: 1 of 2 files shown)


python/python/lancedb/query.py
```diff
@@ -43,6 +43,11 @@
 
 BlobMode = Literal["lazy", "bytes", "descriptions"]
 
+MULTIPLE_WHERE_ERROR = (
+    "where() has already been called on this query; combine predicates with AND "
+    "in a single where() call"
+)
```
wjones127, line 49: nitpick: if a value is only used once, there's no reason to pull it out into a constant. I would just put the error string in the error itself.

wjones127 (commented):
I admit I'm a bit on the fence about this. I agree the current behavior is confusing. But this change feels somewhere between a fix to a footgun and a breaking change. I'm leaning toward merging but mark it as a breaking change, so users are aware.

wjones127:
I'd also add, if we are going to change this behavior, I'd rather change it for all SDKs at once, and not just Python.

wjones127:
Although, I think my preferred solution is to just change the replacement into an AND. There's no obvious reason we couldn't just make it do `({previous}) AND ({new})` when we get a `new` filter. What do you think of that?

wjones127:
Closing in favor of https://github.com/lancedb/lancedb/pull/3585

PR #3516, opened 2026-06-08
fix(python): push down namespace full reads

BubbleCal:
### What is the bug?

Namespace-backed `LanceTable.to_arrow()` full-table reads bypassed the existing `QueryTable` server-side query path and called the lower-level table `to_arrow()` implementation directly. In Geneva/Sophon this could fail while parsing the Arrow IPC response for `hist.get_table().to_arrow()` / `to_pandas()`, even though `hist.get_table().search().to_arrow()` worked.

### What issues or incorrect behavior does the bug cause?

Full-table reads on namespace-backed tables with `QueryTable` pushdown could fail with Arrow IPC parse errors, while query/search reads on the same table succeeded. Since `to_pandas()` delegates through `to_arrow()` for non-blob/native cases, pandas export was affected too.

When `QueryTable` pushdown is enabled, sync and async table `to_arrow()` now construct a plain no-filter, no-limit, all-columns query and execute it through the table-level `_execute_query()` path. `AsyncTable` now preserves namespace context from async namespace connections so async full reads can make the same pushdown decision. Non-namespace tables and namespace tables without `QueryTable` pushdown keep their existing behavior.

### Tests

- `uv run --extra tests --extra dev --no-sync ruff check python/lancedb/table.py python/lancedb/namespace.py python/tests/test_namespace.py`
- `uv run --extra tests --extra dev --no-sync ruff format python/lancedb/table.py python/lancedb/namespace.py python/tests/test_namespace.py`
- `uv run --extra tests --extra dev --no-sync pytest python/tests/test_namespace.py::TestPushdownOperations::test_lance_table_to_arrow_uses_query_pushdown python/tests/test_namespace.py::TestAsyncPushdownOperations::test_async_table_to_arrow_uses_query_pushdown python/tests/test_namespace.py::test_local_table_to_arrow_and_to_pandas_are_unchanged -q`
- `uv run --extra tests --extra dev --no-sync pytest python/tests/test_namespace.py -q`

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -92,6 +92,12 @@ BlobMode = Literal["lazy", "bytes", "descriptions"]
 _VALID_BLOB_MODES = ("lazy", "bytes", "descriptions")
 
 
+def _should_push_down_query_table(
+    namespace_client: Optional[Any], pushdown_operations: set
+) -> bool:
+    return namespace_client is not None and "QueryTable" in pushdown_operations
+
+
 def _validate_blob_mode(blob_mode: BlobMode) -> None:
     if blob_mode not in _VALID_BLOB_MODES:
         modes = ", ".join(repr(mode) for mode in _VALID_BLOB_MODES)
@@ -2333,6 +2339,11 @@ class LanceTable(Table):
         Returns
         -------
         pa.Table"""
+        if _should_push_down_query_table(
+            self._namespace_client, self._pushdown_operations
+        ):
+            return self._execute_query(Query()).read_all()
+
         return LOOP.run(self._table.to_arrow())
 
     def to_polars(self, batch_size=None) -> "pl.LazyFrame":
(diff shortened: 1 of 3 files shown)


python/python/lancedb/table.py
```diff
@@ -2333,6 +2339,11 @@ def to_arrow(self) -> pa.Table:
         Returns
         -------
         pa.Table"""
+        if _should_push_down_query_table(
+            self._namespace_client, self._pushdown_operations
+        ):
+            return self._execute_query(Query()).read_all()
```
Xuanwo, line 2345: This changes full-table reads into a `QueryTable` request built from an empty `Query()`, but `_query_to_namespace_request()` maps `limit=None` to `k=10`. That means namespace-backed `to_arrow()` / non-blob `to_pandas()` can silently return only 10 rows instead of the full table. We need a request/path that preserves unbounded plain-scan semantics, plus a >10 row regression test through the real request conversion layer.

Xuanwo (changes requested):
This needs one correctness fix before merging.

python/python/lancedb/table.py
BubbleCal: Fixed in a041f480. `_query_to_namespace_request()` now maps plain non-vector/non-FTS queries with `limit=None` to `sys.maxsize`, matching the existing remote-table unbounded sentinel, while vector/FTS queries keep the existing default `k=10`. The namespace regression now goes through the real `_execute_server_side_query()` / request conversion layer with 12 rows and asserts both `to_arrow()` and `to_pandas()` send `k == sys.maxsize`.

Xuanwo (approved):

PR #3518, opened 2026-06-09
test(python): add nested field regression matrix tests

### LINKED ISSUE #3406: Add cross-binding nested field regression matrix
Parent: lancedb/lancedb#3401

## Summary

Add a cross-binding regression matrix so nested field support remains consistent across Python sync/async, Rust, Node, local tables, and remote tables.

## Scope

* Cover nested scalar, vector, and FTS index lifecycle where each binding exposes the API.
* Cover create, list, stats, search/filter, append, and optimize behavior.
* Cover field-name edge cases: mixed case, special characters, top-level literal dots, nested special characters, and same-name leaves.
* Reuse the same semantic matrix across bindings instead of adding unrelated one-off tests.

## Acceptance criteria

* Python sync/async, Rust, and Node have nested field smoke coverage.
* Local and remote tables agree on canonical field path behavior.
* The matrix includes `rowId`, `` `row-id` ``, `` `a.b` ``, `MetaData.userId`, and `` `meta-data`.`user-id` ``.
* Tests fail if list/describe/stats regress to leaf-only field names.

nuthalapativarun:
## Summary

Closes #3406

Add a regression matrix in `python/python/tests/test_nested_fields.py` that exercises the full nested field index lifecycle for both the sync and async Python table APIs.  The tests will fail if any implementation regresses to leaf-only field names in `list_indices`, `index_stats`, search, or filter results.

## Test scenarios covered

**Index types:** BTree scalar, IvfPq vector, FTS

**Field-name edge cases (per acceptance criteria):**
- `rowId` — camelCase top-level field
- `` `row-id` `` — hyphenated top-level field (escaped)
- `parent.`\``leaf.name`\`` ` — struct leaf whose name contains a literal dot
- `MetaData.userId` — mixed-case nested path
- `` `meta-data`.`user-id` `` — hyphenated struct with hyphenated leaf

**Lifecycle operations per index type:**
- `create_index` / `create_scalar_index` / `create_fts_index`
- `list_indices` → verify canonical full dotted path (not leaf name)
- `index_stats` → verify row count and index type
- Filtered scan (`WHERE nested.field = value`)
- Vector search via nested embedding column
- FTS search via nested text column
- `add` (append) then re-check index listing
- `optimize` then re-check index listing

**Both sync and async APIs** are covered in parallel test classes.

## Notes

Lance forbids top-level field names that contain a literal `.`, so the `` `a.b` `` acceptance-criterion variant is exercised as a *struct leaf* field (`parent.`\``leaf.name`\``) rather than a top-level column.

### DIFF
--- a/python/python/tests/test_nested_fields.py
+++ b/python/python/tests/test_nested_fields.py
@@ -172,9 +172,7 @@ def _nested_vector_schema() -> pa.Schema:
             pa.field("id", pa.int32()),
             pa.field(
                 "image",
-                pa.struct(
-                    [pa.field("embedding", pa.list_(pa.float32(), DIM))]
-                ),
+                pa.struct([pa.field("embedding", pa.list_(pa.float32(), DIM))]),
             ),
             pa.field(
                 "MetaData",
@@ -405,9 +403,7 @@ class TestNestedVectorIndexSync:
             name="image_emb_idx",
         )
         results = (
-            tbl.search(_vec(0), vector_column_name="image.embedding")
-            .limit(5)
-            .to_list()
+            tbl.search(_vec(0), vector_column_name="image.embedding").limit(5).to_list()
         )
         assert len(results) > 0
 
@@ -548,9 +544,7 @@ class TestNestedScalarIndexAsync:
     async def test_index_stats_canonical_path(self, async_db):
         """index_stats round-trip: create on nested field, verify stats."""
         tbl = await async_db.create_table("t", _nested_scalar_data())
-        await tbl.create_index(
-            "MetaData.userId", config=BTree(), name="meta_uid_idx"
-        )
+        await tbl.create_index("MetaData.userId", config=BTree(), name="meta_uid_idx")
         stats = await tbl.index_stats("meta_uid_idx")
         assert stats is not None
         assert stats.index_type == "BTREE"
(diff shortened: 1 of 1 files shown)


python/python/tests/test_nested_fields.py
```diff
@@ -0,0 +1,686 @@
...
+
+    def test_append_and_list_indices_stable(self, sync_db):
+        """After appending rows the index listing must remain unchanged."""
+        tbl = sync_db.create_table("t", _nested_scalar_data())
+        tbl.create_scalar_index(
+            "MetaData.userId", index_type="BTREE", name="meta_uid_idx"
+        )
+        tbl.add(_nested_scalar_data(nrows=4))
+        col_map = _columns_by_name_sync(tbl)
+        assert col_map["meta_uid_idx"] == ["MetaData.userId"]
```
wjones127, line 340: suggestion: could you add another tests that validates merge_insert works when the `on` column is this camel case nested field and there is a scalar index on that column? I think that's the one thing I can think of that missing here.

wjones127 (approved):
This looks good. I have a suggestion for an additional test case, but if it's turns out to be failing or too complex, happy to defer to a follow up.

python/python/tests/test_nested_fields.py
nuthalapativarun: Thanks for the suggestion! I tried adding this — keying `merge_insert` on the `MetaData.userId` nested path with a BTree index on it — but it currently fails at the lance layer:

```
RuntimeError: lance error: Invalid user input: Schema error: No field named "userId". Valid fields are target."rowId", target."row-id", target."MetaData", target."meta-data", target._rowid, target._rowaddr.
```

It looks like `merge_insert`'s `on` argument doesn't currently resolve dotted nested field paths in lance v8.0.0-beta.6 — it appears to only match against top-level field names. Given that, this is exactly the case you mentioned (failing/too complex) — I'd rather not paper over it with a workaround test, and would suggest tracking it as a follow-up issue against `merge_insert` nested-key support rather than blocking this PR on it. Happy to file that issue if useful!

PR #3520, opened 2026-06-09
feat: support rename_table on LanceNamespaceDatabase

### LINKED ISSUE #3412: feat: support rename_table on LanceNamespaceDatabase
It's unimplemented here:

https://github.com/lancedb/lancedb/blob/20556e23a943274be567e93eab5402e56216444b/rust/lancedb/src/database/namespace.rs#L489-L499

But it's part of the namespace spec: https://lance.org/format/namespace/operations/#operation-list:~:text=%E2%9C%93-,RenameTable,-1

nuthalapativarun:
## Summary

Closes #3412

Implements `rename_table` for `LanceNamespaceDatabase` (sync and async Python) and the Rust `NamespaceDatabase` backend. Previously these raised `NotImplementedError`; this PR delegates to the `LanceNamespace.rename_table` method which is part of the lance-namespace spec.

### Changes

- **`rust/lancedb/src/database/namespace.rs`**: Remove the `NotImplementedError` stub for `rename_table`. Build a `RenameTableRequest` (with `id`, `new_table_name`, and optionally `new_namespace_id`) and call `self.namespace.rename_table(...)`, mirroring the existing `drop_table` pattern.

- **`python/python/lancedb/namespace.py`**: Import `RenameTableRequest` from `lance_namespace`. Replace the `raise NotImplementedError` in both `LanceNamespaceDatabase.rename_table` (sync) and `AsyncLanceNamespaceDatabase.rename_table` (async) with a call to `self._namespace_client.rename_table(request)`.

- **`python/python/tests/test_namespace.py`**: Replace the `test_rename_table_not_supported` test (which checked for `NotImplementedError`) with `test_rename_table`, which:
  1. Creates a table in a namespace
  2. Calls `rename_table` with `cur_namespace_path` and `new_namespace_path`
  3. Asserts the old name is gone from `table_names()`
  4. Asserts the new name appears in `table_names()`
  5. Verifies the renamed table can be opened

## Test plan

- [ ] Existing namespace tests pass in CI (all rely on `lance.namespace.DirectoryNamespace` which requires the full lance package)
- [ ] `test_rename_table` exercises the full rename path: create → rename → verify old gone → verify new present → open
- [ ] Rust build passes with the updated `namespace.rs` (requires Rust toolchain in CI)

### DIFF
--- a/python/python/lancedb/namespace.py
+++ b/python/python/lancedb/namespace.py
@@ -58,6 +58,7 @@ from lance_namespace import (
     ListTablesRequest,
     DescribeNamespaceRequest,
     DropTableRequest,
+    RenameTableRequest,
     ListNamespacesRequest,
     CreateNamespaceRequest,
     DropNamespaceRequest,
@@ -604,9 +605,14 @@ class LanceNamespaceDBConnection(DBConnection):
             cur_namespace_path = []
         if new_namespace_path is None:
             new_namespace_path = []
-        raise NotImplementedError(
-            "rename_table is not supported for namespace connections"
+        cur_table_id = cur_namespace_path + [cur_name]
+        new_namespace_id = new_namespace_path if new_namespace_path else None
+        request = RenameTableRequest(
+            id=cur_table_id,
+            new_table_name=new_name,
+            new_namespace_id=new_namespace_id,
         )
+        self._namespace_client.rename_table(request)
 
     @override
     def drop_database(self):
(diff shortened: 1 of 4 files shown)


wjones127 (approved):
This looks good, assuming the tests are passing!

nuthalapativarun:
Fixed the failing CI: `test_rename_table` was calling `db.rename_table(...)` against the `dir` namespace backend, but `lance-namespace-impls`'s directory implementation doesn't override `rename_table` yet (only the `rest` backend does in lance v8.0.0-beta.6) — it falls through to the default trait method which returns `NotSupported`. The Rust-side wiring in this PR is correct and forward-compatible (it'll work for any namespace backend that implements `rename_table`, e.g. `rest`); I've updated the test to assert the current `dir`-backend behavior with a comment explaining the limitation, so it'll need to be revisited once the `dir` backend gains rename_table support upstream.

PR #3523, opened 2026-06-09
feat: adds isin support to the 'Expr' builder

stumpylog:
The `Expr` build already includes a lot of useful filtering options, `eq, ne, gt/gte, lt/lte, and_, or_, contains, cast`, but is was missing a membership like `isin`.  This PR adds that support, as minimally as possible, allowing easy filtering for membership in a list, without needing to be a series of `where` expressions.

I didn't see anything in CONTRIBUTING.md about needing a feature request or issue first, so I just made the change.  My apologies if I missed that somewhere.

Thanks for the vector store, we're using it now in paperless-ngx.


wjones127 (approved):
This looks well done. Just need to address the lints and we can merge.

stumpylog:
Shoot I've gotten dependent on prek running I guess.  Taken care of!

PR #3524, opened 2026-06-09
feat(python): accept Expr in Table.delete and merge when_not_matched_by_source_delete

stumpylog:
Another little pain point as I was working to integrate with paperless-ngx.  The read path of table.search() or table.query() already accepted an Expr, but write paths Table.delete and merge_insert(...).when_not_matched_by_source_delete did not.  This PR attempts to close that gap, so writes and reads can both use Expr, instead of one side needing to build a string.


python/src/table.rs
```diff
@@ -28,6 +29,12 @@ use pyo3::{
 
 mod scannable;
 
+#[derive(FromPyObject)]
+enum PredicateArg {
+    Expr(PyExpr),
+    Sql(String),
+}
```
wjones127, line 36: praise: I really like this implementation. It's a nice use of pyo3's capabilities.

python/python/lancedb/table.py
```diff
@@ -5209,25 +5211,26 @@ async def _do_merge(
...
+                when_not_matched_by_source_condition_expr=merge._when_not_matched_by_source_condition_expr,
                 timeout=merge._timeout,
                 use_index=merge._use_index,
                 use_lsm_write=merge._use_lsm_write,
                 validate_single_shard=merge._validate_single_shard,
             ),
         )
 
-    async def delete(self, where: str) -> DeleteResult:
+    async def delete(self, where: Union[str, Expr]) -> DeleteResult:
```
wjones127, line 5222: issue(blocking): before I can approve, I'd like to see at least one unit test in Python going through the sync and async Table delete method each.

python/python/lancedb/merge.py
```diff
@@ -62,7 +65,7 @@ def when_not_matched_insert_all(self) -> LanceMergeInsertBuilder:
         return self
 
     def when_not_matched_by_source_delete(
-        self, condition: Optional[str] = None
+        self, condition: Union[str, Expr, None] = None
```
wjones127, line 68: issue(blocking): same here, need at least one test for this.

wjones127 (changes requested):

stumpylog:
Tests added!

wjones127 (approved):
Nicely done!

Issue #3231, opened 2026-04-02
feat: Add blob column support to LanceDB


westonpace:
## Summary

Add first-class blob (large binary object) support to LanceDB, built on top of Lance's Blob V2 storage (Format 2.2). This enables efficient storage and retrieval of multimodal data such as images, audio, video, PDFs, and other large binary payloads.

## Motivation

LanceDB is increasingly used for multimodal AI workloads where users need to store and retrieve large binary objects alongside vector embeddings and structured metadata. Lance already supports Blob V2 at the storage layer, but LanceDB does not yet expose a high-level API for working with blobs. Users currently have to drop down to the Lance layer or store blob data inefficiently as regular binary columns.

## Scope

- **SDKs**: Rust core, Python bindings, and TypeScript bindings (all three).
- **Table type**: Native/local tables. Remote table (LanceDB Cloud) support is out of scope for this issue and should be tracked as a follow-up.

---

## 1. Defining Blob Columns

Users need a way to declare that a column should be treated as a blob column. This should be supported via three mechanisms:

### 1a. Field metadata

Users can set Arrow field metadata to mark a `LargeBinary` field as a blob column. This matches the lance API but is pretty low-level.  I prefer the other options I think.

```python
import pyarrow as pa

schema = pa.schema([
    pa.field("id", pa.int64()),
    pa.field("embedding", pa.list_(pa.float32(), 128)),
    pa.field("image", pa.large_binary(), metadata={"lance-encoding:blob": "true"}),
])
```

### 1b. Arrow extension type

Provide a registered Arrow extension type (e.g. `lancedb.blob`) so users can declare blob columns without manually setting metadata.

```python
import lancedb

schema = pa.schema([
    pa.field("id", pa.int64()),
    pa.field("image", lancedb.blob()),  # returns a type with proper metadata
])
```

### 1c. Pydantic model (Python only)

Extend the existing Pydantic integration to support a `Blob` type annotation.

```python
from lancedb.pydantic import LanceModel, Blob

class MyModel(LanceModel):
    id: int
    embedding: Vector(128)
    image: Blob
```

---

## 2. Writing Blob Data

Users should be able to provide blob data in several ways when inserting rows.

### 2a. Inline binary data

Provide blob data directly as a `LargeBinary` (or `Binary`) Arrow array. Lance will store it using its adaptive strategy (inline for small payloads, packed blocks for medium, dedicated regions for large).

```python
table.add([
[...]

PR #3528, opened 2026-06-10
feat(rust): add blob v2 schema declaration and write path

geruh:
First Rust PR for #3231. Lance already stores blob v2. This adds the LanceDB write side.

```rust
let schema = Schema::new(vec![
    Field::new("id", DataType::Int64, false),
    lancedb::blob("image", true),
]);

let table = db.create_table("photos", schema).execute().await?;

table.add(batch_with_large_binary_image_column).execute().await?;
```

Read/materialize and Python are follow-up PRs.

### Testing

- cargo test -p lancedb --test blob_integration
- cargo test -p lancedb blob:: datafusion::blob_coerce
- cargo test -p lancedb (591 passed)
- cargo clippy --features remote --tests

### DIFF
--- a/rust/lancedb/src/table/datafusion/blob_coerce.rs
+++ b/rust/lancedb/src/table/datafusion/blob_coerce.rs
@@ -1,0 +1,521 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+//! Coerces write-path input into blob v2 struct columns.
+//!
+//! [`coerce_blob_expr`] is invoked from [`super::cast::cast_to_table_schema`].
+
+use std::any::Any;
+use std::sync::{Arc, LazyLock};
+
+use arrow_array::{Array, ArrayRef, StructArray, new_null_array};
+use arrow_schema::{DataType, Field, FieldRef, Fields};
+use datafusion_common::config::ConfigOptions;
+use datafusion_expr::{
+    ColumnarValue, ScalarFunctionArgs, ScalarUDF, ScalarUDFImpl, Signature, Volatility,
+};
+use datafusion_physical_expr::ScalarFunctionExpr;
+use datafusion_physical_plan::PhysicalExpr;
+
+use crate::error::{Error, Result};
+
+static COERCE_BLOB_UDF: LazyLock<Arc<ScalarUDF>> =
+    LazyLock::new(|| Arc::new(ScalarUDF::from(CoerceBlobUdf::new())));
+
+/// Build a projection expression coercing `input_expr` into the blob struct
+/// declared by `table_field`.
+///
+/// The table field (metadata included) is the expression return field. Raw
+/// binary lands in `data`; struct input is normalized to the declared layout.
+pub(super) fn coerce_blob_expr(
+    input_expr: Arc<dyn PhysicalExpr>,
+    input_field: &Field,
+    table_field: &FieldRef,
+    config: &Arc<ConfigOptions>,
+) -> Result<(Arc<dyn PhysicalExpr>, FieldRef)> {
+    match input_field.data_type() {
+        DataType::Binary | DataType::LargeBinary => {}
+        DataType::Struct(children) => {
+            if !children
+                .iter()
+                .any(|c| c.name() == "data" || c.name() == "uri")
+            {
+                return Err(Error::InvalidInput {
+                    message: format!(
+                        "blob struct input for column '{}' must contain a 'data' or 'uri' child",
+                        table_field.name()
+                    ),
+                });
+            }
+        }
+        other => {
+            return Err(Error::InvalidInput {
+                message: format!(
+                    "cannot coerce column '{}' with type {} into a blob v2 struct. \
+                     expected Binary, LargeBinary, or a Struct with a 'data' or 'uri' child",
+                    table_field.name(),
+                    other,
+                ),
+            });
+        }
+    }
+
+    let expr: Arc<dyn PhysicalExpr> = Arc::new(ScalarFunctionExpr::new(
+        &format!("coerce_blob({})", table_field.name()),
+        COERCE_BLOB_UDF.clone(),
+        vec![input_expr],
+        table_field.clone(),
+        config.clone(),
+    ));
+    Ok((expr, table_field.clone()))
+}
+
+/// Scalar UDF rewriting binary or struct input into the declared blob layout.
+#[derive(Debug, Hash, PartialEq, Eq)]
+struct CoerceBlobUdf {
+    signature: Signature,
+}
+
+impl CoerceBlobUdf {
+    fn new() -> Self {
+        Self {
+            signature: Signature::any(1, Volatility::Immutable),
+        }
+    }
+}
+
+impl ScalarUDFImpl for CoerceBlobUdf {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn name(&self) -> &str {
+        "coerce_blob"
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> datafusion_common::Result<DataType> {
+        Err(datafusion_common::DataFusionError::Internal(
+            "coerce_blob is only usable through coerce_blob_expr".into(),
+        ))
+    }
+
+    fn invoke_with_args(
+        &self,
+        args: ScalarFunctionArgs,
+    ) -> datafusion_common::Result<ColumnarValue> {
+        let array = match &args.args[0] {
+            ColumnarValue::Array(array) => array.clone(),
+            ColumnarValue::Scalar(scalar) => scalar.to_array_of_size(args.number_rows)?,
+        };
+        let coerced = coerce_array(&array, args.return_field.as_ref())
+            .map_err(|e| datafusion_common::DataFusionError::External(e.into()))?;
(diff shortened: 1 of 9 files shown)


rust/lancedb/src/table/datafusion/blob_coerce.rs
```diff
@@ -0,0 +1,521 @@
...
+            message: format!(
+                "blob v2 column '{}' must be a struct, table declares {}",
+                field.name(),
+                field.data_type()
+            ),
+        });
+    };
+
+    match column.data_type() {
+        DataType::Binary | DataType::LargeBinary => {
```
wjones127: suggestion: we should also support `BinaryView` here.


```suggestion
        DataType::Binary | DataType::LargeBinary | DataType::BinaryView => {
```

rust/lancedb/src/table/datafusion/blob_coerce.rs
```diff
@@ -0,0 +1,521 @@
...
+        }
+    }
+
+    let expr: Arc<dyn PhysicalExpr> = Arc::new(ScalarFunctionExpr::new(
+        &format!("coerce_blob({})", table_field.name()),
+        COERCE_BLOB_UDF.clone(),
+        vec![input_expr],
+        table_field.clone(),
+        config.clone(),
+    ));
```
wjones127, line 119: suggestion: you might actually be able to get the same effect by composing pre-existing scalar functions like cast and get_field. Not required, but might cut down on the amount of custom code required to implement this.

wjones127 (approved):
This all looks good. I have a few optional suggestions. Looking forward to the future PRs :)

rust/lancedb/src/table/datafusion/blob_coerce.rs
geruh: I looked at doing this with `cast/get_field/named_struct`. i might be missing something, but i don't think it covers the main case where users pass raw bytes on add, and lance's write path seems to expect the blob struct already. i couldn't find a lance helper that reshapes a whole bytes column. `BlobArrayBuilder` looks row processed to me. 

We already use `blob_field` for schema. generic struct coercion also skips missing fields instead of null-filling, and `named_struct` might not handle row nulls the way we need. So for now i kept the udf as the step that wraps bytes into the struct before lance runs. happy to change this if there's a cleaner way and I'm missing somehting.

wjones127: Isn't it just something like: `named_struct('data', col('user_binary_col'), ...)`? Or is that case more complicated than moving the binary column into the `data` field of the expected struct?

geruh: Youre right, we could just do this, and drop the UDF, thanks Will!

westonpace (approved):
I'll merge but I'm copying Claude's review here as well.  Feel free to address any items in follow-ups.  The primary concern Claude has (unresolved review item) is a false positive.

---

Now I have enough context for a thorough review.

---

## Code Review: PR #3528 — `feat(rust): add blob v2 schema declaration and write path`

### Overview

This PR adds the LanceDB write-side support for Lance blob v2 columns. It introduces:
1. A public `blob()` helper to declare a blob field in a schema (`rust/lancedb/src/blob.rs`)
2. Automatic storage-format bump to `>=V2_2` and stable-row-id enablement when a blob column is detected
3. A DataFusion physical-plan coercion step (`blob_coerce.rs`) that reshapes raw `Binary`/`LargeBinary`/`BinaryView` input into the blob struct layout before Lance's write path runs
4. An integration test suite covering the main write scenarios

The design is clean and follows the existing project patterns well. Here are my findings:

---

### Positives

- **Good API surface**: `lancedb::blob("name", nullable)` is minimal and matches the docstring example. Re-export from `lib.rs` makes it discoverable.
- **Well-tested**: Unit tests in `blob.rs` and `blob_coerce.rs` cover all major code paths. The integration test file is comprehensive (null handling, binary/LargeBinary/BinaryView variants, wide-struct layout, overwrite semantics, namespace backend, connection-level override).
- **Correct short-circuit**: `cast_to_table_schema` returns early when schemas match, and the new `coerce_blob_expr` also correctly detects an exact-match input and produces a no-projection result (`exact_blob_input_passes_through_unchanged` test).
- **Storage version logic is conservative**: `ensure_blob_storage_version` only bumps upward, never downward, and is a no-op for non-blob schemas.

---

### Issues and Suggestions

#### 1. Open review thread — replace `ScalarFunctionExpr` with pre-existing scalar functions (unresolved)

`wjones127` has an unresolved thread pointing out that `coerce_blob_expr` could be simplified by composing existing `named_struct` / `cast` / `get_field` scalar functions instead of building `ScalarFunctionExpr` manually. The author acknowledged this in the latest reply. **This thread should be resolved before merging**, either by applying the simplification or by explicitly documenting why it cannot be done.

The current code builds `ScalarFunctionExpr` values manually, which duplicates logic already available via the `datafusion::functions::core` helpers already imported in `cast.rs`. Notably, `cast.rs` already uses `named_struct()` and `get_field()` for struct coercion — the same primitives could be reused here without a separate `blob_coerce.rs` file.

#### 2. `blob_coerce` module is `pub(super)` but not `pub` — fine, but it breaks `blob_coerce` out into its own file while `cast.rs` already handles the general struct-coerce case

The blob coercion path and the generic struct coercion path in `cast.rs` have significant overlap (both use `get_field` + `named_struct` + `CastExpr`). If the `named_struct`-based simplification from thread #2 is applied, the logic may shrink enough to inline into `cast.rs` directly, avoiding the extra file.

#### 3. Stable row id priority ordering — `has_blob_columns` wins over connection-level config

In `listing.rs`:
```rust
.or(has_blob_columns(&data_schema).then_some(true))
```

The chain is `table-level override → connection default → blob auto-enable`. This means a user who explicitly sets `OPT_NEW_TABLE_ENABLE_STABLE_ROW_IDS=false` at the **connection** level will have that preference **silently overridden** by blob auto-enable. The integration test `connection_level_stable_row_id_setting_wins_over_blob_default` actually **expects** the connection-level setting to win, but the code does the opposite — the `.or(...)` chain means connection-level `false` (stored in `self.new_table_config.enable_stable_row_ids`) will dominate `has_blob_columns`.

Wait — let me re-read: `stable_row_ids_override` is the table-level `storage_option`, `self.new_table_config.enable_stable_row_ids` is the connection-level setting. If the connection says `false`, then `self.new_table_config.enable_stable_row_ids = Some(false)`, which resolves to `Some(false)` and the `.or(...)` chain stops — the blob default (`true`) is never reached. So the test is correct and the code is correct. The comment in the code ("table-level override takes precedence over connection config") is now missing from the diff (removed in this PR) — the intent is less clear. **Consider restoring or rewriting the comment** explaining the precedence chain (table override → connection config → blob default).

#### 4. `data_schema` call differs between `listing.rs` and `namespace.rs`

- `listing.rs`: `request.data.arrow_schema()` → returns `Schema` by value
- `namespace.rs`: `request.data.schema()` → returns `Arc<Schema>`

This suggests `CreateTableData` exposes two different methods depending on the context. If both paths exist for the same underlying type, consider checking whether they're semantically equivalent or if there's a risk of divergence.

#### 5. `overwrite_replaces_blob_schema_with_input_schema` — behavior could surprise users

The test documents that a raw `LargeBinary` overwrite drops the blob marker. The docstring on `AddDataMode::Overwrite` was updated to mention this, which is good. However, this means users who do `table.add(batch).mode(Overwrite)` with a plain binary column will silently lose their blob column declaration, potentially breaking subsequent reads. Consider:
- Emitting a warning log in this case
- Or documenting this prominently in the `blob()` function's rustdoc

#### 6. Minor: `blob_coerce.rs` — `typed_null` error message quality

```rust
fn typed_null(data_type: &DataType) -> Result<Arc<dyn PhysicalExpr>> {
    let scalar = ScalarValue::try_from(data_type).map_err(|e| Error::InvalidInput {
        message: format!("cannot build null literal for blob child type {data_type}: {e}"),
    })?;
```

The `ScalarValue::try_from` will only fail for types like `Dictionary` or `RunEndEncoded` that can't be trivially null-constructed — this is an internal invariant that Lance controls. The error message is fine, but there's no test exercising this branch. Since it's an internal path (the declared struct comes from `lance::blob::blob_field`), this is low risk.

#### 7. Test isolation — `namespace_create_applies_blob_defaults` test

```rust
let mut properties = std::collections::HashMap::new();
properties.insert("root".to_string(), tmp.path().to_str().unwrap().to_string());
let db = lancedb::connect_namespace("dir", properties).execute().await?;
```

The `connect_namespace` API is not shown in this diff. If this is a recently added API, ensure it's also covered by documentation. If the namespace backend has different code paths for `data_storage_version`, the test is valuable — but the `ensure_blob_storage_version` call in `namespace.rs` only updates `params` without checking whether it's actually applied, so this test serves as a good regression guard.

---

### Summary

| Area | Status |
|---|---|
| Public API design | Good — minimal and composable |
| Write path correctness | Good — binary/LargeBinary/BinaryView all coerced |
| Storage format auto-bump | Correct and conservative |
| Stable row id precedence | Correct; comment removed — should be restored |
| Test coverage | Very thorough |
| Open review thread (simplify coerce impl) | **Unresolved — needs decision before merge** |
| Overwrite behavior documentation | Mentioned in variant doc; consider log warning |
| Code duplication with cast.rs struct logic | Minor; worth addressing if thread #2 is resolved |

The main blocker is the unresolved review thread about simplifying `coerce_blob_expr`. The rest are suggestions — the code is correct and well-tested.

PR #3529, opened 2026-06-10
fix: support LargeList label list indexes

Xuanwo:
## Summary

This PR extends nested-field regression coverage across Rust local/remote, Python sync/async, and Node so canonical escaped paths stay consistent across scalar, vector, and FTS index lifecycle behavior.

It also aligns LanceDB's LabelList type gate with Lance by accepting `LargeList<primitive>` columns while keeping `List<Struct<...>>` unsupported until Lance defines stable membership semantics for struct labels.

Part of #3406.


rust/lancedb/src/remote/table.rs
```diff
@@ -3914,6 +3922,22 @@ mod tests {
     async fn test_create_index_nested_field_paths() {
```
BubbleCal, line 3922: This only verifies the outbound `create_index` request. The regression also affects the remote response path (`list_indices` / `index_stats`), so we should add a mocked response assertion for these new nested field cases to catch leaf-only name regressions.

BubbleCal (approved):

PR #3535, opened 2026-06-10
perf: drop N+1 in RemoteTable::list_indices

### LINKED ISSUE #3494: perf: drop N+1 in RemoteTable::list_indices via server-version gate

wjones127:
`RemoteTable::list_indices` currently makes one `/index/list/` call plus one `/index/{name}/stats/` call per index just to recover `index_type`.

When the server returns `index_type` directly in the `/index/list/` response, all enriched fields are used and the per-index stats fan-out is skipped entirely. When `index_type` is absent (legacy servers), the existing stats fallback runs as before. This is content-based: no version header required.

## Changes

- `RemoteTable::parse_index_list_response` replaces the old split between enriched and legacy parsers. A single struct deserializes both old and new response shapes, with all fields except `index_name` and `columns` optional. `index_type` acts as the sentinel: present → use enriched fields directly; absent → call `/index/{name}/stats/`.

## Tests

Added `test_list_indices_enriched` covering:
- All enriched fields populated correctly when `index_type` is in the list response
- Optional fields absent from the response deserialize as `None`
- Stats endpoint is **not** called (panics if hit), verifying the fan-out is eliminated

Existing `test_list_indices` and `test_list_indices_nested_field_paths` exercise the legacy path unchanged.

## Depends on

- #3497 (expand `IndexConfig`) — already merged
- Server-side enriched response support

Closes #3494

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -45,6 +45,7 @@ use arrow_array::{RecordBatch, RecordBatchIterator, RecordBatchReader};
 use arrow_ipc::reader::FileReader;
 use arrow_schema::{DataType, SchemaRef};
 use async_trait::async_trait;
+use chrono::{DateTime, Utc};
 use datafusion_common::DataFusionError;
 use datafusion_physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion_physical_plan::{ExecutionPlan, RecordBatchStream, SendableRecordBatchStream};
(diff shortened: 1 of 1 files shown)


brendanclement (approved):
🚀 🚢  Great work. thank you for reducing latency / load!

PR #3541, opened 2026-06-12
feat: skills to connect and update column metadata

dantasse:
Two skills to help people connect and manage their column metadata using a server that implements the [REST API](https://lance.org/format/catalog/rest/)

lancedb-column-metadata was built using the [Claude skill creator](https://claude.com/plugins/skill-creator); without the skill it was usually calling at least one method that didn't exist and usually not setting "replace": "false". So, while the base case is already pretty good, adding this skill improves things somewhat.

lancedb-connect should help with most agentic workflows, because "finding all the things you need to connect to your server" can be the hardest part.

### DIFF
--- a/.agents/skills/lancedb-column-metadata/SKILL.md
+++ b/.agents/skills/lancedb-column-metadata/SKILL.md
@@ -1,0 +1,178 @@
+---
+name: lancedb-column-metadata
+description: Column metadata authoring for LanceDB tables via the REST API. This skill is required for tasks like writing field descriptions, setting tags on columns (field_type, model, project_id, version), classifying columns as embeddings vs labels vs eval metrics, or grouping versioned columns into logical families — because it has the API integration needed to read the schema and persist metadata back. Invoke whenever someone wants to document, annotate, tag, or classify what their table columns ARE. Trigger even without an explicit "LanceDB" mention, as long as the context is column-level documentation or tagging for an ML or vector database table.
+metadata:
+  short-description: Write column descriptions, tags, and logical groupings to a LanceDB table
+---
+
+## Overview
+
+This skill authors column-level metadata for a LanceDB table. It connects to a LanceDB deployment over its REST API, inspects the table schema, generates appropriate metadata, and writes it back.
+
+## Step 0: Establish the connection
+
+Use the `lancedb-connect` skill (invoke it via the Skill tool) to resolve the base URL and auth headers (`x-api-key`, `x-lancedb-database`) for whichever deployment the user is working against — LanceDB Cloud, enterprise/self-hosted, or a local dev server. Skip it only if the connection details are already established in the conversation.
+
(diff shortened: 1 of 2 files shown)


.agents/skills/lancedb-connect/SKILL.md
```diff
@@ -0,0 +1,43 @@
...
+1. **Base URL** — the endpoint
+2. **Headers** — `x-api-key`, and usually `x-lancedb-database`
+
+## Resolution steps
+
+1. If the user already gave a URL and API key (or said which environment they're working against), use that.
+2. Otherwise, look for credentials already available in the environment:
+   - Env vars like `LANCEDB_URI` / `LANCEDB_HOST` / `LANCEDB_API_KEY`
+   - A profile in `~/.lancedb/config.toml` (SDK and CLI config: `http_server_url`, `api_key`, `database`)
+   - A LanceDB endpoint already running or port-forwarded locally (the REST default port is 10024, i.e. `http://localhost:10024`)
```
brendanclement: I don't think the spec has a default port

.agents/skills/lancedb-column-metadata/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+
+Tags are open-ended — use whatever key suffix and value make sense given the user's intent. The tag suffix should describe *what is being classified* (e.g., `field_type`, `model`, `project_id`) and the value describes *how*.
+
+## Step 1: Resolve the table identifier
+
+You need:
+- **Table name** (required) — e.g., `my_table` or `my_namespace.my_table`
+- **Database name** — ask if not provided and not inferable from context; it goes in the `x-lancedb-database` header (or the hostname, for LanceDB Cloud), never in the URL path
+
+The table identifier in the URL path is typically `table_name` for a top-level table, or `namespace.table_name` if the table lives in a namespace. The API accepts a `delimiter` query parameter to parse compound identifiers (default `.`).
```
brendanclement: the default delimiter is $ 

https://lance.org/format/namespace/object-relationship/#object-identifier

.agents/skills/lancedb-connect/SKILL.md
dantasse: ([checking again](https://editor.swagger.io/?url=https://raw.githubusercontent.com/lance-format/lance-namespace/refs/heads/main/docs/src/spec.yaml)) ah I guess it does but it's 2333!

This reminds me to check, I don't think ".lancedb/config.toml" is a thing we ever really talk about, right? so I removed that reference too.

.agents/skills/lancedb-column-metadata/SKILL.md
dantasse: 🤦 yes, of course! I was thinking of struct values which are still . but you're right, $ is correct here (and this is important to help it actually make queries).

brendanclement (approved):

PR #3562, opened 2026-06-19
feat(rust): add blob read and materialization APIs

geruh:
This PR is for the Read path against blob v2. #3528 handles declare + write, and this this adds materialization on local tables.

- blob_columns()
- fetch_blobs(column, row_ids) → bytes
- fetch_blob_files(column, row_ids) → lazy handles
- Pass _rowid from query().with_row_id(). Remote returns NotSupported. (for now)

### Use cases

search, grab row ids, materialize images:

```rust
let row_ids = /* _rowid from hits */;
let images = table.fetch_blobs("image", &row_ids).await?;
```

Large blobs: open handles, read only what you need:

```rust
let handles = table.fetch_blob_files("image", &row_ids).await?;
let bytes = handles[0].as_ref().unwrap().read().await?;
```

Filter then batch fetch: collect ids from a filter, one call.
Multiple blob columns: image and thumbnail independently.
Row ids from before compact: still resolve.

### Alignment note
Lance `read_blobs` drops null rows. We descriptor-take first, read non-null ids, re-expand to match input order. Null and zero-length blobs come back null/None. Bytes path sets `preserve_order(true)`. So I added:

```
TODO(lance): expose selection_index or an aligned execute so we can drop the pre-read.
```

### Tests
`cargo test -p lancedb --test blob_integration`
 -  30 tests covering nulls, reorder, dups, cross-fragment bytes + files, compact, delete, legacy v1 errors.

### DIFF
--- a/rust/lancedb/src/blob.rs
+++ b/rust/lancedb/src/blob.rs
@@ -12,10 +12,11 @@
 use std::sync::Arc;
 
 use arrow_array::builder::LargeBinaryBuilder;
-use arrow_array::{Array, LargeBinaryArray, StructArray, UInt8Array, UInt64Array};
-use arrow_schema::{Field, Schema};
+use arrow_array::{Array, LargeBinaryArray, RecordBatch, StructArray, UInt8Array, UInt64Array};
+use arrow_schema::{DataType, Field, Schema};
 use lance::dataset::{Dataset, WriteParams};
 use lance_arrow::FieldExt;
+use lance_core::datatypes::parse_field_path;
 use lance_encoding::version::LanceFileVersion;
 
 use crate::error::{Error, Result};
@@ -26,6 +27,9 @@ pub use lance::dataset::BlobFile;
 ///
 /// `Struct<data, uri>` with the `lance.blob.v2` marker. Same layout Lance
 /// expects on write.
+///
+/// A blob column may be top-level or nested inside a struct or list. Nested
+/// blobs are addressed by a dotted path (e.g. `info.blob`) in the read APIs.
 ///
 /// ```
 /// use arrow_schema::{DataType, Field, Schema};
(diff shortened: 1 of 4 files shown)


geruh:
@claude review

rust/lancedb/src/blob.rs
```diff
@@ -38,6 +44,16 @@ pub(crate) fn has_blob_columns(schema: &Schema) -> bool {
     schema.fields().iter().any(|field| field.is_blob_v2())
 }
 
+/// Blob v2 column names in `schema`, declaration order preserved.
+pub(crate) fn blob_column_names(schema: &Schema) -> Vec<String> {
```
westonpace, line 101: Do we support blobs in nested fields?  For example, a blob in a struct or list column?

rust/lancedb/tests/blob_integration.rs
```diff
@@ -88,10 +93,10 @@ async fn query_image_struct(table: &Table) -> StructArray {
     let batch = arrow_select::concat::concat_batches(&batches[0].schema(), &batches).unwrap();
     batch
         .column_by_name("image")
-        .expect("image column present")
+        .unwrap()
```
westonpace: Why change these expect into unwrap?  I guess it doesn't really matter but it seems to remove information for no real purpose?

rust/lancedb/tests/blob_integration.rs
```diff
@@ -301,17 +299,12 @@ async fn namespace_create_applies_blob_defaults() -> Result<()> {
     Ok(())
 }
 
-// Overwrite takes the input schema as-is (same as cast skip). Raw binary
-// overwrite drops the blob marker unless the input declares blob v2.
```
westonpace, line 305: Did the behavior change?  Can we document the new behavior?

rust/lancedb/src/blob.rs
```diff
@@ -53,6 +69,191 @@ pub(crate) fn ensure_blob_storage_version(schema: &Schema, params: &mut WritePar
...
+    column: &str,
+    row_ids: &[u64],
+) -> Result<LargeBinaryArray> {
+    ensure_blob_v2_column(dataset.schema(), column)?;
+    if row_ids.is_empty() {
+        return Ok(LargeBinaryBuilder::new().finish());
+    }
+
+    let null_mask = blob_null_mask(dataset, column, row_ids).await?;
+    let non_null_row_ids = non_null_row_ids(row_ids, &null_mask);
```
westonpace, line 247: Why do we need to do this `blob_null_mask` thing?  Is it because we can't distinguish between empty and null blob results otherwise?  You mention above `aligned read API would drop this pass.`  What is this change?  Is there an issue for it?

rust/lancedb/src/table.rs
```diff
@@ -927,6 +959,87 @@ impl Table {
         self.inner.count_rows(filter.map(Filter::Sql)).await
     }
 
+    /// Names of the blob v2 columns in this table, in declaration order.
+    ///
+    /// Empty when the table has no blob columns. Declare columns with
+    /// [`crate::blob::blob`] at create time.
+    ///
+    /// Returns [`Error::NotSupported`] on table types without blob support.
+    pub async fn blob_columns(&self) -> Result<Vec<String>> {
```
westonpace, line 957: Is this method needed?  Couldn't users just evaluate the schema / fields themselves?  We could add an `is_blob` method perhaps that looks at a field and returns if it is a blob column?

I mainly worry that this method will lock us in to "no blobs in nested fields" behavior and not be that useful if we later want nested blob columns.  Also, we don't have equivalent methods for any other data type.

rust/lancedb/src/table.rs
```diff
@@ -927,6 +959,87 @@ impl Table {
...
+        self.inner.blob_columns().await
+    }
+
+    /// Materialize blob bytes for the given row ids.
+    ///
+    /// `row_ids` are stable row ids from a query with `with_row_id`. Output
```
westonpace: We should not assert they are stable row ids here.  They might not be stable.

rust/lancedb/src/table.rs
```diff
@@ -927,6 +959,87 @@ impl Table {
...
+    /// matches `row_ids` in length and order; null blob rows are null entries.
```
westonpace: "null blob rows are null entries" seems a little redundant.  Wouldn't that be implicit?

rust/lancedb/src/table.rs
```diff
@@ -927,6 +959,87 @@ impl Table {
...
+    /// Zero-length blobs also read back as null.
```
westonpace: Why can't we return them as zero length blobs?

rust/lancedb/src/table.rs
```diff
@@ -927,6 +959,87 @@ impl Table {
...
+    ///     .try_collect::<Vec<_>>()
+    ///     .await?;
+    /// let row_ids = hits
+    ///     .iter()
+    ///     .flat_map(|batch| {
+    ///         let ids = batch.column_by_name("_rowid").unwrap();
+    ///         let ids = ids.as_any().downcast_ref::<arrow_array::UInt64Array>().unwrap();
+    ///         ids.values().to_vec()
+    ///     })
+    ///     .collect::<Vec<u64>>();
```
westonpace: I feel like this example triggers at least one copy of the row ids that could be avoided.  Also, in practice, in most cases, I would expect a call to `fetch_blobs` for each returned batch instead of collecting all the row ids and making one combined call.

rust/lancedb/src/table.rs
```diff
@@ -927,6 +959,87 @@ impl Table {
...
+        column: impl AsRef<str>,
+        row_ids: &[u64],
+    ) -> Result<LargeBinaryArray> {
+        self.inner.fetch_blobs(column.as_ref(), row_ids).await
+    }
+
+    /// Open lazy [`BlobFile`] handles for the given row ids.
+    ///
+    /// Same length and order as `row_ids`. Null and zero-length rows are `None`.
+    /// Payload bytes move on [`BlobFile::read`].
```
westonpace: This sentence is a bit confusing to me.  Maybe "Payload bytes are not read fromt he disk until a call to [`BlobFile::read`]."?

westonpace (commented):
Some minor thoughts, no major complaints

rust/lancedb/src/blob.rs
geruh: Yeah, we do now. My local checkout was showing some issues.

rust/lancedb/tests/blob_integration.rs
geruh: restored

rust/lancedb/src/table.rs
geruh: I've been thinking this one through but, after moving in the nested logic yeah it makes more sense to expose a helper to point if it's a blob.

rust/lancedb/src/table.rs
geruh: addressed

rust/lancedb/src/blob.rs
geruh: Yeah I'm still thinking this one through but ultimately, lance core's `read_blobs` skips null rows here: https://github.com/lance-format/lance/blob/1c3d2017f0f98df0c2bb15fbc3e54e83f6ba2c23/rust/lance/src/dataset/blob.rs#L2076

So we might need the alignment, unless I'm missing something

rust/lancedb/src/table.rs
geruh: Inline null and zero-length both use the (Inline, 0, 0) sentinel, so we read both back as null. That matches how we expect blob columns to behave since their are no bytes to materialize.

The open issue is mixed present and empty selections erroring in `read_blobs`,  not the single-row null semantics.

rust/lancedb/src/blob.rs
westonpace: This seems problematic but we can address in the future as long as both the public APIs support nulls (which it looks like they do)

westonpace (approved):

PR #3571, opened 2026-06-24
fix: ensure read freshness provider is built into namespace client

rpgreen:
By default the read freshness provider was not included in the namespace client, preventing the read freshness headers from being included in the request. This prevents checkout_latest() from working as expected when using the namespace client.

This fix ensures the provided is built into the client when the namespace impl and properties are provided.

### DIFF
--- a/python/python/lancedb/namespace.py
+++ b/python/python/lancedb/namespace.py
@@ -373,6 +373,19 @@ def _convert_pyarrow_schema_to_json(schema: pa.Schema) -> JsonArrowSchema:
     return JsonArrowSchema(fields=fields, metadata=meta)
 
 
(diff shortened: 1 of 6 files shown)


jackye1995 (approved):

brendanclement (approved):
Pending CI

rpgreen:
@claude review

brendanclement:
This is fixed for sync connection but not for Async client - claude nailed it. The sync client got the required fixes we just need to apply them to the async client

jackye1995 (approved):
I am doing some further refactoring to close the gap between python and rust namespace integration, but this one should work for now.

PR #3572, opened 2026-06-24
fix(python): route sync namespace ops through native connection

wjones127:
Namespace operations (and tables within a namespace) failed on a sync
connection while the async equivalents worked:

```python
db = lancedb.connect(DB_PATH)
db.list_namespaces()  # ValueError: ... No module named 'lance'
```

## Root cause

The sync `LanceDBConnection` namespace and namespaced-table operations
routed through the Python `lance_namespace` client. Its directory (`"dir"`)
implementation lives in `lance.namespace.DirectoryNamespace` — part of the
optional `pylance` package — so these operations failed with
`No module named 'lance'` on installs without the `pylance` extra. The async
API worked because it goes straight to the native Rust connection, which
already bundles a directory-namespace implementation (`lance-namespace-impls`).

`pylance` only ships the *implementations*; the lightweight `lance-namespace`
package (a core dep) ships just the interface and registry. Requiring
`pylance` here meant pulling in a second copy of the Lance Rust bindings to
do something LanceDB already implements natively.

## Fix

Route all sync namespace + namespaced-table operations through the native
connection (via `LOOP.run(self._conn.<op>(...))`), matching the async API:

- `list_namespaces`, `create_namespace`, `drop_namespace`,
  `describe_namespace`, `list_tables`
- `create_table`, `open_table`, `drop_table` with a `namespace_path`
  (`clone_table`/`drop_all_tables` already used the native path)

`_namespace_conn()` is removed. `_dataset_path` now derives a child-namespace
table's location from the native table's resolved URI, so
`to_lance()`/`to_pandas()` keep working without a namespace client.

The only remaining `pylance`-requiring surface is the **public**
`namespace_client()` method and the `connect_namespace(...)` API for
user-supplied Python namespace implementations — both explicit opt-ins,
unchanged here.

## Tests

The repo already has a "Test without pylance or pandas" CI job. This PR makes
it cover the regression: removed the now-unnecessary `importorskip("lance")`
from `test_branches_preserve_namespace`, and added
`test_local_namespace_lifecycle_without_pylance` exercising the full
namespace + table lifecycle. Verified locally with `pylance` actually
uninstalled (the native tests pass; the `connect_namespace`/`to_lance` tests
correctly skip).

Fixes OSS-1275

### DIFF
--- a/python/python/lancedb/db.py
+++ b/python/python/lancedb/db.py
@@ -914,23 +914,7 @@ class LanceDBConnection(DBConnection):
             raise ValueError("mode must be either 'create' or 'overwrite'")
         validate_table_name(name)
 
-        if namespace_path:
-            return self._namespace_conn().create_table(
-                name,
-                data=data,
-                schema=schema,
-                mode=mode,
-                exist_ok=exist_ok,
-                on_bad_vectors=on_bad_vectors,
-                fill_value=fill_value,
-                embedding_functions=embedding_functions,
-                namespace_path=namespace_path,
-                storage_options=storage_options,
-                data_storage_version=data_storage_version,
-                enable_v2_manifest_paths=enable_v2_manifest_paths,
-            )
-
-        tbl = LanceTable.create(
+        return LanceTable.create(
             self,
             name,
             data,
@@ -942,20 +926,8 @@ class LanceDBConnection(DBConnection):
             embedding_functions=embedding_functions,
             namespace_path=namespace_path,
             storage_options=storage_options,
-        )
-        return tbl
-
-    def _namespace_conn(self) -> DBConnection:
-        """Return a LanceNamespaceDBConnection backed by this connection's
-        directory namespace.  Used to delegate child-namespace operations."""
-        from lancedb.namespace import LanceNamespaceDBConnection
-
-        return LanceNamespaceDBConnection(
-            self.namespace_client(),
-            read_consistency_interval=self.read_consistency_interval,
-            storage_options=self.storage_options,
-            namespace_client_impl=None,
-            namespace_client_properties=None,
+            data_storage_version=data_storage_version,
+            enable_v2_manifest_paths=enable_v2_manifest_paths,
         )
 
     @override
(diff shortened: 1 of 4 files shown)


wjones127:
### Why this doesn't affect custom namespace implementations

The central risk here is custom (user-defined Python) namespaces, so to be explicit:

There are two distinct connection classes, and this PR only touches one of them:

- **`LanceDBConnection`** — what you get from `lancedb.connect(path)`. It's backed by the native Rust connection, which already bundles a directory-namespace implementation (`lance-namespace-impls`). It *never* wraps a user-supplied Python namespace. **All of this PR's changes are confined to this class** — they just route its namespace/table ops through the native Rust connection (via `LOOP.run(self._conn.<op>(...))`) instead of re-deriving a `pylance` `DirectoryNamespace` and going back out through the Python client. Functionally identical, minus the implicit `pylance` dependency.

- **`LanceNamespaceDBConnection`** — what you get from `connect_namespace(...)` / `connect(namespace_client_impl=...)`, i.e. the path that carries a custom Python `LanceNamespace`. **This class is untouched by this PR.** Its `list_namespaces`/`create_table`/etc. still call `self._namespace_client.<op>(request)` exactly as before, and the custom impl is still bridged into Rust via `_connect_namespace_client` → `call_py_method`.

So custom implementations keep running through the same code path, unchanged. Removing `_namespace_conn()` is safe specifically because it only ever wrapped the *native* connection's own directory namespace — never a custom client.

Coverage backs this up: `test_namespace_integration.py` defines `CustomNamespace(LanceNamespace)` and parametrizes its tests over `use_custom=[False, True]` (DirectoryNS / CustomNS), asserting via ops-metrics that Rust actually invokes the Python methods. Those still pass — the local (non-S3) one is `test_namespace_create_schema_only_with_provider`, and the `s3_test`-marked ones run in the Linux integration job.

rpgreen (commented):
Does this remove the need for the workaround in https://github.com/lancedb/lancedb/pull/3571 ?

wjones127:
> #3571

@rpgreen I'm not sure. It's not obvious to me they are related.

PR #3574, opened 2026-06-24
fix(remote): support namespace clients with dynamic headers

jackye1995:
## Summary

Bridge LanceDB dynamic header providers into Lance Namespace dynamic context providers for live remote namespace clients.

Dynamic headers are forwarded as per-request `headers.*` context entries, so `conn.namespace_client()` can authenticate with refreshing providers. `namespace_client_config()` still rejects dynamic providers because a static property map cannot serialize a live provider.

### DIFF
--- a/rust/lancedb/src/remote/db.rs
+++ b/rust/lancedb/src/remote/db.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use http::StatusCode;
 use lance_io::object_store::StorageOptions;
+use lance_namespace_impls::{DynamicContextProvider, OperationInfo};
 use moka::future::Cache;
 use reqwest::header::CONTENT_TYPE;
 
@@ -26,7 +27,9 @@ use crate::remote::util::stream_as_body;
 use crate::table::BaseTable;
 
 use super::ARROW_STREAM_CONTENT_TYPE;
-use super::client::{ClientConfig, HttpSend, RequestResultExt, RestfulLanceDbClient, Sender};
+use super::client::{
+    ClientConfig, HeaderProvider, HttpSend, RequestResultExt, RestfulLanceDbClient, Sender,
+};
 use super::table::RemoteTable;
 use super::util::parse_server_version;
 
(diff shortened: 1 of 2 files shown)


rust/lancedb/src/remote/db.rs
```diff
@@ -194,10 +197,66 @@ pub struct RemoteDatabase<S: HttpSend = Sender> {
...
 
+#[derive(Clone)]
+struct NamespaceHeaderProviderContext {
+    header_provider: Arc<dyn HeaderProvider>,
+}
+
+impl std::fmt::Debug for NamespaceHeaderProviderContext {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("NamespaceHeaderProviderContext")
+            .field("header_provider", &"Some(...)")
```
wkalt, line 213: what does it mean to have "Some(...)" here?

rust/lancedb/src/remote/db.rs
```diff
@@ -194,10 +197,66 @@ pub struct RemoteDatabase<S: HttpSend = Sender> {
...
+                return HashMap::new();
+            }
+        };
+
+        let headers = handle.join();
+
+        match headers {
+            Ok(Ok(headers)) => headers
+                .into_iter()
+                .map(|(key, value)| (format!("headers.{key}"), value))
```
wkalt, line 246: is there any risk of the header provider providing headers that we do not want to propagate? for example trace IDs injected by the user's service mesh or things unrelated to our application?

wkalt (approved):
small comments

rust/lancedb/src/remote/db.rs
jackye1995: LanceDB already treats `ClientConfig.extra_headers` and `HeaderProvider` as arbitrary request headers, not auth-only headers. `apply_dynamic_headers()` forwards every valid provider header on normal remote requests today, and Python documents this for auth, request tracking, custom metadata, etc. This change keeps `namespace_client()` consistent with that existing behavior for the same remote host. If callers do not want a header sent to LanceDB requests, they should not include it in the configured provider.

rust/lancedb/src/remote/db.rs
jackye1995: this is just to redact the detailed content

PR #3576, opened 2026-06-24
feat(rust): re-export arrow and datafusion crates from lancedb

### LINKED ISSUE #3575: Feature: Re-export `arrow_array` and `arrow_buffer` from `lancedb::arrow`
### SDK

Rust

### Description

lancedb currently re-exports `arrow_schema` but **not** `arrow_array`. As a result, downstream
crates that build `RecordBatch` / arrays / builders to feed into `Table::add(...)` must declare
`arrow-array` as a **direct** dependency. That direct dependency's major version must exactly
match the arrow line lancedb was compiled against — but nothing in the API surface makes that
constraint visible, and nothing enforces it. The result is a silent version split in the
dependency graph that surfaces only as confusing trait/type errors.

Concrete failure we just hit on lancedb `0.30.0`: a routine update bumped our direct
`arrow-array` from `58` → `59`, while lancedb `0.30.0` still bundles arrow `58.x`. The build
fails with errors like:

```text
error[E0277]: the trait bound `RecordBatch: Scannable` is not satisfied
   --> src/projector.rs:263:19
    |
263 |         table.add(projection_batch(&tenant_rows)?).execute().await?;
    |               --- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Scannable` is not
    |               |                                   implemented for `RecordBatch`
    |               required by a bound introduced by this call
    = note: there are multiple different versions of crate `arrow_array` in the dependency graph
```

…plus cascading `Schema` / `Field` mismatches (`arrow_schema::schema::Schema` vs
`lancedb::arrow::arrow_schema::Schema`) on every `RecordBatch::try_new(...)` and
`GenericListBuilder::with_field(...)` call, all with the same root note:
*"there are multiple different versions of crate `arrow_array`/`arrow_schema` in the dependency graph."*

The coupling is already implicit — lancedb's `Scannable`, `IntoArrow`, and `RecordBatchReader`
impls are bound to a specific `arrow_array::RecordBatch` identity — so consumers are *forced*
onto lancedb's arrow line regardless. They just discover it via a compile error instead of an
import path.

**Describe the solution you'd like**

In `src/arrow.rs`, alongside the existing re-export:

```rust
pub use arrow_schema;
```

add:

```rust
pub use arrow_array;
pub use arrow_buffer;   // moves in lockstep with the above two
```

This makes `lancedb::arrow::arrow_array::{RecordBatch, Int64Array, …}` and
`lancedb::arrow::arrow_buffer::*` available, so consumers who only touch arrow in order to talk
to lancedb can drop their direct `arrow-array` dependency entirely. The version split that caused
our failure becomes structurally impossible, because there's only one arrow line in the graph.

**Rationale**

1. **Symmetry.** `arrow_schema` is already re-exported; `arrow_array` and `arrow_buffer` are
   released together with it as one workspace and must move in lockstep. The current asymmetry
   (`arrow_schema` yes, `arrow_array` no) is exactly what forces the two-source import pattern
   where drift creeps in.
2. **Low cost.** lancedb already pins a specific arrow line internally; re-exporting it adds no
   new maintenance burden beyond what the trait impls already commit to.
3. **Makes implicit coupling explicit.** The constraint "your arrow must match lancedb's arrow"
   exists today — a re-export just surfaces it through a discoverable import path instead of an
   `E0277`.
4. **Precedent.** DataFusion re-exports arrow (`datafusion::arrow::*`) for exactly this reason,
   and it works well in practice for preventing cross-version drift downstream.

**Additional context**

- One semver note: once `lancedb::arrow::arrow_array` is public, bumping lancedb's internal arrow
  line becomes a visible breaking change for downstream. This is *already* true today via the
  trait impls; the re-export surfaces it rather than hiding it behind a confusing `E0277`, so it's
  strictly an improvement.

wjones127:
lancedb's public API forces downstream crates to construct foreign types — `RecordBatch`/arrays/builders for `Table::add(...)` (arrow), and `datafusion_expr::Expr` for `only_if_expr`/`expr_projection`/merge filters. The required version must exactly match lancedb's internal arrow/datafusion line, but nothing on the API surface makes that visible. Drift surfaces only as confusing trait/type errors:

```text
error[E0277]: the trait bound `RecordBatch: Scannable` is not satisfied
  = note: there are multiple different versions of crate `arrow_array` in the dependency graph
```

This re-exports the crates lancedb already pins, so consumers can rely on a single, guaranteed-matching line via a discoverable import path instead of declaring their own (potentially mismatched) direct dependency.

- `lancedb::arrow::{arrow, arrow_array, arrow_buffer, arrow_cast, arrow_data, arrow_ipc, arrow_ord, arrow_schema, arrow_select}` — previously only `arrow_schema` was re-exported. `arrow-buffer` is promoted from a transitive to a direct dependency.
- `lancedb::datafusion` — `Expr` is a first-class part of the query and merge APIs (`only_if_expr`, `expr_projection`, `QueryFilter::Datafusion`, `when_matched_update_all_expr`), and `ExecutionPlan` is returned from `create_plan`.

This follows DataFusion's own precedent of re-exporting `arrow`. The coupling already exists via the trait/impl bounds — this surfaces it rather than hiding it behind an `E0277`.

Closes #3575

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5313,6 +5313,7 @@ dependencies = [
  "anyhow",
  "arrow",
  "arrow-array",
+ "arrow-buffer",
  "arrow-cast",
  "arrow-data",
  "arrow-ipc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ ahash = "0.8"
 # Note that this one does not include pyarrow
 arrow = { version = "58.0.0", optional = false }
 arrow-array = "58.0.0"
+arrow-buffer = "58.0.0"
 arrow-data = "58.0.0"
 arrow-ipc = "58.0.0"
 arrow-ord = "58.0.0"
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -14,6 +14,7 @@ rust-version.workspace = true
 ahash = { workspace = true }
 arrow = { workspace = true }
 arrow-array = { workspace = true }
+arrow-buffer = { workspace = true }
 arrow-data = { workspace = true }
 arrow-schema = { workspace = true }
 arrow-select = { workspace = true }
--- a/rust/lancedb/src/arrow.rs
+++ b/rust/lancedb/src/arrow.rs
@@ -3,7 +3,19 @@
 
 use std::{pin::Pin, sync::Arc};
 
+// Re-export the arrow crates we depend on so downstream consumers can build
+// `RecordBatch`/arrays/builders against the exact same arrow line lancedb was
+// compiled against, instead of declaring their own (potentially mismatched)
+// direct arrow dependencies. See https://github.com/lancedb/lancedb/issues/3575.
+pub use arrow;
+pub use arrow_array;
+pub use arrow_buffer;
+pub use arrow_cast;
+pub use arrow_data;
+pub use arrow_ipc;
+pub use arrow_ord;
 pub use arrow_schema;
+pub use arrow_select;
 use datafusion_common::DataFusionError;
 use datafusion_physical_plan::stream::RecordBatchStreamAdapter;
 use futures::{Stream, StreamExt, TryStreamExt};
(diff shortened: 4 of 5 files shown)


westonpace (approved):
Eh, fine, but I'd still encourage users to use the C data interface so they don't have to align versions.  I wonder if there are things we can do to make that easier.

PR #3578, opened 2026-06-24
feat(python): blob v2 fetch API

geruh:
Python bindings for blob v2 read on **local** tables. Rust read APIs landed in #3562.

This PR wires `fetch_blob_files`, `fetch_blobs`, v2 query/`to_pandas(blob_mode="bytes")`, and hidden `_rowid` metadata so `fetch_*` works from query hits without exposing `_rowid` in the column list.

**Cloud:** `RemoteTable.fetch_blobs` / `fetch_blob_files` raise `NotImplementedError` until Phalanx ships the server route (separate track; not blocking local merge).

### Primary path: lazy file handles

```python
table = db.create_table("videos", schema=pa.schema([
    pa.field("id", pa.int64()),
    lancedb.blob("video"),
]))
table.add([{"id": 1, "video": open("clip.mp4", "rb").read()}])

hits = table.search().select(["id", "video"]).to_arrow()
handle = table.fetch_blob_files("video", hits)[0]

# seek + partial read — PyAV / decoders can use the handle
handle.seek(frame_offset)
chunk = handle.read_range(0, 65536)
```

`BlobFile` exposes `seek`, `read`, `read_range`, `read_up_to`, and works with `BufferedReader`.

### When you want full bytes

```python
blobs = table.fetch_blobs("video", hits)  # eager materialize, null-aligned
df = table.to_pandas(blob_mode="bytes")   # descriptors → bytes in pandas
```

### `_rowid` (join key, not user `id`)

Fetch needs Lance row ids. For v2 blob queries we auto-inject `_rowid`, stash it in Arrow schema metadata on `to_arrow()`, and drop the visible column unless you pass `.with_row_id(True)`.

v1 legacy blobs (`lance-encoding:blob`) unchanged; fetch on v1 raises the migration error.

## Test plan

- [x] `./scripts/test-blob.sh python` (105 passed in worktree)
- [x] `fetch_blob_files` lazy read, seek, partial read, null alignment, cross-fragment dups
- [x] hybrid query → `fetch_blobs` / `fetch_blob_files`
- [ ] Will re-review after seek/`BlobFile` commit (`d77ab1a6`)


geruh:
@claude review

python/python/tests/test_blob.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def testFetchBlobsRoundTrip():
+    table = _blobTable(
+        "round_trip",
+        [{"id": 1, "image": b"alpha"}, {"id": 2, "image": b"beta"}],
+    )
+    by_id = _rowIdsById(table)
+    blobs = table.fetch_blobs("image", [by_id[1], by_id[2]])
+    assert [blobs[0].as_py(), blobs[1].as_py()] == [b"alpha", b"beta"]
```
wjones127, line 176: issue(blocking): unless I'm missing something, this doesn't feel like the right API. One of the main ideal features of blobs, IIUC, is that you can open them as file handles, and seek within a blob. The key example is this one from the Lance docs, where we pass the handle into a video decoder and can seek to a certain frame: https://lance.org/guide/blob/#example-decode-video-frames-lazily

Is that not a target for this API? It seems like `fetch_blobs` here eagerly reads the full bytes, which is not what we'd want for the video decode example.

wjones127 (changes requested):
First off, I really like the PR description, particularly the example. Thank you!

Before we go too far, I have a big question on the design of the API. It seems different than what we had created in Lance, and in some particularly important ways!

sarwarbhuiyan:
This was the original issues #3231

python/python/tests/test_blob.py
geruh: Yeah, you're right. `fetch_blobs` is not really the main thing here.

Rust already has the right path with `fetch_blob_files` and Lance `take_blobs`. Python right now only exposes `read()`, which pretty much forces you to materialize the entirety of the file. 

So the split rn is:

- `fetch_blob_files` is for treating blobs more like files and reading bytes when needed.

- `fetch_blobs` is for when you explicitly want the whole blob loaded.

This first Python version did not expose the seek/file behavior on `BlobFile`, which is why the Lance PyAV example does not work yet with LanceDB handles.

I can update to add either `seek` or `read_range` in this PR, or do it as a small follow up stacked on this one.

I’m also happy to make the docs and examples lead with `fetch_blob_files` and partial reads instead of making `fetch_blobs(hits)` look like the default path.

python/python/lancedb/_blob.py
```diff
@@ -0,0 +1,306 @@
...
+            return self._inner.read_bytes()
+        return super().read(size)
+
+    def read_range(self, offset: int, length: int) -> bytes:
+        return self._inner.read_range(offset, length)
+
+    def readinto(self, b: WriteableBuffer) -> int:
+        view = memoryview(b).cast("B")
+        chunk = self._inner.read_up_to(len(view))
+        view[: len(chunk)] = chunk
```
wjones127, line 94: issue(non-blocking): this isn't exactly in the spirit of the method. It copies the buffer rather than passing it down. Can address in a follow up.

python/python/lancedb/_blob.py
```diff
@@ -0,0 +1,306 @@
...
+        return hits["_rowid"].to_pylist()
+
+    metadata = hits.schema.metadata or {}
+    raw = metadata.get(_AUTO_ROWID_METADATA_KEY)
+    if raw is None:
+        raise ValueError(
+            "query result has no '_rowid' column or hidden row-id metadata; "
+            "pass fresh blob query results, call .with_row_id(True), or pass "
+            "a list of row ids"
+        )
```
wjones127: issue(blocking): this isn't a very robust method of keeping the row id. If the user sorts or filters the record batch before calling `fetch_blobs`, then this will return the wrong order/selection of ids. This is why I lean towards finding a way to stash the row id in the descriptors instead.

python/python/lancedb/_blob.py
```diff
@@ -0,0 +1,306 @@
...
+
+    row_ids = _deserialize_auto_row_ids(raw)
+    if len(row_ids) != hits.num_rows:
+        raise ValueError("query result hidden _rowid metadata has the wrong length")
```
wjones127: suggestion: this error is cryptic and could be improved.

```suggestion
        raise ValueError("query results were modified (filtered/sliced) after the query — pass unmodified hits, call .with_row_id(True), or pass a list of row ids.'")
```

python/python/lancedb/remote/table.py
```diff
@@ -990,6 +990,22 @@ def migrate_v2_manifest_paths(self):
...
+
+    def fetch_blobs(self, column: str, row_ids) -> pa.LargeBinaryArray:
+        raise NotImplementedError(
+            "fetch_blobs() is not yet supported on the LanceDB Cloud"
+        )
+
+    def fetch_blob_files(self, column: str, row_ids):
+        raise NotImplementedError(
+            "fetch_blob_files() is not supported on LanceDB Cloud; "
+            "use fetch_blobs() to materialize bytes"
```
wjones127: issue(blocking): this error message is misleading. fetch_blobs() is also not supported, so we shouldn't recommend it as an alternative.
```suggestion
            "fetch_blob_files() is not supported on LanceDB Cloud"
```

wjones127 (changes requested):
This is looking better, but I think there's a flaw in stashing the row ids in the schema. Is there a way we can project them into the descriptor struct at query time? I think that would be a more robust solution so it survives even after filtering or sorting.

python/python/lancedb/_blob.py
geruh: Yeah that's fair since right now we are copying into the memview so buffer reader work. We can pass that down later.

python/python/lancedb/_blob.py
geruh: yeah, metadata was definitely in wrong place and this approach was pretty hacky. So I just will add it to the descriptor as an additional field.

python/python/lancedb/_blob.py
geruh: updated thanks

python/python/lancedb/remote/table.py
geruh: updated

wjones127 (approved):
This looks good. Thanks for addressing those comments I had earlier!

PR #3582, opened 2026-06-25
feat(embeddings): add TwelveLabs Marengo multimodal embedding function

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

This PR adds a new opt-in embedding function for [TwelveLabs Marengo](https://docs.twelvelabs.io/), registered as `"twelvelabs"` in the embedding-function registry alongside the existing providers (VoyageAI multimodal, SigLIP, OpenCLIP, ImageBind, etc.).

### What it adds
- `TwelveLabsEmbeddingFunction` (`lancedb.embeddings.twelvelabs`) — Marengo is a multimodal model that maps text, images, audio and video into a single shared embedding space, so you can store media embeddings as a source column and query them with plain text (or vice-versa).
- Supports `marengo3.0` (512-dim, default) and `Marengo-retrieval-2.7` (1024-dim).
- Accepts text (`str`), image URLs (`str`), local image paths (`pathlib.Path`), raw `bytes`, and `PIL.Image.Image` inputs.
- API key is read from `TWELVELABS_API_KEY` and reported via `sensitive_keys()`.
- Docs page under the multimodal embedding functions section, and the `twelvelabs>=1.2.8` SDK added to the optional `embeddings` extra.

### Why it helps this project
LanceDB is a natural store for multimodal/video search, and Marengo gives users a single shared embedding space across text/image/audio/video — letting them do cross-modal retrieval (text query → image/video results) directly in a LanceDB table, with no separate inference plumbing.

### Opt-in / non-breaking
Purely additive: a new registry entry and an optional dependency. No existing defaults or behavior change; the SDK is only imported when the function is actually used (`attempt_import_or_raise`).

### How it was tested
- 9 no-network unit tests (mocked SDK) covering registration, dimensions, default model, unsupported-model error, sensitive-key reporting, missing-API-key error, and that text vs. URL inputs route to the correct SDK fields.
- 1 integration test gated on `TWELVELABS_API_KEY` (skipped without it, marked `slow`) that creates a table, embeds rows via the real Marengo API, and runs a cross-modal text search.
- Verified against the live API: a `marengo3.0` text call returns a 512-dim vector. All 10 tests pass locally; `ruff check` and `ruff format --check` are clean on the changed files.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/python/python/lancedb/embeddings/twelvelabs.py
+++ b/python/python/lancedb/embeddings/twelvelabs.py
@@ -1,0 +1,155 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright The LanceDB Authors
+import io
+import os
+from pathlib import Path
+from typing import ClassVar, List, TYPE_CHECKING, Union
+from urllib.parse import urlparse
+
+import numpy as np
+import pyarrow as pa
+
+from ..util import attempt_import_or_raise
+from .base import EmbeddingFunction
+from .registry import register
+from .utils import IMAGES, TEXT, api_key_not_found_help
+
+if TYPE_CHECKING:
+    import PIL
+
+# Marengo embedding models and their output dimensions.
+TWELVELABS_MODEL_DIMS = {
+    "marengo3.0": 512,
+    "Marengo-retrieval-2.7": 1024,
+}
+
+
+def _is_url(text: str) -> bool:
+    try:
+        parsed = urlparse(text)
+        return bool(parsed.scheme) and bool(parsed.netloc)
+    except Exception:
+        return False
+
+
+@register("twelvelabs")
+class TwelveLabsEmbeddingFunction(EmbeddingFunction):
+    """
+    An embedding function that uses the TwelveLabs Marengo multimodal model.
+
+    Marengo produces embeddings that live in a single space shared across text,
+    image, audio, and video, so a text query can be matched directly against
+    image or audio source columns. See https://docs.twelvelabs.io/ for details.
+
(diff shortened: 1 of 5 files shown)


python/python/lancedb/embeddings/twelvelabs.py
```diff
@@ -0,0 +1,155 @@
...
+            result = resp.image_embedding
+        else:
+            # Local image: Path, raw bytes, or a PIL image.
+            if isinstance(item, Path):
+                image_file = item.read_bytes()
+            elif isinstance(item, bytes):
+                image_file = item
+            elif isinstance(item, PIL_Image.Image):
+                buf = io.BytesIO()
+                item.save(buf, format="JPEG")
```
wjones127, line 134: nitpick: always converting to JPEG will silently drop the alpha channel. Maybe worth a warning or at least documenting that on the class doc string.

wjones127 (changes requested):
It looks like your tests are failing:

```
FAILED python/python/tests/test_twelvelabs_embeddings.py::TestTwelveLabsRegistration::test_missing_api_key_raises - ImportError: Please install twelvelabs
FAILED python/python/tests/test_twelvelabs_embeddings.py::TestTwelveLabsRegistration::test_text_embedding_uses_text_field - ImportError: Please install pillow
FAILED python/python/tests/test_twelvelabs_embeddings.py::TestTwelveLabsRegistration::test_url_input_uses_image_url_field - ImportError: Please install pillow
```

For the embedding tests, it's fine to skip them in CI as long as you do manually test them in an environment that has access. You can use `pytest.importorskip`

PR #3584, opened 2026-06-25
feat: add polars dataframe integration

westonpace:
This PR is part cleanup, part feature, part example.

It removes `IntoArrow` and `IntoArrowStream`.  There was only one redundant call site between the two.  Once we moved everything to `Scannable` these traits no longer serve any purpose.

It adds a `Scannable` impl for a polars DataFrame.  We used to have this at one point for `IntoArrow` so this is more like a regression fix than anything.

It adds an example (and unit test) which ensures we can ingest from a Polars DataFrame and export to one.  LazyFrame support would be a follow-up (though a pretty straightforward one) but we've never had proper LazyFrame support before.


wjones127 (approved):
Looks good, though some formatting fixes you'll need to make before merge.

PR #3585, opened 2026-06-26
fix!: combine repeated where filters with AND instead of replacing

### LINKED ISSUE #2649: bug(python): using .where after a FTS search results in spurious results

wjones127:
BREAKING CHANGE: When passing multiple where clauses to a query, they now stack instead of replacing the previous filter.

Previously, calling `where`/`only_if` more than once on a query silently
replaced the previous filter, so only the last filter was applied. This was
surprising and could return rows that an earlier filter should have excluded.

This implements the alternative suggested in
https://github.com/lancedb/lancedb/pull/3514#issuecomment-4664901580: instead of
rejecting a second filter, repeated filters are combined with a logical AND
(`(previous) AND (new)`).

The combination happens in the Rust core (`QueryBase::only_if` and
`only_if_expr`), so it applies to all SDKs at once (Rust, Python async, and
TypeScript). The Python sync query builder keeps its own filter state, so it
combines filters in the binding layer as well.

SQL string and expression filters are combined within their own representation.
When the two representations are mixed, the expression is lowered to SQL (via
`expr_to_sql_string`) and the filters are combined as SQL strings, so chaining
`where` works regardless of which form each filter takes.

Fixes #2649

## Tests
- Rust: `cargo test --features remote -p lancedb --lib query`
- Python: `uv run --extra tests pytest python/tests/test_query.py`
- TypeScript: `pnpm test __test__/query.test.ts`

### DIFF
--- a/docs/src/js/classes/Query.md
+++ b/docs/src/js/classes/Query.md
@@ -518,6 +518,9 @@ x > 5 OR y = 'test'
 
 Filtering performance can often be improved by creating a scalar index
 on the filter column(s).
+
+Calling this multiple times combines the filters with a logical AND rather
+than replacing the previous filter.
 ```
 
 #### Inherited from
--- a/docs/src/js/classes/VectorQuery.md
+++ b/docs/src/js/classes/VectorQuery.md
@@ -767,6 +767,9 @@ x > 5 OR y = 'test'
 
 Filtering performance can often be improved by creating a scalar index
 on the filter column(s).
+
+Calling this multiple times combines the filters with a logical AND rather
+than replacing the previous filter.
 ```
 
 #### Inherited from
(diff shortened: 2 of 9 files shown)


westonpace (approved):
Seems a reasonable addition though it is technically a backwards incompatible API change.  Though unlikely, it is possible someone was relying on the old behavior.

PR #3597, opened 2026-06-30
feat(remote): monotonic reads via x-lancedb-min-read-version watermark

jackye1995:
## Summary

Adds per-session monotonic reads for remote (LanceDB Cloud/Enterprise) tables, preventing successive reads on a handle from moving *backward* in dataset version when a load balancer routes them to query nodes with differently-cached views.

Each `RemoteTable` handle tracks the highest dataset version it has observed in a read response — surfaced by the server via a new `x-lancedb-version` response header — and sends it back as `x-lancedb-min-read-version` on subsequent reads (`count_rows`, `query`). A query node whose cache is behind that version refreshes before serving; a node already at/beyond it serves from cache at no extra cost.

The watermark is sourced only from reads (always committed dataset versions), so unlike the retired `x-lancedb-min-version` it is unaffected by WAL writes returning WAL entry ids. It is reset on `checkout_latest()`. Both headers are optional and ignored by older peers.

Server-side enforcement lives in LanceDB Enterprise. Targets the `codex/update-lance-9-0-0-beta-8` integration branch to match the Enterprise submodule pin.

### DIFF
--- a/rust/lancedb/src/remote/table.rs
+++ b/rust/lancedb/src/remote/table.rs
@@ -70,18 +70,29 @@ use tokio::sync::RwLock;
 const REQUEST_TIMEOUT_HEADER: HeaderName = HeaderName::from_static("x-request-timeout-ms");
 const MIN_VERSION_HEADER: HeaderName = HeaderName::from_static("x-lancedb-min-version");
 const MIN_TIMESTAMP_HEADER: HeaderName = HeaderName::from_static("x-lancedb-min-timestamp");
+const MIN_READ_VERSION_HEADER: HeaderName = HeaderName::from_static("x-lancedb-min-read-version");
+const VERSION_HEADER: HeaderName = HeaderName::from_static("x-lancedb-version");
 const METRIC_TYPE_KEY: &str = "metric_type";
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/remote/table.rs
```diff
@@ -70,18 +70,29 @@ use tokio::sync::RwLock;
...
     /// version update this, and reads send it as `x-lancedb-min-version`.
     min_version: Option<u64>,
+    /// Highest dataset version observed in a *read* response on this handle.
+    /// Reads send it as `x-lancedb-min-read-version` so a load-balanced query
+    /// node whose cache is behind this version must refresh before serving,
+    /// giving monotonic reads across nodes regardless of which one the load
+    /// balancer routes to. Sourced only from reads (always committed dataset
+    /// versions), never from writes (which may return WAL entry ids), so it is
+    /// unaffected by the WAL/version mismatch that retired `min_version`.
+    min_read_version: Option<u64>,
```
wjones127, line 95: note: shame we can't re-use `min_version`.

wjones127 (approved):

brendanclement (approved):

PR #3598, opened 2026-06-30
fix(python): route sync namespace connections through rust

jackye1995:
Summary:
- Route built-in sync namespace connections through the Rust namespace connector.
- Keep custom namespace clients on the existing Python fallback.
- Preserve namespace-backed to_lance compatibility with lazy Python client construction and add regressions.


rpgreen (approved):
LGTM as long as read consistency does not regress

PR #3606, opened 2026-07-01
fix(python): route local sync namespace operations through rust

jackye1995:
Routes local sync child-namespace operations through the Rust-backed connection instead of the Python namespace-client fallback.

Also keeps lazy namespace-client construction for table-to-Lance conversion and preserves public namespace error mappings.

Validated locally with ruff format/check and targeted namespace pytest.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2145,7 +2145,7 @@ class LanceTable(Table):
         namespace_client = self._namespace_client
         if namespace_client is None:
             conn_uri = getattr(self._conn, "uri", "")
-            if get_uri_scheme(conn_uri) == "namespace":
+            if get_uri_scheme(conn_uri) == "namespace" or self._namespace_path:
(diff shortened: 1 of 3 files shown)


justinrmiller (approved):
lgtm

python/python/lancedb/table.py
```diff
@@ -2145,7 +2145,7 @@ def to_lance(self, **kwargs) -> lance.LanceDataset:
         namespace_client = self._namespace_client
         if namespace_client is None:
             conn_uri = getattr(self._conn, "uri", "")
-            if get_uri_scheme(conn_uri) == "namespace":
+            if get_uri_scheme(conn_uri) == "namespace" or self._namespace_path:
```
wjones127, line 2148: nitpick: it would be worth a comment why there's two possibilities here, and which they come from.

wjones127 (approved):

PR #3607, opened 2026-07-01
fix(rust): honor CreateTable namespace pushdown

jackye1995:
Honors NamespaceClientPushdownOperation::CreateTable for namespace-backed table creation when the request can be represented by namespace create_table.

Falls back to declare_table plus local write for overwrite/exist-ok, custom write options, blobs, and streaming or unknown-size inputs.

Validated locally with cargo fmt, targeted namespace tests, and clippy.

### DIFF
--- a/rust/lancedb/src/database/namespace.rs
+++ b/rust/lancedb/src/database/namespace.rs
@@ -6,14 +6,18 @@
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
 
+use arrow_ipc::writer::StreamWriter;
 use async_trait::async_trait;
+use bytes::Bytes;
+use futures::TryStreamExt;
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/database/namespace.rs
```diff
@@ -270,6 +274,101 @@ impl LanceNamespaceDatabase {
...
+        let schema = stream.schema();
+        let mut buffer = Vec::new();
+        {
+            let mut writer = StreamWriter::try_new(&mut buffer, schema.as_ref())?;
+            while let Some(batch) = stream.try_next().await? {
+                writer.write(&batch)?;
+            }
+            writer.finish()?;
+        }
+        Ok(Bytes::from(buffer))
```
justinrmiller, line 328: If there is a large table passed in this could get very large in memory right? Is there a cap on the size of data in the request?

PR #3612, opened 2026-07-02
fix(python): handle scalar-indexed merge inserts

### LINKED ISSUE #3515: bug(python): Create scalar index on merge column make merge_insert update all fail completely without raising error
### LanceDB version

0.33.0

### What happened?

Create scalar index on merge column make merge_insert update all fail completely without raising error.
Dropping the index will fix it.

This is related to https://github.com/lancedb/lancedb/issues/3177 but I don't need to run optimize to see the error so the root cause is probably the create_scalar_index function

lancedb version: 0.33.0
lance(pylance) version: 0.7.0

```python
import tempfile

import lance
import lancedb
import pyarrow as pa
import pyarrow.compute as pc

print(lance.__version__)
print(lancedb.__version__)

with tempfile.TemporaryDirectory() as tempdir:
    db = lancedb.connect(
        tempdir,
    )

    schema = pa.schema(
        [
            pa.field("vector", pa.list_(pa.float32(), 4), nullable=True),
            pa.field("path", pa.string(), nullable=False),
            pa.field("status", pa.utf8()),
            pa.field("file_size", pa.int64()),
        ]
    )

    # 1. Create table with 1000 rows
    tbl = db.create_table("test", schema=schema)
    records = [{"vector": None, "path": f"/img/{i}.jpg", "status": "pending", "file_size": 1000 + i} for i in range(1000)]
    print(tbl.add(pa.Table.from_pylist(records, schema=schema)))
    print(tbl.count_rows())

    # 2. Create scalar index on the merge key
    tbl.create_scalar_index("path", index_type="BTREE")

    # 3. merge_insert 128 rows
    updates = [{"path": f"/img/{i}.jpg", "vector": [0.1] * 4, "status": "indexed"} for i in range(128)]
    print(tbl.merge_insert("path").when_matched_update_all().execute(pa.Table.from_pylist(updates)))

    arrow = tbl.to_arrow().select(["status"])
    n_indexed = pc.sum(pc.equal(arrow.column("status"), "indexed")).as_py()
    print(f"Indexed: {n_indexed}")  # Expected: 128, Actual: 0

    # 4. Drop index will fix the problem
    tbl.drop_index("path_idx")
    print(tbl.merge_insert("path").when_matched_update_all().execute(pa.Table.from_pylist(updates)))

    arrow = tbl.to_arrow().select(["status"])
    n_indexed = pc.sum(pc.equal(arrow.column("status"), "indexed")).as_py()
    print(f"Indexed: {n_indexed}")  # Expected: 128, Actual: 128

```

Result

```sh
7.0.0
0.33.0
AddResult(version=2)
1000
MergeResult(version=4, num_updated_rows=0, num_inserted_rows=0, num_deleted_rows=0, num_attempts=1)
Indexed: 0
MergeResult(version=6, num_updated_rows=128, num_inserted_rows=0, num_deleted_rows=0, num_attempts=1)
Indexed: 128
```

goutamadwant:
## Summary

Fix `merge_insert(...).when_matched_update_all()` for tables that have a scalar index on the merge key when the key is not one of the leading target fields.

The indexed merge path can silently report zero updated rows in that shape. This keeps the indexed path for schemas where the merge keys are leading fields, and falls back to a full scan otherwise.

Adds a Python regression covering a BTREE scalar index on `path` with partial update data, matching the reported failure mode.

## Test Plan

- `python -m pytest python/tests/test_table.py::test_merge_insert_update_all_with_scalar_index`
- `python -m pytest python/tests/test_table.py::test_merge_insert python/tests/test_table.py::test_merge_insert_subschema python/tests/test_table.py::test_merge_insert_update_all_with_scalar_index`
- `ruff check python/tests/test_table.py`
- `ruff format --check python/tests/test_table.py`
- `cargo fmt --all -- --check`
- `cargo test -p lancedb table::merge::tests::test_merge_insert_use_index`

## Breaking Changes

None.

Fixes #3515

### DIFF
--- a/rust/lancedb/src/table/merge.rs
+++ b/rust/lancedb/src/table/merge.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use arrow_array::RecordBatchReader;
+use arrow_schema::Schema as ArrowSchema;
 use futures::future::Either;
 use futures::{FutureExt, TryFutureExt};
 use lance::dataset::{
@@ -250,6 +251,10 @@ pub(crate) async fn execute_merge_insert(
     }
 
     let dataset = table.dataset.get().await?;
+    let target_schema = ArrowSchema::from(dataset.schema());
+    // The indexed merge path classifies matches from the leading target fields.
+    // Fall back to a full scan when the merge keys are not those fields.
+    let use_index = params.use_index && merge_keys_are_leading_fields(&target_schema, &params.on);
     let mut builder = LanceMergeInsertBuilder::try_new(dataset.clone(), params.on)?;
     match (
         params.when_matched_update_all,
@@ -281,7 +286,7 @@ pub(crate) async fn execute_merge_insert(
     } else {
         builder.when_not_matched_by_source(WhenNotMatchedBySource::Keep);
     }
-    builder.use_index(params.use_index);
+    builder.use_index(use_index);
 
     let future = if let Some(timeout) = params.timeout {
         let future = builder
@@ -309,6 +314,15 @@ pub(crate) async fn execute_merge_insert(
         num_deleted_rows: stats.num_deleted_rows,
         num_attempts: stats.num_attempts,
         num_rows: stats.num_inserted_rows + stats.num_updated_rows,
+    })
+}
+
+fn merge_keys_are_leading_fields(schema: &ArrowSchema, keys: &[String]) -> bool {
+    keys.iter().enumerate().all(|(idx, key)| {
+        schema
+            .fields()
+            .get(idx)
+            .is_some_and(|field| field.name() == key)
     })
 }
 
--- a/python/python/tests/test_table.py
+++ b/python/python/tests/test_table.py
@@ -2204,6 +2204,61 @@ def test_merge_insert(mem_db: DBConnection):
         )
 
 
+def test_merge_insert_update_all_with_scalar_index(tmp_db: DBConnection):
+    schema = pa.schema(
+        [
+            pa.field("vector", pa.list_(pa.float32(), 4), nullable=True),
+            pa.field("path", pa.string(), nullable=False),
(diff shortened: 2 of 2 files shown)


rust/lancedb/src/table/merge.rs
```diff
@@ -281,7 +286,7 @@ pub(crate) async fn execute_merge_insert(
     } else {
         builder.when_not_matched_by_source(WhenNotMatchedBySource::Keep);
     }
-    builder.use_index(params.use_index);
+    builder.use_index(use_index);
```
wjones127, line 289: issue(blocking): this will cause performance regressions in some cases. We should probably fix whatever bug is upstream. I think I'd rather do that.

wjones127: In theory, https://github.com/lance-format/lance/pull/7251, should have fixed this. Is your test failing without that change? if so, we should raise a bug issue in the lance-format/lance repo.

wjones127 (changes requested):

rust/lancedb/src/table/merge.rs
goutamadwant: @wjones127 I checked this locally.

With the LanceDB-side fallback removed, the regression still fails on the currently pinned Lance `v9.0.0-beta.10`: the merge result reports `num_updated_rows=0` instead of updating the expected rows.

I also confirmed that `v9.0.0-beta.10` already contains lance-format/lance#7251, and tested the same regression against Lance main at `dca65b904ce7b6398ff9b3e7aa5c9fbb4951af40` (`9.0.0-beta.11`). It still fails there with the same `num_updated_rows=0` behavior.

So I agree this should be raised upstream in `lance-format/lance`. I’ll open a focused Lance issue with the LanceDB regression and link it back here before changing this PR further. Does that make sense or any suggestion?

PR #3613, opened 2026-07-02
fix(python): fill bad vector values element-wise

### LINKED ISSUE #3026: bug: behaviour of "bad vector fill" doesn't align with user expectation
### LanceDB version

master

### What happened?

The [docs](https://docs.lancedb.com/tables/consistency) incorrectly describe `on_bad_vectors='fill'` as element-wise `NaN` replacement, whereas the actual Rust code replaces the ENTIRE vector.
- Docs claim [1.0, NaN, 3.0] becomes [1.0, 0.0, 3.0] with fill_value=0.0, but the code at table.py:3177-3181 replaces the entire vector with [0.0, 0.0, 0.0] — the is_bad flag is per-vector, not per-element
- Users lose ALL valid elements in partially-bad vectors without knowing it
- Zero fill vectors cause downstream issues: undefined cosine similarity (division by zero) and L2 results clustering near the origin

More context here:
https://github.com/lancedb/docs/issues/134

## Intended behaviour

If we're replacing the whole vector we shouldn't even have the feature (we should just be dropping the bad vectors at that point). In such a failing scenario (where there is a bad vector that cannot be ingested) the user would be better served by us dropping the whole vector and point to the failing record ID, loudly calling it out and potentially spilling the bad vector IDs via a summary to stdout.

_No response_

prrao87:
## Summary

Fix `on_bad_vectors="fill"` so it replaces only invalid or missing vector values instead of replacing the entire vector row.

Fixes #3026.

## Reasoning

The old Python sanitizer detected whether a vector row was bad at row granularity. For `fill`, it then used that row-level flag to replace the whole vector with `[fill_value] * dim`. That meant an input like `[1.0, NaN, 3.0]` became `[0.0, 0.0, 0.0]`, even though the documented and more useful behavior is to preserve valid values and fill only the bad element.

I checked whether this should be a Rust-side fix so TypeScript users would benefit too. Today, Rust core exposes `NaNVectorBehavior::{Error, Keep}` for rejecting or keeping NaN vectors, while the Python `on_bad_vectors` API (`error`, `drop`, `fill`, `null`) is implemented in the Python ingestion sanitizer before data reaches Rust. TypeScript does not expose the Python `on_bad_vectors="fill"` behavior today. Moving this exact behavior to Rust would be a broader cross-language API change, so this PR keeps the fix scoped to the currently affected Python API.

## What changed

- Added a small helper that fills bad vector rows by preserving valid elements, replacing NaN elements with `fill_value`, truncating vectors longer than the expected dimension, and padding short vectors with `fill_value`.
- Kept the existing fast path unchanged: the helper only runs after bad vectors are detected and `on_bad_vectors="fill"` is selected.
- Updated sanitizer and table tests to assert element-wise NaN replacement and short-vector padding for both `create_table` and `add`.

## Validation

- `uv run ruff format .`
- `uv run ruff check .`
- `cd python && uv run --no-sync pytest python/tests/test_util.py::test_handle_bad_vectors_jagged python/tests/test_util.py::test_handle_bad_vectors_nan python/tests/test_table.py::test_create_with_nans python/tests/test_table.py::test_add_with_nans -vv`

Targeted pytest result: `10 passed`.

## Why this fix is Python-side (and not Rust)

The problematic behavior lives in Python’s `on_bad_vectors` sanitizer, before data is handed off to Rust. Rust currently only exposes `NaNVectorBehavior::{Error, Keep}` for add operations, while Python has the richer `on_bad_vectors={"error","drop","fill","null"}` API. TypeScript does not currently expose the Python-style fill behavior, so moving this exact fix into Rust would require designing a broader cross-language bad-vector handling API.

This PR keeps the change scoped to the existing affected surface: Python’s `on_bad_vectors="fill"` path. This way, Python users immediately benefit.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 import asyncio
 import inspect
 import deprecation
-import math
 import warnings
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -4114,7 +4113,7 @@ def _handle_bad_vector_column(
                 raise ValueError(
                     "`fill_value` must not be None if `on_bad_vectors` is 'fill'"
                 )
-            vec_arr = _fill_bad_vector_values(vec_arr, is_bad, dim, fill_value)
+            vec_arr = _fill_bad_vector_values(vec_arr, dim, fill_value)
         else:
             raise ValueError(f"Invalid value for on_bad_vectors: {on_bad_vectors}")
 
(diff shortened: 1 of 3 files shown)


python/python/lancedb/table.py
```diff
@@ -4055,17 +4056,35 @@ def _handle_bad_vector_column(
...
+
+        vector = [] if values[idx] is None else values[idx]
+        filled = [
+            fill_value if isinstance(value, float) and math.isnan(value) else value
+            for value in vector[:dim]
+        ]
+        filled.extend([fill_value] * (dim - len(filled)))
+        values[idx] = filled
+
+    return pa.array(values, type=arr.type)
```
wjones127: suggestion(perf): having to drop down to raw python list to correct this is pretty brutal. It might be worth asking an agent to optimize this so it's run using vectorized pyarrow functions instead of python objects.

wjones127 (changes requested):
It looks like the CI failure is related to your change:

```
python/python/tests/test_table.py::test_on_bad_vectors_fill_preserves_arrow_nested_vector_type
```

python/python/lancedb/table.py
prrao87: Makes sense, that's a good point! Will rework it accordingly 👍🏽

prrao87: Addressed in 192511fe. The fill path no longer calls to_pylist() or loops over Python objects. It now uses PyArrow compute kernels to slice vectors to the target dimension, detect NaNs, combine masks, and fill only missing or invalid positions before casting back to the original list type. The only NumPy operation is a broadcast over Arrow-derived list lengths to build the positional padding mask; vector values remain columnar throughout. 

Trying this to see if CI still breaks, and also trying with a pure PyArrow fix after this.

prrao87:
Okay @wjones127 I think this fix should do it. No numpy, uses pure PyArrow + vectorized ops. The failing CI test predates this PR's fix and was written to guard the Arrow nested-list ingestion/type path. Its `[0, 0]` expectation encoded the old whole-row replacement behavior that this PR intentionally changed.

I updated the test suite to match the corrected element-wise fill semantics. The existing nested Arrow test expected the old whole-row replacement `([NaN, 3] → [0, 0])`, so it now expects `[0, 3]` and explicitly verifies that the output remains `list<float32>`. We also added coverage for short, overlong, null, and sliced vectors across Arrow list types to prevent regressions in both values and schema preservation.

wjones127 (approved):
Looks better. Thank you!

PR #3615, opened 2026-07-02
docs(java): fill Java API reference gaps

### LINKED ISSUE #157: [java]: return Arrow IPC stream instead of raw bytes
Today the user experience of the namespace java sdk is that user has to parse the Arrow stream bytes into Arrow IPC stream class, which is not user friendly; see https://lancedb.github.io/lancedb/java/java/#reading-query-results

The example should be turned into a util so users can just use it directly instead of coding their own logic against it, and included in the docs.

See: https://github.com/lancedb/lancedb/issues/3050

### LINKED ISSUE #160: [java]: creating empty table
In the Java SDK, creating an empty table = create table with an Arrow IPC stream with no data. This is not obvious.

We could also consider adding some util functions to it and documenting it better.

See: https://github.com/lancedb/lancedb/issues/3052

prrao87:
## Summary

This updates the Java API reference to close the documentation gaps that can be fixed from the current Java source and generated namespace API.

The patch adds an empty table example, shows how to wrap returned Arrow IPC query bytes in a reusable `ArrowFileReader` helper, and documents the Java index operations that are currently exposed by the namespace client: vector indexes, scalar indexes, full text search indexes, and listing indexes.

## Issue Links

Fixes https://github.com/lancedb/docs/issues/157
Fixes https://github.com/lancedb/docs/issues/160

Partially addresses https://github.com/lancedb/docs/issues/159 by documenting the index parameters currently exposed by Java. The requested `num_partitions` example is still blocked because `CreateTableIndexRequest` does not expose IVF training parameters yet.

Not included: https://github.com/lancedb/docs/issues/158. The current Java docs and source remain remote namespace oriented, so local DB connection documentation should wait until the Java local DB API is available and can be verified.

## Validation

- Built the Java core module with OpenJDK 17:
  `./mvnw -pl lancedb-core -am -DskipTests compile`
- Checked the Markdown diff:
  `git diff --check -- docs/src/java/java.md`

The Java build succeeds. It still reports pre-existing checkstyle warnings in the namespace client builder, but the Maven build is green.

### DIFF
--- a/docs/src/java/java.md
+++ b/docs/src/java/java.md
@@ -249,6 +249,57 @@ try (BufferAllocator allocator = new RootAllocator();
 }
 ```
 
+### Creating an Empty Table
+
+To create an empty table, send an Arrow IPC stream that contains the table schema and no record batches.
+The schema in the IPC stream becomes the table schema, and rows can be inserted later.
+
+```java
+import org.lance.namespace.model.CreateTableRequest;
+import org.lance.namespace.model.CreateTableResponse;
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.memory.RootAllocator;
+import org.apache.arrow.vector.VectorSchemaRoot;
+import org.apache.arrow.vector.ipc.ArrowStreamWriter;
+import org.apache.arrow.vector.types.FloatingPointPrecision;
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.FieldType;
+import org.apache.arrow.vector.types.pojo.Schema;
+
+import java.io.ByteArrayOutputStream;
+import java.nio.channels.Channels;
+import java.util.Arrays;
+
+Schema schema = new Schema(Arrays.asList(
+    new Field("id", FieldType.nullable(new ArrowType.Int(32, true)), null),
+    new Field("name", FieldType.nullable(new ArrowType.Utf8()), null),
+    new Field("embedding",
+        FieldType.nullable(new ArrowType.FixedSizeList(128)),
+        Arrays.asList(new Field("item",
+            FieldType.nullable(new ArrowType.FloatingPoint(FloatingPointPrecision.SINGLE)),
+            null)))
+));
+
+byte[] emptyTableData;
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Looks good.  Attaching review from Claude

---

Both agents are done. Let me synthesize and present the review.

---

## PR #3615 Review — `docs(java): fill Java API reference gaps`

This PR adds documentation for creating empty tables, vector/scalar/FTS index operations, and listing indexes in the Java namespace client. It also refactors the `ByteArraySeekableByteChannel` helper into a cleaner `ArrowIpc` wrapper class. The Java model class names, setter names, return types, and accessor methods were all verified against the upstream generated source — they are correct.

Three findings survived verification:

---

**1. `docs/src/java/java.md:600` — `openFileReader` falsely declares `throws IOException`**

The `ArrowFileReader(SeekableByteChannel, BufferAllocator)` constructor in Arrow 15.0.0 (the version pinned in `java/pom.xml`) does not throw a checked exception — IO exceptions only surface on `loadRecordBatch()` and similar calls. The new `static ArrowFileReader openFileReader(...) throws IOException` declaration forces every caller to either catch `IOException` or propagate it, for an exception the method body can never actually raise at construction time. The `import java.io.IOException` added at the top exists solely to support this false declaration. The fix is to drop `throws IOException` from the method signature.

**2. `docs/src/java/java.md:570` — Section heading demotion breaks document structure**

`### Reading Query Results` (an `h3` subsection under `## Querying Data`) was changed to `## Reading Query Results` (an `h2` top-level section), making it a peer of major sections like `## Indexing` and `## Querying Data`. "Reading query results" is conceptually part of the querying workflow, not a standalone topic. This likely also breaks any auto-generated table of contents. The heading should revert to `###`.

**3. `docs/src/java/java.md:534` — FTS index calls `createTableScalarIndex` with no explanation**

The Full Text Search index example routes through `createTableScalarIndex` (returning `CreateTableScalarIndexResponse`) while the vector index uses `createTableIndex` (returning `CreateTableIndexResponse`). Both are factually correct — there is no `createTableFtsIndex` method. However, a reader will assume this is a copy-paste error and either file a bug or reach for the wrong method. A one-line note ("FTS indexes are created through `createTableScalarIndex`, not `createTableIndex`") would prevent that confusion.

PR #3618, opened 2026-07-03
fix(python): gemini batching, user agent and variable dims

markmcd:
Carrying over from #2915, this patch introduces:
* Single-API call batching support for Gemini embeddings (up to 100 at a time, the API limit)
* A versioned user agent header for Gemini API calls
* Support for [variable embedding dimension size](https://ai.google.dev/gemini-api/docs/embeddings#control-embedding-size) (Gemini is MRL trained)

### DIFF
--- a/python/python/lancedb/embeddings/gemini_text.py
+++ b/python/python/lancedb/embeddings/gemini_text.py
@@ -15,6 +15,8 @@ from .base import TextEmbeddingFunction
 from .registry import register
 from .utils import TEXT, api_key_not_found_help
 
+EMBEDDING_BATCH_SIZE = 100
+
 
 @register("gemini-text")
 class GeminiText(TextEmbeddingFunction):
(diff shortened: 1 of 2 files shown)


python/python/lancedb/embeddings/gemini_text.py
```diff
@@ -133,24 +136,17 @@ def generate_embeddings(
...
-                contents=content,
-                config=config,
-            )
-            embeddings.append(response.embeddings[0].values)
-
-        return embeddings
+        config = types.EmbedContentConfig(**config_kwargs) if config_kwargs else None
+        response = self.client.models.embed_content(
+            model=self.name,
+            contents=contents,
```
wjones127: question(blocking): we now pass the whole of contents through. Does the gemini have any limit on tokens / request? If so, this might cause 400 errors where there were none before. It might be worth finding some heuristic to split up the requests on.

wjones127 (changes requested):
One question, otherwise looks good.

python/python/lancedb/embeddings/gemini_text.py
markmcd: Ah good point. We do have a limit of 100 in a batch. Fixed in the latest push.

wjones127 (approved):
Thanks! looks good.

Issue #3611, opened 2026-07-02
bug(python): Visual Code Debugger Session is freezes after connect to LanceDB


masalinas:
### LanceDB version

Lancedb version: 0.33.0
Visual Code version: 1.124.2

### What happened?

I have integrated lancedb with minio like this:

```
lancedb_uri = f"s3://{TEST_BUCKET}/vector_indices"

storage_options = {
    "endpoint": os.getenv("MINIO_ENDPOINT_URL"),
    "access_key_id": os.getenv("AWS_ACCESS_KEY_ID"),
    "secret_access_key": os.getenv("AWS_SECRET_ACCESS_KEY"),        
    "allow_http": "true",
    "region": "us-east-1"
}

db_connection = lancedb.connect(
    lancedb_uri,
    storage_options=storage_options
)
```

And afte execute the last line  `lancedb.connect` the Visual Code Debugger session is freezed, is impossible jump to any breakpoint or watch any variable inside visual code.

This is my `launch.json` Visual Code file:

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [        
        {
            "name": "PoC Veradoc llamaindex",
            "type": "debugpy",
            "request": "launch",
            "program": "${file}",
            "args": [
                //"--file-path", "docs/DOC_CD2026-001928329.pdf"
                "--file-path", "docs/sentencia-29-1030x832.jpg",
                //"--file-path", "docs/file-sample_100kB.doc"
            ],
            "console": "integratedTerminal",
            "justMyCode": true,
            "env": {
                "PYDEVD_USE_FRAME_EVAL": "NO"
            }
        }
    ]
}
```

and my `settings.json` Visual Code file is:

```
{
    "python-envs.pythonProjects": [
        {
            "path": ".",
            "envManager": "ms-python.python:venv",
            "packageManager": "ms-python.python:pip"
        }
    ],
}
```

### Are there known steps to reproduce?

The steps are the previous lines, where configure your minio credentials and connect lancedb using these credentials, anymore.

Xuanwo:
[...]

PR #3620, opened 2026-07-03
fix(python): avoid async work in sync reprs

ssrheart425:
## Summary

- keep the existing synchronous `connect()` path unchanged
- make `LanceDBConnection.__repr__` and `LanceTable.__repr__` side-effect-free
- add a regression test that verifies sync reprs do not call the Python background loop

## Root cause

The freeze is caused by debugger rendering, not by `connect()` itself:

1. debugpy stops at a breakpoint and suspends all Python threads.
2. The debugger renders the new `db_connection` local by calling `repr()`.
3. `LanceDBConnection.__repr__` reads `read_consistency_interval`.
4. That property calls `LOOP.run(...).result()`.
5. The `LanceDBBackgroundEventLoop` thread is suspended by the debugger, so `repr()` waits for a thread that cannot run.

This explains why the symptom appears immediately after `connect()`: it is the first point where a connection object exists in locals and is automatically rendered. `LanceTable.__repr__` had the same problem because it also read the connection's consistency interval.

This follows the same principle as #3411: `__repr__` must not trigger async work or I/O that a debugger assumes is lightweight.

## Evidence

I reproduced the behavior with the real LanceDB classes and debugpy 1.8.21 using a DAP client:

- latest `main` (`ff6ff099`): the debugger reported `allThreadsStopped: true`, and evaluating `repr(db_connection)` timed out
- this branch (`5755a5ba`): the same evaluation returned `LanceDBConnection(uri='/tmp/lancedb-debug-repro')` immediately
- setting `PYDEVD_UNBLOCK_THREADS_TIMEOUT=0` also allowed the original repr path to complete, independently confirming that it was waiting on a suspended thread

The regression test creates a connection and table, replaces `LOOP.run` with a function that fails, and verifies that both reprs still work.

## Validation

- `maturin develop --manifest-path python/Cargo.toml`
- `python -m pytest python/python/tests/test_db.py::test_sync_repr_does_not_use_background_loop python/python/tests/test_table.py::test_consistency -q` (`4 passed`)
- `ruff check .`
- `ruff format --check python/python/lancedb/db.py python/python/lancedb/table.py python/python/tests/test_db.py python/python/tests/test_table.py`
- `git diff --check`

Refs #3611.

### DIFF
--- a/python/python/lancedb/db.py
+++ b/python/python/lancedb/db.py
@@ -688,11 +688,7 @@ class LanceDBConnection(DBConnection):
         return cls(None, _inner=inner)
 
     def __repr__(self) -> str:
-        val = f"{self.__class__.__name__}(uri={self._conn.uri!r}"
-        if self.read_consistency_interval is not None:
-            val += f", read_consistency_interval={repr(self.read_consistency_interval)}"
-        val += ")"
-        return val
+        return f"{self.__class__.__name__}(uri={self._conn.uri!r})"
 
     @override
     def serialize(self) -> str:
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -2464,13 +2464,7 @@ class LanceTable(Table):
         return LOOP.run(self._table.count_rows(filter))
 
     def __repr__(self) -> str:
-        val = f"{self.__class__.__name__}(name={self.name!r}"
-        if self._conn.read_consistency_interval is not None:
-            val += ", read_consistency_interval={!r}".format(
-                self._conn.read_consistency_interval
-            )
-        val += f", _conn={self._conn!r})"
-        return val
+        return f"{self.__class__.__name__}(name={self.name!r}, _conn={self._conn!r})"
 
     def __str__(self) -> str:
         return self.__repr__()
--- a/python/python/tests/test_db.py
+++ b/python/python/tests/test_db.py
@@ -62,6 +62,21 @@ def test_basic(tmp_path):
     assert db.open_table("test").name == db["test"].name
 
 
+def test_sync_repr_does_not_use_background_loop(tmp_path, monkeypatch):
+    from lancedb.background_loop import LOOP
+
+    db = lancedb.connect(tmp_path)
+    table = db.create_table("test", data=[{"id": 1}])
+
+    def fail_run(*args, **kwargs):
+        raise AssertionError("repr should not use the Python background loop")
+
+    monkeypatch.setattr(LOOP, "run", fail_run)
+
+    assert repr(db) == f"LanceDBConnection(uri={str(tmp_path)!r})"
+    assert repr(table) == f"LanceTable(name='test', _conn={db!r})"
+
+
 def test_ingest_pd(tmp_path):
     db = lancedb.connect(tmp_path)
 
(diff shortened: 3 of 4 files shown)


wjones127 (changes requested):
This PR changes how `connect()` works, but doesn't really change any of the other places (all the methods on connection and table) that use `LOOP.run()`. Almost every method uses that. Do we have evidence the issue #3611 is specific to `connect()`? Or are we just changing the place where a user will get a frozen debugger?

If we can show this actually does fix the issue entirely, I'm open to merging it. Though I would like a deeper understanding of why. Otherwise I think we need a different solution.

masalinas:
Any schedule for this ticket?. Not being able to debug using lancedb from visual code is critical.

ssrheart425:
Thanks @wjones127 — you were right that the previous test only proved `connect()` no longer called `LOOP.run()` and did not show that #3611 was connect-specific.

I dug further and force-pushed a different fix. The deadlock is caused by debugger rendering rather than `connect()` itself:

- debugpy suspends all threads at the breakpoint and calls `repr()` to render `db_connection`
- `LanceDBConnection.__repr__` read `read_consistency_interval`
- that property called `LOOP.run(...).result()`
- the `LanceDBBackgroundEventLoop` thread was also suspended, so repr waited on a thread that could not run

I reproduced this with the real LanceDB classes and debugpy 1.8.21:

- latest `main` (`ff6ff099`): DAP reported `allThreadsStopped: true`, and `repr(db_connection)` timed out
- updated branch (`5755a5ba`): the same evaluation returned immediately
- `PYDEVD_UNBLOCK_THREADS_TIMEOUT=0` also made the old repr complete, which independently confirms the suspended-thread dependency

That explains why it appears to happen after `connect()`: this is the first point where the connection object exists in locals and VS Code automatically renders it. Ordinary connection/table methods still use `LOOP.run()` while the program is running, but they are not implicitly called by debugger rendering.

The updated PR now keeps the original connect path, makes connection/table reprs side-effect-free, and includes a regression test that fails if either repr calls `LOOP.run()`. This is also consistent with the approach taken in #3411.

@masalinas, could you please test the updated branch with your MinIO + VS Code setup, especially with breakpoints immediately after `lancedb.connect()` and after a normal table operation? That real-environment confirmation would let us close the remaining validation gap.

masalinas:
After install a especial version of protobuf compiler (29.2) in my Ubuntu 22.04. Finally I compiled your fix branch of lanceDB in my computer and test it using my Minio service as usual and this simple sample Python code to set some breackpoints and debug it from my Visual Studio:

```
from datetime import timedelta
import lancedb

db = lancedb.connect(
    "s3://warehouse/v-db/",
    read_consistency_interval=timedelta(seconds=5),
    storage_options={
        "endpoint": "http://localhost:9000",   
        "aws_access_key_id": "minioadmin",
        "aws_secret_access_key": "minioadmin",
        "allow_http": "true",
        "region": "us-east-1",
        "virtual_hosted_style_request": "false",
    }
)

tables = list(db.table_names())

print(tables)
```

I could debug this piece of code from my VS without any problem, so this **fix branch fix/sync-connect-blocking resolve the frozen debugging problem in Visual Studio**. 

Please review and merge this branch asap. Regards

wjones127 (approved):
Nice work investigating this!

PR #3625, opened 2026-07-06
fix(python): bound scanner memory for wide-row bulk ingestion

wjones127:
## Problem

`table.add(dataset)` with a `pyarrow.dataset.Dataset` OOMs the client during bulk ingestion of wide rows (e.g. embedding columns), even against a remote table where the upload itself is streaming.

The cause is in `to_scannable`: a `Dataset` is scanned with pyarrow's default scanner settings (`batch_size=131072` rows, `batch_readahead=16`, `fragment_readahead=4`). pyarrow's internal threads prefetch that read-ahead window independently of LanceDB's backpressure, so for wide rows a large fraction of the dataset is held in memory. On the remote path this is then multiplied across the multipart write partitions (one in-flight batch per partition, up to CPU-core count).

Reproduced on a 10 GB / 1.55M-row dataset with two 768-dim float32 embeddings: peak client RSS ~11.7 GB for the scan alone (6.8 GB after consuming a *single* batch), ~15.4 GB for the full remote `add()`.

## Fix

`to_scannable` now sizes the scanner from an estimate of bytes-per-row derived from the schema:

- **Narrow datasets keep pyarrow's defaults** (empty scanner kwargs) — no throughput regression. The bound only engages above ~410 bytes/row.
- **Wide rows** get a smaller `batch_size` (~16 MiB/batch) and reduced read-ahead (`batch_readahead=2`, `fragment_readahead=1`) so peak in-flight memory stays near a ~1 GiB budget. Read-ahead (not just batch size) has to drop, because pyarrow pins whole row-group buffers.

On the 10 GB dataset this drops peak client RSS to ~1.4 GB, and it stays flat as the dataset grows. The `Dataset`/`LanceDataset` scannables remain rescannable (retry-safe).

## Also: expose `write_parallelism` on `add()`

`AddDataBuilder::write_parallelism` already existed in Rust but was not exposed in Python. This PR forwards it through the async, sync, and remote `add()` methods, so users can cap the number of parallel write partitions (each buffers data in flight) to trade throughput for memory on large uploads.

## Tests

- `test_scannable.py`: bytes-per-row estimation; narrow → defaults; wide → bounded; `Dataset` reader streams bounded batches and stays rescannable.
- `test_table.py`: `write_parallelism` on sync and async `add()`, and that `write_parallelism=0` is rejected.

Fixes ENT-1883

### DIFF
--- a/python/python/lancedb/scannable.py
+++ b/python/python/lancedb/scannable.py
@@ -7,6 +7,7 @@ import sys
 from typing import Callable, Iterator, Optional
 from lancedb.arrow import to_arrow
 import pyarrow as pa
+import pyarrow.compute as pc
 import pyarrow.dataset as ds
 
 from .pydantic import LanceModel
@@ -32,14 +33,32 @@ _WIDE_FRAGMENT_READAHEAD = 1
 # Estimate for variable-width columns (string/binary/list) whose true width is
 # unknown from the schema alone. Only needs to be large enough to flag "wide".
 _VARIABLE_WIDTH_ESTIMATE = 128
+# Rows peeked from a rescannable source to refine the list-length guess for
+# variable-length list columns (e.g. embeddings stored as `list<float32>`
+# instead of `list<float32, N>`), whose per-row width the schema can't tell us.
+_SAMPLE_ROWS = 10
 
 
-def _estimate_field_width(dtype: pa.DataType) -> int:
+def _observed_list_length(sample: pa.ChunkedArray) -> Optional[int]:
+    """Average element count per row observed in a list/large_list sample."""
+    if len(sample) == 0:
+        return None
+    mean = pc.mean(pc.list_value_length(sample)).as_py()
+    return None if mean is None else max(1, round(mean))
+
+
+def _estimate_field_width(
+    dtype: pa.DataType, sample: Optional[pa.ChunkedArray] = None
+) -> int:
     if pa.types.is_fixed_size_list(dtype):
         return dtype.list_size * _estimate_field_width(dtype.value_type)
     if pa.types.is_struct(dtype):
         return sum(
-            _estimate_field_width(dtype.field(i).type) for i in range(dtype.num_fields)
+            _estimate_field_width(
+                dtype.field(i).type,
+                pc.struct_field(sample, [i]) if sample is not None else None,
+            )
+            for i in range(dtype.num_fields)
         )
     if pa.types.is_dictionary(dtype):
         return _estimate_field_width(dtype.value_type)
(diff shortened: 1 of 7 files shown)


python/python/lancedb/table.py
```diff
@@ -4842,7 +4865,12 @@ async def add(
         data = to_scannable(data)
         progress, owns = _normalize_progress(progress)
         try:
-            return await self._inner.add(data, mode or "append", progress=progress)
+            return await self._inner.add(
```
jackye1995, line 4868: Could we route the sanitize path above through the same bounded Dataset scanner as `to_scannable()`? For `mode="overwrite"`, non-default `on_bad_vectors`, and embedding-function tables, `AsyncTable.add()` converts a `pyarrow.dataset.Dataset` / `lance.LanceDataset` inside `_sanitize_data()` / `_into_pyarrow_reader()` before reaching this call. Those dataset branches still call bare `data.scanner().to_reader()`, so these `table.add(dataset)` cases keep PyArrow's default read-ahead and can still hit the wide-row OOM this PR is trying to avoid.

python/python/lancedb/scannable.py
```diff
@@ -11,6 +11,92 @@
...
+    if pa.types.is_fixed_size_binary(dtype):
+        return dtype.byte_width
+    if pa.types.is_boolean(dtype):
+        return 1
+    # Fixed-width scalars (ints, floats, temporal, decimal) expose bit_width;
+    # variable-width types (string, binary, list, map, ...) raise ValueError.
+    try:
+        return max(1, dtype.bit_width // 8)
+    except (ValueError, AttributeError):
+        return _VARIABLE_WIDTH_ESTIMATE
```
jackye1995, line 80: Could we make the variable-length list estimate less optimistic here? For `list<float32>` / `large_list<float32>` columns, the schema does not always include the vector dimension, so the current 128-byte fallback can classify common embedding-shaped rows as "narrow" and leave PyArrow's default batch/read-ahead settings in place. A more robust approach might be to sample a small number of rows or the first scanner batch, compute the observed list lengths for numeric list columns, and use that sampled dimension to estimate bytes per row before deciding whether to bound `batch_size`, `batch_readahead`, and `fragment_readahead`.

wjones127: Yeah, I suppose we can get a small sample from a rescannable source. That would be the best approach.

wjones127: Pushed in 20ebde7f: `to_scannable` now peeks up to 10 rows from `Dataset`/`LanceDataset` sources (tight batch size, no read-ahead, so the peek itself can't trigger the blowup we're fixing) and uses the observed average list length to size variable-length `list`/`large_list` columns, instead of falling back to the flat 128-byte estimate.

jackye1995 (approved):

PR #3629, opened 2026-07-07
fix(python): flatten_columns raises when flatten=False

PranavAchar01:
### Summary

`flatten_columns` raises `ValueError` when called with `flatten=False`, even though `False` should mean "do not flatten". This is reachable from the public API — `Query.to_pandas(flatten=...)` and `to_batches(flatten=...)` type their `flatten` param as `Optional[Union[int, bool]]` and pass it straight to `flatten_columns`.

### Cause

`bool` is a subclass of `int`, so `isinstance(False, int)` is `True`. `flatten=False` skips the `flatten is True` check, falls into the integer branch, and `False <= 0` evaluates to `True`, raising:

```
ValueError: Please specify a positive integer for flatten or the boolean value `True`
```

### Reproduction

```python
import lancedb
db = lancedb.connect("/tmp/db")
t = db.create_table("t", data=[{"id": 1, "vector": [0.1, 0.2]}])
t.search([0.1, 0.2]).to_pandas(flatten=False)   # -> ValueError
```

### Fix

Guard the integer branch with `not isinstance(flatten, bool)` so that `flatten=False` (and `None`) mean "do not flatten". Behavior is otherwise unchanged:

- `flatten=True` → flatten all nested levels
- positive `int` → flatten to that depth
- non-positive `int` (e.g. `0`) → still rejected with `ValueError`

Added a regression test in `tests/test_util.py` covering `None`, `False`, `True`, a positive depth, and `0`.

### DIFF
--- a/python/python/lancedb/util.py
+++ b/python/python/lancedb/util.py
@@ -177,7 +177,10 @@ def flatten_columns(tbl: pa.Table, flatten: Optional[Union[int, bool]] = None):
                 continue
             else:
                 break
-    elif isinstance(flatten, int):
+    # `bool` is a subclass of `int`, so guard against it explicitly: `flatten=False`
+    # (and `None`) must mean "do not flatten" rather than falling into the integer
+    # branch and raising on the `flatten <= 0` check.
+    elif isinstance(flatten, int) and not isinstance(flatten, bool):
         if flatten <= 0:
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
This looks like a good fix. Will merge once CI is passing.

PR #3630, opened 2026-07-07
fix(rust): bound remote insert request size to avoid ingestion timeouts

wjones127:
## Problem

On the remote (LanceDB Cloud) write path, each write partition is uploaded as a **single** `/insert?upload_id=...` request that stays open until the whole partition has been streamed and the server has written it to object storage. For large bulk ingests a partition can be many GB, so a single request can run longer than the client read timeout (default 300s), surfacing as:

```
lancedb.remote.errors.HttpError: operation timed out
```

The server already supports staging **multiple** parts under one `upload_id` (each `/insert` writes a separate transaction that `complete` merges atomically), but the client never used that — it sent one part per partition.

## Change

Split each partition into multiple parts of at most `max_bytes_per_request` (Arrow IPC, LZ4-compressed) bytes, each uploaded as its own `/insert?upload_id=...&upload_part_id=...` request. This bounds how long any single request stays open, independent of total data size or write parallelism.

Key properties:
- **Still streamed, not buffered.** Each part's body is driven through a bounded channel while the request is in flight (`futures::join!` of a producer + the send), so peak memory stays at a couple of batches per partition regardless of the part size. Backpressure from a slow/throttled server still propagates upstream.
- **Correct part accounting.** An empty partition still sends exactly one (schema-only) part so `complete` has a transaction to commit; a size cut landing exactly on the end of input does not emit a trailing empty part.
- **Multipart only.** The single-request (non-multipart) path is unchanged.

## Config

New `ClientConfig::max_bytes_per_request: Option<usize>`, also settable via the `LANCE_CLIENT_MAX_BYTES_PER_REQUEST` environment variable. **Default 1 GiB** (`Some(0)` disables splitting → one request per partition). Python users pick up the default/env automatically through the remote client.

## Tests

- `test_multipart_chunked_splits_into_parts`: a 1-byte budget puts each batch in its own part → N requests, each carrying the shared `upload_id` and a distinct `upload_part_id`.
- `test_multipart_single_part_when_under_budget`: a large budget keeps the partition in a single request.
- Verified end-to-end against a live remote table: a forced-chunked multipart add (many parts) assembles to the correct row count.

Related to ENT-1883.

### DIFF
--- a/rust/lancedb/src/remote/client.rs
+++ b/rust/lancedb/src/remote/client.rs
@@ -54,7 +54,7 @@ pub trait HeaderProvider: Send + Sync + std::fmt::Debug {
 /// into undersized files across parts. The time-based cut
 /// ([`DEFAULT_MAX_REQUEST_DURATION_DIVISOR`]) bounds request duration on slow
 /// uploads, so a large byte budget does not risk the read timeout.
-const DEFAULT_MAX_BYTES_PER_REQUEST: usize = 8 * 1024 * 1024 * 1024;
+const DEFAULT_MAX_BYTES_PER_REQUEST: u64 = 8 * 1024 * 1024 * 1024;
 
 /// The default max request duration is the read timeout divided by this, leaving
 /// headroom for the server to finalize and acknowledge a part before the read
@@ -97,7 +97,7 @@ pub struct ClientConfig {
     /// increase peak memory. Set to `Some(0)` to disable splitting (one request
     /// per partition). You can also set the `LANCE_CLIENT_MAX_BYTES_PER_REQUEST`
     /// environment variable. Defaults to 8 GiB.
-    pub max_bytes_per_request: Option<usize>,
+    pub max_bytes_per_request: Option<u64>,
     /// Maximum wall-clock time to spend uploading a single insert HTTP request.
     ///
     /// Complements [`Self::max_bytes_per_request`]: during a multipart write a
@@ -299,7 +299,7 @@ pub struct RestfulLanceDbClient<S: HttpSend = Sender> {
     // `resolve_max_bytes_per_request` / `resolve_max_request_duration`), where a
     // default has already been applied and `None` means the feature is disabled.
     /// Maximum bytes per insert request. `None` disables request splitting.
-    pub(crate) max_bytes_per_request: Option<usize>,
+    pub(crate) max_bytes_per_request: Option<u64>,
     /// Maximum wall-clock time per insert request. `None` disables the
     /// time-based part cut.
     pub(crate) max_request_duration: Option<Duration>,
@@ -506,11 +506,11 @@ impl RestfulLanceDbClient<Sender> {
 
     /// Resolve the max bytes per insert request from config, environment, or the
     /// default. A value of `0` (from either source) disables request splitting.
-    fn resolve_max_bytes_per_request(passed: Option<usize>) -> Result<Option<usize>> {
+    fn resolve_max_bytes_per_request(passed: Option<u64>) -> Result<Option<u64>> {
         let value = if let Some(value) = passed {
             value
         } else if let Ok(env) = std::env::var("LANCE_CLIENT_MAX_BYTES_PER_REQUEST") {
-            env.parse::<usize>().map_err(|_| Error::InvalidInput {
+            env.parse::<u64>().map_err(|_| Error::InvalidInput {
                 message: format!(
                     "LANCE_CLIENT_MAX_BYTES_PER_REQUEST must be a non-negative integer, got '{}'",
                     env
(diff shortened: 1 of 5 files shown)


rust/lancedb/src/remote/table/insert.rs
```diff
@@ -213,6 +221,173 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
...
+    /// the caller completes the multipart write. An empty partition still sends
+    /// exactly one (schema-only) part so the session has a transaction to
+    /// commit.
+    #[allow(clippy::too_many_arguments)]
+    async fn send_multipart_chunked(
+        client: RestfulLanceDbClient<S>,
+        identifier: String,
+        table_name: String,
+        upload_id: String,
+        branch: Option<String>,
```
wjones127: question: why String and not `&str`? Many of these seem to convert to `&str` later.

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -213,6 +221,173 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
...
+        overwrite: bool,
+        max_bytes: usize,
+        mut input: SendableRecordBatchStream,
+        tracker: Option<Arc<WriteProgressTracker>>,
+    ) -> DataFusionResult<()> {
+        let schema = input.schema();
+
+        // Prime the first batch so an empty partition still sends one part and a
```
wjones127: question: why would an empty partition need to send one part? Shouldn't you just early return if it's empty?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -213,6 +221,173 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
...
+            let mut part_bytes: usize = 0;
+            let mut input_ended = false;
+
+            let input_ref = &mut input;
+            let part_bytes_ref = &mut part_bytes;
+            let input_ended_ref = &mut input_ended;
+            let schema_ref = &schema;
+            let producer = async move {
+                let options = arrow_ipc::writer::IpcWriteOptions::default()
+                    .try_with_compression(Some(CompressionType::LZ4_FRAME))
```
wjones127: suggestion: this function body is very long. Could we break some of these out into helper functions with clear signatures?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -215,6 +233,223 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
...
+        };
+
+        loop {
+            let (part_bytes, input_ended) = self
+                .send_one_part(&schema, max_bytes, max_duration, first, &mut input)
+                .await?;
+
+            if let Some(ref t) = tracker {
+                t.record_bytes(part_bytes);
+            }
```
wjones127: issue(blocking): recording the bytes at the end makes the progress bar look like it's starting and stopping. Can we push that down into the stream?

rust/lancedb/src/remote/client.rs
```diff
@@ -47,6 +47,20 @@ pub trait HeaderProvider: Send + Sync + std::fmt::Debug {
...
 }
 
+/// Default maximum bytes per insert request (8 GiB).
+///
+/// Sized so a multipart part can hold at least one full Lance data file (the
+/// default is 1M rows / 90 GB per file), which keeps fragments from being split
+/// into undersized files across parts. The time-based cut
+/// ([`DEFAULT_MAX_REQUEST_DURATION_DIVISOR`]) bounds request duration on slow
+/// uploads, so a large byte budget does not risk the read timeout.
+const DEFAULT_MAX_BYTES_PER_REQUEST: usize = 8 * 1024 * 1024 * 1024;
```
wjones127: issue: `usize` is probably the wrong type. On 32-bit systems the default overflows, right?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -215,6 +233,238 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
...
+    ) -> DataFusionResult<()> {
+        let schema = input.schema();
+
+        // A part always starts from a batch we already hold: the first batch of
+        // the partition, or the look-ahead batch from the previous part. This
+        // keeps empty partitions from staging a part and stops a size cut that
+        // lands exactly on the end of input from emitting a trailing empty part.
+        let mut first = match input.next().await {
+            Some(batch) => batch?,
+            None => return Ok(()),
```
jackye1995, line 279: Could we keep staging one schema-only part when the whole multipart input is empty, or avoid multipart for known zero-row inputs? This early return means an empty multipart overwrite can create an upload and complete it without any insert request; since `mode=overwrite` is only attached to the per-part insert request, complete has no staged part carrying the overwrite intent. That seems like a regression from the previous schema-only empty insert request behavior.

wjones127: Good point. Let's see if we can initiate the upload only when we have data.

wjones127: Fixed in e642f9f1: `add()` now peeks the input regardless of `write_parallelism` and falls back to the single-request path when it's fully empty, so multipart is never entered for a zero-batch write. Added `test_multipart_write_empty_overwrite_uses_single_partition` to cover the empty-overwrite-with-explicit-parallelism case.

jackye1995 (approved):

PR #3635, opened 2026-07-09
fix: avoid manifest writes for read-only directory namespace opens

### LINKED ISSUE #3633: bug(python): read-only hf:// table open attempts xet write token via directory namespace init
### LanceDB version

0.33.0

### What happened?

Opening an existing Lance table from a Hugging Face bucket with a read-only token can trigger a write-token request:

```text
POST .../xet-write-token
```

This makes a read-only training/data-loading path fail with 403, even though the same table can be read successfully with `lance.dataset(...)` and the same token.

Expected behavior: opening an existing root-level table for read should not attempt to create or update namespace/manifest metadata, and should not require bucket write permission.

Observed behavior: the `lancedb.connect(...).open_table(...)` path can initialize a directory namespace / manifest path and request a Hugging Face Xet write token.

A practical downstream example is `stable-worldmodel`, which calls LanceDB roughly like this:

```python
db = lancedb.connect(
    "hf://buckets/<org>/<bucket>/path/to/db",
    storage_options={
        "region": "us-east-1",
        "virtual_hosted_style_request": "true",
        "token": os.environ["HF_TOKEN"],
    },
)
table = db.open_table("table_name")
```

With a read-only HF token this can fail, while direct Lance reads work:

```python
import lance

ds = lance.dataset(
    "hf://buckets/<org>/<bucket>/path/to/db/table_name.lance",
    storage_options={
        "region": "us-east-1",
        "virtual_hosted_style_request": "true",
        "token": os.environ["HF_TOKEN"],
    },
)
ds.count_rows()
```

Use a Hugging Face bucket containing an existing `.lance` table and a token that can read the bucket but cannot write to it.

```python
import os
import lancedb
import lance

DB_URI = "hf://buckets/<org>/<bucket>/path/to/db"
TABLE = "table_name"
TABLE_URI = f"{DB_URI}/{TABLE}.lance"

storage_options = {
    "region": "us-east-1",
    "virtual_hosted_style_request": "true",
    "token": os.environ["HF_TOKEN"],
}

# This read path succeeds.
ds = lance.dataset(TABLE_URI, storage_options=storage_options)
print(ds.count_rows())

# This path can request .../xet-write-token and fail with 403 for read-only tokens.
db = lancedb.connect(DB_URI, storage_options=storage_options)
table = db.open_table(TABLE)
print(table.to_lance().count_rows())
```

Current workaround:

```python
db = lancedb.connect(
    namespace_client_impl="dir",
    namespace_client_properties={
        "root": DB_URI,
        "manifest_enabled": "false",
        "dir_listing_enabled": "true",
        **{f"storage.{k}": v for k, v in storage_options.items()},
    },
    storage_options=storage_options,
)
table = db.open_table(TABLE)
```

This is too complex for downstream libraries to ask users to configure.

Possible fix: ordinary listing connections should not eagerly initialize a manifest-enabled internal directory namespace for root-level table reads. Either initialize that namespace lazily only when namespace operations are actually used, or ensure the internal namespace used by listing connections has `manifest_enabled=false` unless the user explicitly requested manifest mode.

Xuanwo:
Bumps Lance to v9.0.0-beta.19, which includes lance-format/lance#7687 for side-effect-free DirectoryNamespace read paths.

This fixes root-level read-only table opens that previously could trigger `__manifest` creation through directory namespace construction, including Hugging Face bucket reads with read-only tokens. A LanceDB regression test now covers root listing operations without creating `__manifest`.

Fixes #3633.

### DIFF
--- a/python/python/tests/test_s3_bucket_dots.py
+++ b/python/python/tests/test_s3_bucket_dots.py
@@ -7,7 +7,8 @@ Tests for S3 bucket names containing dots.
 Related issue: https://github.com/lancedb/lancedb/issues/1898
 
 These tests validate the early error checking for S3 bucket names with dots.
-No actual S3 connection is made - validation happens before connection.
+When validation succeeds, eager namespace initialization may still fail later
+because these tests intentionally do not provide real S3 credentials.
 """
 
 import pytest
@@ -20,6 +21,13 @@ BUCKET_WITH_DOTS_AND_AWS_REGION = ("s3://my.bucket.name", {"aws_region": "us-eas
 BUCKET_WITHOUT_DOTS = "s3://my-bucket/path"
 
 
+def assert_not_rejected_for_bucket_dots(connect):
+    try:
+        connect()
+    except ValueError as err:
+        assert "contains dots" not in str(err)
+
+
 class TestS3BucketWithDotsSync:
     """Tests for connect()."""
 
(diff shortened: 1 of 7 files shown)


BubbleCal (approved):

python/python/tests/test_s3_bucket_dots.py
```diff
@@ -20,26 +21,36 @@
 BUCKET_WITHOUT_DOTS = "s3://my-bucket/path"
 
 
+def assert_not_rejected_for_bucket_dots(connect):
+    try:
+        connect()
+    except ValueError as err:
+        assert "contains dots" not in str(err)
```
wjones127, line 28: nitpick: this seems like it is a bit fragile. It might be better to make this an integration test where we use real credentials to connect to minio or localstack. We already have some CI tests that use those dependencies.

wjones127 (approved):

PR #3641, opened 2026-07-10
fix: purge stale session metadata cache on drop_table (#3626)

### LINKED ISSUE #3626: drop_table leaves stale session cache entries — same-URI recreate can panic with "row id missing from index"
## Summary

`Connection::drop_table` removes the table's data but does not purge the shared
`Session` caches (dataset metadata cache / index cache) for the dropped table's
URI. When a table is dropped and a new table is **re-created under the same
URI**, a subsequent mutation (DELETE/UPDATE with stable row ids) can resolve
stale cached metadata belonging to the dropped dataset and panic inside lance:

```
lance/src/dataset/utils.rs:129: row id missing from index
```

## Why the stale entry survives

The most damaging stale entry is the row-id sequence. Its cache key is
version-less (`row_id_sequence/{fragment_id}` — `RowIdSequenceKey` in lance), and
the only namespacing above it is the dataset **URI** prefix
(`GlobalMetadataCache::for_dataset(uri)`), which the comment there notes
"prevents collisions between different datasets". It does not prevent collisions
between two generations of the **same** URI. So after drop + same-URI re-create,
a lookup for `row_id_sequence/0` of the *new* dataset can hit the sequence of the
*old* (dropped) dataset still resident in the shared session, and the mismatch
against the new fragment's physical rows trips the `row id missing from index`
invariant. (Filed the version-less-key angle separately against lance.)

## Reproduction (embedding host, single process, shared session)

1. CREATE table `t` (stable row ids), insert rows, COMPACT, DROP TABLE `t`.
2. CREATE table `t` again at the same URI, insert rows.
3. First DELETE on the new `t` → panic. Deterministic when the stale entry is
   still resident; with concurrent readers priming the cache it reproduces
   within a few cycles.

Observed with lancedb 0.31 / lance 8.0.0.

## Fix that worked for us

- Have the connection own its `Session` explicitly (`ConnectBuilder::session()`),
  so `drop_table` can reach both cache layers deterministically.
- After the drop, purge both caches by URI prefix — using **two
  terminator-precise prefixes** (`{uri}/` and `{uri}?`) so sibling tables that
  share a name prefix (`t` vs `t2`, or a query-string variant) are not evicted.
- Run the purge unconditionally after the drop attempt (also on the timeout
  path), so a timed-out-but-still-progressing drop cannot leave repopulated
  stale entries behind.

After the fix: 30/30 drop → re-create → mutate cycles clean (previously panicked
by cycle 3).

Happy to upstream a PR if the approach sounds right — the ideal fix probably
combines this drop-time purge with a version/generation component in lance's
row-id sequence cache key.

PranavAchar01:
Closes #3626.

## Problem
`Connection::drop_table` (the local `ListingDatabase`) removes a table's files but never purges the shared `Session` caches. The metadata cache is namespaced only by the dataset **URI** prefix (`Session::metadata_cache` → `GlobalMetadataCache::for_dataset(uri)`), and the most damaging entry — the row-id sequence — is keyed version-less. So when a table is dropped and a new table is **re-created under the same URI**, a later mutation (DELETE/UPDATE with stable row ids) can resolve stale cached metadata from the dropped dataset and panic inside lance:

```
lance/src/dataset/utils.rs: row id missing from index
```

## Fix
`ListingDatabase::drop_tables` now invalidates the dropped dataset's cached metadata after removing its files:

```rust
if let Ok(table_uri) = self.table_uri(&name) {
    let uri_prefix = /* table_uri without any ?query suffix */;
    self.session
        .file_metadata_cache()
        .invalidate_prefix(uri_prefix)
        .await;
}
```

- Uses lance's documented public API: `Session::file_metadata_cache()` + `LanceCache::invalidate_prefix`.
- Reuses the same `table_uri(name)` that `create_table`/`open_table` pass to `DatasetBuilder::from_uri`, so the invalidation prefix matches the cache prefix exactly. The `.lance` suffix makes it a safe prefix boundary (won't touch sibling tables); any `?query` suffix is stripped since lance keys off the dataset path.

## Test
`test_drop_table_purges_session_metadata_cache` creates a table, reads it to populate the shared session metadata cache, asserts entries exist under the table's URI, drops it, and asserts the cache has **zero** entries under that URI afterwards. Fails on `main` (entries survive), passes with this change.

## Notes
- The index cache (`Session::index_cache`) is namespaced the same way but isn't exposed publicly for invalidation; the reported panic comes from the metadata cache (row-id sequence), which this addresses. Happy to extend to the index cache if lance exposes an accessor.

### DIFF
--- a/rust/lancedb/src/database/listing.rs
+++ b/rust/lancedb/src/database/listing.rs
@@ -739,6 +739,27 @@ impl ListingDatabase {
                     },
                     _ => Error::from(err),
                 })?;
+
+            // Removing the files leaves the shared session's metadata cache
+            // (keyed by the dataset URI prefix, see `Session::metadata_cache`
+            // and `GlobalMetadataCache::for_dataset`) holding metadata for the
+            // now-deleted dataset. If a table is later re-created under the same
+            // URI, a mutation can resolve a stale row-id sequence and panic
+            // inside lance ("row id missing from index"). Invalidate the dropped
+            // dataset's cached metadata so the session stays consistent.
+            // See https://github.com/lancedb/lancedb/issues/3626.
+            if let Ok(table_uri) = self.table_uri(&name) {
+                // Strip any query string; lance keys the cache off the dataset
+                // path, and the `.lance` suffix makes this a safe prefix.
+                let uri_prefix = match table_uri.split_once('?') {
+                    Some((base, _)) => base,
+                    None => table_uri.as_str(),
+                };
+                self.session
+                    .file_metadata_cache()
+                    .invalidate_prefix(uri_prefix)
+                    .await;
+            }
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


rust/lancedb/src/database/listing.rs
```diff
@@ -752,6 +739,27 @@ impl ListingDatabase {
...
+            if let Ok(table_uri) = self.table_uri(&name) {
+                // Strip any query string; lance keys the cache off the dataset
+                // path, and the `.lance` suffix makes this a safe prefix.
+                let uri_prefix = match table_uri.split_once('?') {
+                    Some((base, _)) => base,
+                    None => table_uri.as_str(),
+                };
+                self.session
+                    .file_metadata_cache()
+                    .invalidate_prefix(uri_prefix)
```
wjones127, line 760: issue(blocking): We are actually planning on deprecating and removing this method. We are working towards expanding the cache backends we support, and not all of them will be able to support this method. https://github.com/lance-format/lance/discussions/7575

Instead, our plan to address this issue will be to add a TTL option to the cache: https://github.com/lance-format/lance/issues/7490

wjones127 (changes requested):

wjones127:
I'm going to close this, as the mechanism is deprecated and we'll need an alternative fix.

PR #3652, opened 2026-07-13
fix(python): preserve zero distance bounds in hybrid search

### LINKED ISSUE #3651: bug(python): sync hybrid search drops zero-valued distance bounds
### LanceDB version

`0.35.0-beta.1` / current `main` (`a548e59d49dfbe2491ce217e937b88d736fc8931`)

### What happened?

The public synchronous hybrid query builder drops an explicitly supplied `0.0`
distance bound when the other endpoint is omitted.

For example, `.distance_range(upper_bound=0.0)` should forward the upper bound to
the vector child query. With L2 distance and the documented half-open range
`[lower_bound, upper_bound)`, no vector can satisfy an upper bound of `0.0`.
Instead, the range is not applied and vector-only rows survive the hybrid query.

The same parameter-transport problem affects `lower_bound=0.0` when no upper bound
is supplied. Non-zero bounds are unaffected. The async query builder follows a
different forwarding path and is not affected by this condition.

The sync hybrid child-builder setup currently uses a truthiness check for the two
optional floats, so a valid `0.0` value is treated the same as an omitted `None`.

Expected: the query below returns zero rows.

Actual: it returns four rows. Their FTS `_score` values are all null, confirming
that the rows come from the vector child after the distance range was dropped.

I am willing to work on a fix and submit a PR with this regression test.

The following is a self-contained black-box pytest case. It uses only public APIs
and creates a real local table, FTS index, hybrid query, reranker, and Arrow result.

```python
import lancedb
import pyarrow as pa
from lancedb.rerankers.rrf import RRFReranker


def test_sync_hybrid_preserves_zero_upper_distance_bound(tmp_path):
    db = lancedb.connect(str(tmp_path / "db"))
    data = pa.table(
        {
            "text": pa.array(["a", "b", "cat", "dog"]),
            "vector": pa.array(
                [[0.1, 0.1], [2.0, 2.0], [-0.1, -0.1], [0.5, -0.5]],
                type=pa.list_(pa.float32(), list_size=2),
            ),
        }
    )
    table = db.create_table("test", data)

    result = (
        .vector([0.0, 0.4])
        .text("elephant")
        .distance_range(upper_bound=0.0)
        .rerank(RRFReranker(return_score="all"))
        .limit(4)
        .to_arrow()
    )

    assert len(result) == 0
```

Run with:

```bash
pytest -q test_zero_hybrid_bound.py
```

Observed on current `main`, deterministically across two runs:

```text
E       assert 4 == 0

Returned row ids: [0, 2, 3, 1]
Returned _score values: [null, null, null, null]
```

u70b3:
## Summary

- preserve explicit `0.0` distance bounds in synchronous hybrid search
- distinguish omitted `None` endpoints from zero-valued endpoints when configuring the vector child query
- add a public end-to-end regression test for a zero upper bound

## Testing

- `cd python && uv run --extra tests pytest python/tests/test_hybrid_query.py -q`
- `uv run --project python ruff format --check python/python/lancedb/query.py python/python/tests/test_hybrid_query.py`
- `uv run --project python ruff check .`

Fixes #3651

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -2634,7 +2634,7 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
             self._vector_query.ef(self._ef)
         if self._bypass_vector_index:
             self._vector_query.bypass_vector_index()
-        if self._lower_bound or self._upper_bound:
+        if self._lower_bound is not None or self._upper_bound is not None:
             self._vector_query.distance_range(
                 lower_bound=self._lower_bound, upper_bound=self._upper_bound
             )
--- a/python/python/tests/test_hybrid_query.py
+++ b/python/python/tests/test_hybrid_query.py
@@ -139,6 +139,20 @@ def test_hybrid_query_distance_range(sync_table: Table):
             assert 0.2 <= dist.as_py() <= 0.5
 
 
+def test_hybrid_query_applies_zero_upper_distance_bound(sync_table: Table):
+    result = (
+        sync_table.search(query_type="hybrid")
+        .vector([0.0, 0.4])
+        .text("elephant")
+        .distance_range(upper_bound=0.0)
+        .rerank(RRFReranker(return_score="all"))
+        .limit(4)
+        .to_arrow()
+    )
+
+    assert len(result) == 0
+
+
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


wjones127 (approved):
Looks reasonable and a good minimal test. Thank you!

PR #3654, opened 2026-07-13
fix(python): preserve phrase semantics in sync queries

### LINKED ISSUE #3653: bug(python): phrase semantics are lost in sync query plans
### LanceDB version

Current `main` at `a548e59d49dfbe2491ce217e937b88d736fc8931`.

### What happened?

Sync FTS phrase semantics are applied only in `to_arrow()`.
`to_query_object()` serializes the original unquoted query, so output-schema,
explain-plan, analyze-plan, and hybrid child-plan paths can execute or describe
a normal term query instead of a phrase query.

`to_arrow()` also mutates the builder's `_query`, causing reranker input and
subsequent behavior to depend on call order.

In addition, sync hybrid `.phrase_query()` documents a default of `True`, but
its runtime default is `None`, so calling it without an argument does not enable
phrase mode.

### Reproduction

```python
from unittest.mock import Mock

from lancedb.query import LanceFtsQueryBuilder

query = (
    LanceFtsQueryBuilder(Mock(), "puppy runs")
    .phrase_query()
    .to_query_object()
)

assert query.full_text_query.query == '"puppy runs"'
```

The assertion currently receives `puppy runs`.

### Expected behavior

- Execution and query-plan paths serialize the same phrase query.
- Only the FTS backend receives query-language quoting.
- Rerankers continue to receive the original user text.
- Calling sync hybrid `.phrase_query()` without an argument enables phrase mode.

### Related work

#2742 reported the ordinary execution problem, which was fixed by #2781.
The query-object and plan paths remain inconsistent.

u70b3:
## Summary

- serialize sync phrase queries consistently for execution and query plans
- restore the documented no-argument hybrid `phrase_query()` behavior
- keep reranker input as the original user text without mutating the builder

Fixes #3653.

## Testing

- `python/.venv/bin/python -m pytest <8 focused test nodes> -q` (`8 passed`)
- `python/.venv/bin/python -m ruff format --check python/python/lancedb/query.py python/python/tests/test_fts.py python/python/tests/test_hybrid_query.py`
- `python/.venv/bin/python -m ruff check .`
- `git diff --check origin/main...HEAD`

The complete hybrid module and the real native FTS phrase test were not completed
in the current PyO3 runtime environment: both stalled in the native
`lancedb.connect()` fixture and were interrupted without an assertion failure.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -1951,8 +1951,7 @@ class LanceFtsQueryBuilder(LanceQueryBuilder):
         Parameters
         ----------
         phrase_query: bool, default True
-            If True, then the query will be wrapped in quotes and
-            double quotes replaced by single quotes.
+            If True, then an unquoted string query will be wrapped in quotes.
 
         Returns
         -------
@@ -1961,6 +1960,18 @@ class LanceFtsQueryBuilder(LanceQueryBuilder):
         """
         self._phrase_query = phrase_query
         return self
+
+    def _query_with_phrase_semantics(self) -> str | FullTextQuery:
+        query = self._query
+        if not self._phrase_query:
+            return query
+        if isinstance(query, str):
+            if not query.startswith('"') or not query.endswith('"'):
+                return f'"{query}"'
+            return query
+        if isinstance(query, PhraseQuery):
+            return query
+        raise TypeError("Please use PhraseQuery for phrase queries.")
 
     def fast_search(self) -> LanceFtsQueryBuilder:
         """
@@ -1986,7 +1997,7 @@ class LanceFtsQueryBuilder(LanceQueryBuilder):
             fragments=self._fragments,
             fragment_ids=self._fragment_ids,
             full_text_query=FullTextSearchQuery(
-                query=self._query, columns=self._fts_columns
+                query=self._query_with_phrase_semantics(), columns=self._fts_columns
             ),
             offset=self._offset,
             fast_search=self._fast_search,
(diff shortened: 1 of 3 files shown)


python/python/lancedb/query.py
```diff
@@ -1962,6 +1961,18 @@ def phrase_query(self, phrase_query: bool = True) -> LanceFtsQueryBuilder:
...
+        query = self._query
+        if not self._phrase_query:
+            return query
+        if isinstance(query, str):
+            if not query.startswith('"') or not query.endswith('"'):
+                return f'"{query}"'
+            return query
+        if isinstance(query, PhraseQuery):
+            return query
+        raise TypeError("Please use PhraseQuery for phrase queries.")
```
wjones127: note: this is an existing issue, but this comment seems unhelpful. I feel like it should at least say the class name it received instead.

python/python/lancedb/query.py
```diff
@@ -1986,7 +1997,7 @@ def to_query_object(self) -> Query:
             fragments=self._fragments,
             fragment_ids=self._fragment_ids,
             full_text_query=FullTextSearchQuery(
-                query=self._query, columns=self._fts_columns
+                query=self._query_with_phrase_semantics(), columns=self._fts_columns
```
wjones127, line 2003: praise: moving this into a helper function is the right move.

wjones127 (approved):

python/python/lancedb/query.py
u70b3: Fixed in 6ffa7882. The TypeError now reports both the accepted inputs and the received class (for example, `got MatchQuery`), and the existing regression test asserts the class name.

PR #3656, opened 2026-07-13
feat(python): add offset caching to StreamingDataset

westonpace:
Introduces a `CachingPermutationReader` that wraps `PermutationReader` and stores fetched rows in a local LanceDB table, keyed by `{table_name}_v{version}__{selection_hash}`.  Cache hits are served locally; misses are forwarded to the remote reader and written back.

A `SharedOffsetCache` (Arc<Mutex<…>>) is shared across all splits so that every split in an epoch can benefit from rows already fetched by another split.  Different column selections automatically get separate cache tables via a stable FNV-1a hash of the selection descriptor, so changing the selection never silently returns stale data.

`StreamingDataset` accepts an optional `cache_dir` parameter; when set it initialises a `SharedOffsetCache` once and wraps each split's `Permutation` with `with_cache()` before iteration begins.


rust/lancedb/src/dataloader/permutation/cache.rs
```diff
@@ -0,0 +1,400 @@
...
+            (Some(hit_batch), None) => {
+                // Drop the _remote_offset column that the cache table stores.
+                Ok(hit_batch.drop_column(REMOTE_OFFSET_COL)?)
+            }
+            (None, Some(miss_batch)) => Ok(miss_batch),
+            (Some(hit_batch), Some(miss_batch)) => {
+                let hit_batch = hit_batch.drop_column(REMOTE_OFFSET_COL)?;
+                let schema = miss_batch.schema();
+                Ok(concat_batches(&schema, &[hit_batch, miss_batch])?)
+            }
```
wjones127, line 362: **issue (blocking):** Requested order is lost here: hits and misses are concatenated in that order, so from epoch 2 on `take_offsets` returns rows in a different order than the caller asked for — note `orig_idx` is recorded at lines 260-267 and then never used. Acceptance: scatter both batches back by `orig_idx` (or build a take-index over the concatenation), plus a test that takes a mix of cached and uncached offsets and asserts the returned rows match the requested offset order exactly.

rust/lancedb/src/dataloader/permutation/cache.rs
wjones127, line 1: issue(blocking): this module, and the PR generally, has no tests. I'd like to see tests for:

1. all-miss and all-hit returns identical batches to the uncached reader impl
2. Partial hit/miss mix returns the correct order (bug I pointed out elsewhere)
3. Re-opening same `cache_dir` in a fresh cache instance still serves hits.
4. Two different selections lance in separate cached tables
5. A python test that iterates a dataset twice and asserts the second epoch is served by cache.

rust/lancedb/src/dataloader/permutation/cache.rs
```diff
@@ -0,0 +1,400 @@
...
+        })
+    }
+
+    /// Return a new reader with the given limit applied.
+    pub async fn with_limit(self, limit: u64) -> Result<Self> {
+        Ok(Self {
+            inner: self.inner.with_limit(limit).await?,
+            cache: self.cache,
+        })
+    }
```
wjones127, line 390: question(blocking): offset and limit aren't in the key `{table}_v{version}__{selection}`. Is that safe? Can we have a test showing that?

python/python/lancedb/streaming.py
```diff
@@ -175,6 +178,16 @@ def __init__(
...
         self._connection_factory = connection_factory
         self._worker_info_override = worker_info_override
+        self._cache_dir = cache_dir
+
+        # Initialise the shared offset cache (if requested).
+        if cache_dir is not None:
+            tbl_version = table.version
+            self._shared_cache = LOOP.run(
+                SharedOffsetCache.create(cache_dir, table.name, tbl_version)
+            )
```
wjones127, line 188: **question (blocking):** The cache and its `Connection` are built eagerly in `__init__` in the parent process, while this class otherwise defers connection setup to the worker via `connection_factory` — what is the intended behaviour with `num_workers > 0` (fork copies the mutex into unrelated processes; spawn has to pickle a PyO3 object)? Acceptance: either construct the cache lazily in `__iter__` per worker, or document and validate the supported worker configuration.

rust/lancedb/src/dataloader/permutation/cache.rs
```diff
@@ -0,0 +1,400 @@
...
+///
+/// * If the table does not exist → returns an empty [`CacheEntry`].
+/// * If the table is corrupted (missing `_remote_offset`) → drops and returns empty.
+/// * Otherwise → reconstructs the `offset_map` from the stored data.
+async fn init_cache_entry(conn: &Connection, table_name: &str) -> Result<CacheEntry> {
+    match conn.open_table(table_name).execute().await {
+        Err(_) => {
+            // Table does not exist yet.
+            Ok(CacheEntry::empty())
+        }
```
wjones127, line 141: **issue (non-blocking):** `Err(_)` treats every `open_table` failure as "not created yet" — a permissions or IO error becomes an empty cache, and then the `create_table` at line 325 fails confusingly on a name that already exists; match on `Error::TableNotFound` and propagate the rest.

rust/lancedb/src/dataloader/permutation/cache.rs
```diff
@@ -0,0 +1,400 @@
...
+    new_fields.push(Arc::new(remote_field));
+    let new_schema = Arc::new(arrow_schema::Schema::new_with_metadata(
+        new_fields,
+        batch.schema().metadata().clone(),
+    ));
+
+    let mut new_columns: Vec<Arc<dyn arrow_array::Array>> = batch.columns().to_vec();
+    new_columns.push(Arc::new(UInt64Array::from(offsets.to_vec())));
+
+    Ok(RecordBatch::try_new(new_schema, new_columns)?)
```
wjones127, line 211: suggestion: use `lance_arrow::RecordBatchExt::try_with_column()`

wjones127 (changes requested):
This seems to be missing some tests. Is that intentional?

PR #3646, opened 2026-07-11
fix(rust): skip embedding functions for empty batches

u70b3:
Fixes #3174
Also fixes #3645

Empty record batches now append correctly typed empty embedding arrays
without invoking embedding providers. This avoids OpenAI requests with
an invalid empty input while preserving source-column validation and
the non-empty execution paths.

As a small cleanup, the single- and multi-embedding code paths now share
a single upfront lookup of their source columns ("input_columns") instead
of each path looking them up independently. Also moves `lance-testing`
from regular dependencies to dev-dependencies where it belongs.

Tests run:
- `cargo fmt --all -- --check`
- `cargo test --quiet -p lancedb --lib empty_batch_skips_embedding_functions`
- `cargo test --quiet -p lancedb --lib empty_batch_still_validates_source_column`
- `cargo test --quiet -p lancedb --lib test_create_empty_table_with_embeddings`
- `cargo check --quiet -p lancedb --features remote --tests --examples`
- `cargo clippy --quiet -p lancedb --features remote --tests --examples`
- `cargo test --quiet -p lancedb --lib`
- `cargo test --quiet --features remote --tests`

PR #3661, opened 2026-07-14
chore(rust): make lance-testing a dev-dependency

### LINKED ISSUE #3645: Feature: lance-testing should be a dev-dependency
### SDK

Rust

### Description

`lance-testing` pulls in crates like `inferno` and is a direct dependency of `lancedb` even though `lance_testing` is only used in test modules.

saitejabandaru-in:
Resolves #3645.

### Description
Currently, `lance-testing` is declared as a direct dependency of the `lancedb` crate, pulling in testing-specific dependencies into release builds of `lancedb`.

Since `lance-testing` is only used within test modules in `connection.rs` and `query.rs`, it can be moved to `[dev-dependencies]` to reduce compilation time and dependency footprint for production builds.

### Changes
- Moved `lance-testing` from `[dependencies]` to `[dev-dependencies]` in `rust/lancedb/Cargo.toml`.

### DIFF
--- a/rust/lancedb/Cargo.toml
+++ b/rust/lancedb/Cargo.toml
@@ -93,8 +93,8 @@ tokenizers = { version = "0.19.1", optional = true }
 semver = { workspace = true }
 
 [dev-dependencies]
-anyhow = "1"
 lance-testing = { workspace = true }
+anyhow = "1"
 tempfile = "3.5.0"
 random_word = { version = "0.4.3", features = ["en"] }
 tokio = { version = "1.23", features = ["io-util", "macros", "net", "rt-multi-thread", "sync"] }


wjones127 (approved):

wjones127:
Thanks for the contribution, @saitejabandaru-in! Unfortunately this one is already taken care of: #3645 was resolved by #3646 (merged July 14), which moved `lance-testing` from `[dependencies]` to `[dev-dependencies]` as a small cleanup alongside its main fix.

On current `main`, `lance-testing` is already only in `[dev-dependencies]`, so the remaining change here is just a cosmetic reordering of two lines within that section. The CI failures you saw (a `goosefs-sdk` compile error in `build-no-lock` and a segfault in the Python `test_sql_query`) were unrelated pre-existing/flaky issues, not caused by this change.

Closing as already resolved — but thanks for spotting the issue and taking the time to send a fix. Hope to see more PRs from you!

PR #3662, opened 2026-07-14
docs: add LanceDB agent skill for portable pipelines

prrao87:
## What the new agent skill covers

We want to help users _easily_ write LanceDB pipelines to bring their data in from other places, no matter whether they use LanceDB OSS or Enterprise.

The `lancedb` set of skills contains guidance for agents on the following:
- Distinguishes local and remote table capabilities.
- Promotes bounded reads using `select()` and `limit()`.
- Prevents accidental full-table materialization.
- Documents correct Python sync/async scan APIs.
- Recommends validated Python schemas and batched ingestion.
- Provides indexing, query-tuning, diagnostics, and maintenance guidance.
- Documents the Enterprise table-name cache issue: avoid immediately reusing a dropped or overwritten table name; write to a fresh name and rename after propagation.
- Adds Python and TypeScript API, pattern, and performance references.
- Adds a heuristic scanner for potentially unsafe Python and TypeScript materialization patterns.

This change only adds agent documentation and tooling: no LanceDB runtime code, Rust code, SDK APIs, dependencies, or CI configuration are modified.

## Context

The LanceDB agent skill was accidentally pushed directly to `main` in `8ea78e3fbcb26718112ab4ddec55a91804b869d3`, bypassing the normal review workflow. That commit was reverted on `main` by `c12a6dce` so the protected branch is back to its prior content.

### DIFF
--- a/.agents/skills/lancedb/SKILL.md
+++ b/.agents/skills/lancedb/SKILL.md
@@ -1,0 +1,79 @@
+---
+name: lancedb
+description: Use when writing, reviewing, debugging, or documenting LanceDB pipelines in Python or TypeScript, especially code that should work across local LanceDB OSS tables and remote LanceDB Enterprise/Cloud tables. Helps avoid non-portable full-table materialization, choose idiomatic query/search patterns, and apply LanceDB performance defaults for ingestion, indexing, filtering, and diagnostics.
+---
+
+# Building LanceDB Pipelines
+
+Use this skill to produce LanceDB pipelines that are portable between local and remote tables (for LanceDB Enterprise/Cloud) and idiomatic for the selected SDK.
+
+## LanceDB Table Modes
+
+LanceDB has two common execution modes:
+
+- **Local table**: embedded, open source, in-process LanceDB. The client opens data from a local path or object storage URI and executes queries in the application process.
+- **Remote table**: LanceDB Enterprise/Cloud table opened through a `db://...` URI. The data may be very large, commonly backed by object storage, and queried through a remote service.
+
+Do NOT assume local-only table helpers exist on remote tables. If the user asks for LanceDB Enterprise, Cloud, `db://...`, production remote access, or a remote table, focus on the remote table path: use `search()` / `query()`, keep reads bounded with `select()` and `limit()`, and avoid table-level full materialization APIs.
+
+## Workflow
+
+1. Identify the SDK: Python, TypeScript, or both.
+2. Identify the table mode: local/embedded OSS, remote Enterprise/Cloud, or portable across both. If the user says "LanceDB Enterprise", choose the remote table path.
+3. Read the matching language branch before writing or changing code:
+   - Python patterns: `references/python/patterns.md`
+   - Python API quick reference: `references/python/api_reference.md`
+   - Python performance guidance: `references/python/performance.md`
+   - TypeScript patterns: `references/typescript/patterns.md`
+   - TypeScript API quick reference: `references/typescript/api_reference.md`
+   - TypeScript performance guidance: `references/typescript/performance.md`
+4. Start with `patterns.md` for the selected SDK. Read `api_reference.md` when choosing method names or return collectors. Read `performance.md` when the task involves ingestion, indexing, filtering, query tuning, diagnostics, or large datasets.
+5. For Python schemas, favor Pydantic models and validate records before writing. Use PyArrow schemas when Arrow-native, streaming, or highly dynamic data makes them materially better suited.
+6. Prefer `search()` or `query()` builders with explicit `select()` and `limit()` for reads.
+7. Avoid table-level full materialization in remote or portable code. This is the main local-vs-remote read pitfall.
+8. After a successful embedded OSS ingestion, call `table.optimize()`. Do not call it for Enterprise/Cloud; remote maintenance is automatic.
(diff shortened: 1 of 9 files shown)


prrao87:
@dantasse I tested this skill in combination with the existing ones, and it seems more robust than the earlier version. I ran it on a raw JSON file -> OSS LanceDB/Enterprise path, and it works with both Claude and GPT models. I'll run more experiments to improve robustness on specific parts in future PRs, but I think this should pair with your existing skills that cover more specialized ops.

.agents/skills/lancedb/references/python/patterns.md
```diff
@@ -0,0 +1,173 @@
...
+```python
+for row in rows:
+    table.add([row])  # one commit + fragment per row
+```
+
+Each `add()` creates a new version and fragment. Pass the whole batch in a single call, or chunk very large inputs:
+
+```python
+table.add(rows)  # single commit
+# for very large inputs, add batches of several thousand rows
```
dantasse, line 152: should we be any more explicit than this? I guess "several thousand rows" is probably detailed enough, right?

.agents/skills/lancedb/references/python/patterns.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+After the final successful write to an embedded OSS table, call
+`table.optimize()`. Skip this for Enterprise/Cloud tables because their
```
dantasse, line 156: oh yeah ok nevermind this will be fine then. (oh wait, huh, do we automatically optimize tables in enterprise?)

.agents/skills/lancedb/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Python: `table.search(...).select([...]).limit(10).to_pandas()`
+- TypeScript: `await table.search(...).select([...]).limit(10).toArray()`
+
+The unsafe pattern is table-level or unbounded collection, plus local-only dataset escape hatches in remote code:
+
+- Python: `table.to_pandas()`, `table.to_arrow()`, `table.to_polars()`; `table.to_lance()` is local/OSS-only dataset access, not materialization
+- TypeScript: `await table.toArrow()`, `await table.query().toArray()` without `limit()`
+
+## Enterprise: never drop-then-reuse the same table name
```
dantasse, line 53: hmm, this seems like a little bit of an edge case to be in the top level doc? should we split it into a lower level "troubleshooting" doc and here just say "don't drop-then-reuse the same table name; if you want to do so, read `no-drop-then-rename.md` for details about why"?

(or am I over-optimizing context here?)

also, is this going to be this way forever? can we like, fix this? 😅  like ... can we have drop_table poke the data plane cache? I'm sure we've thought about this but I'm just curious myself

dantasse (commented):
ok this generally looks cool! a couple questions:
- what about enterprise rest queries? there was that whole thread we had about "should we use python or rest", and I guess if there's no strong reasons either way (or say python is a little faster) we might as well just point to that bc it works for OSS too. but what about when we add some functionality to phalanx before adding it to python? (or other edge cases?) maybe we just try not to do this
- how do we fold in the pre-existing skills (e.g. table metadata, branches)? I assume we'd put some table-of-contents kind of thing: "if you're trying to edit table metadata, open table-metadata.md"

.agents/skills/lancedb/references/python/patterns.md
prrao87: Yes, I think this should work well. I notice for very small datasets, it doesn't over-engineer right now with multiple batching logic, parameters, etc.

.agents/skills/lancedb/references/python/patterns.md
prrao87: Yes, I tested on both OSS and Enterprise and it doesn't add the `table.optimize()` call for the Enterprise code. It only does so in OSS 😄.

.agents/skills/lancedb/SKILL.md
prrao87: I only added this in here because I had major trouble in *rerunning* a pipeline when the table already exists. 😅 
I hope it's not this way forever!

prrao87:
@dantasse here are my thoughts.

> what about enterprise rest queries? there was that whole thread we had about "should we use python or rest", and I guess if there's no strong reasons either way (or say python is a little faster) we might as well just point to that bc it works for OSS too. but what about when we add some functionality to phalanx before adding it to python? (or other edge cases?) maybe we just try not to do this

My take is that Python is easier for agents to write/reason over, because the API surface is right there in the `.venv` for them to look at. I think for this initial rollout and see if we can add REST API-specific skills at a later time. We could  potentially even plug it into our `openapi.yml` (or whatever that file's called) to inspect the latest API as a reference to find what it needs if the user explicitly instructs the agent to use the REST endpoints rather than Python.

> how do we fold in the pre-existing skills (e.g. table metadata, branches)? I assume we'd put some table-of-contents kind of thing: "if you're trying to edit table metadata, open table-metadata.md"

We don't need to do that as agents are smart enough to look inside every `*/SKILL.md` and only scan its frontmatter, and we let the agent figure out what it wants. So the frontmatter description should be really good.

dantasse (approved):
ok, sounds good! I think let's merge this and then we can iterate on it.

(as for adding the skills into this one, I just mean to do that so that we can avoid making people install more than one skill. but once this is merged I can try it out and see how it works both ways.)

prrao87:
Great! Fixed the broken YAML frontmatter, should be good now I think.

PR #3670, opened 2026-07-15
fix: avoid AttributeError in JinaEmbeddings image input for str/Path

chuenchen309:
## What

`JinaEmbeddings._generate_image_input_dict()` crashes with `AttributeError: 'function' object has no attribute 'urlparse'` on any image given as a URL string, local path string, or `pathlib.Path` — i.e. every documented `jina-clip-v1` image-embedding use case except raw `bytes`.

## Why

```python
from urllib.parse import urlparse
...
parsed = urlparse.urlparse(image)
```

`urlparse` is imported as a function, then called as if it were the `urllib.parse` module (`urlparse.urlparse(...)`). The module-level `is_valid_url()` a few lines above does it correctly (`urlparse(text)`), which is why this reads as a typo rather than intentional. Fixed to `urlparse(str(image))` — `str()` is needed because `urlparse()` only accepts `str`/`bytes` and raises a different `AttributeError` on a raw `Path`.

## Testing

Added `test_jina_generate_image_input_dict_local_path`, which fails with the original `AttributeError` before the fix and passes after, covering both a `str` path and a `pathlib.Path`. Verified locally (built the Rust extension, ran red→green, then the full `test_embeddings.py` file: 15 passed / 8 skipped, no regressions) and with `ruff check`/`ruff format`.

---
Disclosure: this PR was drafted with AI assistance (Claude); I reviewed, tested, and take responsibility for the change.

### DIFF
--- a/python/python/lancedb/embeddings/jinaai.py
+++ b/python/python/lancedb/embeddings/jinaai.py
@@ -87,7 +87,7 @@ class JinaEmbeddings(EmbeddingFunction):
         if isinstance(image, bytes):
             image_dict = {"image": base64.b64encode(image).decode("utf-8")}
         elif isinstance(image, (str, Path)):
-            parsed = urlparse.urlparse(image)
+            parsed = urlparse(str(image))
             # TODO handle drive letter on windows.
             PIL_Image = attempt_import_or_raise("PIL.Image", "pillow")
             if parsed.scheme == "file":
--- a/python/python/lancedb/embeddings/jinaai.py
+++ b/python/python/lancedb/embeddings/jinaai.py
@@ -88,11 +88,12 @@ class JinaEmbeddings(EmbeddingFunction):
             image_dict = {"image": base64.b64encode(image).decode("utf-8")}
         elif isinstance(image, (str, Path)):
             parsed = urlparse(str(image))
-            # TODO handle drive letter on windows.
             PIL_Image = attempt_import_or_raise("PIL.Image", "pillow")
             if parsed.scheme == "file":
                 pil_image = PIL_Image.open(parsed.path)
-            elif parsed.scheme == "":
+            elif parsed.scheme == "" or (os.name == "nt" and len(parsed.scheme) == 1):
+                # A Windows drive letter parses as a one-character scheme
+                # ("C:\\img.png" -> scheme="c"), so treat it as a local path.
                 pil_image = PIL_Image.open(image if os.name == "nt" else parsed.path)
             elif parsed.scheme.startswith("http"):
                 pil_image = PIL_Image.open(io.BytesIO(url_retrieve(image)))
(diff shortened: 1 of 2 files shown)


python/python/lancedb/embeddings/jinaai.py
```diff
@@ -87,7 +87,7 @@ def _generate_image_input_dict(image: Union[str, bytes, "PIL.Image.Image"]) -> D
         if isinstance(image, bytes):
             image_dict = {"image": base64.b64encode(image).decode("utf-8")}
         elif isinstance(image, (str, Path)):
-            parsed = urlparse.urlparse(image)
+            parsed = urlparse(str(image))
```
wjones127, line 90: issue(blocking): we should watch for whether this works on windows. Sometimes drive letters can confuse URL parsers.

wjones127 (changes requested):
Changes seem reasonable, but I think we will need a rebase so we can see if it is passing in Windows CI.

python/python/lancedb/embeddings/jinaai.py
chuenchen309: Rebased onto `main` (`b091d97d`) so Windows CI can run — and you were right to flag the drive letter, it was broken. I checked what `urlparse` actually returns:

```
'C:\\images\\a.png'    scheme='c'   path='\\images\\a.png'
'C:/images/a.png'      scheme='c'   path='/images/a.png'
'images\\a.png'        scheme=''    path='images\\a.png'
'\\\\server\\share\\a.png'  scheme=''    path='\\\\server\\share\\a.png'
```

A drive letter parses as a one-character scheme, so `C:\...` matched neither `""` nor `file` nor `http*` and fell through to `raise NotImplementedError("Only local and http(s) urls are supported")`. Relative paths and UNC paths were already fine — the existing `image if os.name == "nt"` branch covers those.

So this PR now has a second commit routing a one-character scheme to the local-path branch on Windows, which also retires the `# TODO handle drive letter on windows.` right above it:

```python
elif parsed.scheme == "" or (os.name == "nt" and len(parsed.scheme) == 1):
    # A Windows drive letter parses as a one-character scheme
    # ("C:\\img.png" -> scheme="c"), so treat it as a local path.
    pil_image = PIL_Image.open(image if os.name == "nt" else parsed.path)
```

I kept the `os.name == "nt"` guard deliberately: on POSIX, `c:foo` is a legal filename, and treating a one-character scheme as a path there would silently open `foo` instead of erroring.

Worth noting this also means the existing test in this PR would have failed on Windows before the second commit — `tmp_path` is absolute, so it carries a drive letter. That's the case your CI should now exercise.

I could not verify the Windows branch locally (Linux only); the `urlparse` results above are platform-independent, but the `os.name == "nt"` path itself is only reachable on Windows, so the CI run is the real check.

`ruff format --check` and `ruff check` are clean on both files. In `test_embeddings.py`, `test_jina_generate_image_input_dict_local_path` passes; five unrelated tests in that file fail in my checkout with `Table.add() got an unexpected keyword argument 'write_parallelism'`, which is a stale compiled `_lancedb` in my environment rather than anything in this diff — they fail identically without it.

AI-assisted with agentic coding tools; all changes were reviewed and tested by me.

wjones127 (approved):

PR #3675, opened 2026-07-15
feat: support distributed analyze plan metrics in clients

jackye1995:
Adds client-side support for analyze_plan distributed metrics modes across Rust, Python, and TypeScript clients. Defaults to aggregate for backward compatibility and sends the remote distributed_metrics parameter only when a non-default mode is requested.


docs/src/js/classes/Query.md
```diff
@@ -33,14 +33,20 @@ protected inner: Query | Promise<Query>;
 ### analyzePlan()
 
 ```ts
-analyzePlan(): Promise<string>
+analyzePlan(distributedMetrics): Promise<string>
```
LuQQiu: Does it break the customers api? Can they not provide this? Make it option?

LuQQiu (approved):

nodejs/__test__/table.test.ts
```diff
@@ -2774,7 +2774,7 @@ describe("when calling analyzePlan", () => {
     queryVec = Array(2)
       .fill(1)
       .map(() => Math.random());
-    const plan = await table.query().nearestTo(queryVec).analyzePlan();
+    const plan = await table.query().nearestTo(queryVec).analyzePlan("full");
```
LuQQiu: We cannot keep old way while supporting this new parameter?

docs/src/js/classes/Query.md
jackye1995: Updated the TypeScript signature to make `distributedMetrics` optional and regenerated the docs, so `analyzePlan()` remains the documented/default call and `analyzePlan("full")` is optional.

nodejs/__test__/table.test.ts
jackye1995: Updated the Node test to keep coverage for the old `analyzePlan()` call and add separate coverage for `analyzePlan("full")`.

PR #3676, opened 2026-07-16
ci(python): add win_arm64 wheel build

vortex-captain:
## Summary

Adds a Windows/ARM64 (`aarch64-pc-windows-msvc`) Python wheel build to CI and the
release workflow, folding the new target into the existing `windows` job as a
matrix leg (`x86_64` + `arm64`). The wheel builds as `cp39-abi3`, same as the
existing `win_amd64` build. Verified end-to-end on GitHub's `windows-11-arm`
runner.

## Why tests are skipped on the arm64 leg

Both legs build the wheel, but tests run on `x86_64` only. Several test
dependencies have **no `win_arm64` wheel on PyPI**, so the `[tests]` install
cannot resolve on Windows/ARM64:

| Dependency | Note |
| --- | --- |
| **pyarrow** | core runtime dep — `import lancedb` fails without it, so the whole suite cannot run |
| **pylance** | required by a subset of tests |
| **datafusion** | required by a subset of tests |
| **polars** | the `tests` extra pins `polars<=1.3.0`; arm64 wheels exist only from `1.18.0` |

For the same reason the arm64 wheel is built but **not published** yet: it is
uploaded under an `unpublished-` artifact name so the release `publish` job
(which downloads `wheels-*`) does not ship an uninstallable package.

## Validation (native `aarch64-pc-windows-msvc`)

- **CI (this PR):** the `python.yml` arm64 leg and the release build both succeed
  and produce `lancedb-0.35.0b2-cp39-abi3-win_arm64.whl`. The Linux/Mac/x86 test
  jobs pass.
- **Local run of the CI-built wheel** — the `unpublished-wheels-windows-arm64`
  artifact downloaded from this PR's release build (with a prebuilt `pyarrow`
  win_arm64 wheel + newer `polars`; `pylance`/`datafusion` tests skipped via
  `importorskip`): `pytest -m "not slow and not s3_test"` → **797 passed, 86
  skipped**, no wheel-attributable failures.
- **Rust suite** (`cargo test --features remote --tests -p lancedb`): **709 passed**.

## Future steps

Once `pyarrow` (and ideally `pylance`/`datafusion`) publish `win_arm64` wheels,
two `TODO(win_arm64)` markers point to the follow-ups:

1. **`python.yml`** — drop the `matrix.config.platform == 'x86_64'` guard on
   `run_tests` so the suite also runs on the arm64 leg.
2. **`pypi-publish.yml`** — rename the arm64 artifact from
   `unpublished-wheels-windows-arm64` to `wheels-windows-arm64` so it is included
   in releases.

from Microsoft

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -88,12 +88,32 @@ jobs:
           path: target/wheels/lancedb-*.whl
           if-no-files-found: error
   windows:
+    name: "Python windows-${{ matrix.config.platform }}"
     timeout-minutes: 90
-    runs-on: windows-latest
+    runs-on: ${{ matrix.config.runner }}
+    strategy:
+      matrix:
+        config:
+          - platform: x86_64
+            runner: windows-latest
+            # Published: the publish job downloads artifacts matching `wheels-*`.
+            artifact: wheels-windows-x86_64
+          - platform: arm64
+            runner: windows-11-arm
+            # TODO(win_arm64): the `unpublished-` prefix intentionally keeps this
+            # wheel OUT of the release. The publish job downloads `wheels-*`, so
+            # `unpublished-*` is not picked up. The wheel is still uploaded (so it
+            # can be downloaded for local validation), it just isn't published,
+            # because pyarrow (a core runtime dependency) has no win_arm64 wheel
+            # on PyPI yet, so `pip install lancedb` cannot resolve pyarrow on
+            # win_arm64. Once pyarrow ships win_arm64 wheels, rename this to
+            # `wheels-windows-arm64` to include it in the release.
+            artifact: unpublished-wheels-windows-arm64
     env:
       # link.exe is single-threaded and the long pole on Windows builds. Use
       # rustc's bundled lld-link instead.
       CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_LINKER: rust-lld
+      CARGO_TARGET_AARCH64_PC_WINDOWS_MSVC_LINKER: rust-lld
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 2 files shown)


vortex-captain:
Hi @wjones127, could you please take a look at this PR when you have a chance? Thanks!

Psychlist1972:
Hi @wjones127

Sorry for the direct tag. We're new to the repo here, and working to get this ready for our arm64 partners who need lancedb. Is there anything else we should do to help move the PR forward?

Thanks!

Pete
Microsoft

Xuanwo:
Hi, @vortex-captain and @Psychlist1972, thank you for working on this. However, as stated in the PR, what is the expectation here? Merging a non-working PR seems pointless. Perhaps we can continue after Arrow has a win-arm64 release first.

vortex-captain:
> Hi, @vortex-captain and @Psychlist1972, thank you for working on this. However, as stated in the PR, what is the expectation here? Merging a non-working PR seems pointless. Perhaps we can continue after Arrow has a win-arm64 release first.

Hi @Xuanwo, thanks for the feedback! Our initial thought was to first get the Python win_arm64 CI passing and producing unpublished wheels. That would allow us to test locally using private PyArrow wheels, giving us an opportunity to identify and address issues and improve stability before the official PyArrow win_arm64 release. Would that approach make sense?

wjones127:
> > Hi, @vortex-captain and @Psychlist1972, thank you for working on this. However, as stated in the PR, what is the expectation here? Merging a non-working PR seems pointless. Perhaps we can continue after Arrow has a win-arm64 release first.
> 
> Hi @Xuanwo, thanks for the feedback! Our initial thought was to first get the Python win_arm64 CI passing and producing unpublished wheels. That would allow us to test locally using private PyArrow wheels, giving us an opportunity to identify and address issues and improve stability before the official PyArrow win_arm64 release. Would that approach make sense?

TBH it shouldn't be that much trouble to build the LanceDB wheels locally on win_arm64. I don't think you need our repo to do that for you. I'm inclined to agree that we shouldn't be building these until PyArrow has the necessary wheels.

I'm going to close this PR, and suggest you continue with locally built wheels for testing.

PR #3684, opened 2026-07-17
feat(remote): add RemoteTable fetch_blobs HTTP client

geruh:
Remote half of the blob read path. #3578 did local Python. This makes `RemoteTable` hit the server.

- `fetch_blobs(column, row_ids or hits)` → bytes over `POST /v1/table/{id}/fetch_blobs/`
- `blob_columns()` from the cached schema (describe already has the metadata, no extra route)
- search then `fetch_blobs` works. row identity rides inside the blob descriptor so you do not need a public `_rowid`
- `fetch_blob_files` still `NotSupported` on remote. use `fetch_blobs` for full bytes for now. Range is a follow up

Accepts Binary / LargeBinary / BinaryView on the way back. Empty `row_ids` short-circuits. Version + branch go in the request body same as other read calls.

### Example

```python
db = lancedb.connect(uri="db://my-project", api_key=...)
table = db.open_table("clips")

hits = table.search(query_vec).select(["id", "video"]).limit(10).to_arrow()
# hits is just id + video. row ids are stashed on the descriptor
blobs = table.fetch_blobs("video", hits)  # null-aligned, same length as hits
```

Or pass ids yourself:

```python
blobs = table.fetch_blobs("video", [10, 20, 30])
```

### Testing

- `cargo test -p lancedb --features remote --lib`
- `cargo test -p lancedb --features remote --test blob_integration`
- `pytest python/tests/test_remote_db.py -k remote_blob`
- live e2e against a local 0.5.0 remote server (search → fetch, nulls, nested path, old server gate)

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -52,7 +52,6 @@ from ._blob import (
     finalize_blob_query_table,
     replace_v2_blob_columns_with_bytes,
     replace_v2_blob_columns_with_bytes_sync,
-    supports_blob_auto_row_id,
     validate_blob_mode,
 )
 from .types import BlobMode, QueryProjection
@@ -1277,10 +1276,7 @@ class LanceQueryBuilder(ABC):
         return self._with_row_id is True
 
     def _blob_auto_row_id_enabled(self) -> bool:
-        if not supports_blob_auto_row_id(self._table):
-            return False
         return blob_auto_row_id_for_scan(
-            self._table,
             self._table.schema,
             self._columns,
             with_row_id=self._with_row_id,
@@ -2771,7 +2767,7 @@ class AsyncQueryBase(object):
         )
 
     async def _maybe_add_blob_row_id(self) -> None:
-        if self._table is None or not supports_blob_auto_row_id(self._table):
+        if self._table is None:
             self._blob_auto_row_id = False
             self._blob_paths = ()
             return
@@ -2779,7 +2775,6 @@ class AsyncQueryBase(object):
         req = self._inner.to_query_request()
         schema = await self._table.schema()
         self._blob_auto_row_id = blob_auto_row_id_for_scan(
-            self._table,
             schema,
             req.select,
             with_row_id=self._with_row_id,
@@ -3031,7 +3026,6 @@ class AsyncQueryBase(object):
 
         schema = await self._table.schema()
         blob_auto_row_id = blob_auto_row_id_for_scan(
-            self._table,
             schema,
             query.columns,
             with_row_id=self._with_row_id,
@@ -3875,10 +3869,9 @@ class AsyncHybridQuery(AsyncStandardQuery, AsyncVectorQueryBase):
         req = fts_query._inner.to_query_request()
         blob_auto_row_id = False
         blob_paths: tuple[str, ...] = ()
-        if self._table is not None and supports_blob_auto_row_id(self._table):
+        if self._table is not None:
             schema = await self._table.schema()
             blob_auto_row_id = blob_auto_row_id_for_scan(
-                self._table,
                 schema,
                 req.select,
                 with_row_id=self._with_row_id,
--- a/python/python/lancedb/remote/table.py
+++ b/python/python/lancedb/remote/table.py
@@ -20,6 +20,7 @@ from typing import (
 import warnings
 
 from lancedb import __version__
+from lancedb._blob import BlobFile
 
 from lancedb._lancedb import (
     AddColumnsResult,
@@ -1037,22 +1038,22 @@ class RemoteTable(Table):
         )
 
     def blob_columns(self) -> list[str]:
-        raise NotImplementedError(
-            "blob_columns() is not yet supported on the LanceDB Cloud"
-        )
+        return LOOP.run(self._table.blob_columns())
 
-    def fetch_blobs(self, column: str, row_ids) -> pa.LargeBinaryArray:
-        raise NotImplementedError("fetch_blobs() is not supported on LanceDB Cloud")
+    def fetch_blobs(
+        self, column: str, row_ids: Union[list[int], pa.Table]
+    ) -> pa.LargeBinaryArray:
+        return LOOP.run(self._table.fetch_blobs(column, row_ids))
 
     def fetch_blob_ranges(self, column: str, requests) -> pa.LargeBinaryArray:
         raise NotImplementedError(
             "fetch_blob_ranges() is not supported on LanceDB Cloud"
         )
 
-    def fetch_blob_files(self, column: str, row_ids):
-        raise NotImplementedError(
-            "fetch_blob_files() is not supported on LanceDB Cloud"
-        )
+    def fetch_blob_files(
+        self, column: str, row_ids: Union[list[int], pa.Table]
+    ) -> "list[Optional[BlobFile]]":
+        return LOOP.run(self._table.fetch_blob_files(column, row_ids))
 
     def head(self, n=5) -> pa.Table:
         """
--- a/python/python/tests/test_remote_db.py
+++ b/python/python/tests/test_remote_db.py
@@ -1020,7 +1020,7 @@ def query_test_table(query_handler, *, server_version=Version("0.1.0")):
             request.send_header("Content-Type", "application/json")
             request.send_header("phalanx-version", str(server_version))
             request.end_headers()
-            request.wfile.write(b"{}")
+            request.wfile.write(b'{"version": 1, "schema": {"fields": []}}')
         elif request.path == "/v1/table/test/query/":
             content_len = int(request.headers.get("Content-Length"))
             body = request.rfile.read(content_len)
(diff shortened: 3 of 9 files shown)


geruh:
@claude review

python/python/lancedb/remote/table.py
```diff
@@ -231,6 +232,10 @@ def to_pandas(self, blob_mode: BlobMode = "lazy", **kwargs):
         """to_pandas() is not yet supported on LanceDB cloud."""
         raise NotImplementedError("to_pandas() is not yet supported on LanceDB cloud.")
 
+    def to_lance(self, **kwargs):
+        """to_lance() is not supported on LanceDB cloud tables."""
+        raise NotImplementedError("to_lance() is not supported on LanceDB cloud tables")
+
```
wjones127: question: why is this added? It seems like an unrelated change.

python/python/lancedb/query.py
```diff
@@ -108,6 +108,16 @@ def _unsupported_blob_pandas_error(reason: str) -> RuntimeError:
...
 
 
+_NATIVE_PANDAS_UNAVAILABLE = (
+    "this query shape cannot use Lance native pandas conversion"
+)
+
+
+def _native_pandas_failure_reason(native_error: Optional[Exception]) -> str:
+    reason = "" if native_error is None else str(native_error)
+    return reason or _NATIVE_PANDAS_UNAVAILABLE
```
wjones127: nitpick: same here, this seems like a totally unrelated drive-by change.

python/python/tests/test_remote_db.py
```diff
@@ -1792,3 +1792,106 @@ def handler(request):
...
+        elif request.path == "/v1/table/test/fetch_blobs/":
+            content_len = int(request.headers.get("Content-Length", 0))
+            body = json.loads(request.rfile.read(content_len))
+            assert body["column"] == "image"
+            assert body["row_ids"] == [10, 20, 30]
+            data = pa.table(
+                {"image": pa.array([b"alpha", None, b"gamma"], type=pa.large_binary())}
+            )
+            request.send_response(200)
+            request.send_header("Content-Type", "application/vnd.apache.arrow.file")
```
wjones127: question: what made you choose `file` over `stream` content type?

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use crate::Error;
+use crate::blob::BlobFile;
+use crate::error::Result;
+use crate::remote::client::{HttpSend, RequestResultExt};
+
+use super::RemoteTable;
+
+impl<S: HttpSend> RemoteTable<S> {
+    pub(super) async fn blob_columns_impl(&self) -> Result<Vec<String>> {
```
wjones127, line 23: question: why is this a separate endpoint from the describe, which provides the schema? It seems like this information should already be encoded on the schema.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -0,0 +1,126 @@
...
+        let mut stream = self.read_arrow_stream(&request_id, response).await?;
+
+        let mut chunks: Vec<Arc<dyn Array>> = Vec::new();
+        while let Some(batch) = stream.try_next().await? {
+            let blob_col = batch.column_by_name(column).ok_or_else(|| Error::Http {
+                source: format!("fetch_blobs response is missing the '{column}' column").into(),
+                request_id: request_id.clone(),
+                status_code: None,
+            })?;
+            if !matches!(blob_col.data_type(), DataType::LargeBinary) {
```
wjones127: suggestion: even if we always return LargeBinary now, it might be prudent to support all of the binary types (Binary, LargeBinary, BinaryView). That way we can change the serve implementation without breaking older clients.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -0,0 +1,126 @@
...
+            let blobs = arrow::compute::concat(&chunk_refs)?;
+            // Each chunk was checked as LargeBinary above.
+            blobs
+                .as_any()
+                .downcast_ref::<LargeBinaryArray>()
+                .expect("concat of LargeBinary columns yields LargeBinary")
+                .clone()
+        };
+        // Same length/order contract as local fetch_blobs.
+        if blobs.len() != row_ids.len() {
```
wjones127: question: Xuanwo was saying blob APIs currently don't have this contract when there are nulls? Does that apply here or is that at a different level?

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -0,0 +1,126 @@
...
+                    request_id: request_id.clone(),
+                    status_code: None,
+                });
+            }
+            chunks.push(blob_col.clone());
+        }
+        let blobs = if chunks.is_empty() {
+            LargeBinaryArray::from(Vec::<Option<&[u8]>>::new())
+        } else {
+            let chunk_refs: Vec<&dyn Array> = chunks.iter().map(AsRef::as_ref).collect();
```
wjones127: suggestion: maybe we could return `Vec<ArrayRef>` and leave it up to the caller to decide whether they want to concatenate? In PyArrow, this could be converted to `ChunkedArray`, which is what a column of `Table` is.

wjones127 (changes requested):
There seem to be some unrelated changes here, those don't seem like they belong here.

Other feedback I'd like to discuss before approving, though solving them aren't hard blockers.

python/python/lancedb/remote/table.py
geruh: yeah you're right I removed

python/python/lancedb/query.py
geruh: Yep, same thing here. Both changes are out now.

python/python/tests/test_remote_db.py
geruh: Yeah mostly because that helper does not actually stream today. It buffers the whole response and reads it as an IPC file, same as /query/.

I agree stream is the better format, but I think that should be fixed once for all the remote routes instead of only here.

Let me know what you think.

rust/lancedb/src/remote/table/blobs.rs
geruh: Yeah I think you're right I kind of went off on a tangent here. 

The blob marker is already in the schema from describe, and `RemoteTable` already has that schema cached.

rust/lancedb/src/remote/table/blobs.rs
geruh: Added this ty

rust/lancedb/src/remote/table/blobs.rs
geruh: In Lance they were skipping null rows, but LanceDB now we restore them before returning the result. Remote uses the same helper, and the client checks the returned length too.

But we were accounting for an empty blob bug that was fixed in https://github.com/lance-format/lance/pull/7903.

It's worth noting that while testing the bump it's breaking take_blobs since we return a `vec<option<BlobFIle>` now and `ReadBlob::data` becomes `Option<Bytes>`, which is the contract blob_null_mask hand-rolls in blob.rs today. The bump should let us delete that pass, which is what the TODO(lance): aligned read API would drop this pass comment there is waiting for.

I'll do that in a followup.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -0,0 +1,126 @@
...
+            .post_read(&format!("/v1/table/{}/blobs/columns/", self.identifier))
+            .json(&body);
+        let (request_id, response) = self.send(request, true).await?;
+        let response = self.check_table_response(&request_id, response).await?;
+        let body = response.text().await.err_to_http(request_id.clone())?;
+        serde_json::from_str::<Vec<String>>(&body).map_err(|e| Error::Http {
+            source: format!("Failed to parse blob_columns response: {e}").into(),
+            request_id,
+            status_code: None,
+        })
```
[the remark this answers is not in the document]
geruh: blob_columns is schema based now

rust/lancedb/src/remote/table/blobs.rs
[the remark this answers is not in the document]
geruh: yeah ts will be done in a followup left ts out to filter out the heavy logic for review.

rust/lancedb/src/remote/table/blobs.rs
geruh: I left it out of this pr because the `BaseTable::fetch_blobs` is `LargeBinaryArray` for local too from #3578, so flipping it is a rust and python api change. 

server only sends one batch rn anyway so the concat is basically a no-op. happy to revisit if you think `ChunkedArray` is worth a follow up

python/python/tests/test_remote_db.py
wjones127: > Yeah mostly because that helper does not actually stream today.

That's a separate issue. I think we're talking about two different things here:

1. Whether the API returns `Content-Type` `application/vnd.apache.arrow.file` or `application/vnd.apache.arrow.stream`. Plus, correspondingly, whether we serialize in [the file format](https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format) or [the stream format](https://arrow.apache.org/docs/format/Columnar.html#ipc-streaming-format). Those are slightly different on the wire, and the stream is slightly smaller and more appropriate here (because random access isn't of interest).
2. Whether the client consumes the content in a streaming fashion or buffers the whole thing.

You are talking about (2); and you're right: we can change that at any time.

But I'm talking about (1). This is baked into the server behavior. If you starting returning `file`, we have to support that in the client going forward to maintain client-server compatability.

rust/lancedb/src/remote/table/blobs.rs
wjones127: Understood. Just know I cringe a little whenever we concatenate arrays. It can be a real performance killer in queries when we are spending a lot of time doing memcpy for concat.

wjones127 (changes requested):
Are the server changes merged yet? I didn't find them. If they aren't, I'd still push to use `stream` over `file` for IPC format.

python/python/tests/test_remote_db.py
geruh: Yeah you;re right that does seem like the better shape i was comparing to my test setup

wjones127 (approved):
Thanks. Looks good now, once you get the merge conflicts addressed.

PR #3691, opened 2026-07-20
feat(fts): add block size configuration

BubbleCal:
## What changed

- add `block_size` to Python FTS configuration and the deprecated local/remote helpers
- add `blockSize` to the TypeScript FTS options and propagate it through the NAPI binding
- serialize the value as `block_size` for remote index creation
- document the existing Rust builder API and generate the TypeScript API reference
- add local, remote, metadata, search, and invalid-value regression coverage

## Why

Lance supports configuring the number of documents per compressed FTS posting block, but LanceDB's Python and TypeScript APIs did not expose the setting. This made the experimental FTS V3 layout unavailable through those clients and allowed the value to be dropped before index creation.

## How it works

The default remains `128`. Supported values are `128` and `256`; selecting `256` uses the experimental FTS V3 format. Invalid values are rejected by the Lance builder and surfaced as Python or JavaScript errors.

## Validation

- `cargo check --quiet --features remote --tests --examples`
- `cargo +1.94.0 clippy --quiet --features remote --tests --examples -- -D warnings`
- targeted Rust local and remote index tests
- Rust doctests: 34 passed
- Python Ruff checks, doctest, and targeted local/remote tests: 5 passed
- TypeScript build, Biome lint, generated docs, and targeted Jest tests: 9 passed
- `git diff --check`

## Limitations

The Java client remains unchanged because its external remote REST model does not currently expose `block_size`.

### DIFF
--- a/docs/src/js/interfaces/FtsOptions.md
+++ b/docs/src/js/interfaces/FtsOptions.md
@@ -43,6 +43,19 @@ The following tokenizers are available:
 
 ***
 
+### blockSize?
+
+```ts
+optional blockSize: 128 | 256;
+```
+
+Number of documents per compressed posting block.
+
+The default is 128. Supported values are 128 and 256. A value of 256 uses
+the experimental FTS V3 format and may introduce breaking changes.
+
+***
+
 ### language?
 
 ```ts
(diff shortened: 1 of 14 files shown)


jackye1995 (approved):
looks good to me

PR #3703, opened 2026-07-22
feat: support batched blob range reads

Xuanwo:
## Summary

Lance can now plan multiple byte ranges for the same blob in one `read_blob_ranges` operation, but LanceDB users currently cannot expose a complete set of logical ranges to that planner.

This complements `BlobFile`: file-like consumers such as PyAV can continue to discover ranges dynamically, while callers that already know the ranges for a batch can submit them together.

## Motivating example

A training table may store a large video blob together with a small application-level clip index:

```text
video: blob
clips: [{offset, length}, ...]
```

The caller can select the videos and clips for a batch, obtain their row IDs from the query, and read all of the selected windows together:

```python
rows = (
    table.search()
    .select(["clips"])
    .with_row_id(True)
    .limit(64)
    .to_arrow()
    .to_pylist()
)

requests = []
for row in rows:
    clip = sample_clip(row["clips"])
    requests.append(
        (row["_rowid"], clip["offset"], clip["length"])
    )

chunks = table.fetch_blob_ranges("video", requests)
```

Here, `_rowid` comes from the LanceDB query, while `offset` and `length` come from the application's clip index and are relative to that row's video blob. The caller describes only the logical reads; Lance still handles validation, source grouping, coalescing, scheduling, and byte backpressure.

Lance v10.0.0-beta.5 returns one logical result per blob selector or range request and explicitly distinguishes null blobs from valid empty values. LanceDB consumes that aligned result contract directly and only adds a cardinality check for unresolved row IDs.

This PR exposes batched blob-range reads on local Rust and Python tables. Results preserve request identity, duplicates, null slots, and valid empty ranges while allowing Lance to execute the physical reads out of order. Scheduler buffer sizing remains an internal Lance concern, so the LanceDB API does not expose `io_buffer_size`.

Cloud tables continue to report this operation as unsupported until there is a corresponding remote API.


wjones127:
It's not clear to me how users would use this API. BlobFile is the way it is because users pass it into things that expect file handles. The main example being things like this demo:

https://lance.org/guide/blob/#example-decode-video-frames-lazily

```python
blob = ds.take_blobs("video", indices=[0])[0]
if blob is None:
    raise ValueError("video blob is null")

start_ms, end_ms = 500, 1000

with av.open(blob) as container:
    stream = container.streams.video[0]
```

I feel like this API means the user now has to create an adapter that takes those handles and coalesces requests into batches that can be passed to `read_blob_ranges`. **But wouldn't it be better if we did that coalescing ourselves in our own IO scheduler?** It seems like it would save the user a lot of trouble.

If I'm wrong about the use place, please correct me. Maybe there's some reason a user knows exactly which ranges they want from each blob.

Xuanwo:
Thanks, that's a fair question. I think the missing context is that this API isn't intended to replace the PyAV-style `BlobFile` workflow.

cc @AyushExel for the workload you want in training.

The motivating workload is a training loader that already knows all the video windows needed for the next batch. For example, each row could contain a video blob and a small clip index:

```text
video: blob
clips: [{offset, length}, ...]
```

The caller could then do:

```python
rows = (
    table.search()
    .select(["clips"])
    .with_row_id(True)
    .limit(64)
    .to_arrow()
    .to_pylist()
)

requests = []
for row in rows:
    clip = sample_clip(row["clips"])
    requests.append(
        (row["_rowid"], clip["offset"], clip["length"])
    )

chunks = table.fetch_blob_ranges("video", requests)
```

Here, the row ID comes from the LanceDB query, while the offset and length come from the application's clip index and are relative to that video blob. The caller only describes the logical reads; Lance still handles grouping, coalescing, and scheduling.

I agree that `BlobFile` is still the better abstraction when something like PyAV discovers ranges dynamically. These two APIs are meant for different, complementary access patterns. I've updated the PR description to make that clearer.

rust/lancedb/src/blob.rs
```diff
@@ -145,91 +169,57 @@ pub(crate) fn ensure_blob_v2_column(
...
-    for segment in &path[1..] {
-        current = current
-            .column_by_name(segment)
-            .and_then(|c| c.as_any().downcast_ref::<StructArray>())
-            .ok_or_else(not_struct)?;
+fn ensure_all_row_ids_resolved(column: &str, requested: usize, resolved: usize) -> Result<()> {
+    if requested == resolved {
+        return Ok(());
+    }
+    if resolved < requested {
```
wjones127, line 176: **praise:** Good simplification — collapsing the null-mask/non-null-take dance into a single aligned take plus a shared under/over-resolved check reads much cleaner than the old two-pass approach.

wjones127 (approved):
Looks good!

PR #3711, opened 2026-07-24
fix(remote): surface masked merge_insert stream errors under HTTP2 (#2339)

### LINKED ISSUE #2339: Bug: HTTP2 can mask errors in a stream of data

sanskar-singh-2403:
## Summary

Fixes #2339. `merge_insert()` on the remote client could mask the real cause of a mid-stream input error, reporting only:

> stream error sent by user: unexpected internal error

## Root cause

There were two divergent streaming-write code paths in the remote client:

- `add()` uses `RemoteInsertExec`, which streams the request body through a `tokio::sync::oneshot` error side-channel and drains it before reporting the HTTP result. If the input stream errors mid-body, the original error is recovered.
- `merge_insert()` used a legacy path (`send_streaming` -> `reader_as_body`) that piped arrow `Some(Err(e))` straight into the HTTP2 request body. Hyper swallows body-stream errors under HTTP2 (see hyperium/hyper#2547), so the original error was lost and only the generic transport error surfaced.

## Fix

Consolidate both write paths onto the side-channel mechanism instead of patching the legacy path:

- Generalize `RemoteInsertExec` into `RemoteWriteExec`, carrying a `WriteOp` enum (`Insert { overwrite }` | `MergeInsert { query, timeout }`) that selects the endpoint, query params, request-timeout header, and response parsing. The executor returns a `WriteResult` enum (`Add` | `Merge`) with typed accessors, and `with_new_children` still resets the result so the rescannable retry loop is unaffected.
- Route `merge_insert()` through `RemoteWriteExec`. The public API only accepts a `RecordBatchReader` (not rescannable), so the reader is buffered into a `Vec<RecordBatch>` before the retry loop to preserve the previous retry-on-retryable-status behaviour. This mirrors what the old `send_streaming(with_retry=true)` path already did.
- Remove the now-unused `send_streaming` / `reader_as_body` / `buffer_reader` / `make_reader` helpers. Multipart stays insert-only (the server has no multipart merge_insert endpoint), so that hot path is behaviorally unchanged.

## Testing

- Added `test_merge_insert_input_error_surfaces_original`, which drives an erroring input through the single-request `merge_insert` path and asserts the original error (`boom`) is surfaced rather than the masked HTTP error. Confirmed it fails without the side-channel drain (it then reports a masked `500 ... request or response body error`).
- Full suite green: `cargo test -p lancedb --lib --features remote` -> 694 passed, 0 failed. Includes the existing `test_merge_insert_retries_on_409`, confirming retry behaviour is preserved.

### DIFF
--- a/rust/lancedb/src/remote/table/insert.rs
+++ b/rust/lancedb/src/remote/table/insert.rs
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
-//! DataFusion ExecutionPlan for inserting data into remote LanceDB tables.
+//! DataFusion ExecutionPlan for streaming writes (add / merge_insert) to
+//! remote LanceDB tables.
 
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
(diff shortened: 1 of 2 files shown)


sanskar-singh-2403:
@wjones127 @westonpace can we kindly review this?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -464,24 +513,24 @@ impl<S: HttpSend + 'static> PartRequestCtx<'_, S> {
...
-                write!(
-                    f,
-                    "RemoteInsertExec: table={}, overwrite={}",
-                    self.table_name, self.overwrite
-                )
+                let op = match &self.op {
+                    WriteOp::Insert { overwrite } => format!("insert, overwrite={}", overwrite),
+                    WriteOp::MergeInsert { .. } => "merge_insert".to_string(),
+                };
+                write!(f, "RemoteWriteExec: table={}, op={}", self.table_name, op)
```
wjones127: nitpick: when formatting, try to avoid allocating temporary strings:


```suggestion
                write!(f, "RemoteWriteExec: table={}, op=")?;
                let op = match &self.op {
                    WriteOp::Insert { overwrite } => write!(f, "insert, overwrite={}", overwrite),
                    WriteOp::MergeInsert { .. } => write!(f, "merge_insert"),
                };
```

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -152,14 +183,32 @@ impl<S: HttpSend + 'static> RemoteInsertExec<S> {
...
             .unwrap_or_else(|e| e.into_inner())
             .clone()
+        {
+            Some(WriteResult::Add(r)) => Some(r),
+            _ => None,
+        }
+    }
+
+    /// Get the merge result after execution, if this exec ran a merge_insert.
+    #[allow(dead_code)]
```
wjones127: question: remind me—why are these marked as dead code?

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -1297,6 +1406,65 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_merge_insert_input_error_surfaces_original() {
```
wjones127, line 1408: question: does this fix the multi-part path? Or just the single-request path?

wjones127 (approved):
Initially, I thought this refactor seemed a bit aggressive. But it actually looks pretty reasonable so I'm in favor of keeping it.

Deniwn22 (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

rust/lancedb/src/remote/table/insert.rs
sanskar-singh-2403: This fixes the single-request path (no upload_id), which is where the HTTP2 masking happened - the error side-channel in stream_as_http_body recovers the original DataFusion error. The multi-part path (send_one_part) already surfaced input errors correctly: a Some(Err(e)) aborts the body and returns the original error directly, so there was nothing to mask there. That path keeps its own regression test (test_multipart_input_error_surfaces_original). Note merge_insert never takes the multipart path - there's no multipart merge_insert endpoint, and it's guarded as a programming error in execute().

rust/lancedb/src/remote/table/insert.rs
sanskar-singh-2403: Stale - leftover from an earlier revision where these weren't wired up yet. Both are now called via downcast in remote/table.rs (add_result in add_single_partition, merge_result in the merge_insert commit path). Removed the attributes; `cargo check --features remote` is clean without them.

rust/lancedb/src/remote/table/insert.rs
sanskar-singh-2403: Good call, applied. Writes each arm straight into the formatter instead of building a temporary String.

sanskar-singh-2403:
> Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

Which line did you have in mind? The comment came through without a line anchor, so I want to make sure I'm looking at the right spot.

For context, Rust doesn't have finally - cleanup is handled via Drop or explicit calls on each exit path. The one place that needs server-side cleanup on an early return is the multipart write lifecycle in remote/table.rs (add_multipart): every error path already calls abort_multipart_write before returning or retrying, and the parallel insert tasks run in a JoinSet that aborts remaining tasks on drop. Happy to add coverage if you're pointing at a path I've missed.

rust/lancedb/src/remote/table/insert.rs
wjones127: Did you make this change? It looks the same to me in the diff.

sanskar-singh-2403: @wjones127  I am really sorry, i was working on another branch and committed it there, it is fixed now, sorry again, 🙏

wjones127: No worries! Thanks for making that fix.

wjones127 (approved):

PR #3713, opened 2026-07-24
fix: deflake test_read_consistency_interval

### LINKED ISSUE #3712: bug: flaky test_read_consistency_interval breaks Rust CI on main
`table::tests::test_read_consistency_interval` fails intermittently on the `linux` job of the Rust workflow. It broke CI on `main` at 0bc08160 ([run 29958259215](https://github.com/lancedb/lancedb/actions/runs/29958259215/job/85234925637)), even though that commit only touches Python:

```
thread 'table::tests::test_read_consistency_interval' panicked at rust/lancedb/src/table.rs:4098:21:
assertion `left == right` failed
  left: 1
 right: 0
```

## Cause

The `Some(100)` case configures a 100ms `read_consistency_interval` on `table2`, then asserts that `table2` still reads stale (0 rows) right after `table1` commits a write:

```rust
let table2 = conn2.open_table("my_table").execute().await.unwrap();   // seeds cache at t0

assert_eq!(table1.count_rows(None).await.unwrap(), 0);
assert_eq!(table2.count_rows(None).await.unwrap(), 0);

table1.add(data).execute().await.unwrap();
assert_eq!(table1.count_rows(None).await.unwrap(), 1);

// ...
assert_eq!(table2.count_rows(None).await.unwrap(), 0);               // line 4098
```

`table2`'s cache timestamp is set once, when the table is opened, and reads that hit within the interval do not refresh it. So the assertion at line 4098 only holds if the intervening open/count/commit/count sequence finishes within 100ms of real wall-clock time. On a loaded CI runner it does not, the TTL expires, `count_rows` performs a synchronous refresh, and the test sees 1 row.

Reproduces deterministically with a 120ms stall inserted before `table1.add`.

## Fix

Use the existing `background_cache::clock` mock instead of racing the wall clock: pin the clock after `table2` is opened and advance it explicitly in place of `tokio::time::sleep`. Same approach as #3547.

wjones127:
`test_read_consistency_interval` asserted that a table opened with a 100ms `read_consistency_interval` still read stale data immediately after a concurrent write. The cache timestamp is set when the table is opened and reads within the interval do not refresh it, so that assertion only held if the intervening open/count/commit/count sequence finished within 100ms of real wall-clock time. On a loaded CI runner it did not: the TTL expired, `count_rows` refreshed synchronously, and the test failed with `left: 1, right: 0`. This broke the Rust workflow on `main` at 0bc08160 (a Python-only commit).

This pins the `background_cache` mock clock once `table2` has seeded its cache, and advances it explicitly in place of `tokio::time::sleep`, so the test controls when the interval elapses. Same approach as #3547. With the clock pinned there is no real sleep left to be imprecise, so the `cfg(not(target_os = "windows"))` guard is dropped and the test now runs on Windows too.

Verified by inserting a stall before the write: 120ms reproduces the original failure deterministically, and with this change the test still passes with a 500ms stall.

Fixes #3712

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -4049,10 +4049,10 @@ mod tests {
         Box::new(RecordBatchIterator::new(vec![Ok(batch)], schema))
     }
 
-    // Windows does not support precise sleep durations due to timer resolution limitations.
-    #[cfg(not(target_os = "windows"))]
     #[tokio::test]
     async fn test_read_consistency_interval() {
+        use crate::utils::background_cache::clock;
+
         let intervals = vec![
             None,
             Some(0),
@@ -4079,6 +4079,12 @@ mod tests {
             let conn2 = conn2.execute().await.unwrap();
             let table2 = conn2.open_table("my_table").execute().await.unwrap();
 
+            // Freeze the consistency clock now that `table2` has seeded its cache, so the
+            // interval only elapses when this test advances it. Otherwise the write and
+            // count_rows calls below race the real 100ms interval, which a loaded CI
+            // runner loses. Must come after open_table: creating the cache clears the mock.
+            clock::pin();
+
             assert_eq!(table1.count_rows(None).await.unwrap(), 0);
             assert_eq!(table2.count_rows(None).await.unwrap(), 0);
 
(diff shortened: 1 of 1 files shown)


westonpace (approved):
Very cool!

PR #3717, opened 2026-07-25
fix: preserve row count in MetadataEraserExec for zero-column batches

vivek-bharathan:
SELECT COUNT(*) FROM t WHERE <predicate> — and any query that plans an empty-projection scan — panics the executing query task:

InvalidArgumentError("must either specify a row count or at least one column")

Root cause

MetadataEraserExec wraps every LanceDB table scan to strip schema-level metadata, rebuilding each batch in execute():

RecordBatch::try_new(schema.clone(), batch.columns().to_vec()).unwrap()

RecordBatch::try_new infers the row count from the columns. COUNT(*) with a filter is planned with an empty projection, so the scan emits zero-column batches — there are no columns to infer a length from, try_new returns Err, and the .unwrap() panics.

(This is specific to the empty-projection case: COUNT(*) with no filter is answered from statistics and never scans, and COUNT(<col>) projects a column — both already work.)

### DIFF
--- a/rust/lancedb/src/table/datafusion.rs
+++ b/rust/lancedb/src/table/datafusion.rs
@@ -12,7 +12,7 @@ pub mod udtf;
 
 use std::{collections::HashMap, sync::Arc};
 
-use arrow_array::RecordBatch;
+use arrow_array::{RecordBatch, RecordBatchOptions};
 use arrow_schema::Schema as ArrowSchema;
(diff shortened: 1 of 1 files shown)


LuQQiu (approved):
Can we have a SQL level end to end test for the COUNT(*) kind of workload?

PR #3724, opened 2026-07-26
feat(python): add setup_time property to StreamingDataset

### LINKED ISSUE #3692: Feature: StreamingDataset should report "setup" times
### SDK

Python

### Description

The StreamingDataset has a number of "setup" phases.  It needs to apply the filter to find the matching row ids and it needs to shuffle these row ids and assign them to splits.  It should report how much time is spent at each stage of setup.  This can help users to know, for example, if they should invest in more indexing so that they can speed up their filtering.

2dmurali:
## Summary
- Adds `setup_time` property to `StreamingDataset` that reports seconds
  spent building the permutation table during `__init__`
- Measures filter configuration + split/shuffle execution in a single
  `time.perf_counter()` window
- Property is immutable after construction, serializes naturally via
  existing `__getstate__`/`__setstate__`

## Test plan
- [x] `test_setup_timing_without_filter` — positive value without filter
- [x] `test_setup_timing_with_filter` — positive value with filter
- [x] `test_setup_timing_empty_filter_result` — ValueError on zero rows
- [x] `test_setup_time_survives_pickle` — survives DataLoader serialization

## Files changed:
- python/python/lancedb/streaming.py — added setup_time property + timing in __init__                                                                                                                                                                                             
- python/python/tests/test_elastic_dataloader.py — added 4 test cases

Closes #3692

### DIFF
--- a/python/python/lancedb/streaming.py
+++ b/python/python/lancedb/streaming.py
@@ -206,6 +206,10 @@ class StreamingDataset(IterableDataset):
 
         # Build the permutation table once, deterministically.
         builder = permutation_builder(table)
+
+        # Time the full setup: filter configuration + split/shuffle execution.
+        # Set once during __init__; immutable thereafter.
+        t0 = time.perf_counter()
         if filter is not None:
             builder = builder.filter(filter)
         if shuffle:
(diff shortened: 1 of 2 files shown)


2dmurali:
@wjones127 Could you take a look at this? Tagging you since you handled the earlier streaming changes.

PR #3728, opened 2026-07-27
fix(node): reject non-string Arrow metadata

### LINKED ISSUE #3729: bug(node): makeArrowTable accepts non-string Arrow schema metadata
## SDK

Node.js / TypeScript

## Description

`makeArrowTable([], { schema })` can accept malformed Arrow schema metadata at runtime even though schema metadata is required to be a `Map<string, string>`.

This is observable when schemas come from plain JavaScript, unsafe casts, or another installed Apache Arrow version. A string key paired with a non-string value is accepted; a metadata entry with both a non-string key and value is also accepted.

## Black-box reproduction

```ts
import { makeArrowTable } from "@lancedb/lancedb";
import { Field, Int32, Schema } from "apache-arrow";

const metadata = new Map<unknown, unknown>([["source", 42]]) as unknown as Map<
  string,
  string
>;
const schema = new Schema(
  [new Field("value", new Int32(), true)],
  metadata,
);

makeArrowTable([], { schema }); // currently succeeds
```

## Expected behavior

- Valid string key/value metadata is accepted and preserved.
- Metadata with a non-string key, a non-string value, or both is rejected with the existing validation error:
  `Expected metadata, if present, to be a Map<string, string> but it had non-string keys or values`

## Black-box acceptance coverage

Exercise the empty-table conversion boundary against Apache Arrow 15, 16, 17, and 18:

| Metadata entry | Expected result |
| --- | --- |
| `["source", "fixture"]` | Accepted and preserved |
| `[42, "fixture"]` | Rejected |
| `["source", 42]` | Rejected |
| `[42, false]` | Rejected |

## Implementation

The fix and regression coverage are in PR #3728.

u70b3:
## Summary

- validate Arrow metadata keys and values independently at runtime
- reject malformed foreign schemas before constructing a local Arrow schema
- cover valid and invalid metadata entries across Arrow 15–18

## Testing

- `node_modules/.bin/jest --runInBand __test__/arrow.test.ts -t "schema metadata"`
- `node_modules/.bin/jest --runInBand __test__/arrow.test.ts`
- `node node_modules/@biomejs/biome/bin/biome format --write lancedb/sanitize.ts __test__/arrow.test.ts`
- `pnpm lint`
- `pnpm build`
- `pnpm run docs`

Fixes #3729

### DIFF
--- a/nodejs/__test__/arrow.test.ts
+++ b/nodejs/__test__/arrow.test.ts
@@ -991,6 +991,37 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
 
         expectValidMapField(roundTripped.schema.fields[0]);
       });
+
+      it("preserves string schema metadata", function () {
+        const metadata = new Map([["source", "fixture"]]);
+        const schema = new Schema(
+          [new Field("value", new Int32(), true)],
+          metadata,
+        );
+
+        expect(makeEmptyTable(schema).schema.metadata.get("source")).toBe(
+          "fixture",
+        );
+      });
+
+      it.each([
+        ["non-string keys", new Map<unknown, unknown>([[42, "fixture"]])],
+        ["non-string values", new Map<unknown, unknown>([["source", 42]])],
+        [
+          "non-string keys and values",
+          new Map<unknown, unknown>([[42, false]]),
+        ],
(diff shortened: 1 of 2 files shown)


wjones127 (approved):
Thanks for the fix!

PR #3730, opened 2026-07-27
feat(python): expose AsyncTable.to_lance

### LINKED ISSUE #1387: Feature: to_lance method in asynchronous table python api
### SDK

Python

### Description

Currently, in python api,  "to_lance" method that returns a pyarrow dataset is only supported in synchronous table. 
I was wondering if "to_lance" method can be exposed for asynchronous tables as well.

For using third party tools like duckdb, pyarrow dataset is preferred over pyarrow tables as "Arrow Datasets  allow for full column selection and filter pushdowns,  unlike Arrow tables which are eagerly loaded into memory." 

Since the newly introduced lance v2 is supported only in asynchronous tables , and since asynchronous tables  do not have a to_lance method, we are not able to evaluate features like filter pushdowns, when using lance v2 tables against duckdb.

u70b3:
## Summary

- expose the existing async Lance dataset conversion as `AsyncTable.to_lance`
- preserve table version, branch, and refreshed storage options when opening the dataset
- route internal async pandas/query paths through the public API
- cover normal tables, checked-out versions, branches, and forwarded dataset options

## Testing

- `cd python && uv run --no-sync pytest python/tests/test_table.py -q`
- `cd python && uv run --no-sync pytest python/tests/test_query.py -q`
- `cd python && uv run --no-sync pytest --doctest-modules python/lancedb/table.py -q`
- `uv run --project python --no-sync ruff format --check python/python/lancedb/table.py python/python/lancedb/query.py python/python/tests/test_table.py`
- `uv run --project python --no-sync ruff check .`

Fixes #1387

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -3041,7 +3041,7 @@ class AsyncQueryBase(object):
             if blob_mode == "bytes"
             else {}
         )
-        dataset = await self._table._to_lance()
+        dataset = await self._table.to_lance()
         scanner = dataset.scanner(
             **_scanner_kwargs_for_query(
                 query,
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -4650,7 +4650,24 @@ class AsyncTable:
         """
         return AsyncQuery(self._inner.query(), self)
 
-    async def _to_lance(self, **kwargs) -> lance.LanceDataset:
+    async def to_lance(self, **kwargs) -> lance.LanceDataset:
+        """Return the Lance dataset backing this table.
+
+        Parameters
+        ----------
+        **kwargs
+            Forwarded to [`lance.dataset`][lance.dataset].
+
+        Returns
+        -------
+        lance.LanceDataset
+            The Lance dataset at this table handle's version and branch.
+
+        Examples
+        --------
+        >>> async def get_lance_dataset(table):
+        ...     return await table.to_lance()
+        """
         try:
             import lance
         except ImportError:
@@ -4700,7 +4717,7 @@ class AsyncTable:
             return (await self.to_arrow()).to_pandas(**kwargs)
         if blob_mode == "bytes" and blob_v2_column_paths(schema):
             return await self.query().to_pandas(blob_mode=blob_mode, **kwargs)
-        return (await self._to_lance()).to_pandas(blob_mode=blob_mode, **kwargs)
+        return (await self.to_lance()).to_pandas(blob_mode=blob_mode, **kwargs)
 
     async def to_arrow(self) -> pa.Table:
         """Return the table as a pyarrow Table.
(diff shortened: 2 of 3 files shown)


u70b3:
## API decision and validation scope

The maintainer guidance in #1387 proposed opening the dataset from the table URI and
`storage_options`. Since then, `AsyncTable` has gained a private `_to_lance()`
implementation that follows this approach and is already used internally by async
pandas/query paths.

This PR promotes that existing behavior to a public API instead of introducing a new
dataset access path. The resulting public behavior is:

- the returned dataset targets the version currently checked out by the table handle;
- branch-scoped handles return a dataset checked out to the same branch and version;
- the latest storage options are used when opening the dataset;
- additional keyword arguments are forwarded to `lance.dataset()`.

The private method is renamed rather than retained as a compatibility wrapper because it
was not a public API, and all internal call sites are updated in this PR.

This change does not add a separate remote endpoint or modify the Rust/PyO3 bindings. The
implementation continues to open the Lance dataset directly from the URI and storage
options, as proposed in #1387. Tests cover local tables, checked-out versions, branches,
and forwarded dataset options.

A live LanceDB Cloud environment has not been exercised: this contributor environment
does not have a Cloud test database or credentials configured, and secret-dependent
remote tests are skipped for pull requests from forks. Therefore, this PR does not claim
that the Cloud data path has been independently validated. Maintainer feedback or
internal Cloud validation would be appreciated, especially on whether the version/branch
semantics and direct URI-based behavior are appropriate as the public
`AsyncTable.to_lance()` contract.

u70b3:
## Follow-up: local remote-path validation

I also exercised this change through a local Cloud-compatible remote fixture rather than
calling `AsyncTable.to_lance()` only on a native/local connection.

The fixture used:

- a real `connect_async("db://test", api_key=..., host_override=...)` connection;
- the real Rust/PyO3 remote table implementation, including the `x-api-key` and
  `x-lancedb-database` headers;
- a minimal local REST control plane implementing the table `describe` and branch-list
  endpoints;
- a real Lance dataset and branch as the data plane returned by `describe.location`.

The following scenarios passed on commit `3b6863e3`:

- main table: `to_lance(default_scan_options={"with_row_id": True})`;
- projection and filter pushdown (`id >= 2`, returning `[2, 3]`);
- a table handle pinned to version 1;
- a remote branch handle targeting branch `exp` at version 2;
- branch isolation (3 rows on main and 4 rows on `exp`).

The fixture made 8 authenticated remote requests to:

- `/v1/table/cloud_table/describe/`
- `/v1/table/cloud_table/branches/list/`

This validates the SDK's remote control-plane-to-Lance-data-plane handoff and the public
version/branch behavior without production credentials.

It is still not a live LanceDB Cloud test: the returned dataset location was a local path,
so hosted authentication, server-provided object-store locations, and any credential
vending required to read those locations remain outside this validation. A live internal
Cloud smoke test would still be valuable for that final deployment-specific layer.

wjones127 (approved):

PR #3731, opened 2026-07-27
fix: distinguish corrupt table from missing in open_table

### LINKED ISSUE #3127: bug: table_names() lists table with corrupt/incomplete .lance directory that open_table() cannot open
### LanceDB version

0.19.0 (also present in current main — listing.rs unchanged)

### What happened?

`table_names()` returns a table name because the `.lance` directory exists on disk, but `open_table()` fails with `ValueError: Table 'conversations' was not found` because the internal dataset files are missing or corrupt.

This happens when a process is interrupted between `drop_table()` and `create_table()` (e.g., during a schema migration that does drop+recreate). The result is a `.lance` directory with no valid manifest inside.

**Root cause:** `ListingDatabase::table_names()` ([listing.rs:752](https://github.com/lancedb/lancedb/blob/main/rust/lancedb/src/database/listing.rs#L752)) does a `read_dir` + filter for `*.lance` extensions — pure directory listing, no validation. `open_table()` → `NativeTable::open_with_params` → `DatasetBuilder::from_uri().load()` actually reads the manifest and fails with `DatasetNotFound`, which becomes `TableNotFound` → Python `ValueError`.

Code that uses the common pattern `if name in db.table_names(): db.open_table(name)` will fail unexpectedly.

### Steps to reproduce

```python
import lancedb
import pyarrow as pa
import shutil
import os

db = lancedb.connect("/tmp/lancedb_test")

# Create a table normally
schema = pa.schema([pa.field("id", pa.string())])
db.create_table("test_table", schema=schema)

# Simulate interrupted drop+recreate: remove internal files but keep directory
table_dir = "/tmp/lancedb_test/test_table.lance"
for item in os.listdir(table_dir):
    item_path = os.path.join(table_dir, item)
    if os.path.isdir(item_path):
        shutil.rmtree(item_path)
    else:
        os.remove(item_path)

# table_names() says it exists
print("test_table" in db.table_names(limit=100))  # True

# open_table() disagrees
db.open_table("test_table")  # ValueError: Table 'test_table' was not found
```

### Expected behavior

Either:
1. `table_names()` should validate that `.lance` directories contain a loadable dataset (at minimum check for `_versions/` or a manifest file), or
2. `open_table()` should give a distinct error (e.g., `TableCorrupted`) rather than `TableNotFound`, so callers can distinguish "never existed" from "exists but broken" and recover appropriately.

Option 2 is probably better — it's cheap and doesn't add I/O cost to `table_names()`.

### Related

- #2727 — `table_names()` default `limit=10` causes `__contains__` to miss tables (different bug, same method)
- #2963 — NFS-related `NotFound` errors during concurrent writes (similar symptom, different cause)

joaquinhuigomez:
`table_names()` lists any `*.lance` directory, but `open_table()` maps every `DatasetNotFound` to `TableNotFound`, so a corrupt or partially-written table looks identical to one that never existed (#3127). This takes the issue's Option 2: on `DatasetNotFound`, check the parent listing for the table's `.lance` entry — the same predicate `table_names()` uses — and return a new `TableCorrupted` error when the directory is present. The check runs only on the error path, and any failure in the recheck falls back to the previous `TableNotFound` behavior.

Tests cover the reporter's empty-dir repro, a deleted-manifest case, true absence (still `TableNotFound`), and an end-to-end list-then-open assertion; the three new corrupt-case tests fail without the src change. `cargo test -p lancedb --lib` 732 passed, clippy/fmt clean, `cargo check --workspace --all-targets` clean (both language bindings end in wildcard error arms).

Two notes for review: `Error` isn't `#[non_exhaustive]`, so the new variant is technically semver-breaking for exhaustive matchers (pre-1.0, and the alternative — changing `TableNotFound`'s shape — breaks more); and on the Python side corrupt tables now surface as `RuntimeError` rather than `ValueError`, which is the intended distinction but worth a maintainer's eye. `open_from_namespace` was left unchanged since namespace listings come from a server-side registry, not directory globbing.

Closes #3127

### DIFF
--- a/rust/lancedb/src/error.rs
+++ b/rust/lancedb/src/error.rs
@@ -18,6 +18,10 @@ pub enum Error {
     InvalidInput { message: String },
     #[snafu(display("Table '{name}' was not found"))]
     TableNotFound { name: String, source: BoxError },
+    #[snafu(display(
+        "Table '{name}' exists but could not be loaded (it may be corrupt or incomplete): {source}"
+    ))]
+    TableCorrupted { name: String, source: BoxError },
     #[snafu(display("Database '{name}' was not found"))]
     DatabaseNotFound { name: String },
     #[snafu(display("Database '{name}' already exists."))]
--- a/rust/lancedb/src/connection.rs
+++ b/rust/lancedb/src/connection.rs
@@ -456,6 +456,10 @@ impl Connection {
     ///
     /// # Returns
     /// Created [`TableRef`], or [`Error::TableNotFound`] if the table does not exist.
+    /// If the table's storage is present but holds no readable dataset (for example a
+    /// `<name>.lance` directory left behind by an interrupted drop and re-create, which
+    /// [`Self::table_names`] still lists) this returns [`Error::TableCorrupted`]
+    /// instead.
     pub fn open_table(&self, name: impl Into<String>) -> OpenTableBuilder {
         OpenTableBuilder::new(
             self.internal.clone(),
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -48,6 +48,7 @@ use crate::connection::NamespaceClientPushdownOperation;
 use crate::DistanceType;
 use crate::data::scannable::{PeekedScannable, Scannable, estimate_write_partitions};
 use crate::database::Database;
+use crate::database::listing::LANCE_FILE_EXTENSION;
 use crate::database::read_freshness::TableFreshness;
 use crate::embeddings::{EmbeddingDefinition, EmbeddingRegistry, MemoryRegistry};
 use crate::error::{Error, Result};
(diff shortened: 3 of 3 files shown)


rust/lancedb/src/error.rs
```diff
@@ -18,6 +18,10 @@ pub enum Error {
     InvalidInput { message: String },
     #[snafu(display("Table '{name}' was not found"))]
     TableNotFound { name: String, source: BoxError },
+    #[snafu(display(
+        "Table '{name}' exists but could not be loaded (it may be corrupt or incomplete): {source}"
+    ))]
+    TableCorrupted { name: String, source: BoxError },
```
wjones127, line 24: suggestion: it would be nice to map this error to a nice error in Python. Right now, they will just be converted to `RuntimeError`:

https://github.com/lancedb/lancedb/blob/98a52267a2d192e0d5d2d5c62ba874247ddd4896/python/src/error.rs#L105

wjones127 (approved):
This makes sense, although perhaps you want to take it a step further and map it to a good error in Python? If not, we can file a follow up issue.

This also makes me think we should create a repair command on databases.

PR #3732, opened 2026-07-27
fix: tolerate a transient TableNotFound in wait_for_index's list_indices poll

ChilePiquin:
## Summary
- `wait_for_index()`'s poll loop called `list_indices().await?`, so a single transient `TableNotFound` (e.g. a replica read racing ahead of manifest/cache propagation) immediately aborted the whole `wait_timeout` with a misleading "table not found" error, even though the table indisputably exists (the caller just created/wrote to it).
- `index_stats()` already treats a 404 as "not indexed yet" (`Ok(None)`) and the poll loop retries on that; `list_indices()` had no equivalent tolerance.
- This surfaced in a Sophon staging investigation where `create_index(..., wait_timeout=...)` failed near-instantly with `Table '<uuid>' was not found` right after the table was created, which turned out to be this exact asymmetry rather than a real missing table.
- Fix: in the poll loop, catch `Error::TableNotFound` from `list_indices()` and retry (bounded by the existing overall `timeout`) instead of propagating immediately. A table that's genuinely gone still surfaces as `Error::Timeout` once the deadline passes.

## Test plan
- [x] Added `test_wait_for_index_tolerates_transient_list_indices_not_found`, which mocks `list_indices` 404-ing on the first poll and succeeding after, asserting `wait_for_index` completes instead of erroring.
- [x] `cargo test --features remote -p lancedb` — all 717 tests pass.
- [x] `cargo fmt -p lancedb` and `cargo clippy --features remote -p lancedb --tests -- -D warnings` — clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3421,8 +3421,8 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
 name = "fsst"
-version = "9.1.0-beta.7"
-source = "git+https://github.com/lance-format/lance.git?tag=v9.1.0-beta.7#4b6a69a1279c62600bd00c858d2002f53c801bd4"
+version = "9.1.0-beta.8"
+source = "git+https://github.com/lance-format/lance.git?tag=v9.1.0-beta.8#5107a99e3f3912851c8cbb3822bd5b4cbf7fda2f"
 dependencies = [
  "arrow-array",
  "rand 0.9.5",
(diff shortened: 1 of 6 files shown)


jackye1995:
> a single transient TableNotFound (e.g. a replica read racing ahead of manifest/cache propagation) immediately aborted the whole wait_timeout with a misleading "table not found" error, even though the table indisputably exists (the caller just created/wrote to it).

hmm this really bugs me, this looks like a serious bug to me that we should investigate and fix, instead of trying to handle it

PR #3736, opened 2026-07-28
feat(table): checkpoint_lsm, flush_lsm, compact_lsm, get_lsm_stats

hamersaw:
Converge a table's LSM write path into its base table, and inspect it.

`checkpoint_lsm` is `flush` then `compact`, repeated until the fresh tier is empty — and the loop runs **client-side**. Putting it on the server would mean a background task, which means a single-flight intent, an intent that leaks on panic, a bounded-iteration policy, an "is it done" observable, and a story for every way a client can vanish mid-operation. None of that exists in this shape: each request does a bounded unit of work and reports what is left, so completion is *carried in the responses* rather than inferred from a shared counter that cannot distinguish "converged" from "hasn't started yet".

Best-effort by construction. Nothing is frozen, so `converged` means L0 was empty as of the last pass. It is idempotent, abandonable at any point with zero consequence, and safe to run on a cadence — an already-converged table costs one round trip and zero compaction passes, because `flush` reports `generations_remaining` and the loop is never entered.

## The failure taxonomy is the load-bearing part

Five distinct conditions used to arrive at a client as one 503. `Error::LsmRoute` carries a classification read from the response body's namespace error code **at the point of receipt** — before any generic helper folds the body into a string and keeps only the status.

| condition | wire | client action |
|---|---|---|
| contention (latch held / pool saturated) | 429, code 21 | retry with backoff |
| owning node draining | 503, code 19 `InvalidTableState` | **stop** |
| fenced / no slot / transport | 503, code 17 | retry with backoff |
| registry entry vanished | 404 | re-issue from `flush` (capped) |
| table being dropped / not WAL-backed | 409 / 400 | stop |

Draining is terminal because the drain gate is a one-way latch — retrying spins until the deadline to report a failure that was knowable on the first response. Transport retry is disabled on these routes for the same reason: it treats every 503 alike and would burn its budget before the classifier ever saw the body.

`get_lsm_stats` returns `Option<LsmStats>`, matching `get_lsm_write_spec` — `None` only when the table has no LSM write path, since a struct of zeros would read as measurements.

Python bindings mirror all four, preserving per-bucket detail rather than flattening to a table-level summary.

## Testing

Six new unit tests against the mocked endpoint, plus the taxonomy round-trip:
- flush into an empty L0 issues **zero** compact calls (asserts the call count — `generations_consumed: 0` is also true of a loop that ran a pointless pass)
- the loop drives compact until the server reports zero remaining
- **contention is not draining**: a 429 retries and converges; asserts the retry count
- a draining node stops after **exactly one** request, no retries
- stats round-trips fully populated; `include_generation_rows` off by default
- every `(status, code)` pair classifies correctly, including unparseable 503 bodies falling back to *retryable* rather than terminal

`cargo test -p lancedb --features remote --lib`: 723 passed.

## Notes for review

- Depends on the sibling lance change returning `SealedGeneration` from `force_seal_active` only at the *server* level — no lance API is used here.
- The branch is based on `codex/update-lance-10-0-0-beta-5`, so it carries one extra commit (`chore: update lance dependency to v10.0.0-beta.5`) that is not part of this change.


rust/lancedb/src/table/checkpoint.rs
```diff
@@ -0,0 +1,340 @@
...
+            Ok(()) => {}
+            Err(e) => match checkpoint_fault(e, attempt)? {
+                CheckpointControl::Retry | CheckpointControl::ReissueFromFlush => {
+                    backoff(attempt).await;
+                    continue;
+                }
+            },
+        }
+
+        let Some(stats) = table.get_lsm_stats(false).await? else {
```
xuanyu-z: await? will error out that bypassing the previous error handling, better to also handling through ```checkpoint_fault```. ALos may consider ```Retry``` and ```ReissueFromFlush``` not using the same budget

xuanyu-z (commented):
Just for my understanding, what is the use cases in enterprise we want to support?

xuanyu-z (approved):

hamersaw:
> what is the use cases in enterprise we want to support?

We need to be able to disable the WAL on tables. We can only do that once we have flushed all current data from the WAL to the lance base table. So `get_lsm_stats` is a way to check WAL usage -- it should be very useful even outside of the checkpoint flow and then `checkpoint_lsm` is a client-driven WAL data flush. So once that completes we should no longer have data in the WAL and can disable.

Issue #3708, opened 2026-07-24
Feature: StreamingDataset two-stage I/O


westonpace:
### SDK

Python

### Description

Random IOPS are a problem for cloud storage.  Other data loaders have worked around this challenge by downloading from cloud storage in large sequential blocks and writing these blocks to NVMe.  Then they read from NVMe in a random order.

This places two limitations on shuffling.  First, we must create splits from sequential blocks, not from individual rows.  Second, when shuffling the individual rows we probably want to limit how far the rows can move.  This reduces the number of chunks that must be downloaded before we can start emitting data.

It also splits our pipeline into 4 stages.  An initial sequential I/O stage (cloud storage to local storage) then a random I/O stage (local storage to in-memory) then the standard CPU & GPU stages.

However, it should allow us to achieve reasonable shuffling while still achieving a high rows/second.

We also should setup an evaluation to compare this approach to the clumped shuffling approach.

sanskar-singh-2403:
hey @westonpace can i work on this?

Thanks for laying this out. I dug through the current torch data loader to
ground a proposal against what already exists. Sharing my understanding and a
suggested approach, would appreciate a sanity check before I start.

## Current state (as I read it)

- `LanceDataset.__iter__` (python/lance/torch/data.py) runs:
  `sampler -> _buffer_arrow_batches -> [optional CachedDataset] -> to_tensor`.
- `ShardedBatchSampler` scans in order and takes every Nth row. Its own
  docstring notes the reservoir approach "does not perfectly randomize the
  stream." This looks like the "clumped shuffling" baseline referenced here.
- `CachedDataset` already spills the stream to a local Arrow IPC file, but only
  for sequential re-reads across epochs, not random reads.
- On the Rust side, `lance-datafusion/src/spill.rs` (`create_replay_spill`)
  already buffers to memory then spills to a temp file with a replayable
  reader. Seems like a natural building block for the NVMe stage.

## Proposed approach

Introduce the 4-stage pipeline as described:

1. Stage 1 (sequential, cloud -> NVMe): download row ranges in large sequential
   blocks and land them on local disk. Blocks become the shuffle unit.
2. Stage 2 (random, NVMe -> memory): a bounded/windowed shuffle that only reads
   from a limited number of resident blocks at a time, capping how far a row can
   travel and how many blocks must be present before emitting starts.
3/4. Existing CPU (to_tensor) and GPU stages unchanged.

Open questions I would like direction on before coding:

[...]

PR #3737, opened 2026-07-28
feat(dataloader): add py1e bounded shuffle planner

sanskar-singh-2403:
Add a bounded, block-oriented shuffle planner modelled on MosaicML Streaming's py1e algorithm, in support of the two-stage I/O pipeline (issue #3708).

py1e spreads each block's rows across a range of approximately shuffle_block_size rows and sorts by the resulting noisy positions, bounding how far any row can travel from its origin. This keeps the resident working set to a small sliding window of blocks so that cloud -> local NVMe downloads stay large and sequential while random reads only ever hit the local copy.

Py1ePlanner::plan returns a BlockPlan describing the contiguous physical row ranges to download (in layout/fetch order) and the bounded emission order. Includes unit tests covering plan validity and non-identity shuffling.

Register the new module via pub mod py1e in permutation.rs.

### DIFF
--- a/rust/lancedb/src/dataloader/permutation/py1e.rs
+++ b/rust/lancedb/src/dataloader/permutation/py1e.rs
@@ -1,0 +1,446 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+//! A bounded, block-oriented shuffle planner modelled on MosaicML Streaming's
+//! `py1e` algorithm.
+//!
+//! # Motivation
+//!
+//! Random IOPS are expensive on cloud storage.  The two-stage I/O pipeline (see
+//! issue #3708) works around this by downloading data in large *sequential
+//! blocks* (cloud -> local NVMe) and only performing *random* reads against the
+//! local copy.  To make that viable the shuffle must be **bounded**: a row may
+//! only travel a limited distance from its original position, so that only a
+//! small, sliding window of blocks has to be resident before emission can
+//! begin.
+//!
+//! `py1e` achieves this by spreading each block's rows across a range of
+//! approximately `shuffle_block_size` rows and then sorting by the resulting
+//! (noisy) positions.  Compared to a global shuffle this trades a small amount
+//! of shuffle quality for a bounded working set and balanced, sequential
+//! downloads.
+//!
+//! # What this module produces
+//!
+//! [`Py1ePlanner::plan`] returns a [`BlockPlan`] describing, for a single split:
+//!
+//! * `blocks` -- the contiguous physical row ranges to download, in *download
+//!   order*.  We call this contiguous position space the "layout": it is the
+//!   order in which blocks are fetched and laid out on local storage.  Each
+//!   block is an internally contiguous range of the split's row offsets, so
+//!   downloading one block is a single sequential read.
+//! * `order` -- the emission order: for each output slot, the original row
+//!   offset (within the split) to emit.
+//! * `order_block` -- parallel to `order`, the index into `blocks` of the block
+//!   that holds each emitted offset.  A consumer downloads blocks lazily in
+//!   download (layout) order and can evict a block as soon as all of its rows
+//!   have been emitted (tracked via [`BlockPlan::block_sizes`]).
+//!
+//! The plan is pure integer/float math over block sizes -- it needs **no
+//! data** -- so it can be computed cheaply up front and then drive a streaming
+//! reader.
+//!
+//! This follows the *construction* of MosaicML's `py1e` (see
+//! <https://docs.mosaicml.com/projects/streaming/en/stable/dataset_configuration/shuffling.html>).
+//! In MDS the bounded shuffle runs independently within each *canonical node*
+//! and never moves a row across a canonical-node boundary, which is what keeps
+//! the shuffle deterministic under elastic resumption and rescaling.  Here each
+//! split *is* its own canonical node: we read one split at a time and never
+//! shuffle rows across split boundaries, so no cross-node behaviour has been
+//! removed or specialised away.  We do not reproduce NumPy's PCG64 bitstream;
+//! determinism is provided by a seeded `SmallRng` instead.
+
+use rand::{Rng, SeedableRng, rngs::SmallRng};
+
+use crate::{Error, Result};
+
+/// Configuration for the [`Py1ePlanner`].
+#[derive(Debug, Clone)]
+pub struct Py1eConfig {
+    /// The download unit: the maximum number of contiguous rows in a single
+    /// block.  Blocks are the granularity at which data is fetched from cloud
+    /// storage, so larger blocks mean larger (cheaper) sequential reads but a
+    /// larger minimum working set.
+    pub block_size: u64,
+    /// The shuffle window: the approximate maximum distance (in rows) that any
+    /// single row may travel from its original position.  For a high-quality shuffle this should
+    /// be several times larger than `block_size`.
+    pub shuffle_block_size: u64,
+    /// Seed for the shuffle.  The caller (for example `StreamingDataset`) is
+    /// responsible for folding the epoch and any run-level base seed into this
+    /// value so that each epoch produces a different ordering.
+    pub seed: u64,
+}
+
+impl Py1eConfig {
+    fn validate(&self) -> Result<()> {
+        if self.block_size == 0 {
+            return Err(Error::InvalidInput {
+                message: "block_size must be greater than 0".to_string(),
+            });
+        }
+        if self.shuffle_block_size == 0 {
+            return Err(Error::InvalidInput {
+                message: "shuffle_block_size must be greater than 0".to_string(),
+            });
+        }
+        Ok(())
+    }
+}
+
+/// A concrete shuffle plan for a single split.
+///
+/// See the [module docs](self) for the meaning of each field.  All row offsets
+/// are relative to the start of the split (i.e. in `0..num_rows`).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlockPlan {
+    /// Contiguous physical row ranges `[begin, end)` (end exclusive), in the
+    /// order they should be downloaded.  `blocks[i]` is the `i`-th block in
+    /// download (layout) order.
+    pub blocks: Vec<(u64, u64)>,
+    /// Emission order: `order[k]` is the original split-relative row offset to
+    /// emit at output slot `k`.  This is a permutation of `0..num_rows`.
+    pub order: Vec<u64>,
+    /// `order_block[k]` is the index into [`blocks`](Self::blocks) of the block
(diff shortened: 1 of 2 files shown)


rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+//!   layout order and can evict a block as soon as all of its rows have been
+//!   emitted (tracked via [`BlockPlan::block_sizes`]).
+//!
+//! The plan is pure integer/float math over block sizes -- it needs **no
+//! data** -- so it can be computed cheaply up front and then drive a streaming
+//! reader.
+//!
+//! This is a faithful port of the *construction* of MosaicML's `py1e` (see
+//! <https://docs.mosaicml.com/projects/streaming/en/stable/dataset_configuration/shuffling.html>)
+//! specialised to a single canonical node (each StreamingDataset split is read
```
westonpace: Hmm, our splits should be equivalent to canonical nodes so in theory we might want to support the same.

However, I'm not sure how we could shuffle across a split without messing up elastic determinism.  If a row starts in split 5 and we shuffle it into split 12 that might be fine in some cases (when splits 5 and 12 are assigned to the same rank) and not be fine in other cases (when splits 5 and 12 are assigned to different ranks).

Does MDS really shuffle across canonical nodes?  I don't think we've actually specialized anything here have we?

sanskar-singh-2403:
@westonpace @wjones127 just a friendly reminder for reviewing this, Thanks!

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+#[derive(Debug, Clone)]
+pub struct Py1eConfig {
+    /// The download unit: the maximum number of contiguous rows in a single
+    /// block.  Blocks are the granularity at which data is fetched from cloud
+    /// storage, so larger blocks mean larger (cheaper) sequential reads but a
+    /// larger minimum working set.
+    pub block_size: u64,
+    /// The shuffle window: the approximate maximum distance (in rows) that any
+    /// single row may travel from its original position.  Corresponds to
+    /// MosaicML's `shuffle_block_size`.  For a high-quality shuffle this should
```
westonpace:
```suggestion
    /// single row may travel from its original position.  For a high-quality shuffle this should
```
It's good to mention Mosaic in the rationale / history (file-level comment) but let's avoid mentioning them on field / method parameters.

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// be several times larger than `block_size`.
+    pub shuffle_block_size: u64,
+    /// Base seed, held constant across an entire training run.  Controls the
+    /// run-level block ordering.
+    pub seed: u64,
+    /// Current epoch.  Combined with `seed` so each epoch produces a different
+    /// ordering.
+    pub epoch: u64,
```
westonpace: I think epoch & seed mixing should happen at a higher level.  For example, the `StreamingDataset` should be the class to merge the two together.  So we can drop epoch from this struct.

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+}
+
+/// A concrete shuffle plan for a single split.
+///
+/// See the [module docs](self) for the meaning of each field.  All row offsets
+/// are relative to the start of the split (i.e. in `0..num_rows`).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlockPlan {
+    /// Contiguous physical row ranges `(begin, end)` (end exclusive), in the
```
westonpace:
```suggestion
    /// Contiguous physical row ranges `[begin, end)` (end exclusive), in the
```

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// order they should be downloaded.  `blocks[i]` is layout block `i`.
```
westonpace: What is "layout" or "layout block" mean in this context?

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+    pub blocks: Vec<(u64, u64)>,
+    /// Emission order: `order[k]` is the original split-relative row offset to
+    /// emit at output slot `k`.  This is a permutation of `0..num_rows`.
+    pub order: Vec<u64>,
+    /// `order_block[k]` is the index into [`blocks`](Self::blocks) of the block
+    /// that contains `order[k]`.
+    pub order_block: Vec<u32>,
+    /// Total number of rows in the split (== `order.len()`).
+    pub num_rows: u64,
```
westonpace: If this is equal to `order.len()` do we need to store it separately?

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// position jitter).
+    fn seeds(&self) -> (u64, u64) {
+        // Mix the epoch into the run seed with a large odd constant so distinct
+        // (seed, epoch) pairs are extremely unlikely to collide.
+        let epoch_seed = self
+            .config
+            .seed
+            .wrapping_add(self.config.epoch.wrapping_mul(0x9E37_79B9_7F4A_7C15));
+        (self.config.seed, epoch_seed)
+    }
```
westonpace: As above, let's do epoch mixing elsewhere and get rid of this.

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+        let (run_seed, epoch_seed) = self.seeds();
+
+        // 2. Shuffle the block ordering once for the whole run, then again per
+        //    epoch.  The resulting order of `physical_spans` is the "layout"
+        //    order == download order.
+        let mut run_rng = SmallRng::seed_from_u64(run_seed);
+        shuffle_slice(&mut physical_spans, &mut run_rng);
+
+        let mut epoch_rng = SmallRng::seed_from_u64(epoch_seed);
+        shuffle_slice(&mut physical_spans, &mut epoch_rng);
```
westonpace: Again, no mixing of run & epoch seeds.  Let's just do one shuffle based on seed.

rust/lancedb/src/dataloader/permutation/py1e.rs
```diff
@@ -0,0 +1,459 @@
...
+        let mut base_offsets: Vec<u64> = Vec::with_capacity(num_rows as usize);
+        let mut base_block: Vec<u32> = Vec::with_capacity(num_rows as usize);
+        // `span_layout_bounds[i]` = (layout_begin, layout_end) for block i, i.e.
+        // the half-open range of layout positions occupied by that block.
+        let mut span_layout_bounds: Vec<(u64, u64)> = Vec::with_capacity(blocks.len());
+
+        let mut layout_pos: u64 = 0;
+        for (block_idx, &(begin, end)) in blocks.iter().enumerate() {
+            let mut within: Vec<u64> = (begin..end).collect();
+            shuffle_slice(&mut within, &mut epoch_rng);
```
westonpace: What is this shuffle doing?

westonpace (changes requested):
Looks like a good start.  I have a few questions & suggestions

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: You're right, MDS does not shuffle across canonical nodes, and neither does this. py1e runs the bounded shuffle independently within each canonical node and never moves a row across a node boundary, which is exactly what preserves elastic determinism when splits 5 and 12 land on different ranks. Here a split is its own canonical node: we plan one split at a time and never shuffle across split boundaries, so nothing has been specialized away. I've reworded the module doc so it doesn't read as if we dropped cross-node behavior; it says the bounded shuffle is scoped to a single canonical node (== one split) by construction.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: Done, dropped the Mosaic reference on the field. Kept it in the file-level rationale as you suggested.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: Done. Removed epoch from Py1eConfig and the seeds() helper. The struct now takes a single seed and the caller (StreamingDataset) is responsible for folding epoch and any run-level base seed into it. Doc comment on seed says as much.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: Done.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: "Layout" is the contiguous position space that results after we shuffle the block download order: it's the order blocks are fetched and laid out on local storage. "Layout block i" is just blocks[i], the i-th block in that download order. I've expanded the module doc and the blocks field doc to define this instead of assuming it.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: Good catch, no. Dropped the num_rows field and replaced it with a num_rows() method returning order.len(). Removes the chance of the two drifting.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: Done, seeds() is gone. Single seed now, epoch folding moves up to the caller.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: Done. There's now a single SmallRng seeded from config.seed driving the block-order shuffle. The second (epoch) shuffle pass is removed.

rust/lancedb/src/dataloader/permutation/py1e.rs
sanskar-singh-2403: That's the intra-block shuffle: it permutes the row offsets within a single block so rows from the same block don't emit in physical (on-disk) order. It's the within-block half of the shuffle; the bounded cross-block spreading is added separately as the position jitter in step 4. I've added a comment saying this at the loop. It now draws from the same single rng as everything else.

sanskar-singh-2403:
@westonpace just a friendly reminder for your review here, thanks!

PR #3742, opened 2026-07-28
feat: create_index returns a Job handle

wkalt:
IndexBuilder::execute now returns a Job with wait and cancel methods. Local tables build the index synchronously and return an already-done job. Remote tables read the job id the server returns from create_index and track it through the /v1/jobs API: wait polls describe until the job reaches a terminal state and cancel posts a cancellation. Servers that return no job id yield a done job, so behavior against older servers is unchanged. The job id is not exposed on the handle.

The Python and TypeScript bindings keep their current signatures and discard the handle; exposing Job there is left to follow-ups.


rust/lancedb/src/remote/job.rs
```diff
@@ -0,0 +1,86 @@
...
+        })?;
+        value
+            .get("job_state")
+            .and_then(|state| state.as_str())
+            .map(str::to_string)
+            .ok_or_else(|| Error::Http {
+                source: format!("job description has no job_state: {}", body).into(),
+                request_id,
+                status_code: None,
+            })
```
wjones127: suggestion: using serde `Derive` on a struct would probably handle some of this parsing for you.

rust/lancedb/src/remote/job.rs
```diff
@@ -0,0 +1,86 @@
...
+    job_id: String,
+}
+
+impl<S: HttpSend> RemoteJob<S> {
+    pub fn new(client: RestfulLanceDbClient<S>, job_id: String) -> Self {
+        Self { client, job_id }
+    }
+
+    /// One `/v1/jobs/describe` round trip, returning the job state string.
+    async fn describe_state(&self) -> Result<String> {
```
wjones127: suggestion: should this be an enum instead? Could have an "other" state for ones that aren't known by the client.

rust/lancedb/src/error.rs
```diff
@@ -40,6 +40,10 @@ pub enum Error {
     Runtime { message: String },
     #[snafu(display("Timeout error: {message}"))]
     Timeout { message: String },
+    #[snafu(display("Job failed: {message}"))]
+    JobFailed { message: String },
+    #[snafu(display("Job was cancelled"))]
+    JobCancelled,
```
wjones127: suggestion: is there at least a job id that could be exposed here? That way if it's unknown why the job was cancelled we might be able to find the answer in logs.

rust/lancedb/examples/simple.rs
```diff
@@ -118,8 +118,12 @@ async fn create_empty_table(db: &Connection) -> Result<LanceDbTable> {
 
 async fn create_index(table: &LanceDbTable) -> Result<()> {
     // --8<-- [start:create_index]
-    table.create_index(&["vector"], Index::Auto).execute().await
+    table
+        .create_index(&["vector"], Index::Auto)
+        .execute()
+        .await?;
```
wjones127: question(blocking): what does this `.await` wait for? The job to start? or the completion of the job?

My initial idea was that `Job` itself implemented `IntoFuture`, but that might make an awkward API like:


```rust
    table
        .create_index(&["vector"], Index::Auto)
        .execute()
        .await?
        .await?;
```

(double await)

Alternatively the API now is:

```rust
    table
        .create_index(&["vector"], Index::Auto)
        .execute()
        .await?
        .wait()
        .await?;
```

rust/lancedb/src/job.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Job")
+            .field("done", &self.handle.is_none())
+            .finish()
+    }
+}
+
+impl Job {
+    /// A job whose operation finished before the handle was created.
+    pub(crate) fn done() -> Self {
```
wjones127: nitpick: `done()` almost sounds like a property ("is it done?"). Maybe


```suggestion
    pub(crate) fn new_done() -> Self {
```

rust/lancedb/src/job.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/// A handle to an operation that may still be running.
+///
+/// Local tables perform operations such as index creation synchronously, so
+/// their jobs are already finished when returned. Remote tables may run the
+/// operation in the background; [`Job::wait`] blocks until it finishes.
+pub struct Job {
+    handle: Option<Box<dyn JobHandle>>,
+}
```
wjones127, line 30: thought: part of me wonders whether we need both layers. Maybe we should just return `Box<dyn JobHandle>`, and local operations just have a job handle where `wait()` returns immediately and `cancel()` errors.

rust/lancedb/src/remote/table.rs
wjones127, line 1: suggestion: should we have a test that has two tasks one calling `wait()` and then another starting a bit later concurrently calling `cancel()`. What happens to the `wait()` future? Does that error? (This might be defined more by server behavior, I suppose.)

wjones127 (changes requested):
I'd like to understand a little more expected usage, and have various non-blocking suggestions.

rust/lancedb/src/remote/table.rs
wkalt: yeah, defined by the server I think. We should add this test in enterprise though.

rust/lancedb/examples/simple.rs
wkalt: .await here is awaiting submission of the job in the remote SDK. That is not the same as start or completion necessarily. In general I have been thinking of the usual calling style as less pipelined, like

```
let job = table.create_index(...).execute().await?

job.wait().await?;
// ... or
job.cancel().await?;
```

Chaining cancel() doesn't make much sense and having styles for the two seems a bit odd to me. What do you think, do you prefer the first?

rust/lancedb/src/job.rs
wkalt: Returning `Box<dyn JobHandle>` seems like a less clean and less extensible public API than returning a concrete `Job` to me. What do you see as the advantage?

rust/lancedb/src/remote/job.rs
wkalt: updated

rust/lancedb/src/remote/job.rs
wkalt: updated

rust/lancedb/src/error.rs
wkalt: updated, thanks

wkalt:
@wjones127 thanks for the review, updated/responded to your comments

rust/lancedb/src/job.rs
wkalt: updated

cmccabe:
In the past we talked about having the remote API block in a similar fashion to the local API. In other words, making non-blocking an option, while blocking was the default. This PR seems to go in a different direction by continuing the existing behavior of treating enterprise as a special case with different rules from local.

I also don't understand the thinking behind not exposing the job ID, can you clarify? It seems like something users would certainly want. Or did I misunderstand and the ID will be accessible through a future API?

rust/lancedb/src/job.rs
```diff
@@ -0,0 +1,102 @@
...
+/// task, so only the first call reports its outcome.
+struct SpawnedJob {
+    task: Mutex<Option<JoinHandle<Result<()>>>>,
+    abort: AbortHandle,
+}
+
+#[async_trait]
+impl JobHandle for SpawnedJob {
+    async fn wait(&self) -> Result<()> {
+        let Some(task) = self.task.lock().await.take() else {
```
wjones127: **issue (blocking):** `self.task.lock().await.take()` means only the first caller of `wait()` on a given `Job` ever observes the real outcome — every subsequent (or concurrently racing) caller gets a spurious `Ok(())` even if the build actually failed or was cancelled; acceptance criterion: add a test with two tasks both awaiting the same `Job::wait()` (or one `wait()` racing a `cancel()`) and make every caller see the correct terminal result, e.g. by storing the result behind a `tokio::sync::watch`/`OnceCell` instead of consuming the `JoinHandle`.

python/python/lancedb/remote/table.py
```diff
@@ -538,6 +539,32 @@ def create_index(
             )
         )
 
+    def create_index_async(
```
wjones127, line 542: **question (non-blocking):** Local `Table.create_index_async` exposes `replace`, but this remote counterpart doesn't — is that intentional (matching `submit_create_index` in Rust, which never reads `index.replace` for remote tables), or a gap that should be closed for parity?

python/python/tests/test_index.py
```diff
@@ -84,6 +84,13 @@ async def binary_table(db_async):
     )
 
 
+@pytest.mark.asyncio
+async def test_create_index_async_returns_done_job(some_table: AsyncTable):
+    job = await some_table.create_index_async("id", config=BTree())
+    await job.wait()
+    await job.cancel()
```
wjones127, line 93: **nitpick:** Add an assertion that the index actually exists after `wait()` so this test can't pass if index creation silently no-ops.

```suggestion
    job = await some_table.create_index_async("id", config=BTree())
    await job.wait()
    assert len(await some_table.list_indices()) == 1
    await job.cancel()
```

rust/lancedb/src/job.rs
wjones127: I was seeing them as roughly equivalent to the user, and `Box<dyn JobHandle>` just seemed like strictly less code.

nodejs/__test__/table.test.ts
```diff
@@ -851,7 +851,10 @@ describe("When creating an index", () => {
   afterEach(() => tmpDir.removeCallback());
 
   it("should create a vector index on vector columns", async () => {
-    await tbl.createIndex("vec");
+    const job = await tbl.createIndexAsync("vec");
+    expect(job.id).toBeNull();
+    await job.wait();
+    await job.cancel();
```
wjones127, line 858: question(blocking): is the behavior of cancel after completion to silently pass? Maybe worth a comment here that's what's going on. It does seem consistent with some other systems that cancelling something that finished silently succeeds.

rust/lancedb/src/remote/job.rs
```diff
@@ -0,0 +1,110 @@
...
+    InProgress,
+    #[serde(rename = "CANCELLED")]
+    Cancelled,
+    #[serde(rename = "FAILED")]
+    Failed,
+    #[serde(rename = "DONE")]
+    Done,
+    /// A state this client version does not know; treated as still running.
+    #[serde(other)]
+    Other,
```
wjones127: suggestion: I wonder if this could somehow be:


```suggestion
    #[serde(other)]
    Other(String),
```

so we an print it.

rust/lancedb/src/job.rs
wjones127: I'm open to multiple ways to solve this, but I think the current behavior seems wrong. We could either make it impossible for multiple calls to `wait()` to happen or else support multiple callers.

wjones127 (changes requested):

nodejs/__test__/table.test.ts
wkalt: yes, that's intended. I'll make sure it's documented, thanks.

rust/lancedb/src/remote/job.rs
wkalt: updated

rust/lancedb/src/job.rs
wkalt: added in https://github.com/lancedb/lancedb/pull/3742/changes/070c09669bacbf6ee72f28f3a7100347686d8a08. Multiple callers are now supported and the same result is returned to all via a tokio::sync::watch.

python/python/lancedb/remote/table.py
wkalt: updated in https://github.com/lancedb/lancedb/pull/3742/changes/62ce3d5cc26b8745996caed09f24c5672e042036

python/python/tests/test_index.py
wkalt: updated in https://github.com/lancedb/lancedb/pull/3742/changes/e743bc0f34d5a7d7348465b3669c947ac48c7bd5

rust/lancedb/src/job.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+impl SpawnedJob {
+    fn new(task: JoinHandle<Result<()>>) -> Self {
+        let abort = task.abort_handle();
+        let (tx, outcome) = watch::channel(None);
+        tokio::spawn(async move {
+            let outcome = match task.await {
+                Ok(Ok(())) => Outcome::Succeeded,
+                Ok(Err(err)) => Outcome::Failed(err.to_string()),
```
wjones127: **issue (non-blocking):** Local job failures now always collapse to a generic `Error::JobFailed{message}`, losing the original error's variant (previously `Ok(result) => result` preserved it) — necessitated by `Outcome` needing `Clone`. Consider `Outcome::Failed(Arc<Error>)` instead of stringifying, so callers can still match on the underlying error kind.

wjones127 (approved):
On piece of non-blocking feedback. Nice work on this!

rust/lancedb/src/job.rs
wkalt: I'm a little bit stumped on this because the remote SDK won't be able to do that with the same fidelity as local. I wonder if we may need some kind of error protocol in the public API to address this correctly.

wkalt: actually, I guess this same problem must exist in a lot of places in the SDK so it shouldn't be a design blocker here. We can afford some divergence and standardize it all later.

wkalt: Added more info to the failure message and attached the source error for the local case.

PR #3747, opened 2026-07-29
test(python): fail when the API reference misses a public symbol

wjones127:
`docs/src/python/python.md` is the entire Python API reference and is maintained by hand, so a new public class or function stays invisible until someone notices. #3746 found around fifty symbols in that state, including branch management. This adds a test so it fails loudly next time.

The test derives the public surface from the sources with `ast` — `__all__` where a module defines one, top-level class and function definitions otherwise — follows re-exports so a symbol counts as documented under any name it is importable by, and asserts each one is reachable from the reference page. Names deliberately left out go in an `INTENTIONALLY_UNDOCUMENTED` set with a reason, which turns "we don't document this" into a reviewable line rather than an oversight. A second test flags entries on the page that no longer match any source.

Static parsing rather than importing means the test needs no optional dependencies (`lancedb.streaming` needs torch, which the `tests` extra does not install), and it matches how griffe reads the sources when the docs are actually built.

Stacked on #3746, and based on that branch rather than `main` so the diff here is just the test. It cannot pass without the entries #3746 adds, so merge that one first, then retarget this to `main` **before** deleting the `docs/python-api-reference-gaps` branch — deleting the base branch of an open PR closes it rather than retargeting it.

## Example

Deleting the `Branches` entry from the reference page:

```
E   AssertionError: 1 public name(s) in lancedb.table are missing from docs/src/python/python.md:
E       lancedb.table.Branches
E
E   Add a `::: <path>` line to the matching section of that page, or add the name
E   to INTENTIONALLY_UNDOCUMENTED in this test with a reason.
```

## Testing

Checked the test fails for each case it is meant to catch — a symbol dropped from the page, a newly added public class, and a stale entry pointing at nothing — and that adding a name to an automodule'd `__all__` correctly still passes without a docs edit.

## Not included

The docs build itself is still unchecked on PRs: the workflow only runs on pushes to `main`, so a broken cross-reference in a docstring is caught by neither this test nor CI. #3707 is addressing that.

### DIFF
--- a/python/python/tests/test_api_reference.py
+++ b/python/python/tests/test_api_reference.py
@@ -1,0 +1,249 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+"""Guard against the Python API reference drifting from the public API.
+
+`docs/src/python/python.md` is the whole Python API reference and is maintained
+by hand, so a new public class or function is simply absent from the docs until
+someone remembers to list it. This test fails when that happens.
+
+Everything is derived from the source with `ast` rather than by importing, both
+because the docs are built the same way (griffe reads the sources) and because
+some modules need optional dependencies that CI does not install.
+"""
+
+import ast
+from functools import cache
+from pathlib import Path
+from typing import Optional
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+PACKAGE_ROOT = REPO_ROOT / "python" / "python" / "lancedb"
+REFERENCE_PAGE = REPO_ROOT / "docs" / "src" / "python" / "python.md"
+
+pytestmark = pytest.mark.skipif(
+    not REFERENCE_PAGE.exists(),
+    reason="needs a source checkout; docs/ is not shipped in the wheel",
+)
+
+# Modules whose public surface belongs in the reference. Modules absent from this
+# list (internal helpers such as `lancedb.util` or `lancedb._blob`) are not checked.
+CHECKED_MODULES = [
+    "lancedb",
+    "lancedb.context",
+    "lancedb.db",
+    "lancedb.embeddings",
+    "lancedb.exceptions",
+    "lancedb.expr",
+    "lancedb.index",
+    "lancedb.merge",
+    "lancedb.namespace",
+    "lancedb.otel",
+    "lancedb.permutation",
+    "lancedb.pydantic",
+    "lancedb.query",
+    "lancedb.remote",
+    "lancedb.rerankers",
+    "lancedb.schema",
+    "lancedb.streaming",
+    "lancedb.table",
+]
+
+# Public names deliberately kept out of the reference. Each entry is a decision,
+# not an oversight -- add to this list only with a reason.
+INTENTIONALLY_UNDOCUMENTED = {
+    # Concrete implementations documented through their abstract base class.
+    "lancedb.LanceDBConnection",
+    "lancedb.RemoteDBConnection",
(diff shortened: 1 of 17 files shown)


python/python/tests/test_api_reference.py
```diff
@@ -0,0 +1,249 @@
...
+REFERENCE_PAGE = REPO_ROOT / "docs" / "src" / "python" / "python.md"
+
+pytestmark = pytest.mark.skipif(
+    not REFERENCE_PAGE.exists(),
+    reason="needs a source checkout; docs/ is not shipped in the wheel",
+)
+
+# Modules whose public surface belongs in the reference. Modules absent from this
+# list (internal helpers such as `lancedb.util` or `lancedb._blob`) are not checked.
+CHECKED_MODULES = [
```
wjones127: suggestion: now we have to remember to add new modules here. Why not make an ignore list instead?

python/python/tests/test_api_reference.py
```diff
@@ -0,0 +1,249 @@
...
+    "lancedb.pydantic.get_extras",
+    "lancedb.pydantic.is_nullable",
+    "lancedb.pydantic.model_to_dict",
+    "lancedb.schema.blob_column_paths",
+    "lancedb.schema.blob_v2_column_paths",
+    "lancedb.schema.is_blob_like_field",
+    "lancedb.schema.is_blob_v2_field",
+    "lancedb.schema.schema_has_blob_field",
+    "lancedb.table.has_nan_values",
+    "lancedb.table.sanitize_create_table",
```
wjones127: suggestion: can we make some of these private (underscore) methods now?

python/python/tests/test_api_reference.py
wjones127: Done — inverted it. Every module under `lancedb/` is checked now unless it is underscore-prefixed or listed in `UNCHECKED_MODULES`, so a new module is covered by default.

Worth doing on its own merits: the inversion immediately caught two things the allowlist was hiding. `StaticHeaderProvider` and `OAuthProvider` in `lancedb.remote.header` are documented example implementations of `HeaderProvider`, constructed directly in `test_remote_db.py`, and were entirely absent from the reference — they are now exported from `lancedb.remote` next to the base class they implement. And seven module-level helpers in the voyageai and jinaai backends, which are module-local and are now private.

python/python/tests/test_api_reference.py
wjones127: Yes — most of the list is gone now, split between private and documented.

Made private (all confirmed package-internal, no callers outside `lancedb` and its tests): the five blob helpers in `schema`, `has_nan_values` and `sanitize_create_table` in `table`, `ensure_vector_query` in `query`, `lang_mapping` in `index`, and `get_extras` / `is_nullable` / `model_to_dict` in `pydantic`. Privatising `lang_mapping` also let me drop the mkdocstrings `filters` workaround it needed in #3746.

Two I did not privatise, because they turned out to be misfiled rather than internal — `Permutations` is a fully documented public class that tests construct directly, and `Transforms` was already rendered on the page, so my listing it was just wrong. Both are documented now.

What is left is the genuinely deliberate set: implementations reached through an abstract base, query base classes covered by `inherited_members`, wire types, `sanitize_uri` (undocumented and docstring-less but in `lancedb.__all__` for compatibility), and `FixedSizeListMixin`. Entries are keyed by definition site now so one covers every alias, and `test_intentionally_undocumented_is_accurate` fails if an entry no longer exists, is actually rendered, or is not exported by anything checked — so the list cannot quietly rot.

One caveat worth flagging: these renames are technically breaking for anyone importing the old names, though none are documented or exported from `lancedb`.

PR #3749, opened 2026-07-30
feat: compute column sizes in bytes

dantasse:
There was a request to show the size of each column in the UI. Since we already have bytes_on_disk, this seems cheap to compute, so this just computes it.

For nested fields, it reports the size for each field. e.g. if your fields are `name` and `geo` (which includes `lat` and `lon`), you could get:
- name: 1mb
- geo: 2mb
- geo.lat: 1mb
- geo.lon: 1mb

but the table's total size is 3mb (sum of top-level fields). This seems right to me (e.g. it is correct for each field), you just have to be careful not to naively sum all fields to get the table size. As the table size is already computed (elsewhere), it seems like we won't accidentally do that.

## Note
This PR changes the python return type from a dict to a dataclass, which would be a breaking change - can change it back to a dict if you prefer.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -2096,6 +2096,23 @@ impl Table {
     }
 
     /// Retrieve statistics on the table
+    ///
+    /// See [`TableStatistics`] for the full breakdown, including
+    /// [`TableStatistics::column_bytes`] for the on-disk size of each column.
+    ///
+    /// ```
+    /// # use lancedb::Table;
+    /// # async fn column_sizes(table: &Table) -> Result<(), Box<dyn std::error::Error>> {
+    /// let stats = table.stats().await?;
+    /// println!("{} bytes total", stats.total_bytes);
+    /// if let Some(column_bytes) = &stats.column_bytes {
+    ///     for (path, bytes) in column_bytes {
+    ///         println!("  {path}: {bytes} bytes");
+    ///     }
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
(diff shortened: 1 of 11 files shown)


LuQQiu (approved):
looks like this is beneficial, can we make Python and other API also expose this info?

dantasse:
@LuQQiu good idea, thanks! I've added it to Python/TS/Rust here.

dantasse:
(So I guess, take another look if you'd like)

rust/lancedb/src/table.rs
```diff
@@ -3454,6 +3472,19 @@ impl BaseTable for NativeTable {
         let ds_stats = Arc::new(ds_clone).calculate_data_stats().await?;
         let total_bytes = ds_stats.fields.iter().map(|f| f.bytes_on_disk).sum::<u64>() as usize;
 
+        let bytes_by_id: HashMap<u32, u64> = ds_stats
```
jackye1995, line 3475: I don’t think we should expose per-column sizes to the UI yet. This relies on `calculate_data_stats()`, which opens every data file and reads its metadata. Even though it does not scan the data pages, the I/O grows with the number of data files, which will be a problem at scale.

We can calculate the table’s total size from the file statistics already stored in the manifest without opening the data files. Until per-column sizes can also be obtained from manifest-level or pre-aggregated metadata, I think we should avoid supporting this in the UI.

dantasse: Hmm ok, good point. I'll close this for now, we can pick it up later.

PR #3750, opened 2026-07-30
feat(remote): add seekable blob range reads

geruh:
## Summary

- Implements Cloud `fetch_blob_files`: returns real seekable `BlobFile` handles over HTTP Range instead of `NotSupported`.
- Completes the second Cloud blob read verb after #3684 (`fetch_blobs` = eager whole bytes; this = lazy / partial / sequential reads).
- Same public handle API as local (`read_range`, `read_up_to`, `seek`, `tell`, `close`), so one code path works for local and Cloud.

Large blobs (video, audio, PDFs) should not require downloading the whole object to inspect a header or stream a slice. After search, callers open a handle and read only what they need:

```python
hits = table.search(vec).select(["id", "video"]).limit(5).to_arrow()

with table.fetch_blob_files("video", hits)[0] as f:
    header = f.read_range(0, 256)
    f.seek(keyframe_offset)
    chunk = f.read_up_to(1 << 20)
```

### Behavior

- Handle creation probes size with `bytes=0-0` (bounded concurrency, input order preserved).
- `204` → null (`None`); `416` with `bytes */0` → valid empty blob; other `416` → error.
- `read_range` validates `Content-Range` and body length; OOB ranges fail with `invalid_input` before the request (aligned with Lance).
- `read_up_to` reuses one open-ended Range response across sequential reads; `seek` drops it.
- Servers older than 0.5.0 get a clear `NotSupported` (does not suggest `fetch_blobs`, which they also lack).

## Testing

- `cargo test --features remote -p lancedb remote_blob`
- `cargo test --features remote -p lancedb test_blob`
- `cargo clippy --features remote --tests --examples` (no new warnings from this change)

### DIFF
--- a/python/src/table.rs
+++ b/python/src/table.rs
@@ -426,9 +426,11 @@ pub struct PyBlobFile {
 impl PyBlobFile {
     fn read_bytes(self_: PyRef<'_, Self>) -> PyResult<Py<PyBytes>> {
         let inner = self_.inner.clone();
-        let bytes = block_on(async move { inner.read().await })
+        let py = self_.py();
+        let bytes = py
+            .detach(move || block_on(async move { inner.read().await }))
             .map_err(|e| PyRuntimeError::new_err(format!("blob read failed: {e}")))?;
-        Ok(PyBytes::new(self_.py(), bytes.as_ref()).unbind())
+        Ok(PyBytes::new(py, bytes.as_ref()).unbind())
     }
 
     pub fn read(self_: PyRef<'_, Self>) -> PyResult<Bound<'_, PyAny>> {
@@ -444,24 +446,32 @@ impl PyBlobFile {
 
     fn close(self_: PyRef<'_, Self>) -> PyResult<()> {
         let inner = self_.inner.clone();
-        block_on(async move { inner.close().await })
+        self_
+            .py()
+            .detach(move || block_on(async move { inner.close().await }))
             .map_err(|e| PyRuntimeError::new_err(format!("blob close failed: {e}")))
     }
 
     fn is_closed(self_: PyRef<'_, Self>) -> bool {
         let inner = self_.inner.clone();
-        block_on(async move { inner.is_closed().await })
+        self_
+            .py()
+            .detach(move || block_on(async move { inner.is_closed().await }))
     }
 
     fn seek(self_: PyRef<'_, Self>, position: u64) -> PyResult<()> {
         let inner = self_.inner.clone();
-        block_on(async move { inner.seek(position).await })
+        self_
+            .py()
+            .detach(move || block_on(async move { inner.seek(position).await }))
             .map_err(|e| PyRuntimeError::new_err(format!("blob seek failed: {e}")))
     }
 
     fn tell(self_: PyRef<'_, Self>) -> PyResult<u64> {
         let inner = self_.inner.clone();
-        block_on(async move { inner.tell().await })
+        self_
+            .py()
+            .detach(move || block_on(async move { inner.tell().await }))
             .map_err(|e| PyRuntimeError::new_err(format!("blob tell failed: {e}")))
     }
 
@@ -475,16 +485,20 @@ impl PyBlobFile {
             .checked_add(length as u64)
             .ok_or_else(|| PyValueError::new_err("offset + length overflowed"))?;
         let inner = self_.inner.clone();
-        let bytes = block_on(async move { inner.read_range(offset..end).await })
+        let py = self_.py();
+        let bytes = py
+            .detach(move || block_on(async move { inner.read_range(offset..end).await }))
             .map_err(|e| PyRuntimeError::new_err(format!("blob read_range failed: {e}")))?;
-        Ok(PyBytes::new(self_.py(), bytes.as_ref()).unbind())
+        Ok(PyBytes::new(py, bytes.as_ref()).unbind())
     }
 
     fn read_up_to(self_: PyRef<'_, Self>, length: usize) -> PyResult<Py<PyBytes>> {
         let inner = self_.inner.clone();
-        let bytes = block_on(async move { inner.read_up_to(length).await })
-            .map_err(|e| PyRuntimeError::new_err(format!("blob read failed: {e}")))?;
-        Ok(PyBytes::new(self_.py(), bytes.as_ref()).unbind())
+        let py = self_.py();
+        let bytes = py
+            .detach(move || block_on(async move { inner.read_up_to(length).await }))
+            .map_err(|e| PyRuntimeError::new_err(format!("blob read_up_to failed: {e}")))?;
+        Ok(PyBytes::new(py, bytes.as_ref()).unbind())
     }
 }
 
--- a/rust/lancedb/src/blob.rs
+++ b/rust/lancedb/src/blob.rs
@@ -9,6 +9,7 @@
 //!
 //! Blob tables require Lance file format >= 2.2 and stable row ids at create.
 
+use std::ops::Range;
 use std::sync::Arc;
 
 use arrow_array::LargeBinaryArray;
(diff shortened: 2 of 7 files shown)


geruh:
@claude review

rust/lancedb/src/blob.rs
```diff
@@ -17,10 +18,200 @@ use arrow_schema::{DataType, Field, Schema};
...
+    pub fn size(&self) -> u64 {
+        match &self.inner {
+            BlobFileInner::Native(file) => file.size(),
+            #[cfg(feature = "remote")]
+            BlobFileInner::Remote(file) => file.size(),
+        }
+    }
+
+    /// Physical offset, or zero for a remote logical resource.
+    pub fn position(&self) -> u64 {
```
wjones127: **issue (blocking):** `position()` → 0, `kind()` → `Dedicated` (line 198), and `data_path()` → a synthetic `lancedb-cloud/...` path (line 188) are fabricated for remote handles, so a caller branching on layout or resolving the path silently gets nonsense; `uri()` already returns `None` honestly, and since `BlobFile` is a new public type these signatures can't be widened later without breaking callers. Acceptance: either these three return an `Option`/`Result` that is `None`/`Err` for remote handles, or — if you'd rather keep the shape — the fabricated values are documented as the contract *and* pinned by a test so they're intentional rather than incidental.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
...
+            state: Mutex::new(RemoteBlobState::default()),
+            size,
+            path,
+        }
+    }
+
+    pub(crate) async fn close(&self) -> lance_core::Result<()> {
+        let mut state = self.state.lock().await;
+        state.sequential_response = None;
+        state.closed = true;
```
wjones127: **issue (blocking):** Nothing tests the closed-handle path: the five `closed` guards (here, 173, 187, 256, 268) all return `invalid_input("blob file is already closed")` and no test calls `close()` then reads, and the PR's claimed OOB behavior at line 125 (`range.end > self.size` → `invalid_input`, "aligned with Lance") is also unpinned. Acceptance: tests asserting (a) `read_range`/`read_up_to`/`read`/`seek`/`tell` each error after `close()` and `is_closed()` is true, and (b) `read_range` beyond `size` errors with `invalid_input` and sends no request.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
...
+            return Ok(Bytes::new());
+        }
+
+        // Remove the retained response from shared state before awaiting. Failed or
+        // cancelled reads leave the committed cursor unchanged and force the next
+        // read to open a fresh response.
+        let mut sequential_response = state.sequential_response.take();
+        let mut cursor = state.cursor;
+        let mut output = BytesMut::with_capacity(target_len);
+        while output.len() < target_len {
```
wjones127, line 226: **issue (non-blocking):** The state mutex is held across every network await in this loop, so `close()`, `tell()`, `is_closed()`, and even an unrelated `read_range` (via `ensure_open`) all block behind an in-flight sequential read — notably a caller can't `close()` out of a hung read; worth a doc note on the handle if serialized-by-design, or splitting the retained response out from the cursor state if not.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
...
+        }
+        Ok(bytes)
+    }
+
+    pub(crate) async fn read_ranges(
+        &self,
+        ranges: &[Range<u64>],
+    ) -> lance_core::Result<Vec<Bytes>> {
+        let mut output = Vec::with_capacity(ranges.len());
+        for range in ranges {
```
wjones127: **suggestion:** `read_ranges` is one serial round trip per range while `probe_blob_files` right below already does bounded concurrency — consider reusing `buffered(BLOB_SIZE_PROBE_CONCURRENCY)` here; also note this public verb currently has no test of its own.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -111,17 +424,595 @@ impl<S: HttpSend> RemoteTable<S> {
...
+                    client: self.client.clone(),
+                    path,
+                    version,
+                    branch: self.branch.clone(),
+                    freshness,
+                });
+                (requester, logical_path)
+            })
+            .collect();
+        probe_blob_files(probes).await
```
wjones127: **question (non-blocking):** Opening N handles costs N HTTP round trips before a single payload byte is read, and a caller who then reads everything pays 2N vs `fetch_blobs`' N — is there (or could there be) a bulk size/metadata response we could use instead, or is the sync infallible `size()` inherited from the native handle what forces the eager probe?

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
...
+fn parse_unsatisfied_content_range(value: &str) -> Option<u64> {
+    value.strip_prefix("bytes */")?.parse().ok()
+}
+
+fn validate_partial_response(
+    response: &Response,
+    expected: Range<u64>,
+    size: u64,
+) -> lance_core::Result<()> {
+    if response.status() != StatusCode::PARTIAL_CONTENT {
```
wjones127, line 316: **question (non-blocking):** This accepts only 206 with a fully-numeric `Content-Range`, so a proxy/CDN that legally ignores `Range` and answers `200` with the whole body, or answers `bytes 0-0/*` with an unknown total, fails the read outright — is 206 with a concrete total guaranteed end-to-end through the Cloud edge, or should we tolerate those two shapes?

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
...
+        if let Some(version) = self.version {
+            request = request.query(&[("version", version)]);
+        }
+        if let Some(branch) = &self.branch {
+            request = request.query(&[("branch", branch)]);
+        }
+        let (request_id, response) = self.client.send_with_retry(request, None, true).await?;
+        // Empty blobs answer the `bytes=0-0` size probe with 416 and
+        // `Content-Range: bytes */0`. Preserve that response for the probe handler.
+        if response.status() == StatusCode::RANGE_NOT_SATISFIABLE {
```
wjones127: **suggestion:** The comment scopes this to the `bytes=0-0` probe, but the bypass applies to every request, so a genuine 416 on a read path (e.g. blob shrank since the probe) skips `check_response` and loses the typed error/server body — consider a `probe: bool` (or a separate `request_probe`) so only the probe opts out.

rust/lancedb/src/remote/table.rs
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright The LanceDB Authors
 
-mod blobs;
+pub mod blobs;
```
wjones127, line 4: **suggestion:** Making this `pub` exposes `RemoteBlobFile` as a public type whose every method is `pub(crate)` — unusable to callers but now part of the API surface; a `pub(crate) fn BlobFile::from_remote(..)` (or `#[doc(hidden)]`) would let the module stay private.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
...
+        state.sequential_response = None;
+        state.closed = true;
+        Ok(())
+    }
+
+    pub(crate) async fn is_closed(&self) -> bool {
+        self.state.lock().await.closed
+    }
+
+    async fn ensure_open(&self) -> lance_core::Result<()> {
```
wjones127: **nitpick:** The `closed` check is spelled out five times (here plus 173, 187, 256, 268); a `fn ensure_open(&self)` on `RemoteBlobState` returning `lance_core::Result<()>` would let both the locked and unlocked callers share one message.

python/src/table.rs
```diff
@@ -426,9 +426,11 @@ pub struct PyBlobFile {
 impl PyBlobFile {
     fn read_bytes(self_: PyRef<'_, Self>) -> PyResult<Py<PyBytes>> {
         let inner = self_.inner.clone();
-        let bytes = block_on(async move { inner.read().await })
+        let py = self_.py();
```
wjones127, line 429: **praise:** Right call detaching the GIL here rather than leaving `block_on` bare — these methods now do network IO, and it fixes the pre-existing GIL hold on local reads too.

wjones127 (changes requested):
This looks good overall. There's just a few minor things to address before this is ready to merge.

rust/lancedb/src/blob.rs
geruh: yeah i agree. those values were misleading. position, kind, and data_path now return Option, and remote returns None. removed the synthetic path and added a test

rust/lancedb/src/remote/table/blobs.rs
geruh: Cool updated. close now covers all ops and verifies no extra request. OOB and inverted ranges also fail before HTTP.

rust/lancedb/src/remote/table/blobs.rs
geruh: makes sense, moved closed outside the mutex. close now returns during an read, and the reader checks again before updating state.

tell() can still wait since it needs the cursor lock.

rust/lancedb/src/remote/table/blobs.rs
geruh: done!

rust/lancedb/src/remote/table/blobs.rs
geruh: yeah, still one probe per handle. size is sync and open needs to distinguish null, empty, and sized blobs.

bulk size metadata would be a separate server change.

rust/lancedb/src/remote/table/blobs.rs
geruh: keeping it strict. 200 can mean Range was ignored and the whole blob was returned. added a test for that.

rust/lancedb/src/remote/table/blobs.rs
geruh: fixed! only size probes preserve 416 and data reads go through normal error handling.

rust/lancedb/src/remote/table.rs
geruh: changed it to `pub(crate)` and removed the path state.

rust/lancedb/src/remote/table/blobs.rs
geruh: factored into ensure_open(), including another check after taking the lock.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -111,17 +424,595 @@ impl<S: HttpSend> RemoteTable<S> {
...
-        let message = if self.server_version.support_blobs() {
-            "fetch_blob_files is not supported on LanceDB Cloud yet. \
-             Use fetch_blobs for full bytes"
+        if !self.server_version.support_blobs() {
+            return Err(Error::NotSupported {
+                message: "fetch_blob_files requires LanceDB Cloud server 0.5.0 or newer".into(),
+            });
+        }
+        if row_ids.is_empty() {
+            return Ok(Vec::new());
```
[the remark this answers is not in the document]
geruh: fixed. empty row-ID lists now return before the server-version check in both blob read paths. `empty_row_id_lists_bypass_server_version_gate` covers both.

rust/lancedb/src/remote/table/blobs.rs
```diff
@@ -1,21 +1,334 @@
 // SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
geruh: yeah, this is still missing. Once all of this lands node will have it....

wjones127 (approved):
Nice work!

PR #3756, opened 2026-07-31
fix(python): add on_bad_vectors="keep" to allow NaN vectors

### LINKED ISSUE #3153: bug(python): NaN and other float edge cases no longer allowed in tables.
### LanceDB version

0.30.0

### What happened?

We are writing because we noticed a behavior change between `lancedb==0.29.2` and `lancedb==0.30.0` when adding Arrow data that contains `NaN` values in a floating-point list column.

In our case, Hyrax stores result tensors in a fixed-size list column and writes them with `Table.add()`. We have a regression test that writes values including `np.nan`, `np.inf`, `-np.inf`, and `-0.0`. With `lancedb==0.29.2`, that write succeeds and the values round-trip correctly. With `lancedb==0.30.0`, the same `Table.add()` call raises:

```text
ValueError: lance error: LanceError(Arrow): Compute error: Vector column contains NaN values
```

The minimal reproduction we extracted is below.

## Precise behavior change

- `lancedb==0.29.2`: `Table.add()` accepts a `pyarrow.Table` whose fixed-size-list float column contains `NaN`.
- `lancedb==0.30.0`: the same `Table.add()` call raises a `ValueError` complaining that the vector column contains `NaN` values.

We were relying on the older behavior, so we pinned `<0.30.0` for now.

Would you mind confirming whether this behavior change was intentional?

We checked the `python-v0.30.0` release notes here and could not find a note about this specific change:
<https://github.com/lancedb/lancedb/releases/tag/python-v0.30.0>

If this was intentional, we would be very grateful for guidance on the recommended migration path for data that may legitimately contain `NaN` values.

Thank you very much for your time.

```
from pathlib import Path
import shutil
import tempfile

import lancedb
import numpy as np
import pyarrow as pa

root = Path(tempfile.mkdtemp(prefix="lancedb-nan-repro-"))
schema = pa.schema([pa.field("data", pa.list_(pa.float32(), 2))])
empty = pa.table({"data": pa.array([], type=schema.field("data").type)}, schema=schema)
batch = pa.table(
    {
        "data": pa.array(
            [np.array([np.nan, 1.0], dtype=np.float32)],
            type=schema.field("data").type,
        )
    },
    schema=schema,
)

print(f"lancedb={lancedb.__version__}")
try:
    table = lancedb.connect(str(root / "db")).create_table("t", empty, mode="overwrite")
    table.add(batch)
    print("table.add(...) succeeded")
except Exception as exc:
    print(f"table.add(...) raised {type(exc).__name__}: {exc}")
finally:
    shutil.rmtree(root, ignore_errors=True)
```

leepokai:
Closes #3153

Adding data with NaN values in a float vector column started failing in 0.30.0 with `Vector column contains NaN values`. The Rust core already supports keeping them via `NaNVectorBehavior::Keep`, but the Python bindings never exposed it, so every Python write took the reject path.

This adds `"keep"` to `OnBadVectorsType` and plumbs it through `AsyncTable.add` to the Rust add builder. Per the discussion on the issue, `"keep"` only keeps NaN and other float edge cases — wrong-size vectors still raise, since @mtauraso asked for bad sizes to keep erroring.

Verified with the repro from the issue: the default still raises, and `on_bad_vectors="keep"` round-trips `[nan, 1.0]` unchanged. New tests cover the NaN-kept, wrong-size-still-errors, and unclassified-vector-column cases; `pytest python/tests/test_table.py` passes (136 tests). `cargo clippy`, `cargo fmt`, and `ruff` are clean.

Two notes for reviewers:
- The wrong-size test asserts `pytest.raises((ValueError, RuntimeError))` because that `ValueError` is raised lazily while Arrow consumes the record-batch reader, so it surfaces wrapped as a `RuntimeError`. I followed the repo's existing idiom rather than papering over it; happy to translate the exception instead if you prefer.
- Remote tables take the same `on_bad_vectors` argument, but NaN handling there is enforced server-side, so a remote table may still reject NaN even with `"keep"`. I had no remote endpoint to verify that parity.

### DIFF
--- a/python/python/lancedb/_lancedb.pyi
+++ b/python/python/lancedb/_lancedb.pyi
@@ -227,6 +227,7 @@ class Table:
         mode: Literal["append", "overwrite"],
         progress: Optional[Any] = None,
         write_parallelism: Optional[int] = None,
+        on_nan_vectors: Optional[Literal["error", "keep"]] = None,
     ) -> AddResult: ...
     async def update(
         self, updates: Dict[str, str], where: Optional[str]
--- a/python/python/lancedb/db.py
+++ b/python/python/lancedb/db.py
@@ -331,7 +331,9 @@ class DBConnection(EnforceOverrides):
             schema that's specified.
         on_bad_vectors: str, default "error"
             What to do if any of the vectors are not the same size or contains NaNs.
-            One of "error", "drop", "fill".
+            One of "error", "drop", "fill", "null", "keep". With "keep", vectors
+            containing NaNs are kept unchanged; wrong-size vectors still raise
+            an error.
         fill_value: float
             The value to use when filling vectors. Only used if on_bad_vectors="fill".
         storage_options: dict, optional
@@ -1498,7 +1500,9 @@ class AsyncConnection(object):
             schema that's specified.
         on_bad_vectors: str, default "error"
             What to do if any of the vectors are not the same size or contains NaNs.
-            One of "error", "drop", "fill".
+            One of "error", "drop", "fill", "null", "keep". With "keep", vectors
+            containing NaNs are kept unchanged; wrong-size vectors still raise
+            an error.
         fill_value: float
             The value to use when filling vectors. Only used if on_bad_vectors="fill".
         storage_options: dict, optional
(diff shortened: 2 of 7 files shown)


leepokai:
Closing this — I won't be able to follow up on it. Sorry for the noise.

PR #3757, opened 2026-07-31
chore: update lance dependency to v10.1.0-beta.1

LuQQiu:
Updates the Lance Rust workspace dependencies and Java lance-core version to v10.1.0-beta.1.

Includes a compatibility fix for the Lance file writer API by using the explicit V2_1 writer creation path for permutation shuffle spill files.

Triggered by https://github.com/lance-format/lance/releases/tag/v10.1.0-beta.1


jackye1995 (approved):

rust/lancedb/src/remote/table/insert.rs
```diff
@@ -49,7 +49,7 @@ pub(crate) enum WriteOp {
 /// The parsed server response for a completed write, discriminated by the
 /// operation that produced it.
 #[derive(Debug, Clone)]
-pub(crate) enum WriteResult {
+pub enum WriteResult {
```
wjones127: question: why these visibility changes?

cmccabe (approved):

rust/lancedb/src/remote/table/insert.rs
LuQQiu: These were only to satisfy clippy after merging upstream/main.

remote::table` is still declared as `pub(crate) mod table`, so changing these nested items from `pub(crate)` to `pub` does not expose them outside the crate. The effective visibility is still capped by the parent module.

Clippy started failing with `clippy::redundant_pub_crate` for these items because they live inside a crate-private module. Since we run clippy with `-D warnings`, I changed them to `pub` as suggested.

No intended API or behavior change.

LuQQiu: This is my local clippy though, let me try revert it and see whether the PR clippy complains

wjones127: Okay that's fine. Thanks for explaining.

PR #3768, opened 2026-08-03
fix: hybrid query offset/nprobes and merge_insert delete-condition reset bugs

### LINKED ISSUE #3765: bug(python): hybrid search silently ignores .offset()
## Bug

## Root cause

## Repro

```python
import lancedb

table = db.create_table(
    "t",
    [
    ],
)

full = (
    .vector([0.0, 0.4])
    .text("dog")
    .limit(4)
    .with_row_id()
    .to_arrow()
)

offset_result = (
    .vector([0.0, 0.4])
    .text("dog")
    .offset(2)
    .limit(2)
    .with_row_id()
    .to_arrow()
)

```

## Environment

## Fix

I've opened a PR that:

### LINKED ISSUE #3766: bug(python): hybrid search minimum_nprobes(0) silently no-ops instead of raising
## Bug

`LanceHybridQueryBuilder.minimum_nprobes(0)` silently no-ops instead of raising the same validation error that a plain vector query raises for the same input.

## Root cause

In `python/python/lancedb/query.py`, `LanceHybridQueryBuilder._create_query_builders()`:

```python
if self._minimum_nprobes:
    self._vector_query.minimum_nprobes(self._minimum_nprobes)
if self._maximum_nprobes is not None:
    self._vector_query.maximum_nprobes(self._maximum_nprobes)
```

`self._minimum_nprobes` is checked for truthiness, while the very next line correctly checks `is not None` for `self._maximum_nprobes`. Since `0` is falsy in Python, calling `.minimum_nprobes(0)` on a hybrid query sets `self._minimum_nprobes = 0`, but the `if self._minimum_nprobes:` check then skips forwarding it to the underlying vector sub-query entirely — so the value is silently dropped instead of being validated.

For comparison, on a **plain** vector query, `LanceVectorQueryBuilder.minimum_nprobes(0).to_list()` correctly raises:
```
ValueError: minimum_nprobes must be greater than 0
```
(validated in `rust/lancedb/src/query.rs`, `VectorQuery::minimum_nprobes`, and covered by `test_invalid_nprobes_sync` in `python/tests/test_query.py`).

For a **hybrid** query, the equivalent call silently does nothing — no error, and the value is simply discarded, which is surprising for anyone passing `0` intentionally (e.g. programmatically, from a config value) expecting either the value to take effect or a clear error.

## Repro

```python
import lancedb

table = db.create_table(
    "t",
    [
    ],
)

# Expected: raises ValueError("minimum_nprobes must be greater than 0"),
# matching the plain vector-query behavior.
# Actual: silently succeeds, minimum_nprobes(0) is dropped.
table.search(query_type="hybrid").vector([0.0, 0.4]).text("dog").minimum_nprobes(0).to_arrow()
```

## Environment

## Fix

I've opened a PR that changes the check to `is not None` (matching `maximum_nprobes`'s existing check) and adds a regression test in `python/python/tests/test_hybrid_query.py`.

### LINKED ISSUE #3767: bug(python): when_not_matched_by_source_delete() doesn't reset a previously-set condition
## Bug

`LanceMergeInsertBuilder.when_not_matched_by_source_delete()` does not clear a previously-set condition when called again with no argument (or a different type of argument). The builder is left with a stale condition instead of widening the delete to "unconditional," contradicting its own docstring.

## Root cause

In `python/python/lancedb/merge.py`:

```python
    self, condition: Union[str, Expr, None] = None
) -> LanceMergeInsertBuilder:
    """
    ...
    condition: str or :class:`~lancedb.expr.Expr` or None, default None
        If None then all such rows will be deleted.  Otherwise the
        condition will be used as a filter to limit what rows are deleted.
    """
    self._when_not_matched_by_source_delete = True
    if isinstance(condition, Expr):
        self._when_not_matched_by_source_condition_expr = condition._inner
    elif condition is not None:
        self._when_not_matched_by_source_condition = condition
    return self
```

If a builder is configured with `.when_not_matched_by_source_delete("a > 2")` and then reconfigured with `.when_not_matched_by_source_delete()` (no args, meant per the docstring to mean "delete all unmatched rows"), neither branch fires for the second call, so `self._when_not_matched_by_source_condition` still holds the stale `"a > 2"` string from the first call. The builder has no way to widen the delete condition back to unconditional once a string/Expr condition has been set — the "reset to None" case is a no-op instead of clearing the earlier value.

This is inconsistent with every other setter on the same builder (e.g. `when_matched_update_all(where=...)`), which unconditionally overwrite their state field on each call — "last call wins."

## Repro

```python
import lancedb

table = db.create_table("t", data=[{"a": 1, "b": "a"}, {"a": 2, "b": "b"}, {"a": 3, "b": "c"}])
new_data = [{"a": 2, "b": "x"}, {"a": 4, "b": "z"}]

result = (
    table.merge_insert("a")
    .when_matched_update_all()
    .when_not_matched_insert_all()
    .when_not_matched_by_source_delete("a > 2")   # condition set
    .when_not_matched_by_source_delete()          # intended: widen to unconditional
    .execute(new_data)
)

# Expected: num_deleted_rows == 2 (rows a=1 and a=3, both unmatched by source)
# Actual: num_deleted_rows == 1 (still filtered by the stale "a > 2" condition)
print(result.num_deleted_rows)
```

## Environment

## Fix

I've opened a PR that makes each call to `when_not_matched_by_source_delete()` set both the string-condition and Expr-condition fields (one to the new value, the other to `None`), so a later call with a different condition type — or no condition at all — always overwrites, rather than accumulating, the effective condition. Adds a regression test in `python/python/tests/test_table.py`.

Adityaj0:
## Summary

Three small, independent correctness bugs in the Python bindings, each found while auditing `lancedb/query.py` and `lancedb/merge.py`, and each with a regression test.

- `LanceHybridQueryBuilder` (sync hybrid search) silently ignored `.offset()` — it was never forwarded to the vector/FTS sub-queries or applied when slicing the final combined/reranked result, so `.offset(N)` behaved identically to `.offset(0)` with no error. Fixes #3765.
- `LanceHybridQueryBuilder._create_query_builders()` checked `self._minimum_nprobes` for truthiness instead of `is not None` (unlike the adjacent `maximum_nprobes` check, which is already correct). Since `0` is falsy, `.minimum_nprobes(0)` on a hybrid query silently no-opped instead of raising the same `ValueError` a plain vector query raises for the same input. Fixes #3766.
- `LanceMergeInsertBuilder.when_not_matched_by_source_delete()` didn't clear a previously-set condition when called again with no argument (or a different condition type) — the docstring says `None` means "delete all unmatched rows," but a stale condition from an earlier call remained in effect. Fixes #3767.

## Changes

- `python/python/lancedb/query.py`: forward `self._offset` into the per-sub-query fetch limit (`limit + offset`) and apply it when slicing the final result; fix the `minimum_nprobes` truthiness check.
- `python/python/lancedb/merge.py`: `when_not_matched_by_source_delete()` now always overwrites both the string-condition and Expr-condition fields, so the latest call wins (consistent with every other setter on the builder).
- Regression tests added in `python/python/tests/test_hybrid_query.py` and `python/python/tests/test_table.py`.

## Test plan

- [x] `uv run --extra tests pytest python/tests/test_hybrid_query.py -vv` — 14 passed
- [x] `uv run --extra tests pytest python/tests/test_query.py -q -m "not slow and not s3_test"` — 85 passed
- [x] `uv run --extra tests pytest python/tests/test_table.py -q -m "not slow and not s3_test" --deselect python/tests/test_table.py::test_merge_insert` — 127 passed (the deselected test is a pre-existing, unrelated timing-sensitive failure that reproduces identically on `main` without this change)

### DIFF
--- a/python/python/lancedb/merge.py
+++ b/python/python/lancedb/merge.py
@@ -92,8 +92,10 @@ class LanceMergeInsertBuilder(object):
         self._when_not_matched_by_source_delete = True
         if isinstance(condition, Expr):
             self._when_not_matched_by_source_condition_expr = condition._inner
-        elif condition is not None:
+            self._when_not_matched_by_source_condition = None
+        else:
             self._when_not_matched_by_source_condition = condition
+            self._when_not_matched_by_source_condition_expr = None
         return self
 
     def use_index(self, use_index: bool) -> LanceMergeInsertBuilder:
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -2235,6 +2235,7 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
             reranker=self._reranker,
             limit=self._limit,
             with_row_ids=True,
+            offset=self._offset,
         )
         return self._finish_hybrid_results(results)
 
@@ -2256,6 +2257,7 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
         reranker,
         limit: int,
         with_row_ids: bool,
+        offset: Optional[int] = None,
     ) -> pa.Table:
         if norm == "rank":
             vector_results = LanceHybridQueryBuilder._rank(vector_results, "_distance")
@@ -2332,7 +2334,7 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
             score_i = results.column_names.index("_score")
             results = results.set_column(score_i, "_score", original_scores)
 
-        results = results.slice(length=limit)
+        results = results.slice(offset=offset or 0, length=limit)
 
         if not with_row_ids:
             results = results.drop(["_rowid"])
@@ -2679,8 +2681,12 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
 
         # Apply common configurations
         if self._limit:
-            self._vector_query.limit(self._limit)
-            self._fts_query.limit(self._limit)
+            # The final offset/limit window is sliced out of the combined,
+            # reranked results, so each sub-query must fetch enough rows to
+            # cover the skipped prefix as well as the window itself.
+            sub_query_limit = self._limit + (self._offset or 0)
+            self._vector_query.limit(sub_query_limit)
+            self._fts_query.limit(sub_query_limit)
         if self._columns:
             self._vector_query.select(self._columns)
             self._fts_query.select(self._columns)
@@ -2697,7 +2703,7 @@ class LanceHybridQueryBuilder(LanceQueryBuilder):
             self._fts_query.phrase_query(True)
         if self._distance_type:
             self._vector_query.metric(self._distance_type)
-        if self._minimum_nprobes:
+        if self._minimum_nprobes is not None:
             self._vector_query.minimum_nprobes(self._minimum_nprobes)
         if self._maximum_nprobes is not None:
             self._vector_query.maximum_nprobes(self._maximum_nprobes)
--- a/python/python/tests/test_hybrid_query.py
+++ b/python/python/tests/test_hybrid_query.py
@@ -123,6 +123,44 @@ async def test_async_hybrid_query_default_limit(table: AsyncTable):
     assert texts.count("a") == 1
 
 
+def test_hybrid_query_offset(sync_table: Table):
+    # The offset window of a hybrid query must be a suffix of the same query
+    # run without an offset -- it must not be silently ignored.
+    full = (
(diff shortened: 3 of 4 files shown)


Adityaj0:
Splitting this into three separate PRs, one per issue, per feedback: #3769 (offset), #3770 (minimum_nprobes), #3771 (merge_insert condition reset). Closing this one in favor of those.

PR #3769, opened 2026-08-03
fix: hybrid search silently ignores .offset()

### LINKED ISSUE #3765: bug(python): hybrid search silently ignores .offset()

Adityaj0:
## Summary

`LanceHybridQueryBuilder` (sync hybrid search, `table.search(query_type="hybrid")`) silently ignored `.offset()`. `self._offset` was never forwarded to the vector/FTS sub-queries and never applied when slicing the final combined/reranked result, so `.offset(N)` behaved identically to `.offset(0)` — no error, just wrong pagination.

Fixes #3765

## Changes

- `_create_query_builders()`: each sub-query now fetches `limit + offset` rows so there's enough data to slice the correct window out of after combining/reranking.
- `_combine_hybrid_results()` / `to_arrow()`: the final table is sliced with `offset=self._offset` instead of always starting at 0.

## Test plan

- [x] New regression test `test_hybrid_query_offset` in `python/python/tests/test_hybrid_query.py`
- [x] `uv run --extra tests pytest python/tests/test_hybrid_query.py -vv` — 13 passed


Adityaj0:
@wjones127 can you review this too?

wjones127 (approved):
looks good!

PR #3776, opened 2026-08-03
fix(python): consume Lance redirect host fix

anonx3247:
## Summary

- consume the Lance S3 redirect Host fix from lance-format/lance#8205
- make protoc available in isolated Python source builds

## Verification

- cargo check -p lancedb-python


anonx3247:
Closing this tag-based integration branch; the underlying fix is tracked in lance-format/lance#8205.

PR #3777, opened 2026-08-03
fix(listing): paginate table listing instead of enumerating the database

wjones127:
`ListingDatabase::table_names` and `list_tables` listed every table directory under the database prefix and only then applied `limit` and `page_token`. What a request cost was set by the size of the database rather than the size of the page, so asking for the first table in a large database meant one list round trip per thousand tables. Past a few hundred thousand tables that exceeds any client timeout.

Both methods now list through `ObjectStore::read_dir_stream`, which pushes the resume position and the page size into the object store's list request. Asking for ten tables is one request no matter how many tables the database holds. Stores with no paginated list API — the local filesystem, in-memory, S3 Express — fall back to a full listing, which is what every store did before.

Measured against a database of 100,000 tables on S3, listing the first ten went from 11.2s to 0.09s and walking the whole database in pages of 1000 went from 20 minutes to 22s. Those numbers come from LanceDB Enterprise's REST endpoint, where a fix for a per-request connection rebuild landed alongside this change; the round trips are what this PR removes.

## Breaking changes

**Tables come back in storage-key order rather than sorted by name.** The two differ only when one table name is a prefix of another: given tables `users` and `users-archive`, `users-archive` now comes first, because the `-` of `users-archive.lance` sorts below the `.` of `users.lance`. This is the order object stores list in, and a page cannot be both pushed down into the list request and re-sorted by name afterwards. Paging still visits every table exactly once.

**`list_tables` no longer skips a table at each page boundary.** It returned the first name of the *next* page as `page_token`, but `page_token` resumes *after* the name it is given, so a client paging through a database silently lost one table per page. The token is now the last name of the page. Paging with the old code over tables `a` through `e` with `limit=2` returned `["a", "b", "d", "e"]`; it now returns all five.

## Depends on

lance-format/lance#8120, which adds `read_dir_stream`. It is not in a lance release yet, so the first commit here temporarily points the lance dependencies at that PR cherry-picked onto the `v10.1.0-beta.1` tag. Revert that commit to the release tag before merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3422,7 +3422,7 @@ checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 [[package]]
 name = "fsst"
 version = "10.1.0-beta.1"
-source = "git+https://github.com/lance-format/lance.git?tag=v10.1.0-beta.1#68f4d4c1d0c4871b067557c61fc405078f1ab3b7"
+source = "git+https://github.com/lance-format/lance.git?rev=b7ef2bae3c7aeb85da637a8c3bc6643ab7d81564#b7ef2bae3c7aeb85da637a8c3bc6643ab7d81564"
 dependencies = [
  "arrow-array",
  "rand 0.9.5",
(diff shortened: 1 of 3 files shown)


wjones127:
Fixed in https://github.com/lancedb/lancedb/pull/3979

PR #3779, opened 2026-08-04
feat: computed columns defined by SQL expressions

wkalt:
A computed column is defined by a SQL expression rather than by values written at insert time. Declaring one stores the expression and adds the column empty; refresh fills it.

```rust
table.add_columns().computed("doubled", "x * 2").execute().await?;
table.refresh_column("doubled").await?;                    // blocking, returns rows_filled
table.refresh_column_async("doubled").await?.wait().await?; // Job handle, like create_index
```
```python
table.add_columns(computed={"doubled": "x * 2"})
table.refresh_column("doubled")
```
```typescript
await table.addColumns({ computed: [{ name: "doubled", valueSql: "x * 2" }] });
await table.refreshColumn("doubled");
```

`NewColumnTransform::SqlExpressions` already evaluates an expression, but discards it -- rows appended afterward read null with no way to fill them.

The expression is the whole definition: its result type and input columns are derived from it, so a caller writes neither. Expressions that don't parse, unknown input columns, and taken names are rejected at declare time. Inputs are recorded from the expression as written, before simplification -- the optimizer can fold a referenced column out of `true OR x > 0` entirely, and the guard protecting the stored SQL has to see every column the text names. A nested input like `metadata.age` is recorded by path and read through its root.

Refresh makes two passes over each fragment. The first scans only the unfilled live rows and evaluates the expression over them, which yields the exact fill count and the staging decision -- a fragment where nothing would change stages nothing, which is what lets an expression yielding NULL settle after one pass instead of restaging forever. The second streams the fragment's physical rows into a standalone column file a batch at a time, published in one DataReplacement, so peak memory is bounded by a scan batch rather than the fragment's column. The expression is evaluated by the refresh itself, never through a projection alias -- so no internal name can collide with a user column -- and only over the rows being filled: deleted rows and rows already holding a value have their inputs masked to null first, so a poison value in a row nobody is filling (10 / x behind a deleted x = 0) cannot fail the refresh. The contract stays row-level: a row without a value gains one, every other row keeps what it has, and refresh is idempotent without observing input mutation. The contract is still row-level, though -- a row without a value gets one, a row that has one keeps it, because the computed values are merged with the current ones on the is-null mask. So refresh is idempotent, picks up whatever was appended since the last one, and does not observe input mutation: once a row is filled, changing what the expression reads leaves the stored result alone. A row counts as filled only when it gains a value, so an expression yielding NULL settles after one pass rather than reporting the same rows every time.

The first pass is also what finds work a manifest read would miss: a row rewrite (an update, or a compaction folding an unfilled fragment into a filled one) materializes the column as null behind a covering data file, so coverage alone would read those rows as filled and leave them null permanently -- the test for that case fails against a coverage-only implementation. The cost is that a refresh with nothing to do pays a filtered scan per fragment rather than a manifest read.

Because an expression names its inputs as text, renaming, retyping or dropping a column some declaration reads is now refused, as is retyping a computed column itself -- a cast rewrites the field without its metadata, silently destroying the declaration. Declarations are also immutable through update_field_metadata: editing computed_column.* keys would bypass declare-time validation, fabricate a declaration on a plain column, or move a binding out from under a refresh. These are the behaviour changes to existing operations. Nullability changes are unaffected, and a computed column can always be dropped along with its own binding.

Local tables only. The remote protocol has no representation for a stored expression and the server persists declarations under a different vocabulary, so both operations refuse with a message naming the reason rather than guessing at a wire format.

Refresh has both of create_index's forms: a blocking call returning `{rows_filled, version}`, and `refresh_column_async` returning the same `Job` handle `create_index` uses (`refreshColumnAsync` in TypeScript, `AsyncJob` in async Python) -- local jobs run as an in-process task; a future remote implementation returns the server's job. Invalid input is reported by the submitting call, not by the job. Two concurrent refreshes of one column conflict at commit on lance's (fragment, field) CAS -- the loser fails cleanly and a rerun finds nothing left to fill; cancelling mid-refresh can leave staged-but-unreferenced files for cleanup to collect, the same as an aborted index build.

Adds `ColumnNotFound`, `ColumnAlreadyExists`, `NotAComputedColumn`, `InvalidExpression`. The Python `Table` base class gains abstract `refresh_column` / `refresh_column_async` and a keyword-only `computed` on `add_columns`, which is source-breaking for a third-party subclass.

Staleness after an input changes is deliberately not modelled -- that needs a record of what the inputs were at fill time, and the API says so rather than implying a freshness guarantee it hasn't got.


rust/lancedb/src/table/computed_columns.rs
```diff
@@ -0,0 +1,491 @@
...
+/// Field metadata key holding the column's inputs, as a JSON array of names.
+pub const INPUTS_META_KEY: &str = "computed_column.inputs";
+
+/// A computed column's declaration, as read back from field metadata.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ComputedColumn {
+    /// Name of the computed column.
+    pub name: String,
+    /// The SQL expression that defines it.
+    pub expression: String,
```
jackye1995: Should this be typed enum, given this could also be a udf that is not expressed as SQL?

wkalt:
Superseded by the stacked series #3937 → #3938 → #3939 → #3941, which carries this feature split into reviewable layers (declaration, refresh, job handle, remote SDK) plus the hardening from the review rounds since.

@jackye1995 your typed-enum question is resolved in #3937: `ComputedColumnKind` is a `#[non_exhaustive]` enum with the kind persisted under its own metadata key, so a UDF-backed kind is a later additive variant — and an unrecognized kind reads back as a computed column this version cannot evaluate rather than as a plain column.

PR #3782, opened 2026-08-05
fix(node): validate inferred types across records

weimingdiit:
## What does this PR do?

Fixes inferred schema validation for Node.js records containing multiple rows.

The implementation:

- Compares inferred Arrow types by their semantic representation instead of
  object identity.
- Throws an error when a field changes type across records.
- Adds regression tests for consistent and inconsistent inferred types.

## Why is this needed?

Previously, identical inferred types from different records were different objects and failed an identity comparison. More importantly, the resulting `Error` was never thrown, so incompatible values could pass schema inference silently.

## Testing

Added tests covering:

- Consistent numeric types across multiple records.
- Inconsistent numeric/string types across records.

### DIFF
--- a/nodejs/__test__/arrow.test.ts
+++ b/nodejs/__test__/arrow.test.ts
@@ -449,6 +449,20 @@ describe.each([arrow15, arrow16, arrow17, arrow18])(
         );
       });
 
+      it("will accept consistent inferred types across records", function () {
+        const table = makeArrowTable([{ value: 1 }, { value: 2 }]);
+
+        expect(table.getChild("value")?.type.toString()).toEqual(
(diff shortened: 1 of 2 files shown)


weimingdiit:
@claude review

wjones127:
Thanks for submitting a PR! I think I'm going to close this in favor of https://github.com/lancedb/lancedb/pull/3786, which has a more robust fix for this.

weimingdiit:
> Thanks for submitting a PR! I think I'm going to close this in favor of #3786, which has a more robust fix for this.

@wjones127  Thanks for reviewing this! That makes sense, and I agree that #3786 is the better fix.

PR #3790, opened 2026-08-05
fix: get table size from metadata, not files

dantasse:
Some issues:
- file_size_bytes is optional in the manifest, so if it's not there (old writer I guess) it'll under-report the table size.
- it changes results a little bit from the old way by including per-file footers and metadata (probably not a big difference at real scale)


nodejs/src/table.rs
```diff
@@ -1043,7 +1043,10 @@ impl From<lancedb::index::IndexStatistics> for IndexStatistics {
 
 #[napi(object)]
 pub struct TableStatistics {
-    /// The total number of bytes in the table
+    /// The total size, in bytes, of the table's data files
+    ///
+    /// Read from the manifest, so this excludes index files, deletion files,
+    /// overlay files, and manifests.
```
wjones127: suggestion: there's no reason you couldn't get index file and overlay file sizes. Those are both in the manifest. The deletion files and manifest size is harder.

rust/lancedb/src/table.rs
```diff
@@ -3441,9 +3440,10 @@ impl BaseTable for NativeTable {
         let num_rows = self.count_rows(None).await?;
         let num_indices = self.list_indices().await?.len();
         let ds = self.dataset.get().await?;
-        let ds_clone = (*ds).clone();
-        let ds_stats = Arc::new(ds_clone).calculate_data_stats().await?;
-        let total_bytes = ds_stats.fields.iter().map(|f| f.bytes_on_disk).sum::<u64>() as usize;
+        // Sizes come from the manifest. Summing per-field `bytes_on_disk` instead
+        // would open every data file to read its column metadata, which costs one
+        // IO per fragment and reports 0 for legacy v1 storage.
+        let total_bytes = ds.manifest().summary().total_files_size as usize;
```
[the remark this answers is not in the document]
wjones127: Realistically, this is not a huge deal. Any dataset written to with a recent version of Lance is going to have this. I wouldn't worry about it, especially given it's already approximate for different reasons.

wjones127 (changes requested):
I like the approach, but I think we could do a little better by at least including index size (that one is important), and additionally including overlays.

nodejs/src/table.rs
dantasse: ah yeah, good point. updated!

wjones127 (approved):
This is looking good to me!

PR #3897, opened 2026-08-07
feat(python): add backpressure to StreamingDataset post-transform queue

westonpace:
Rename prefetch_batches → io_queue_depth and introduce transform_queue_depth as a symmetric pair: both express "number of batches to buffer per split at this pipeline stage."  The old names are still accepted as keyword arguments but log a deprecation warning redirecting callers to the new names.

transform_queue_depth caps how many transform-result batches can accumulate per split in the post-transform queue.  Without this limit a slow consumer (e.g. a GPU training step) causes cooked rows to pile up unboundedly.  The backpressure check in _try_submit_tx counts both already-cooked rows and rows expected from in-flight transforms; it skips proactive transform submission when the combined total reaches the limit.  The reactive _ensure_cooked path bypasses the check so the consumer never stalls.

### DIFF
--- a/python/python/lancedb/streaming.py
+++ b/python/python/lancedb/streaming.py
@@ -61,7 +61,7 @@ class StreamingDataset(IterableDataset):
 
     Internally ``__iter__`` runs a two-stage pipeline:
 
-    - **Stage 1 (I/O)**: one thread pool with ``num_splits * prefetch_batches``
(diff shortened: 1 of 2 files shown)


hamersaw (approved):
Interesting read to dive through this! Looks reasonable to me.

PR #3911, opened 2026-08-10
feat(lsm): require recorded index catch-up, as an explicit activation

xuanyu-z:
> Stacked on #3780. Blocked only on #3922 (`lance` → `v11.0.0-beta.6`), so CI
> stays red until that lands.

## Missing coverage must mean "not known to be covered"

#3780 caps the SSTable exclusion watermark at an index's recorded catch-up when
there is one, and silently ignores the case where there is none. On a table that
requires catch-up, an absent entry means the index is *not* known to hold the
compacted rows — and the LSM base arm reads base through the index
(`fast_search`, no brute-force tail), so dropping that SSTable loses those rows
for that query.

```rust
Some(caught_up) => watermark = watermark.min(caught_up),
None if catchup_required => watermark = 0,   // retain everything
None => {}
```

`catchup_required` reads the manifest feature bit directly, and requires both
words: a half-set manifest is treated as legacy, which is the conservative side.
Without the bit the field is not maintained at all, so absence carries no
information and behaviour is unchanged.

## Activation, as a table-level entry point

`Table::require_mem_wal_index_catchup()` performs the one-way switch, separate
from `set_lsm_write_spec`: a table carrying the bit retains every generation
until something records catch-up, so it has to follow the deployment of whatever
repairs coverage, not the creation of the table.

This is a convenience, not the only path — a writer holding the dataset calls
the equivalent on `DatasetMemWalExt`, which is what the WAL pod does. Lance
enforces the preconditions either way: the MemWAL index must exist, and the
table must not already carry `compacted_sstables` from before this protocol,
since those numbers cannot be validated.

## Still correct after the Lance rework

lance-format/lance#8481 replaced the transmitted `IndexCatchupAdvance` with a
position derived at commit time from the version a transaction read. That
changed how a writer earns coverage; it did not change what a reader may
conclude from its absence. The rule here, and the field it reads, are unchanged.

## Tests

Existing `exclusion_watermarks` unit tests carry the new argument. Coverage
against a real dataset follows once #3922 lands and this can build.

### DIFF
--- a/rust/lancedb/src/table.rs
+++ b/rust/lancedb/src/table.rs
@@ -637,6 +637,15 @@ pub trait BaseTable: std::fmt::Display + std::fmt::Debug + Send + Sync {
             message: "set_lsm_write_spec is not supported on this table type".into(),
         })
     }
+    /// Switch this table to required index catch-up, one way.
+    ///
+    /// The default implementation returns `NotSupported`. Implementations
+    /// that support the MemWAL LSM write path must override this.
+    async fn require_mem_wal_index_catchup(&self) -> Result<()> {
+        Err(Error::NotSupported {
+            message: "require_mem_wal_index_catchup is not supported on this table type".into(),
+        })
+    }
     /// Remove the [`LsmWriteSpec`] from this table.
     ///
     /// This is a no-op if no spec is currently set.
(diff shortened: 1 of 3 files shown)


hamersaw:
@xuanyu-z overall looks great! My only question is on making this the default in the future. When we decide to make that switch do we just set this flag on `set_lsm_write_spec` and deprecate the function? I guess, what is stopping us from doing that right now? I think before we merge we should have this story figured out.

hamersaw (approved):

PR #3931, opened 2026-08-13
feat: refresh materialized views

wkalt:
A declared view holds no rows; refresh computes them. It pins one source
version, brings the view to exactly the definition's result at that version,
and records the version as a watermark in the view's schema metadata.

It is incremental when it can reconcile what changed: appended rows are
computed and appended, and rows the source deleted or updated are found by
the lance delta and evicted by their __source_row_id provenance, the updated
ones recomputed in the same commit. Compaction rearranges rows without changing
them, so its outputs cost nothing -- which is what keeps routine background
compaction from rebuilding the view. A vacuumed watermark, a
delta the transaction-log walk cannot classify, a Legacy-storage source, or
more staged ids than a fixed cap all fall back to a rebuild; rebuilding an
indexed view swaps every fragment in one Update, so readers never see it
unindexed or empty.

Concurrent refreshes serialize at commit -- each carries the
same sentinel row id in its inserted-rows filter, so the loser lands
nothing. On the append path the watermark moves in a follow-up commit, so a
crash between the two re-appends those rows. Bumps lance
to v11.0.0-beta.19 for the delta reader.

### DIFF
--- a/rust/lancedb/src/lib.rs
+++ b/rust/lancedb/src/lib.rs
@@ -211,7 +211,9 @@ pub use function::FunctionVersion;
 pub use job::Job;
 use lance_index::vector::ApproxMode as LanceApproxMode;
 use lance_linalg::distance::DistanceType as LanceDistanceType;
-pub use materialized_view::{MaterializedView, MaterializedViewDefinition};
+pub use materialized_view::{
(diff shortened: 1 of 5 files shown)


wkalt:
superseded by https://github.com/lancedb/lancedb/pull/4010. Same content, reraised PR to unstick gatekeeper.

PR #3932, opened 2026-08-13
test: differential refresh harness for materialized views

wkalt:
Example tests pin behaviors; the refresh contract is a property: after any
sequence of source mutations, a view maintained by default refreshes equals
the definition evaluated against the source directly, and so does a forced
rebuild. This drives every mutation sequence up to length three -- appends,
deletes, updates crossing the filter, compactions, unrelated column adds --
over an identity and a filtered view shape, checking against an oracle that
shares nothing with the refresh path: a plain column scan with the filter
applied in Rust. The oracle runs after every step because a later
rebuild-forcing mutation silently heals an incremental error; end-state
checks miss exactly the transient bugs that matter. A length-four sweep runs behind
ignore.

Named regressions additionally assert the refresh mode, which value
comparison cannot: a wrongly rebuilding classifier still matches the
oracle, so the append, unrelated-column and compaction cases pin that the
incremental path actually ran.

### DIFF
--- a/rust/lancedb/src/materialized_view.rs
+++ b/rust/lancedb/src/materialized_view.rs
@@ -10,6 +10,10 @@
 //! table. Queries, indexes and search work on the view unchanged.
 
 pub mod refresh;
+
+#[cfg(test)]
+mod differential;
+
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


jackye1995 (approved):
Looks good!

PR #3933, opened 2026-08-13
feat(python): materialized view bindings

wkalt:
Exposes materialized views to Python in both the async and sync clients:
create_materialized_view / open_materialized_view / list_materialized_views
on the connections, and MaterializedView / AsyncMaterializedView handles
carrying the parsed definition and refresh(full=, source_version=), which
returns the typed refresh result. select accepts column names, (alias,
expression) pairs, or a dict of the same; the definition reads back off the
stored schema, so a reopened handle needs no side channel. Remote
connections raise NotImplementedError up front rather than failing deep in
a request, matching the computed-column convention.

### DIFF
--- a/docs/src/python/python.md
+++ b/docs/src/python/python.md
@@ -102,6 +102,12 @@ listing a storage directory.
 
 ::: lancedb.job.AsyncJob
 
+## Materialized Views (Synchronous)
+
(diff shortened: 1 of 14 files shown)


jackye1995 (approved):
Looks good to me!

PR #3937, opened 2026-08-13
feat: declare computed columns by SQL expression

wkalt:
add_columns().computed("doubled", "x * 2") stores the expression in field
metadata and commits the column empty; a later refresh fills it. Type and
inputs are derived from the expression.

The declaration stays authoritative for its lifetime: writes that would give
the column a value (append, update, merge, SQL insert), schema changes that
would break the stored expression or reshape its output, metadata edits,
volatile expressions, declaration metadata arriving through any path but the
validated declare call, and LSM write specs in either order against latest
committed state are all refused. The LSM check also refuses on the mem-wal
catch-up feature flag, which outlives unset and marks retained SSTable rows.
Simultaneous declare/install interleavings conflict at commit via lance's
mem-wal rule (lance#8539). Local tables only.

---


wkalt:
The commit-time exclusion is now upstream: lance-format/lance#8539 (merged as c3ae07a9d) makes a MemWAL-index CreateIndex and a schema-changing Merge mutually conflicting in both rebase directions, so the simultaneous declare/install interleaving fails at commit rather than publishing both. It reaches this repo with the next lance release bump.

LuQQiu:
@wkalt, do we expect computed-column expressions to stay small? Since the SQL expression is persisted in field metadata, it becomes part of the schema stored in every manifest version that carries the column. That seems fine for expressions like `x * 2`, but I wonder if we should add/document a reasonable size expectation or guard to avoid accidentally bloating manifests with large generated expressions. it's a temporary midterm solution that we may convert it to its own file in the future?

LuQQiu (approved):

wkalt:
@LuQQiu I think it should be ok. The expression needs to take column inputs and produce one output; it shouldn't get too crazy. There isn't a length cap but I don't think the exposure is worse than what we have for column names. Most complexity will go into the UDF (python code) when complexity is warranted and that code is stored out of line.

PR #3941, opened 2026-08-13
feat: computed columns on remote tables

wkalt:
LanceDB Cloud and Enterprise support computed columns through the REST API,
so declaration dispatches per backend: local tables plan the expression
themselves, remote ones send {name, computed} entries for the server to
plan. A remote refresh is the server's backfill job -- refresh_column_async
submits it and returns a handle whose successful wait establishes a
read-freshness baseline on the submitting handle, unless a checkout has
pinned the handle by the time the job completes; the blocking form refuses
rather than invent a fill count the server does not report.

Declaration entries are built from the namespace client's AddColumnsEntry
model (lance-namespace 0.11.0, via the lance beta.13 pin), so the payload
shape is compile-checked against the published contract.

---

### DIFF
--- a/docs/src/js/classes/Table.md
+++ b/docs/src/js/classes/Table.md
@@ -79,8 +79,9 @@ input leaves the value computed at fill time; recomputing means dropping
 the column and declaring it again. While a declaration reads a column,
 that column cannot be renamed, retyped or dropped.
 
-Computed columns are local-only: LanceDB Cloud and Enterprise reject a
-declaration.
+On LanceDB Cloud and Enterprise the expression is planned by the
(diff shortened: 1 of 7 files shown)


wkalt:
The `{name, computed}` declaration shape is now part of the published namespace contract: lance-format/lance-namespace#360 adds the `computed` member to `AddColumnsEntry` (same semantics the enterprise server ships: declared all-null, expression persisted as the binding, type and inputs inferred server-side) and widens the backfill route description beyond UDF-backed columns. Generated clients are regenerated in the same change.

LuQQiu (approved):

PR #3957, opened 2026-08-17
feat: expose approx_mode on vector queries

leohoare:
`ApproxMode` landed in the Rust core in #3549, but the Python and Node.js bindings were never plumbed, so `approx_mode` is unreachable from either SDK. Both were built mirroring the existing `ef` plumbing.


leohoare:
@wkalt or @Xuanwo could I please get a review of this PR? 🙇🏻 
Very excited to get ACORN filtering inside a python service we're running.

PR #3959, opened 2026-08-17
feat: materialized views

wkalt:
A materialized view is a table whose contents are defined by a query over one source table and maintained by refresh rather than by writes. The query algebra is projection expressions, an optional filter and an optional row limit over a single source -- no joins, aggregation or ordering. Expressions are DataFusion expressions typed against the source schema; volatile functions are rejected at declaration, since a definition has to evaluate identically across refreshes or incremental maintenance would mix results from different evaluations. Views are local-only and refreshed in-process; there is no auto-refresh.

Declaration: create_materialized_view(name, source) resolves the definition against the source schema -- output types and input columns come from the DataFusion planner, never the caller -- and commits an empty table carrying it as kind-tagged JSON in schema metadata. The tag lets a kind added later read back as a view this version cannot refresh rather than as a plain table. Views open and list as ordinary tables, so queries, indexes and search apply unchanged.

Every view row records the source row that produced it in __source_row_id, which is the source's stable _rowid. Sources must therefore have stable row ids, checked at declaration because the property cannot be enabled after a table exists. That provenance is what makes maintenance exact: it survives source compactions, updates and deletes, and it is the key refresh evicts by.

Refresh pins one source version and brings the view to exactly the definition's result at that version, recording the version as a watermark in the view's schema metadata. It is incremental when it can reconcile what changed: rows the source added are computed and appended, and rows it removed or changed are evicted by provenance id, the changed ones recomputed in the same commit so a reader never sees the view missing them. A compaction rearranges rows without changing them, so its outputs cost nothing -- which is what keeps routine background compaction from rebuilding the view, and is only sound because provenance survives the rewrite. A vacuumed watermark, or a delta the transaction-log classifier cannot separate from a compaction, rebuilds from scratch; rebuilding an indexed view swaps all fragments in one commit that retains index definitions, so readers never see it unindexed or empty. The watermark is stamped in a follow-up commit, so a crash between the two leaves the view correct but visibly unstamped and the next refresh rebuilds. Refreshes of one view are serialized in-process by a lock, and across processes by the commit: the provenance ids ride along as an inserted-rows filter, so two refreshes that planned the same rows conflict and only one lands.

A view inherits the metadata that describes its columns and none that governs how a table is written, so blob markers carry through at every depth while declarations its always-nullable fields would contradict are stripped. Embedding configuration and column definitions are rewritten to the names the view gives those columns, and dropped where it does not project both ends of a function.

This is the whole feature on one diff, for context. It is split for review into a stack, each PR based on the one above it:

- #3930 declarations -- the definition, its metadata, and the stable-row-id gate
- #3931 refresh -- the incremental/rebuild engine and provenance writes
- #3932 a differential harness -- full == incremental == oracle over an op alphabet
- #3933 Python bindings
- #3935 TypeScript bindings

The delta reader refresh uses to find deleted row ids landed in lance-format/lance#8589 and ships in v11.0.0-beta.19, which #3931 pins.


wkalt:
closing - components are merged

PR #3963, opened 2026-08-18
feat: resolve retry config from environment variables

### LINKED ISSUE #2350: RetryConfig env var support
The documentation suggests that retry config can be specified by environment variables (i.e. `LANCE_CLIENT_MAX_RETRIES`), but this does not appear to be supported in code.

https://github.com/lancedb/lancedb/blob/16851389ea4e52fa48ef8198b3c76295c7b06d4c/python/python/lancedb/remote/__init__.py#L59

jsnider3:
`RetryConfig`'s six environment variables are documented in the Rust, Python, and TypeScript APIs but were never read. Each name appears exactly once in the codebase — in a doc comment, never in an `env::var` call — so setting any of them had no effect:

- `LANCE_CLIENT_MAX_RETRIES`
- `LANCE_CLIENT_CONNECT_RETRIES`
- `LANCE_CLIENT_READ_RETRIES`
- `LANCE_CLIENT_RETRY_BACKOFF_FACTOR`
- `LANCE_CLIENT_RETRY_BACKOFF_JITTER`
- `LANCE_CLIENT_RETRY_STATUSES`

`ResolvedRetryConfig::try_from` went straight from the `Option` field to a hardcoded default. The timeout equivalents (`LANCE_CLIENT_TIMEOUT`, `LANCE_CLIENT_CONNECT_TIMEOUT`, `LANCE_CLIENT_READ_TIMEOUT`, `LANCE_CLIENT_CONNECTION_TIMEOUT`) plus `LANCE_CLIENT_MAX_BYTES_PER_REQUEST` and `LANCE_CLIENT_MAX_REQUEST_DURATION` are all wired up already; retry config was the only group that wasn't.

This was noticed in #2349, which filed #2350 as follow-up.

## Approach

Each field resolves with the same precedence the timeout path already uses — an explicit config value wins, then the environment, then the default — following the shape of `get_timeout` and `resolve_max_bytes_per_request` in `client.rs`.

`LANCE_CLIENT_RETRY_STATUSES` parses as a comma-separated list, tolerating whitespace and trailing commas. A blank value yields an empty list, disabling status-based retries, matching what passing an empty vector already does.

Because the resolution lives in the shared core's `TryFrom<RetryConfig>`, the Python, TypeScript, and Java bindings pick it up with no binding changes — they already pass a `RetryConfig` down to this same conversion. No public API surface changes; behavior now matches the documentation that was already published.

## Two panics become errors

Both were reachable before this change, and reading values from the environment widens the exposure, so I fixed them here rather than leaving a new input path pointed at a panic:

- `StatusCode::from_u16(status).unwrap()` panicked on an out-of-range status code. This was already reachable through the public `RetryConfig.statuses` field.
- A negative or non-finite backoff factor or jitter panics later inside `Duration::from_secs_f32` when `next_sleep_time` computes the sleep. These are now rejected at resolution time, from either source.

These are behavior changes rather than pure additions, so I'm happy to split them into a separate PR if you'd rather keep this one to just the env plumbing.

## Tests

Added to the existing `mod tests` in `retry.rs`, reusing the `ENV_MUTEX` / `lock_env()` idiom from `client.rs`:

- defaults when nothing is set
- all six variables read from the environment
- an explicit config value beating the environment, with unset siblings still falling through to defaults
- whitespace and trailing commas in the status list
- a blank status list disabling status-based retries
- an `rstest` case table of nine invalid values, asserting the error names the variable to fix and echoes the offending value
- rejection of invalid values passed directly through the config struct

Verified with `cargo fmt --all --check`, `cargo clippy --quiet --features remote --tests --examples` (clean), and `cargo test --quiet --features remote -p lancedb --lib remote::` (360 passed, 0 failed). Ran the suite repeatedly to confirm the env-mutating tests don't race the `client.rs` tests that also build a `ResolvedRetryConfig`.

Closes #2350

### DIFF
--- a/rust/lancedb/src/remote/retry.rs
+++ b/rust/lancedb/src/remote/retry.rs
@@ -226,6 +226,215 @@ mod tests {
             }
         ));
     }
+    // Serializes the env-var-mutating tests below: cargo test runs tests in
+    // parallel, but these tests read and write the same process-global env
+    // vars (`LANCE_CLIENT_*_RETRIES`, `LANCE_CLIENT_RETRY_*`), so they would
+    // race without this.
+    static ENV_MUTEX: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+    fn lock_env() -> std::sync::MutexGuard<'static, ()> {
+        ENV_MUTEX.lock().unwrap_or_else(|e| e.into_inner())
+    }
+
+    /// Clears every retry env var so a test starts from a known state.
+    fn clear_retry_env() {
+        for var in [
+            "LANCE_CLIENT_MAX_RETRIES",
+            "LANCE_CLIENT_CONNECT_RETRIES",
+            "LANCE_CLIENT_READ_RETRIES",
+            "LANCE_CLIENT_RETRY_BACKOFF_FACTOR",
+            "LANCE_CLIENT_RETRY_BACKOFF_JITTER",
+            "LANCE_CLIENT_RETRY_STATUSES",
+        ] {
+            // SAFETY: This is only called in tests, under `lock_env()`.
+            unsafe {
+                std::env::remove_var(var);
+            }
+        }
+    }
+
+    /// Sets a retry env var. Callers must hold the `lock_env()` guard.
+    fn set_env(var: &str, value: &str) {
+        // SAFETY: This is only called in tests, under `lock_env()`.
+        unsafe {
+            std::env::set_var(var, value);
+        }
+    }
+
+    #[test]
+    fn test_resolve_retry_defaults_when_unset() {
+        let _guard = lock_env();
+        clear_retry_env();
+
+        let resolved: ResolvedRetryConfig = RetryConfig::default().try_into().unwrap();
+
+        assert_eq!(resolved.retries, 3);
+        assert_eq!(resolved.connect_retries, 3);
+        assert_eq!(resolved.read_retries, 3);
+        assert_eq!(resolved.backoff_factor, 0.25);
+        assert_eq!(resolved.backoff_jitter, 0.25);
+        assert_eq!(
+            resolved.statuses,
+            vec![
+                reqwest::StatusCode::CONFLICT,
+                reqwest::StatusCode::TOO_MANY_REQUESTS,
+                reqwest::StatusCode::INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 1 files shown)


jsnider3:
Pushed 09495aa addressing both points from the gate review.

**The backoff panic was real, and my original description of the fix was wrong.** Validating the factor and jitter at resolution time only closed the negative/non-finite *input* path. The exponential itself overflows: `retries` is a `u8`, and with the default factor of 0.25, `Duration::from_secs_f32` starts panicking at 66 request failures.

| `backoff_factor` | panics at `request_failures` |
| --- | --- |
| 0.25 (default) | 66 |
| 1.0 | 64 |
| 100.0 | 58 |

Adding `LANCE_CLIENT_MAX_RETRIES` is exactly what makes that reachable from outside the process, so it belongs in this PR rather than a follow-up. `next_sleep_time` now saturates at `MAX_RETRY_SLEEP` instead of panicking, with a regression test covering 66, 128, and 255 failures plus one below the threshold to confirm normal backoff is untouched.

One judgement call worth your input: I set `MAX_RETRY_SLEEP` to one hour. The value is arbitrary in the sense that any configuration reaching it is already absurd — the default factor and retry count peak at about two seconds, so nothing sane comes near it. If you would rather saturate at `Duration::MAX` to preserve the current behavior exactly for the 14-to-65 failure range that works today, say the word and I will change it.

**On test isolation**, I took a different route than suggested. The environment is now read through a `lookup` closure threaded into `ResolvedRetryConfig::resolve`, with `try_from` supplying the real `std::env::var`. Every parsing and precedence test resolves through an injected map and touches no process-global state. Two tests still set real environment variables, purely to confirm `try_from` is wired to the environment at all.

I avoided subprocess isolation because it seemed heavier than the problem warrants, and I did not want to introduce a test-injection framework that diverges from the `ENV_MUTEX` / `lock_env()` idiom already established in `client.rs`. Happy to go further if you disagree.

The concern was well founded, though: `table.rs` has transport-retry tests that assert retry budgets against a default `RetryConfig`, and they run in parallel with these in the same test binary. Once `try_from` reads the environment, a transiently-set variable could have changed a sibling test's budget.

Verified with `cargo fmt --all --check`, `cargo clippy --quiet --features remote --tests --examples` (clean), and `cargo test --quiet --features remote -p lancedb --lib remote::` — 363 passed, 0 failed, run three times.

PR #3969, opened 2026-08-20
feat: support remote tables in the data loader

jackye1995:
`StreamingDataset`, `PermutationBuilder`, and `Permutation` now work against a `RemoteTable` (LanceDB Cloud and Enterprise). Rows are addressed by `_rowid` exactly as `Table::take_row_ids` does — a supported enterprise query path — so the loader's fetch was already the take path; it just was never allowed to run.

```python
db = lancedb.connect("db://my-db", api_key=..., host_override=...)
ds = StreamingDataset(db.open_table("training"), world_size=8, rank=r)
```

Three things had to change beyond removing the guard in `PermutationBuilder.__init__`.

### Ask for the row id one way

The build scanned `Select::Columns(["_rowid"])` while the fetch used the `with_row_id` flag. Only the flag is accepted for remote tables — the view layer rejects `_rowid` in a select list outright — and `include_row_id` appends it last, which `Splitter` already assumes. `Splitter::project` no longer needs to know about `_rowid` at all.

### Bound the schema probe

`PermutationReader::output_schema` reads the schema off a query plan, and building a plan on a remote table *executes* the query (`create_plan` → `execute_query`). With no limit that is `k = isize::MAX`, so asking a remote table for its output schema pulled the whole table over HTTP and discarded it — once per assigned split, on every epoch, since `StreamingDataset.__iter__` constructs a `Permutation` per split.

### Read the base table

Both the native LSM scanner and the server reject `with_row_id` on a MemWAL table, because the LSM scanner exposes `_rowaddr` rather than a stable `_rowid`. `load_batch` sets `with_row_id`, so **the data loader is broken today on any MemWAL-enabled table, local or remote.** A permutation is defined over `_rowid`, so it now sets `use_lsm(false)` on every base-table read. Consequence worth knowing: rows still in an uncompacted MemWAL are not part of the permutation.

That forced one adjacent fix — `use_lsm(false)` had to stop disabling namespace pushdown. It asks for exactly the base-table read a pushdown performs, and a table that *does* carry a MemWAL spec already forces local execution in `can_execute_namespace_query`, so only `use_lsm(true)` needs to.

### Tests

Mocked-endpoint tests pin the wire shapes (build scan sends `columns: []` + `with_row_id` + `use_lsm: false`; the take sends `_rowid IN (...)`; the schema probe sends `k: 1`), plus a permutation over a MemWAL table, the hash and calculated split projections, and end-to-end `StreamingDataset` and `permutation_builder` runs against a mock server in Python.

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -255,8 +255,8 @@ class Permutations:
 
     Attributes
     ----------
-    base_table: LanceTable
-        The base table that the permutations are based on.
+    base_table: Table
+        The base table that the permutations are based on.  May be a remote table.
     permutation_table: LanceTable
         The permutation table that defines the splits.
     split_names: list[str]
@@ -289,7 +289,7 @@ class Permutations:
     {'train': 0, 'test': 1}
     """
 
-    def __init__(self, base_table: LanceTable, permutation_table: LanceTable):
+    def __init__(self, base_table: Table, permutation_table: LanceTable):
         self.base_table = base_table
         self.permutation_table = permutation_table
 
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -1041,6 +1041,10 @@ class Permutation:
         The returned value is passed through the permutation's current transform,
         so `with_format` and `with_transform` affect this method in the same way
         they affect iteration.
+
+        Offsets must be distinct.  A repeated offset resolves to one row rather than
+        two, which would silently misalign a batch, so it raises instead.  Sampling
+        with replacement is therefore not supported here.
         """
         self._ensure_open()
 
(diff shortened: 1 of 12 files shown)


jackye1995:
Closing in favour of a smaller change; superseded by a fresh PR from `jack/remote-table-dataloader`.

The scope here grew well past what enabling remote tables actually requires. Much of it was built on a premise that turned out to be wrong: I read `_rowid cannot be used in a select statement, pass `with_row_id` instead` as a universal server rule, but it lives in `sophon-view`'s non-materialized view path and does not apply to plain tables — for those, `columns: ["_rowid"]` goes through lance's projection planner, which handles the system column. The machinery for keeping `_rowid` off the wire, and the guards that came with it, were solving a problem the normal table path does not have.

The replacement keeps only what closes a concrete failure: the Python type guard, a bounded schema probe, `use_lsm(false)` plus a base-only row count for MemWAL tables, and making identity-reader construction fallible so a transient remote error is not a panic across the binding boundary.

PR #3970, opened 2026-08-20
feat: support remote tables in the data loader

jackye1995:
`StreamingDataset`, `PermutationBuilder`, and `Permutation` now work against a `RemoteTable` (LanceDB Cloud and Enterprise), which unblocks benchmarking the loader against the enterprise cluster cache.

```python
db = lancedb.connect("db://my-db", api_key=..., host_override=...)
ds = StreamingDataset(db.open_table("training"), world_size=8, rank=r)
```

Rows are addressed by `_rowid` exactly as before — `PermutationReader::load_batch` already built the same `_rowid IN (...)` filter `Table::take_row_ids` sends, so the loader's fetch was always the take path. It just was never allowed to run.

### What changes

**The Python guard.** `PermutationBuilder.__init__` rejected anything without `_inner`, so a `RemoteTable` raised `TypeError` before reaching the PyO3 layer — which already unwraps one via `_table._inner`.

**A bounded schema probe.** `PermutationReader::output_schema` reads the schema off a query plan, and building a plan on a remote table *executes* the query (`create_plan` → `execute_query`). With no limit that is `k = isize::MAX`, so asking a remote table for its output schema pulled the whole table over HTTP and discarded it — once per assigned split, on every epoch, since `StreamingDataset.__iter__` constructs a `Permutation` per split. Now bounded at one row. Native tables never execute the plan, so it costs them nothing.

One row rather than zero, deliberately: lance gates its limit node on `self.limit.unwrap_or(0) > 0`, so `Some(0)` means *no limit*.

**Base-table reads for MemWAL.** A permutation is an addressing of rows by stable `_rowid`, and the LSM scanner exposes `_rowaddr` instead — both backends reject `with_row_id` on a table with a MemWAL write spec. Since `load_batch` needs `_rowid` back to restore the requested order, the loader errors on any such table today, locally too. Every base-table read now sets `use_lsm(false)`.

`count_rows` had no way to say that, and the server rejects the call outright on a WAL table (its own comment: *"lancedb's `count_rows` has no `use_lsm` to mirror"*), so `BaseTable` gains `count_base_rows` — default delegates to `count_rows`, so no implementor changes, and `RemoteTable` overrides it. Consequence, stated in the docstrings: rows in an uncompacted MemWAL are not part of the permutation.

**Fallible identity construction.** `PermutationReader::identity` resolved `inner_new` with `unwrap`. That was near-total against a local dataset, but construction counts the base table — an HTTP round trip for a remote one — so a transient network or auth failure became a panic across the PyO3 boundary.

### Tests

Mocked-endpoint tests pin the wire shapes (base-only count sends `use_lsm: false`, a plain count does not; the schema probe sends `k: 1`), plus end-to-end `permutation_builder` and `StreamingDataset` runs against a mock server in Python — the former torch-free, so it runs wherever the suite does.

### DIFF
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -41,21 +41,21 @@ class PermutationBuilder:
     The permutation is stored in memory and will be lost when the program exits.
     """
 
-    def __init__(self, table: LanceTable):
+    def __init__(self, table: Table):
         """
         Creates a new permutation builder for the given table.
 
         By default, the permutation builder will create a single split that contains all
         rows in the same order as the base table.
+
+        Both local tables and remote tables (LanceDB Cloud and Enterprise) are
+        supported.  Against a remote table the permutation is built by reading every
+        row id over HTTP, so expect the build to cost roughly 8 bytes per row of
+        transfer, and note that it reads the base table only: rows still sitting in a
+        MemWAL that have not been compacted are not part of the permutation, because
+        the LSM scanner does not expose the stable ``_rowid`` a permutation is
+        defined over.
         """
-        if not hasattr(table, "_inner"):
-            raise TypeError(
-                f"PermutationBuilder requires a local LanceTable, "
-                f"got {type(table).__name__}. "
-                "The permutation API is not supported on remote tables. "
-                "Remote tables connect to LanceDB Cloud or Enterprise and do not have "
-                "direct access to the underlying Lance dataset needed for permutations."
-            )
         self._async = async_permutation_builder(table)
 
     def split_random(
@@ -231,7 +231,14 @@ class PermutationBuilder:
         return LOOP.run(do_execute())
 
 
-def permutation_builder(table: LanceTable) -> PermutationBuilder:
+def permutation_builder(table: Table) -> PermutationBuilder:
+    """
+    Creates a new permutation builder for the given table.
+
+    Accepts both local tables and remote tables (LanceDB Cloud and Enterprise); see
+    [PermutationBuilder][lancedb.permutation.PermutationBuilder] for what building a
+    permutation over a remote table costs.
+    """
     return PermutationBuilder(table)
 
 
@@ -248,8 +255,8 @@ class Permutations:
 
     Attributes
     ----------
-    base_table: LanceTable
-        The base table that the permutations are based on.
+    base_table: Table
+        The base table that the permutations are based on.  May be a remote table.
     permutation_table: LanceTable
         The permutation table that defines the splits.
     split_names: list[str]
@@ -282,7 +289,7 @@ class Permutations:
     {'train': 0, 'test': 1}
     """
 
-    def __init__(self, base_table: LanceTable, permutation_table: LanceTable):
+    def __init__(self, base_table: Table, permutation_table: LanceTable):
         self.base_table = base_table
         self.permutation_table = permutation_table
 
--- a/python/python/lancedb/permutation.py
+++ b/python/python/lancedb/permutation.py
@@ -48,13 +48,8 @@ class PermutationBuilder:
         By default, the permutation builder will create a single split that contains all
         rows in the same order as the base table.
 
-        Both local tables and remote tables (LanceDB Cloud and Enterprise) are
-        supported.  Against a remote table the permutation is built by reading every
-        row id over HTTP, so expect the build to cost roughly 8 bytes per row of
-        transfer, and note that it reads the base table only: rows still sitting in a
-        MemWAL that have not been compacted are not part of the permutation, because
-        the LSM scanner does not expose the stable ``_rowid`` a permutation is
-        defined over.
+        Rows sitting in an LSM that have not been flushed to the base table have no
+        row id yet, so they are not part of the permutation.
         """
         self._async = async_permutation_builder(table)
 
@@ -232,13 +227,6 @@ class PermutationBuilder:
 
 
 def permutation_builder(table: Table) -> PermutationBuilder:
-    """
-    Creates a new permutation builder for the given table.
-
-    Accepts both local tables and remote tables (LanceDB Cloud and Enterprise); see
-    [PermutationBuilder][lancedb.permutation.PermutationBuilder] for what building a
-    permutation over a remote table costs.
-    """
     return PermutationBuilder(table)
 
 
@@ -256,7 +244,7 @@ class Permutations:
     Attributes
     ----------
     base_table: Table
-        The base table that the permutations are based on.  May be a remote table.
+        The base table that the permutations are based on.
     permutation_table: LanceTable
         The permutation table that defines the splits.
     split_names: list[str]
--- a/python/python/lancedb/streaming.py
+++ b/python/python/lancedb/streaming.py
@@ -75,16 +75,7 @@ class StreamingDataset(IterableDataset):
     Parameters
     ----------
     table:
-        LanceDB table to stream from.  Local tables and remote tables (LanceDB Cloud
-        and Enterprise) are both supported.
-
-        Against a remote table every fetch is one HTTP request taking
-        ``read_batch_size`` rows by row id, so ``prefetch_batches`` is what hides the
-        round trip: raise it if ``prefetch_queue_depth`` keeps touching zero.
-        Construction is more expensive too — the permutation is built by reading every
-        row id over HTTP, roughly 8 bytes per row, once per rank.  Rows that are still
-        in an uncompacted MemWAL are not included; see
-        [PermutationBuilder][lancedb.permutation.PermutationBuilder].
+        LanceDB table to stream from.
     num_splits:
         Number of fixed splits to partition the table into.  Must be divisible
         by ``world_size``.  When used with DataLoader workers it must also be
--- a/python/python/lancedb/streaming.py
+++ b/python/python/lancedb/streaming.py
@@ -75,7 +75,16 @@ class StreamingDataset(IterableDataset):
     Parameters
     ----------
     table:
-        LanceDB table to stream from.
+        LanceDB table to stream from.  Local tables and remote tables (LanceDB Cloud
+        and Enterprise) are both supported.
+
+        Against a remote table every fetch is one HTTP request taking
+        ``read_batch_size`` rows by row id, so ``prefetch_batches`` is what hides the
+        round trip: raise it if ``prefetch_queue_depth`` keeps touching zero.
+        Construction is more expensive too — the permutation is built by reading every
+        row id over HTTP, roughly 8 bytes per row, once per rank.  Rows that are still
+        in an uncompacted MemWAL are not included; see
+        [PermutationBuilder][lancedb.permutation.PermutationBuilder].
     num_splits:
         Number of fixed splits to partition the table into.  Must be divisible
         by ``world_size``.  When used with DataLoader workers it must also be
--- a/rust/lancedb/src/dataloader/permutation/builder.rs
+++ b/rust/lancedb/src/dataloader/permutation/builder.rs
@@ -242,6 +242,13 @@ impl PermutationBuilder {
         // First pass, apply filter and load row ids.  A permutation addresses rows by
         // stable `_rowid`, which the MemWAL LSM scanner does not expose, so it always
         // reads the base table.
+        //
+        // This scan's row order is load-bearing: `Shuffler` permutes positions rather
+        // than values, so the permutation is a function of (scan order, seed).  Every
+        // rank builds its own permutation and they have to agree, or splits overlap and
+        // rows are trained twice or not at all.  Locally that holds because a scan
+        // reads fragments in manifest order; remotely it depends on the server
+        // returning a stable order for the same table version.
         let mut rows = self
             .base_table
             .query()
--- a/rust/lancedb/src/dataloader/permutation/builder.rs
+++ b/rust/lancedb/src/dataloader/permutation/builder.rs
@@ -239,16 +239,12 @@ impl PermutationBuilder {
 
     /// Builds the permutation table and stores it in the given database.
     pub async fn build(self) -> Result<Table> {
-        // First pass, apply filter and load row ids.  A permutation addresses rows by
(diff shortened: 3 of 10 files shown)


rust/lancedb/src/dataloader/permutation/builder.rs
```diff
@@ -238,8 +239,14 @@ impl PermutationBuilder {
...
     /// Builds the permutation table and stores it in the given database.
     pub async fn build(self) -> Result<Table> {
-        // First pass, apply filter and load row ids
-        let mut rows = self.base_table.query().select(Select::columns(&[ROW_ID]));
+        // First pass, apply filter and load row ids.  A permutation addresses rows by
+        // stable `_rowid`, which the MemWAL LSM scanner does not expose, so it always
+        // reads the base table.
+        let mut rows = self
+            .base_table
+            .query()
```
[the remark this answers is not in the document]
jackye1995: Real finding, thanks — the mechanism checks out: `Shuffler::shuffle_batch` permutes positions (`indices = (0..num_clumps)`, shuffled, then `take_record_batch`), and the scan has no `order_by`, so the permutation is a function of (scan order, seed) rather than (row-id set, seed). With each rank building its own permutation, disagreeing ranks would overlap splits and train some rows twice while never touching others.

Deliberately not fixing this client-side. The determinism guarantee is going to be made server-side, and since remote support in the data loader is not a released feature yet, landing the client change ahead of it is acceptable — there is no user relying on the current behaviour to break.

I have recorded the dependency at the scan site so the contract the server needs to satisfy is written down next to the code that assumes it, rather than being implicit in the positional shuffle.

python/python/lancedb/permutation.py
```diff
@@ -41,21 +41,21 @@ class PermutationBuilder:
...
         By default, the permutation builder will create a single split that contains all
         rows in the same order as the base table.
+
+        Both local tables and remote tables (LanceDB Cloud and Enterprise) are
+        supported.  Against a remote table the permutation is built by reading every
+        row id over HTTP, so expect the build to cost roughly 8 bytes per row of
+        transfer, and note that it reads the base table only: rows still sitting in a
+        MemWAL that have not been compacted are not part of the permutation, because
+        the LSM scanner does not expose the stable ``_rowid`` a permutation is
+        defined over.
```
westonpace: This all seems like extraneous information?

I'd expect all features to work on both tables by default.  We should only need to comment when a feature is specific to one or the other?

Also, both local and remote tables need to scan in the row ids so this is nothing new for remote tables.  The comment seems to say this will be special.

The LSM thing we could probably include as a brief "will not include rows sitting in the LSM because the _rowid is unstable"

python/python/lancedb/permutation.py
```diff
@@ -231,7 +231,14 @@ async def do_execute():
...
 
 
-def permutation_builder(table: LanceTable) -> PermutationBuilder:
+def permutation_builder(table: Table) -> PermutationBuilder:
+    """
+    Creates a new permutation builder for the given table.
+
+    Accepts both local tables and remote tables (LanceDB Cloud and Enterprise); see
+    [PermutationBuilder][lancedb.permutation.PermutationBuilder] for what building a
+    permutation over a remote table costs.
```
westonpace: Again, nothing special needs to be said here right?

python/python/lancedb/permutation.py
```diff
@@ -248,8 +255,8 @@ class Permutations:
 
     Attributes
     ----------
-    base_table: LanceTable
-        The base table that the permutations are based on.
+    base_table: Table
+        The base table that the permutations are based on.  May be a remote table.
```
westonpace:
```suggestion
        The base table that the permutations are based on.
```

python/python/lancedb/streaming.py
```diff
@@ -75,7 +75,16 @@ class StreamingDataset(IterableDataset):
...
+        LanceDB table to stream from.  Local tables and remote tables (LanceDB Cloud
+        and Enterprise) are both supported.
+
+        Against a remote table every fetch is one HTTP request taking
+        ``read_batch_size`` rows by row id, so ``prefetch_batches`` is what hides the
+        round trip: raise it if ``prefetch_queue_depth`` keeps touching zero.
+        Construction is more expensive too — the permutation is built by reading every
+        row id over HTTP, roughly 8 bytes per row, once per rank.  Rows that are still
+        in an uncompacted MemWAL are not included; see
+        [PermutationBuilder][lancedb.permutation.PermutationBuilder].
```
westonpace: This information is true for local tables as well.  Also, I'm not sure that we should give special advice to `prefetch_batches` and `prefetch_queue_depth` here.  All of these parameters matter.

rust/lancedb/src/dataloader/permutation/builder.rs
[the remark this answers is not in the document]
westonpace: Why wouldn't this be a client side change to add `order_by(['_rowid'])`?

rust/lancedb/src/dataloader/permutation/builder.rs
```diff
@@ -238,8 +239,21 @@ impl PermutationBuilder {
...
+        // First pass, apply filter and load row ids.  A permutation addresses rows by
+        // stable `_rowid`, which the MemWAL LSM scanner does not expose, so it always
+        // reads the base table.
+        //
+        // This scan's row order is load-bearing: `Shuffler` permutes positions rather
+        // than values, so the permutation is a function of (scan order, seed).  Every
+        // rank builds its own permutation and they have to agree, or splits overlap and
+        // rows are trained twice or not at all.  Locally that holds because a scan
+        // reads fragments in manifest order; remotely it depends on the server
+        // returning a stable order for the same table version.
```
westonpace: What does the LSM scanner not expose?  Is the `_rowid` null?  Or can it change without a version change?  A permutation is expected to be pinned to a specific version so there is no reliance on a stable row id feature.

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -97,8 +97,13 @@ impl PermutationReader {
         Self::inner_new(base_table, Some(permutation_table), split).await
     }
 
-    pub async fn identity(base_table: Arc<dyn BaseTable>) -> Self {
-        Self::inner_new(base_table, None, 0).await.unwrap()
+    /// A reader over the base table in storage order, with no permutation.
+    ///
+    /// Fallible: construction counts the base table, which for a remote table is an
+    /// HTTP round trip, so a transient network or auth failure must surface as an
+    /// error rather than a panic across a binding boundary.
```
westonpace: Technically a `count_rows` could fail even on a native table.  So I don't think we need to call out remote tables specially here.

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -203,6 +209,9 @@ impl PermutationReader {
             filter: Some(QueryFilter::Datafusion(col(ROW_ID).in_list(in_list, false))),
             select: selection,
             with_row_id: true,
+            // The permutation is defined over stable `_rowid`s, which the MemWAL LSM
+            // scanner does not expose — and which it rejects `with_row_id` for.
+            use_lsm: Some(false),
```
westonpace: We are specifically querying for specific row ids here, why do we need to disable LSM?  Is this a performance fix (skip LSM if we know it won't match any rows)?

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -487,7 +500,21 @@ impl PermutationReader {
...
+        //
+        // One row, not zero: lance reads a limit of `Some(0)` as *no limit* rather than
+        // no rows (`Scanner`: `self.limit.unwrap_or(0) > 0` gates the limit node), so a
+        // zero here would scan the whole table.  Native tables never execute the plan,
+        // so the bound costs them nothing.
+        table
+            .query()
+            .select(selection)
+            .limit(1)
+            .use_lsm(false)
```
westonpace: We probably don't really need `use_lsm(false)` here right?

rust/lancedb/src/dataloader/permutation/reader.rs
```diff
@@ -487,7 +500,21 @@ impl PermutationReader {
...
         let table = Table::from(self.base_table.clone());
-        table.query().select(selection).output_schema().await
+        // `output_schema` reads the schema off a query plan, and building a plan on a
+        // remote table executes the query.  Unbounded that is a full table scan over
+        // HTTP for every `Permutation` constructed — once per split, per epoch.
+        //
+        // One row, not zero: lance reads a limit of `Some(0)` as *no limit* rather than
+        // no rows (`Scanner`: `self.limit.unwrap_or(0) > 0` gates the limit node), so a
+        // zero here would scan the whole table.  Native tables never execute the plan,
+        // so the bound costs them nothing.
```
westonpace:
```suggestion
        // limit(1) because some table types may require executing the
        // query to determine the output schema
```

These comments get a little extraneous

westonpace (changes requested):
This `use_lsm=false` stuff is pretty confusing.  Do we document somewhere how LSM affects row ids?  Does it allow row ids to change without the version changing?

Also, this PR is adding a lot of comment bloat and I think most of it is extraneous.  Can we trim it down?

rust/lancedb/src/dataloader/permutation/builder.rs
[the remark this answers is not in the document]
jackye1995: My goal here is to make the server return deterministic results for a given table version, which would remove the need for an `order_by` on the client entirely. Would you be okay with that direction, or do you think making that guarantee server side would be challenging?

rust/lancedb/src/dataloader/permutation/builder.rs
jackye1995: You're right that "stable" was the wrong word — I was using it to mean "doesn't change", which collides with the stable row id feature, and a version-pinned permutation doesn't rely on that. Dropped it.

What I actually meant is narrower: rows that are still in the LSM have not been flushed to the base table, so they have no row id assigned yet. Not null, and not changing underneath us — simply not there to reference. That is why they are outside the data loader's view. The comment now says that.

Worth saying that this isn't especially well defined at the moment. Pinned to a specific version we could reference the MemWAL index to read the WAL as of that point in time and include the data as it was then. That feels quite niche for now, but it does mean we need more concrete definitions on this front rather than leaning on an informal notion of what a version contains.

rust/lancedb/src/dataloader/permutation/reader.rs
jackye1995: Not a performance fix. Rows still in the LSM have no row id yet, so they cannot match a row id take at all — excluding them is about what the data loader can see, not about skipping work. Reworded the comment to say that.

rust/lancedb/src/dataloader/permutation/reader.rs
jackye1995: Agreed, removed — the probe reads no rows so it has no reason to say anything about the LSM.

python/python/lancedb/permutation.py
jackye1995: Trimmed to just the LSM note. You're right that the rest was extraneous: features should work on both table types by default, and scanning row ids is nothing new for remote.

python/python/lancedb/permutation.py
jackye1995: Nothing special — docstring removed.

python/python/lancedb/permutation.py
jackye1995: Applied.

python/python/lancedb/streaming.py
jackye1995: Removed. This was all true for local tables too, and singling out `prefetch_batches` / `prefetch_queue_depth` over the other parameters wasn't warranted.

rust/lancedb/src/dataloader/permutation/reader.rs
jackye1995: Applied.

rust/lancedb/src/dataloader/permutation/reader.rs
jackye1995: Coming back to this one — I removed it, and it turned out to be needed after all, so I have put it back.

A `Permutation` keeps its selection as a dict, which PyO3 turns into `Select::Dynamic`, and the LSM scanner rejects that shape outright (`lsm.rs`: `Select::Dynamic(_) | Select::Expr(_) => unsupported("dynamic column projection")`). So on a MemWAL table `Permutation.schema`, `column_names`, `num_columns`, `shape`, and the empty-offset fallback all raised:

```
not supported: the MemWAL LSM scanner does not support dynamic column projection;
set use_lsm(false) to read the base table only
```

The reason it is needed is narrower than the reads: not that the LSM would return wrong rows, but that it cannot serve this projection shape at all. Metadata now reads the same base-table view the permutation reads do, and there is a test that builds a permutation over a MemWAL table and asks for the schema with a dynamic selection — it fails without the flag.

Worth noting my remote mock only probed `Select::All`, so nothing I had written could have caught this.

jackye1995: Correcting my last note here: I have taken it back out, so this stays as you suggested.

Checking main, `output_schema` has never pinned the base table, so permutation metadata over a MemWAL table already goes through the LSM scanner there and hits the same dynamic-projection limit. It is a pre-existing gap rather than something this change introduces, and it belongs in its own change covering the MemWAL metadata path properly rather than being smuggled into remote table support.

jackye1995:
Superseded by a cleaner PR from `jack/remote-table-loader`.

The MemWAL handling here was spread across four `use_lsm(false)` call sites plus a new `BaseTable::count_base_rows` and a `RemoteTable` override, which is what most of the review discussion ended up being about. Replacing all of it with a single pre-check — refuse a table that has an LSM write spec, since rows not yet flushed to the base table have no row id for a permutation to reference — removes that surface entirely and is a better answer anyway: reading around those rows would have dropped them from training without saying so.

`rust/lancedb/src/table.rs` and `rust/lancedb/src/remote/table.rs` are untouched in the replacement.

PR #3972, opened 2026-08-20
fix(python): accept tuple and Arrow vector queries

tandede:
## Summary

Sync `Table.search()` declares tuple, `pyarrow.Array`, and `pyarrow.ChunkedArray` query vectors as supported, and the async query path already accepts them. The sync query builder only classified lists and NumPy arrays as vectors, however, so these inputs were routed through embedding/FTS resolution and then rejected. Arrow and tuple inputs also did not contribute a query dimension when selecting among multiple vector columns.

This change:

- recognizes the supported tuple and Arrow containers as sync vector queries for both `auto` and explicit `vector` query types;
- converts them to the existing internal `float32` NumPy representation at the sync builder boundary;
- infers dimensions from tuple, primitive Arrow, and list-valued Arrow queries so schemas with differently sized vector columns select the matching column.

The regression test uses two vector columns with dimensions 2 and 3. It verifies tuple, Arrow array, and chunked Arrow queries through both sync query modes and confirms the dimension-2 column is selected.

## Test plan

- `pytest python/tests/test_table.py::test_sync_search_accepts_supported_vector_containers -q` (6 passed)
- `pytest python/tests/test_util.py python/tests/test_query.py python/tests/test_table.py -q -m "not slow and not s3_test"` (308 passed, 1 skipped, 1 deselected)
- `pytest python/tests -q -m "not slow and not s3_test" --ignore=python/tests/docs` (981 passed, 25 skipped, 65 deselected)
- `ruff format --check python`
- `ruff check python`
- `cargo fmt --all --check`


tandede:
Hi maintainers, I’m following up on this PR in case it was missed. The PR checks and security scan have passed, and the automated gatekeeper recommends approval. The tuple and Arrow vector-query support is ready for a maintainer review, and I’d be happy to address any feedback. Thank you!

PR #3985, opened 2026-08-21
feat: add first-class function wire contracts

Xuanwo:
## Problem

Enterprise Function-backed computed columns need a stable SDK contract before Sophon catalog and execution endpoints can be added. The existing `Job` API can only represent unit terminal results, and there is no shared Rust/Python wire definition for immutable Function versions, applications, bindings, or refresh results.

## Behavior

This introduces remote-only canonical Function values in Rust and Python, evolves `Job<T = ()>` to decode typed remote terminal results while keeping local spawned operations unit-typed, and fixes the cross-language contract with shared JSON golden fixtures. Unknown fields and discriminator values remain forward-decodable, while canonical output contains only fields known to the client. Function models contain secret names only.

Sophon remains the sole owner of catalog persistence, environment bake, secret resolution, execution, and publication. This PR does not add authoring/catalog endpoints, local execution, refresh runners, or live Sophon E2E coverage.


python/python/lancedb/functions.py
```diff
@@ -0,0 +1,269 @@
...
+        return self.dict(exclude_none=True, exclude_defaults=True)
+
+    def _copy(self, *, update: Mapping[str, Any]):
+        update = {name: _freeze_value(value) for name, value in update.items()}
+        if _PYDANTIC_V2:
+            return self.model_copy(update=update)
+        return self.copy(update=update)
+
+    def to_canonical_json(self) -> str:
+        return json.dumps(
```
BubbleCal, line 137: `json.dumps` does not produce the same number representation as serde_json. For an `ApplicationInput.value` of `1e-7`, Python emits `1e-07` while Rust emits `1e-7`, so the canonical bytes and any digest or golden derived from them diverge across languages. Please use a shared canonical JSON specification or restrict the value grammar, and add numeric cross-language goldens.

python/python/lancedb/job.py
```diff
@@ -44,18 +46,20 @@ async def status(self) -> str:
...
         if it was cancelled, and `TimeoutError` if `timeout` elapses first.
         """
         if self._inner is None:
-            return
+            return cast(T, None)
         if timeout is None:
-            await self._inner.wait()
-        else:
-            await asyncio.wait_for(self._inner.wait(), timeout.total_seconds())
+            return cast(T, await self._inner.wait())
```
BubbleCal: `cast` only changes the static type. `_lancedb.Job.wait()` still wraps `lancedb::Job<()>` and returns `None`, so a future `AsyncJob[FunctionVersion]` would type-check but return `None` at runtime. Please carry a typed or raw result decoder across the PyO3 boundary, or defer the Python generic surface until that path exists, with a typed wait test.

python/python/lancedb/functions.py
```diff
@@ -0,0 +1,269 @@
...
+
+    name: str
+    version: str
+    artifact: FunctionArtifact
+    signature: FunctionSignature
+    runtime: PythonRuntimeSpec
+    runtime_digest: str
+    environment_digest: str
+    required_secrets: tuple[str, ...] = ()
+    created_at: str
```
BubbleCal, line 248: The fixture already contains `resource_defaults`, and the current design makes those defaults part of immutable `FunctionVersion`, but both Python and Rust ignore the field and canonical output drops it. Please model it in both languages and include it in the shared golden, or explicitly scope this type as partial rather than canonical.

python/python/lancedb/functions.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class RefreshColumnResult(_RemoteValue):
+    """Terminal result of a remote Function-column refresh Job."""
+
+    rows_assigned: int
+    rows_failed: int
+    rows_remaining: int
+    source_version: int
+    published_version: Optional[int]
```
BubbleCal: In Pydantic v2, `Optional[int]` without a default is still required. Rust explicitly accepts a missing `published_version` as `None` and omits it from canonical JSON, so Rust output with no published version cannot be decoded by Python. Please set the default to `None` and add a missing-field cross-language fixture.

python/python/lancedb/functions.py
```diff
@@ -0,0 +1,269 @@
...
+    nullable: bool
+
+
+class OutputMapping(_RemoteValue):
+    result_field: str
+    output_name: str
+    output_field_id: int
+    output_ordinal: int
+    arrow_type: str
+    nullable: bool
```
BubbleCal, line 329: The current design says each persisted output mapping includes `assignment_flag_id`, but this wire model omits it while exposing `output_field_id` and the binding revision. If clients intentionally must not see flag IDs, please document that boundary; otherwise include the field in both models and goldens before freezing the binding contract.

python/python/lancedb/functions.py
Xuanwo: Agreed. Slice 1 will restrict the canonical literal grammar to stable JSON integers, strings, booleans, nulls, arrays, and objects, and reject floating-point literals in both Rust and Python until a language-neutral numeric representation is defined. I will cover the same rejected payload from a shared fixture.

python/python/lancedb/job.py
Xuanwo: Agreed. Because Slice 1 does not add typed submit endpoints or a PyO3 result bridge, the Python generic surface is premature and can make a unit result look typed. I will defer Job[T] on Python while retaining the Rust typed remote-result primitive; Python generics can be added when the runtime result crosses PyO3.

python/python/lancedb/functions.py
Xuanwo: The current contract intentionally keeps scheduling resources, priority, concurrency, and retry policy on the Job request rather than in immutable FunctionVersion identity. The resource_defaults entry is stale fixture vocabulary, not a canonical field. I will remove it from this fixture, keep forward-field coverage explicit, and document the identity boundary rather than model the field.

python/python/lancedb/functions.py
Xuanwo: Agreed. I will make published_version genuinely optional in Python and cover missing-field canonical round-trip in both Rust and Python, including the supported Pydantic majors.

python/python/lancedb/functions.py
Xuanwo: The current transition contract intentionally has no assignment flag ID: Lance does not yet expose the cell flag used by the future design, and Slice 1 uses NULL-as-unassigned with fragment work units. I will document that assignment state is outside the client OutputMapping wire model; CellFlag adoption is a later migration and should not freeze a placeholder ID here.

rust/lancedb/src/remote/job.rs
```diff
@@ -104,40 +135,39 @@ impl<S: HttpSend> RemoteJob<S> {
...
     }
 
-    async fn wait(&self) -> Result<()> {
+    async fn wait(&self) -> Result<T> {
         let mut interval = INITIAL_POLL_INTERVAL;
         loop {
-            let description = self.describe().await?;
-            match description.job_state {
-                JobState::Done => return Ok(()),
+            let (request_id, description) = self.describe().await?;
```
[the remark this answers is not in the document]
Xuanwo: Agreed. I will separate lifecycle polling from result decoding: the backend returns a raw successful terminal envelope, Job<()> ignores any additive payload, and typed Job<T> decodes it. The unit remote fixture will contain an object result so the resume-by-ID behavior is covered directly.

python/python/lancedb/functions.py
```diff
@@ -0,0 +1,269 @@
...
+    @classmethod
+    def from_json(cls, payload: str):
+        if _PYDANTIC_V2:
+            return cls.model_validate_json(payload)
+        return cls.parse_raw(payload)
+
+    def _known_dict(self) -> dict[str, Any]:
+        if _PYDANTIC_V2:
+            return self.model_dump(exclude_none=True, exclude_defaults=True)
+        return self.dict(exclude_none=True, exclude_defaults=True)
```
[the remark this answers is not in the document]
Xuanwo: Agreed. I will replace reliance on Pydantic-specific exclude_defaults behavior with an explicit recursive known-field encoder that omits default-factory values consistently. The shared goldens will be exercised under Pydantic 1.10 and the current Pydantic 2 environment.

docs/src/python/python.md
```diff
@@ -54,6 +54,22 @@ listing a storage directory.
 
 ::: lancedb.LsmWriteSpec
 
+## Functions and Jobs
```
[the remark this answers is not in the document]
Xuanwo: Agreed on the public API visibility gap. I will add directives for every symbol in lancedb.functions.__all__ and a regression that keeps the hand-maintained reference synchronized with the exported surface.

BubbleCal (approved):

PR #4010, opened 2026-08-21
feat: refresh materialized views

wkalt:
A declared view holds no rows; refresh computes them. It pins one source
version, brings the view to exactly the definition's result at that version,
and records the version as a watermark in the view's schema metadata.

It is incremental when it can reconcile what changed: appended rows are
computed and appended, and rows the source deleted or updated are found by
the lance delta and evicted by their __source_row_id provenance, the updated
ones recomputed in the same commit. Compaction rearranges rows without changing
them, so its outputs cost nothing -- which is what keeps routine background
compaction from rebuilding the view. A vacuumed watermark, a
delta the transaction-log walk cannot classify, a Legacy-storage source, or
more staged ids than a fixed cap all fall back to a rebuild; rebuilding an
indexed view swaps every fragment in one Update, so readers never see it
unindexed or empty.

Concurrent refreshes serialize at commit -- each carries the
same sentinel row id in its inserted-rows filter, so the loser lands
nothing. On the append path the watermark moves in a follow-up commit, so a
crash between the two re-appends those rows. Bumps lance
to v11.0.0-beta.19 for the delta reader.

### DIFF
--- a/rust/lancedb/src/lib.rs
+++ b/rust/lancedb/src/lib.rs
@@ -211,7 +211,9 @@ pub use function::FunctionVersion;
 pub use job::Job;
 use lance_index::vector::ApproxMode as LanceApproxMode;
 use lance_linalg::distance::DistanceType as LanceDistanceType;
-pub use materialized_view::{MaterializedView, MaterializedViewDefinition};
+pub use materialized_view::{
+    MaterializedView, MaterializedViewDefinition, RefreshMaterializedViewResult, RefreshMode,
+};
 /// Re-export of the [`metrics`](https://docs.rs/metrics) crate facade. Enable
 /// the `metrics` feature to publish LanceDB's internal metrics; install any
 /// `metrics`-compatible recorder to collect them. See also [`metrics_otel`] for
--- a/rust/lancedb/src/materialized_view.rs
+++ b/rust/lancedb/src/materialized_view.rs
@@ -9,6 +9,7 @@
 //! metadata; a kind added later reads back as unrefreshable, not as a plain
 //! table. Queries, indexes and search work on the view unchanged.
 
+pub mod refresh;
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -13,20 +13,20 @@ categories = ["database-implementations"]
...
-lance-namespace = { "version" = "=11.0.0-beta.18", "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
-lance-namespace-impls = { "version" = "=11.0.0-beta.18", default-features = false, "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
-lance-table = { "version" = "=11.0.0-beta.18", "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
-lance-testing = { "version" = "=11.0.0-beta.18", "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
-lance-datafusion = { "version" = "=11.0.0-beta.18", "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
-lance-encoding = { "version" = "=11.0.0-beta.18", "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
-lance-arrow = { "version" = "=11.0.0-beta.18", "tag" = "v11.0.0-beta.18", "git" = "https://github.com/lance-format/lance.git" }
+lance = { "version" = "=11.0.0-beta.19", default-features = false, "tag" = "v11.0.0-beta.19", "git" = "https://github.com/lance-format/lance.git" }
```
jackye1995, line 16: Is this version bump required? Ideally we always use he auto version bump so that it covers all the necessary module updates.

jackye1995 (approved):
Approve this for now to unblock downstream. But I'm not quite aligned on how we are doing the incremental update, it feels pretty fragile. We need a 2x2 matrix across all different transactions to be able to derive a change data feed between 2 versions of a lance table. That probably deserves a whole project on its own.

PR #4013, opened 2026-08-21
feat: return typed refresh job results

Xuanwo:
## Problem

`refresh_column_async` returned a unit-result job even though durable refresh jobs carry a canonical terminal result. Python callers could not obtain row counts or source and published versions through the public `Job` API, and local and remote refresh jobs exposed different result semantics.

## Behavior

`refresh_column_async` now returns `Job[RefreshColumnResult]` for local and remote tables. The general typed-job bridge binds each endpoint to its public result model while preserving unit-result jobs and existing status, wait, cancel, and timeout behavior. A local no-op refresh reports no published version.

The Node.js API continues to resolve `wait()` as `void`; its binding erases the Rust result type internally to preserve the existing public contract.

## Ownership and integration boundary

LanceDB owns the language-neutral `Job<T>` contract and language-binding decode. Sophon owns production and durable persistence of terminal payloads. Sophon #7348 and #7378 now publish the canonical refresh result for Function-backed and expression-backed refresh jobs, respectively. The remote client fixture matches the merged server schema; live deployment and end-to-end demo acceptance remain separate rollout checks.


python/python/lancedb/_lancedb.pyi
```diff
@@ -228,7 +228,7 @@ class Job:
     async def wait(self) -> None: ...
     async def cancel(self) -> None: ...
 
-class FunctionJob:
+class TypedJob:
```
wkalt: why can't we call it Job? Prefixing with "Typed" seems kinda weird to me

Xuanwo: updated.

BubbleCal (approved):

PR #4028, opened 2026-08-22
fix(python): apply offset when combining async hybrid results

### LINKED ISSUE #3765: bug(python): hybrid search silently ignores .offset()

### LINKED ISSUE #4027: Async hybrid query with .offset() repeats and drops rows across pages
### LanceDB version

`main` (0.38.0-beta.3), reproduced against the published 0.37.1 wheel as well

### What happened?

Paginating an **async** hybrid query with `.offset()` returns rows that were already on the previous page, and silently drops rows that belong on the current one.

```python
full   = await table.query().nearest_to(vec).nearest_to_text("dog").limit(4).with_row_id().to_arrow()
page1  = await table.query().nearest_to(vec).nearest_to_text("dog").limit(2).with_row_id().to_arrow()
page2  = await table.query().nearest_to(vec).nearest_to_text("dog").offset(2).limit(2).with_row_id().to_arrow()

full["_rowid"].to_pylist()    # [3, 0, 2, 1]
page1["_rowid"].to_pylist()   # [0, 3]
page2["_rowid"].to_pylist()   # [3, 1]   <- rowid 3 repeats from page 1, rowid 2 never appears
```

Expected `page2` to equal `full[2:]`, i.e. `[2, 1]`.

`.offset()` on a hybrid query pushes the offset down into **both** sub-queries (`HybridQuery::offset` → `inner_vec` + `inner_fts`), so each of the vector and FTS sides independently skips its own first N rows *before* the two are fused and reranked. `AsyncHybridQuery.to_batches` then calls `_combine_hybrid_results(..., limit=self._inner.get_limit())` without passing `offset=`, so the combined table is sliced from position 0 and the sub-query limits are never raised to `limit + offset`. The window ends up taken from the wrong place on both sides.

The **sync** path had exactly this bug and it was fixed in #3769. The async twin was not updated — and this was known at the time: the author of #3765 explicitly scoped it out, noting "the async path ... applies offset independently to each sub-query before combining/reranking ... That's a separate, Rust-level concern and is out of scope for this issue". As far as I can tell no follow-up issue was ever filed, so I'm filing one.

Only the async hybrid path is affected. Sync hybrid is correct since #3769, and plain vector/FTS queries are unaffected.

Yes — build any table with both a vector column and an FTS index, then compare `limit(4)` against `limit(2)` and `offset(2).limit(2)` as above. It reproduces deterministically; a regression test covering it is in the linked PR.

winklemad:
Fixes #4027

## Summary

`AsyncHybridQuery` (`table.query().nearest_to(...).nearest_to_text(...)`) paginates incorrectly when `.offset()` is used: the second page repeats rows from the first page and silently drops others.

`offset()` on a hybrid query pushes the offset down into *both* sub-queries (`HybridQuery::offset` in `python/src/query.rs` forwards to `inner_vec` and `inner_fts`), so each sub-query independently skips its own first `offset` rows before the results are fused. `AsyncHybridQuery.to_batches` then called `_combine_hybrid_results(..., limit=self._inner.get_limit())` without an `offset`, so the reranked table was sliced starting at position 0 and the sub-query limits were never raised to cover the skipped prefix.

On the 4-row fixture in `test_hybrid_query.py`, with `_rowid` ordering `[3, 0, 2, 1]`:

| query | before | after |
| --- | --- | --- |
| `.limit(2)` | `[0, 3]` | `[0, 3]` |
| `.offset(2).limit(2)` | `[3, 1]` | `[2, 1]` |

Row `3` was returned on both pages and row `2` was never returned at all.

This is the async counterpart of #3769 (`Fixes #3765`), which fixed the same bug in the synchronous `LanceHybridQueryBuilder`. #3765 explicitly deferred the async path; this PR closes that gap and reuses the `offset` parameter that #3769 already added to `_combine_hybrid_results`. The synchronous path is unaffected — it was fixed in #3769.

## Changes

`python/python/lancedb/query.py`, `AsyncHybridQuery.to_batches`:

- Each sub-query now fetches `limit + offset` rows and its own offset is reset to 0, so the fused result contains the full prefix the window is sliced out of.
- The combined, reranked table is sliced with `offset=` instead of always starting at 0.

Both halves are needed: raising the sub-query limits without the final slice still returns page 1, and slicing without raising the limits still misses rows.

`nodejs` has no equivalent hybrid combine path, so there is no SDK parity gap here.

## Test plan

- [x] New regression test `test_async_hybrid_query_offset` in `python/python/tests/test_hybrid_query.py`, mirroring the sync `test_hybrid_query_offset`. It asserts the offset window is a suffix of the un-offset result *and* that page 1 + page 2 together cover every row exactly once (a row-count-only assertion would pass even with duplicates).
- [x] `pytest python/tests/test_hybrid_query.py` — 16 passed
- [x] `pytest python/tests/test_rerankers.py` — 9 passed, 11 skipped
- [x] `pytest python/tests/test_query.py` — 86 passed
- [x] `pytest --doctest-modules python/lancedb/query.py` — 13 passed
- [x] `ruff format --check` / `ruff check` — clean
---

## Scope, after review

@lancedb-gatekeeper raised three points. Two were mine and are fixed in `04d07c2`; the third is deliberately left alone and I'd like a maintainer's call on it.

**Fixed — effective limit was read from the FTS child only.** `HybridQuery::get_limit()` (`python/src/query.rs:1159`) returns `self.inner_fts.inner.current_request().limit`, so an FTS-first hybrid with no explicit `.limit()` yielded `None`, skipped the widening branch and passed `limit=None` to the combiner — returning the union of both candidate lists instead of the documented default of 10. The limit is now derived from both children with a `DEFAULT_HYBRID_LIMIT = 10` fallback, so construction order no longer matters.

**Fixed — `explain_plan()` / `analyze_plan()` described a different query than the one that ran.** Both built their children straight from `self._inner`, bypassing the limit/offset rewrite in `to_batches`, and reported `skip=2, fetch=2` while execution used `skip=0, fetch=4`. Child preparation now lives in one `_create_child_queries()` helper used by all three.

> **Visible change to `explain_plan()` output:** because the plan is now built from the real execution children, which carry `with_row_id()`, the two `ProjectionExec` lines gain a `_rowid` column. The doctest is updated to match. This is the diagnostic becoming truthful rather than the assertion being weakened — it is still an exact-match comparison.

**Not fixed here — RRF candidate-pool invariance.** Widening each sub-query to `limit + offset` does change the candidate pool between page requests, so the fused ranking can shift and pagination can still repeat rows. That's a real problem, but it is exactly what the merged sync path does today:

```python
# LanceHybridQueryBuilder (sync), merged in #3769
sub_query_limit = self._limit + (self._offset or 0)
```

Making the pool invariant means choosing a contract — a fixed candidate pool, or an explicit cursor — and that ought to apply to sync and async together rather than letting the two paths diverge. I've asked in the review thread which way you'd prefer, and I'm happy to do it here or in a follow-up covering both paths.

So, to be precise about what this PR delivers: it makes `.offset()` take effect on the async hybrid path and makes the diagnostics honest. It does not make hybrid pagination stable across pages under reranking — that needs the contract decision above.

### DIFF
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -3885,6 +3885,19 @@ class AsyncHybridQuery(AsyncStandardQuery, AsyncVectorQueryBase):
         fts_query.with_row_id()
         vec_query.with_row_id()
 
+        limit = self._inner.get_limit()
+        offset = req.offset or 0
+        if offset:
+            # offset() pushes the offset down into both sub-queries, which would
+            # make each of them skip its own first `offset` rows. The window has
+            # to be taken out of the combined, reranked results instead, so
+            # fetch the skipped prefix here too and slice it off afterwards.
+            if limit:
+                fts_query.limit(limit + offset)
+                vec_query.limit(limit + offset)
+            fts_query.offset(0)
+            vec_query.offset(0)
+
         fts_results, vector_results = await asyncio.gather(
             fts_query.to_arrow(timeout=timeout),
             vec_query.to_arrow(timeout=timeout),
@@ -3896,8 +3909,9 @@ class AsyncHybridQuery(AsyncStandardQuery, AsyncVectorQueryBase):
             norm=self._norm,
             fts_query=fts_query.get_query(),
             reranker=self._reranker,
-            limit=self._inner.get_limit(),
+            limit=limit,
             with_row_ids=True,
+            offset=offset,
         )
         if (
             not self._user_requested_row_id()
--- a/python/python/lancedb/query.py
+++ b/python/python/lancedb/query.py
@@ -78,6 +78,10 @@ if TYPE_CHECKING:
 T = TypeVar("T", bound="LanceModel")
 AnalyzePlanDistributedMetrics = Literal["aggregate", "per_worker", "full"]
 
+# Number of rows a hybrid query returns when no limit was set on it. This
+# mirrors the default the Rust query builder applies to its sub-queries.
+DEFAULT_HYBRID_LIMIT = 10
+
 
 @runtime_checkable
 class _LanceScanner(Protocol):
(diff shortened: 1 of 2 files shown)


winklemad (commented):
Thanks — this is a genuinely useful review, and I've verified all three against the source rather than taking them on trust.

**Effective limit (`:3888`) — confirmed, my bug.** `HybridQuery::get_limit()` at `python/src/query.rs:1159` reads `self.inner_fts.inner.current_request().limit`, i.e. the FTS child only. So an FTS-first hybrid with no explicit `.limit()` returns `None`, my `if limit:` branch is skipped, and `limit=None` reaches the combiner. Your 18-rows-instead-of-10 result follows directly. I'll derive the effective limit independent of construction order, including the documented default of 10, and add FTS-first coverage.

**explain/analyze divergence (`:3890`) — confirmed, my bug.** `explain_plan` and `analyze_plan` build children straight from `self._inner`, so they bypass the rewrite I put in `to_batches` and report a different query than the one that executes. I'll factor child preparation into one helper used by all three.

**Candidate pool (`:3896`) — you're right, and I want to flag something before fixing it.** Widening to `limit + offset` does change the pool between pages, so RRF can reorder and pagination still repeats. But this is exactly what the merged sync path does today:

```python
# LanceHybridQueryBuilder, sync
sub_query_limit = self._limit + (self._offset or 0)
```

I deliberately mirrored that, on the reasoning that an async fix diverging from merged sync behaviour would be the wrong call to make unilaterally. So the invariant-pool problem isn't specific to this PR — sync has it too, and fixing it properly means changing behaviour on a path that's already released.

That's a scope decision I don't think is mine to make, so: would you prefer I

1. keep this PR at parity with sync (fix the two above only) and open a separate issue for pool invariance covering **both** paths, or
2. solve the invariant pool here for async, accepting that async and sync then differ until sync is updated, or
3. fix both paths in this PR?

My preference is (1) — the two confirmed bugs above are self-contained and clearly wrong today, whereas pool invariance needs a deliberate contract (fixed candidate pool vs. an explicit cursor) that ought to apply to both paths at once. I'll also scope the PR description down so it doesn't claim more than it delivers.

Happy to go whichever way you'd rather. I'll push the two confirmed fixes in the meantime.

wjones127 (approved):

winklemad:
Thanks @wjones127!

Issue #3983, opened 2026-08-21
Windows: background asyncio event loop hangs in GetQueuedCompletionStatus (IOCP) after moderate async connect/table volume in one process (0.37.1)


brunosrz:
### Describe the bug

On Windows, after a moderate number of `connect_async`/table open-close cycles run inside a single long-lived process, both `lancedb`'s internal background event loop thread (`LanceDBBackgroundEventLoop`, `background_loop.py`) and the awaiting caller's own asyncio event loop hang forever inside `GetQueuedCompletionStatus` (`_overlapped.GetQueuedCompletionStatus`, Windows `ProactorEventLoop`/IOCP). The process never returns; only killing it externally recovers.

This reproduces reliably running our full `pytest` unit suite (roughly 300 test functions, many of which call `lancedb.connect_async` and open/close tables) — it hangs at ~4% progress. It does **not** reproduce in isolated test files run individually, and it does **not** reproduce with a standalone script that just opens/closes ~300 `connect_async` connections sequentially outside of pytest (see repro script below) — the trigger appears to require additional concurrency/async-context patterns from the pytest+pytest-asyncio harness, not raw connection count alone.

### Reproduction

We could not yet reduce this to a minimal standalone script (see "What we tried" below) — filing now because:
1. We rigorously re-confirmed the bug is present in the current latest stable release (`0.37.1`), and disappears entirely on `0.36.0`, via a clean A/B (same codebase, same test suite, only the `lancedb` version swapped via `uv run --with`).
2. The failure is 100% deterministic on `0.37.1` in our real project's test suite and 100% clean on `0.36.0`, run repeatedly.
3. We have full dual-thread stack traces at the moment of the hang (below), which point at IOCP specifically, and a concrete architectural hypothesis (below) that may help narrow this down faster than us continuing to guess blindly at a minimal repro.

**Standalone script that does *not* reproduce it** (included for anyone trying to narrow this down — rules out plain connection volume as the sole trigger):

```python
import asyncio, tempfile, shutil
from pathlib import Path

async def open_close_once(db_dir, index):
    import lancedb, pyarrow as pa
    db = await lancedb.connect_async(str(db_dir))
    schema = pa.schema([pa.field("id", pa.string()), pa.field("vector", pa.list_(pa.float32(), 4))])
    table = await db.create_table(f"t{index}", schema=schema)
    await table.add([{"id": "1", "vector": [0.1, 0.2, 0.3, 0.4]}])
    await table.count_rows()
[...]

PR #4032, opened 2026-08-23
feat(python): release process-lifetime resources without exiting the process

### LINKED ISSUE #4031: Feature: release process-lifetime resources without exiting the process
### SDK

Python

### Description

**The gap**

Using the Python bindings once creates resources that live for the entire process:

- the Tokio runtime and its worker threads (`python/src/runtime.rs`)
- the background asyncio event loop thread (`python/python/lancedb/background_loop.py`)
- the embedding `ThreadPoolExecutor`

There is currently no way to release any of them short of exiting the process. The one mechanism that does recycle them is the POSIX `fork()` handler — `pthread_atfork` in `runtime.rs`, `os.register_at_fork` in `background_loop.py`. That means on POSIX it only ever happens as a side effect of forking, never on request, and **on Windows, where there is no `fork()`, there is no path at all**.

**Measured cost** (Windows 11, Python 3.13, single `connect_async` + `create_table` + vector search):

| | threads | RSS |
|---|---|---|
| baseline, `lancedb` imported | 13 | 92.4 MB |
| after using LanceDB | 25 | 140.9 MB |
| after releasing (prototype) | 16 | 140.1 MB |

So a process that touches LanceDB once carries **+12 OS threads for the rest of its life**, and a release path recovers 9 of them. Memory is not the story here — only ~0.8 MB comes back; threads are.

For a long-lived host that uses LanceDB in bursts — an agent runtime, a notebook kernel, a test suite, a service that re-indexes periodically — that is a permanent cost for an occasional workload.

**Proposed API**

Opt-in, with no behavior change unless called:

```python
import lancedb
from lancedb.background_loop import reset_background_loop

db = await lancedb.connect_async("./data")
# ... burst of LanceDB work ...

reset_background_loop()  # release runtime, background loop and embedding executor
```

In the same spirit of deterministic release, two smaller gaps in the existing surface:

- `AsyncConnection` and `AsyncTable` define `__enter__`/`__exit__` but no `__aenter__`/`__aexit__`, so `async with` does not work on them despite the names.
- The sync `LanceDBConnection` and `Table` have no `close()`/`is_open()` at all — only their async counterparts do.

**Two limitations we cannot address from `lancedb`**

Recorded here for context rather than as separate reports:

1. Of the 12 threads above, 3 are not reclaimable from this repository. They belong to `lance-core`'s global `CPU_RUNTIME` (`rust/lance-core/src/utils/tokio.rs`), which the code itself documents as never reclaimed, sets `thread_keep_alive` to effectively infinite, and whose atfork reset is a no-op on Windows.
2. `Session` defaults to a 6 GB index cache plus a 1 GB metadata cache, and one is created per connection when not passed explicitly. `Session::file_metadata_cache()` is public and `LanceCache::clear()` exists, so the metadata half is clearable — but `index_cache` is `pub(crate)` in the `lance` crate, so the larger of the two cannot be cleared from here today.

Happy to send a PR for the parts that are in scope for this repository.

brunosrz:
## What

Adds an opt-in way to release the resources that the Python bindings currently hold for the entire life of the process:

- `lancedb.background_loop.reset_background_loop()` — releases the background asyncio event loop, the embedding executor, and the Tokio runtime together, since all three exist for the same reason
- `lancedb._lancedb.reset_runtime()` — the Rust half, usable on its own
- `async with` support on `AsyncConnection` and `AsyncTable`, which define `__enter__`/`__exit__` today but no async equivalents, so `async with` does not work on them despite the names
- `close()`, `is_open()` and `with` support on the sync `LanceDBConnection` and `LanceTable`, which had no way to release their resources short of dropping the object and waiting for the GC
- `Session.clear_metadata_cache()`

Nothing changes unless a caller opts in.

The sync `close()`/`is_open()` are added as concrete methods on the two local implementations, not declared on the `DBConnection`/`Table` base classes. `DBConnection` declares neither today, and its implementations already disagree — `RemoteDBConnection.close()` is `async` while `AsyncConnection.close()` is sync — so promoting them to the base class is a public-interface decision I left to you. Both new methods just delegate to the async objects these classes already wrap, so they inherit the existing idempotency.

## Why

Using LanceDB once creates a Tokio runtime with its worker threads, a background event loop thread, and an embedding thread pool. None of them are ever released. The only mechanism that recycles them today is the POSIX `fork()` handler, so on POSIX it happens only as a side effect of forking, and on Windows — where there is no `fork()` — there is no path at all.

Measured on Windows 11 with Python 3.13, for a single `connect_async` plus `create_table` plus a vector search:

| | threads | RSS |
|---|---|---|
| baseline, `lancedb` imported | 13 | 92.1 MB |
| after using LanceDB | 25 | 139.8 MB |
| after `reset_background_loop()` | 16 | 138.9 MB |

A process that touches LanceDB once carries +12 OS threads for the rest of its life; this recovers 9 of them. Memory is not the point — only ~0.8 MB comes back. The target is a long-lived host that uses LanceDB in bursts: an agent runtime, a notebook kernel, a test suite, a service that re-indexes periodically.

Those numbers are a measurement I took, not an assertion in the test suite. Counting OS threads portably needs `psutil`, and adding a test dependency for one number seemed a worse trade than reporting it here; the tests assert the observable consequences instead (old loop closed, old executor shut down, no orphaned threads).

## What this does not reclaim

Stated explicitly so the API does not promise more than it delivers, and documented in the docstring as well:

- **3 of the 12 threads stay.** They belong to `lance-core`'s global `CPU_RUNTIME` (`rust/lance-core/src/utils/tokio.rs`), which that code documents as never reclaimed, gives an effectively infinite `thread_keep_alive`, and whose atfork reset is a no-op on Windows. Out of reach from this repository.
- **Only the metadata half of the session cache is clearable.** `Session::file_metadata_cache()` is public and `LanceCache::clear()` exists, so `clear_metadata_cache()` works. The index cache — the larger one, 6 GB by default — is `pub(crate)` in the `lance` crate and has no public accessor, so it cannot be cleared from here.

## How the Rust side stays safe

The runtime slot is a raw `AtomicPtr` holding an `Arc`'s own pointer, so the slot owns exactly one strong reference. Three properties carry the design:

**Readers are quiesced per generation, not globally.** Readers register in `ACTIVE_READERS[READER_EPOCH & 1]` and leave through the same slot. A retirement swaps the pointer out and only then bumps the epoch, so any reader that observed the old pointer necessarily read the pre-bump epoch and is therefore counted in the slot being drained. That ordering is load-bearing: with bump-before-swap a reader could take the old pointer while registered in the new slot, which would be a use-after-free. Because no reader can join the retired slot after the bump, the set of readers the drain waits on is finite and already closed, so the drain terminates. An earlier revision waited on a single global counter, which readers of the *new* generation could keep above zero indefinitely.

**Every failure degrades to a leak, never to a hang or a reclaim.** If the drain exceeds its timeout, the retired generation is left unreclaimed rather than freed — reclaiming would risk use-after-free, and dropping would risk running `Runtime::drop` on one of that runtime's own workers, which Tokio forbids. The same reasoning governs the reaper: when it gives up, it forgets its clone instead of dropping it, precisely so the last reference does not fall to a wedged task running on a worker of the runtime being destroyed. Leaking is the pre-existing behavior, so the worst case is the status quo.

**Final disposal never happens inline.** A dedicated one-shot thread per retired generation polls until it is the sole owner, then shuts the runtime down explicitly from there, never from a Tokio worker and never from whatever thread called reset.

Lock acquisition is ordered `RETIRE_LOCK` then `INSTALL_LOCK`, the reader window takes no lock at all, and both locks are RAII so an unwind cannot strand them. `atfork_child` remains atomic stores only — no allocation, no TLS, no `Arc` reclamation.

## Tests

Rust:

- readers of a new generation cannot block the drain of a retired one, exercised with a thread hammering `get_runtime()` throughout the reset
- a drain that times out leaves the generation unreclaimed instead of hanging or freeing it
- the reaper shuts a generation down once it is the sole owner, and gives up by leaking rather than dropping — verified through a `Weak` that still upgrades afterwards
- the install lock is released when runtime construction panics, and the module stays usable
- the reader window stays balanced across an unwind
- retirement leaks rather than drops when thread spawn fails
- concurrent readers never observe a freed pointer, under repeated resets

Python:

- LanceDB keeps working after a reset; repeated resets leak no threads; the old loop is genuinely closed; a join timeout raises without swapping the loop; concurrent resets do not orphan a thread
- the old embedding executor is shut down rather than abandoned
- resetting repeatedly while work keeps arriving does not deadlock, run in a subprocess the test can actually time out and kill
- `async with` releases the connection and the table, including when the block raises
- the sync connection and table close, refuse use afterwards, are idempotent, and work as `with` blocks
- `clear_metadata_cache()` shrinks a populated session (5470 to 208 bytes in a local run), and is a harmless no-op on an untouched one
- a POSIX-only subprocess test covering fork while the reset lock is held

## Note on history

This supersedes #4001, which I closed. That PR proposed the same underlying API but justified it as a fix for #3983, a Windows IOCP hang I reported. Further investigation showed the hang was not attributable to LanceDB and I could not reproduce it, so I retracted that report and closed the issue. The capability gap described above is real and measurable on its own, and I would rather propose it on that basis than leave it resting on a premise I no longer believe.

Closes #4031


brunosrz:
@Xuanwo This PR has been open for over a week without any maintainer review — could you take a look when you have a moment? The gatekeeper approved it (`K-approved`) and the reader-admission race it had flagged was fixed with deterministic regression coverage.

PR #4034, opened 2026-08-23
fix(python): don't crash inferring schema for an all-null vector column

### LINKED ISSUE #4033: create_table crashes with IndexError on an all-null vector column
### Description

`create_table(data=...)` without an explicit schema raises `IndexError: index out of bounds` when a `list`/`large_list` column whose name contains `vector`/`embedding` has all-null values.

```python
import tempfile, lancedb, pyarrow as pa
data = pa.table({
    "id": pa.array([1, 2], type=pa.int64()),
    # valid: vector column is typed list<float32>, embeddings not computed yet
    "vector": pa.array([None, None], type=pa.list_(pa.float32())),
})
db.create_table("t", data=data)
# IndexError: index out of bounds
```

`_modal_list_size` (`python/python/lancedb/table.py`) infers the vector dimension with `pc.mode(pc.list_value_length(arr))[0]`. pyarrow's `mode` skips nulls, so an all-null column yields a zero-length result and the `[0]` subscript raises.

The sibling helper `_infer_vector_dim` in the same file performs the identical `pc.mode(...)` and already guards the empty case with `if len(lengths) == 0: return None`, so this is a case that was meant to be handled — `_modal_list_size` just missed it.

An all-null vector column is a realistic input: you often create the table with the schema in place and the embeddings filled in later.

### Expected

`create_table` succeeds and leaves the column as a variable-length list (the dimension is genuinely unknowable when every value is null), rather than crashing.

I have a fix (return `None` from `_modal_list_size` when all-null and skip the fixed-size dimension at the call site, mirroring `_infer_vector_dim`) plus a regression test, and will open a PR.

winklemad:
Fixes #4033

`create_table(data=...)` without an explicit schema raised `IndexError: index out of bounds` when a `vector`/`embedding`-named list column was all-null. `_modal_list_size` infers the dimension with `pc.mode(pc.list_value_length(arr))[0]`, and pyarrow's `mode` skips nulls, so an all-null column produces a zero-length result and the `[0]` subscript blows up.

The sibling `_infer_vector_dim` already guards the identical `pc.mode(...)` with a `len == 0` check and returns `None`; `_modal_list_size` just didn't. This makes `_modal_list_size` return `Optional[int]` and return `None` when every value is null, and the call site then leaves the column as a variable-length list instead of inventing a fixed size — the dimension is genuinely unknowable when there are no non-null rows.

Added a regression test that creates a table from an all-null `list<float32>` vector column and asserts it succeeds, the column stays a `list<float32>`, and the rows are queryable. It fails on `main` with the `IndexError` and passes with this change.

### DIFF
--- a/python/python/lancedb/table.py
+++ b/python/python/lancedb/table.py
@@ -4573,6 +4573,12 @@ def _infer_target_schema(
             # Use the most common length of the list as the dimensions
             dim = _modal_list_size(peeked.column(i))
 
+            if dim is None:
+                # Every list entry is null (e.g. embeddings not computed yet),
+                # so the dimension is unknowable. Leave the column as a
+                # variable-length list instead of guessing a fixed size.
+                continue
+
             # Determine target type based on value type
             if pa.types.is_floating(field.type.value_type):
                 target_type = pa.list_(pa.float32(), dim)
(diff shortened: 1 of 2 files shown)


winklemad:
Friendly ping on this one. Small fix with a regression test, CI is green, and it mirrors the guard `_infer_vector_dim` already has for the same empty case. Happy to rebase or adjust if anything is needed. Would appreciate a review when someone has a moment.

PR #4038, opened 2026-08-24
fix(python): coerce JSON columns before writes

### LINKED ISSUE #4035: bug(python): add() cannot write a json column from Utf8 or list-of-dict input (casts to large_binary instead of JSONB-encoding)
> **Disclaimer:** this issue was written and filed by an AI coding agent (Claude Code) on behalf of a user who hit it in their own project. Everything below was run in clean containers rather than reasoned about — please treat the diagnosis with the scepticism that deserves. Happy to run further probes or test a patch.

### LanceDB version

0.38.0b5 (also reproduced on 0.36.0 and 0.37.1). pyarrow 25.0.1, Python 3.12, Linux aarch64, local filesystem.

### What happened?

`add()` cannot write a `pa.json_()` column from `Utf8` input. Since `list-of-dict` has no way to express an `arrow.json` array, that documented input type is unusable for json columns entirely.

```
add(pa.array(['{"k": 2}'], type=pa.json_()))   OK    <- #3144, fixed by #3976
add(pa.array(['{"k": 2}']))                    FAIL
add([{"id": "c", "j": '{"k": 2}'}])            FAIL
```

Both failures are:

```
lance error: Append with different schema: `j` should have type json but type was large_binary
```

Note `large_binary` is a type the caller never mentioned — it is lancedb's own cast showing through.

### Repro

```python
import asyncio, tempfile
import lancedb, pyarrow as pa

SCHEMA = pa.schema([pa.field("id", pa.string()), pa.field("j", pa.json_(), nullable=True)])

async def main():
    db = await lancedb.connect_async(tempfile.mkdtemp())
    t = await db.create_table("t", data=pa.Table.from_pylist(
        [{"id": "a", "j": '{"k": 1}'}], schema=SCHEMA))

    # works — typed arrow.json input (the #3976 passthrough)
    await t.add(pa.table({"id": pa.array(["b"]), "j": pa.array(['{"k": 2}'], type=pa.json_())}))

    # fails — same JSON text, Utf8 input
    await t.add(pa.table({"id": pa.array(["c"]), "j": pa.array(['{"k": 3}'])}))

asyncio.run(main())
```

Replacing the last call with `await t.add([{"id": "c", "j": '{"k": 3}'}])` fails identically.

### Cause

The live passthrough for JSON input is in Python, in `_align_field_types`
(`python/python/lancedb/table.py`), added by #3976. It is gated on the **input**
field being `arrow.json`:

```python
if (
    _field_extension_name(field) == "arrow.json"
    and _field_extension_name(target_field) == "lance.json"
):
    new_fields.append(field)
    continue
```

That covers callers who already built an `arrow.json` array. A `Utf8` input fails
the gate, so it falls through to the ordinary cast, which sees `Utf8` vs the
table's `LargeBinary` storage and inserts a plain cast. The JSON text becomes raw
UTF-8 bytes with no JSONB encoding, and lance-core rejects the append.

The cast itself is silently legal at the pyarrow layer, which is why nothing
complains earlier:

```python
pa.Table.from_pylist([{"j": '{"k": 2}'}], schema=pa.schema([pa.field("j", pa.large_binary())]))
# -> [{'j': b'{"k": 2}'}]
```

(An earlier Rust-side fix for the same gate, #3291 against `build_field_exprs`,
was closed unmerged — so `_align_field_types` is the only place this decision is
made today.)

### This gate is shared by `add()` and `merge_insert()`

Worth flagging for whoever picks this up, so it does not get fixed only halfway.
Both write paths reach the same helper:

```
_sanitize_data
  └─ _cast_to_target_schema
       └─ _align_field_types        <- the arrow.json gate lives here

call sites: create_table, merge_insert (sync), merge_insert (async), add
```

So teaching that gate to accept `Utf8` for a `lance.json` target — by converting
it to `arrow.json` and letting lance encode, as it already does for typed input —
would fix `add()` and `merge_insert()` together.

Confirmed on 0.38.0b5 that `merge_insert` takes the same `Utf8` input through the
same gate — but where `add()` is rejected downstream, merge_insert has no
equivalent check, so the raw bytes are written and every subsequent
`json_extract` over the table raises `InvalidJsonb`:

```
merge_insert into a pa.json_() column, 0.38.0b5

  typed arrow.json         wrote, scan OK
  Utf8                     wrote, SCAN BROKEN   <- InvalidJsonb, permanently
  list-of-dict             wrote, SCAN BROKEN
  declared-schema table    wrote, scan OK
```

One bad row is enough — the failure is on scan, so it takes down every
`json_extract` over the whole table, not just the row that was written. Fixing
the gate therefore closes a silent-corruption path as well as a rejection.

### Why this looks like a gap rather than intended strictness

- **`list-of-dict` is a documented input type for `add()`**, alongside `pandas.DataFrame` and `pyarrow.Table`/`RecordBatch`. For a json column, only the arrow types can work, because the other two have nowhere to put an extension type. So the documented surface cannot express a valid write.
- **`add()` has no `schema=` parameter** (only `mode`, `on_bad_vectors`, `fill_value`, `progress`, `write_parallelism`), so a caller passing dicts has no way to declare intent either.
- **The sibling write path already does this coercion.** lance-format/lance#8669 taught `update()` to accept a plain string expression for a json column and JSONB-encode it, so `update(updates={"j": '{"k": 2}'})` works while the equivalent `add` does not.

### Expected

`add()` accepts `Utf8` input for a `lance.json` column and routes it through the JSONB encoder, the way `update()` now does — which would also make `list-of-dict` usable for json columns.

### Workaround

Build the arrow type explicitly and pass it as a `pyarrow.Table`. Any construction that carries the type works — `pa.array(vals, type=pa.json_())`, `pa.Table.from_pylist(rows, schema=DECLARED_SCHEMA)`, `pa.table(..., schema=...)`, `.cast(...)`, or `ExtensionArray.from_storage`:

```python
await table.add(pa.Table.from_pylist(records, schema=DECLARED_SCHEMA))
```

Two cautions for anyone reaching for that:

- The schema must be the one **declared in code**. Feeding back `await table.schema()` returns `large_binary` + `lance.json` metadata, and building input from it reintroduces exactly the raw-bytes write described above.
- If the schema also contains `lancedb.blob()` columns, they need the opposite treatment — declared-schema input breaks them and they must go in as `large_binary`. See #3759.

### Related

- #3144, fixed by #3976 — the `arrow.json`-input half of this.
- #3759 — all-null batches for extension columns; same `add()` coercion layer.
- lance-format/lance#8669 — the equivalent `Utf8` → JSONB coercion, added for `update()`.

LH-kevin:
## Summary

- promote compatible Utf8, LargeUtf8, StringView, and null JSON input to Arrow's logical `arrow.json` representation when writing to a `lance.json` column
- preserve already-typed `arrow.json` input and LargeUtf8 storage width
- apply JSON coercion before normal `add()` writes, including `mode="overwrite"` so an existing JSON column is not silently replaced by plain Utf8
- reuse the same coercion in the schema-alignment path used by standard `merge_insert()`, preventing raw UTF-8 bytes from being written into JSONB-backed columns
- convert logical JSON before LSM merge schema admission so LSM `merge_insert()` also stores valid JSONB
- support PyArrow versions without `pa.json_` by attaching the equivalent `ARROW:extension:name=arrow.json` metadata
- normalize legacy StringView JSON at the shared pre-cast boundary so sanitizing `add()` and `merge_insert()` also work on supported PyArrow 16.x releases
- preserve null semantics and existing field metadata

Fixes #4035

## Why

LanceDB stores JSON columns as JSONB-backed `lance.json`.

Plain Utf8 input and ordinary list-of-dict input do not carry the `arrow.json` extension type. Without promotion, these values can fall through to the physical `large_binary` storage representation instead of reaching Lance as logical JSON.

For `add()`, this can result in a schema mismatch. For `merge_insert()`, raw UTF-8 bytes can be committed successfully but later fail JSON queries with `InvalidJsonb`.

`add(mode="overwrite")` had a related failure mode: the write could succeed while replacing the existing logical JSON column with plain Utf8, after which `json_extract` no longer worked.

This change keeps compatible JSON input logical as `arrow.json`, allowing Lance's existing JSONB encoder to perform the physical encoding.

## Compatibility

The write contract covered by this change is:

- existing typed `arrow.json` input remains unchanged
- Utf8 input is accepted for `lance.json`
- LargeUtf8 input is accepted without narrowing its storage width
- StringView input is normalized to logical JSON on PyArrow versions that expose it
- list-of-dict input works through the normal documented `add()` path
- all-null JSON batches preserve null semantics
- overwrite with compatible string JSON preserves the existing JSON column semantics
- PyArrow versions without `pa.json_` use Arrow extension metadata on Utf8/LargeUtf8 instead of falling back to raw binary casting
- legacy StringView input is materialized to Utf8 immediately before the shared schema cast on releases such as PyArrow 16.1, where direct StringView-to-Utf8 casting is unsupported

The legacy fallback was also verified separately under PyArrow 18.1.0 against a persisted `lance.json` table: `add(Utf8)` and standard `merge_insert(Utf8)` both produced queryable JSONB, with null semantics preserved.

## Tests

Added regression coverage for:

- `add()` with plain Arrow Utf8 JSON
- `add()` with LargeUtf8 JSON
- `add()` with explicit StringView input
- `add()` with list-of-dict JSON
- explicit `mode="append"` and default `mode=None`
- sanitizing `add()` with legacy StringView input
- `add(mode="overwrite")` followed by `json_extract`
- all-null JSON batches
- standard `merge_insert()` with Utf8 JSON
- standard `merge_insert()` with legacy StringView input
- LSM `merge_insert()` with Utf8 JSON
- post-write `json_extract` queries to verify valid JSONB
- preservation of null semantics
- LargeUtf8 storage preservation
- typed `arrow.json` behavior
- PyArrow-without-`pa.json_` fallback behavior
- metadata-only schema promotion
- end-to-end legacy fallback add/null/merge paths
- adjacent Blob V2 regressions after integration with current `main`

## Validation

Against current `main` (`1da58768`), with PR head `481e13d9`:

- reproduced the legacy StringView failure under actual PyArrow 16.1.0: `pa.string_view()` is available, `pa.json_` is absent, and both sanitizing `add()` and standard `merge_insert()` failed at the shared cast with `ArrowNotImplementedError`
- moved legacy StringView materialization to the shared `_cast_to_target_schema` pre-cast boundary, before `Table.cast()`
- re-ran the same public paths under actual PyArrow 16.1.0: sanitizing `add()`, merge update, and merge insert all succeeded, and post-write `json_extract` returned the expected rows
- targeted JSON / Blob / LSM tests — `92 passed, 1 skipped`
- `python/tests/test_table.py`, `python/tests/test_blob.py`, and `python/tests/test_merge_insert_lsm.py` — `271 passed, 1 skipped, 32 warnings`
- `make check` — passed
- repository `make typecheck` — `0 errors, 0 warnings, 0 informations`

## Relationship to #4067

#4067 overlaps the top-level plain-string and all-null JSON write coercion and moves that handling into the shared Rust write boundary.

This PR still covers write paths not handled by that change, notably LSM `merge_insert()` JSON conversion before exact schema admission. It also contains the Python-side overwrite regression/fix requested in review and explicit compatibility coverage for supported PyArrow versions without `pa.json_`.


python/python/lancedb/table.py
```diff
@@ -448,6 +448,64 @@ def _field_extension_name(field: pa.Field) -> Optional[str]:
...
+
+    is_string_like = (
+        pa.types.is_string(field.type)
+        or pa.types.is_large_string(field.type)
+        or (hasattr(pa.types, "is_string_view") and pa.types.is_string_view(field.type))
+    )
+    if is_string_like or pa.types.is_null(field.type):
+        return pa.field(
+            field.name,
+            pa.json_(),
```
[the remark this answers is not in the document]
LH-kevin: Fixed in 601aebe8. _coerce_json_field now preserves LargeUtf8 storage by constructing pa.json_(pa.large_string()) instead of narrowing to Utf8. Added a focused regression asserting the backing storage remains large_string. Full table + LSM regression suite passes (192 tests), along with make check and make typecheck.

python/python/lancedb/table.py
```diff
@@ -460,15 +518,12 @@ def _align_field_types(
...
-        # JSONB. Lance must see arrow.json so it can perform the JSONB encoding.
-        if (
-            _field_extension_name(field) == "arrow.json"
-            and _field_extension_name(target_field) == "lance.json"
-        ):
-            new_fields.append(field)
+        # LanceDB exposes stored JSON columns as lance.json (JSONB-backed
+        # LargeBinary). Keep logical JSON input as arrow.json so Lance performs
+        # the JSONB encoding instead of relabeling raw UTF-8 bytes as JSONB.
+        json_field = _coerce_json_field(field, target_field)
```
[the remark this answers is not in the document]
LH-kevin: Fixed in 601aebe8. The LSM merge path now calls Lance's existing lance_arrow::json::convert_json_columns() before exact align_batch_schema, so logical arrow.json input is encoded to JSONB-backed lance.json before MemWAL admission. Added an LSM upsert regression covering both update and insert and verifying logical JSON readback. Full table + LSM regression suite passes (192 tests), along with make check and make typecheck.

python/python/lancedb/table.py
```diff
@@ -448,6 +448,69 @@ def _field_extension_name(field: pa.Field) -> Optional[str]:
...
 
+def _coerce_json_field(field: pa.Field, target_field: pa.Field) -> Optional[pa.Field]:
+    """Promote compatible JSON input to Arrow's logical JSON type."""
+    if _field_extension_name(target_field) != "lance.json":
+        return None
+
+    if _field_extension_name(field) == "arrow.json":
+        return field
+
+    if not hasattr(pa, "json_"):
```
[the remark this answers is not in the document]
LH-kevin: Fixed in d211f543 and now available on the updated PR head 16f507a5.

On PyArrow versions without `pa.json_`, compatible Utf8/LargeUtf8/null input is promoted by attaching `ARROW:extension:name=arrow.json` while preserving string width (null normalizes to Utf8). `_coerce_json_scannable` also treats metadata-only field changes as real schema changes.

Added fallback regression coverage including end-to-end add/null/merge write paths with post-write `json_extract`, and verified separately under PyArrow 18.1.0 against a persisted `lance.json` table: both `add(Utf8)` and standard `merge_insert(Utf8)` produce queryable JSONB, with null semantics preserved.

After integrating the implementation with current upstream via an additive merge, the related suite is green: 259 passed, 1 skipped; targeted JSON/Blob/LSM coverage is 83 passed, 1 skipped; `make check` and repository `make typecheck` both pass.

python/python/lancedb/table.py
```diff
@@ -618,6 +618,79 @@ def _field_extension_name(field: pa.Field) -> Optional[str]:
...
+    if _field_extension_name(target_field) != "lance.json":
+        return None
+
+    if _field_extension_name(field) == "arrow.json":
+        return field
+
+    is_string_like = (
+        pa.types.is_string(field.type)
+        or pa.types.is_large_string(field.type)
+        or (hasattr(pa.types, "is_string_view") and pa.types.is_string_view(field.type))
```
wjones127, line 661: **issue (blocking):** The PR claims StringView input is normalized to logical JSON, but no test constructs a `pa.string_view()` array — please add a parametrized case alongside the existing Utf8/LargeUtf8 tests so this branch is actually exercised.

python/python/lancedb/table.py
```diff
@@ -5645,6 +5714,8 @@ async def add(
             )
         _register_optional_converters()
         data = to_scannable(data)
+        if mode in (None, "append"):
```
wjones127: **question (blocking):** `add(mode="overwrite")` goes through `sanitize_create_table(data, None, ...)` at line 5698, which never sees the real table schema fetched above — so it looks like overwrite with Utf8 JSON input against an existing `lance.json` column still hits the raw-bytes bug from #4035. Is overwrite intentionally out of scope here, or does it need the same coercion?

rust/lancedb/src/table/merge/lsm.rs
```diff
@@ -683,6 +684,7 @@ pub(crate) async fn execute_lsm_merge_insert(
         if batch.num_rows() == 0 {
             continue;
         }
+        let batch = convert_json_columns(&batch).map_err(|e| Error::Arrow { source: e })?;
```
wjones127, line 687: **question (non-blocking):** `convert_json_columns` only promotes fields that already carry the `arrow.json` extension metadata — can you confirm the batches admitted here still have that metadata attached by the time they reach this function (vs. it being stripped somewhere in the LSM replay/admission path), since that's what the new merge-insert-LSM test is relying on?

wjones127 (changes requested):

python/python/lancedb/table.py
LH-kevin: Added explicit `pa.string_view()` coverage on the updated head. `test_add_coerces_string_json` is now parametrized across Utf8, LargeUtf8, StringView (when available), and pylist input, with both default and explicit append modes. The focused test passes with 8 cases, and the latest-main related suite is green (`269 passed, 1 skipped`).

python/python/lancedb/table.py
LH-kevin: Confirmed — this was a real same-contract bug, not intentionally out of scope. I reproduced `add(mode="overwrite")` with Utf8 JSON input: the write completed, but the stored column became plain string and subsequent `json_extract` failed. The updated head now applies `_coerce_json_scannable(data, schema)` for overwrite as well, using the existing table schema before the inner add call. I added `test_add_overwrite_coerces_string_json`, which verifies the post-overwrite JSON is still queryable with `json_extract`. This fix is in `59e371af`; the latest-main related suite passes (`269 passed, 1 skipped`), along with `make check` and `make typecheck`.

rust/lancedb/src/table/merge/lsm.rs
LH-kevin: Confirmed. On the Python merge path, `_sanitize_data` reaches `_cast_to_target_schema`, where `_align_field` promotes compatible Utf8 input to `arrow.json`. The resulting `reordered_schema` is then used both for the `RecordBatchReader` schema and for each emitted `RecordBatch`, so the extension metadata is still attached when the batch crosses into `execute_lsm_merge_insert`. I also verified this directly on the updated head: both the reader field and the actual emitted batch field report `arrow.json` before the Rust call. The LSM regression then exercises update + insert through `convert_json_columns()` and verifies queryable JSON readback. Latest-main JSON/Blob/LSM targeted coverage is green (`92 passed, 1 skipped`).

LH-kevin:
@wjones127 I addressed the three review points and pushed the updated head (`48ed4fa1`):

- added explicit `pa.string_view()` coverage
- reproduced and fixed `add(mode="overwrite")` losing JSON semantics, with a post-write `json_extract` regression
- verified `arrow.json` metadata is preserved into the LSM conversion boundary

I also merged current `main` (`577fb483`) and revalidated the related suite: `92 passed, 1 skipped` targeted; `269 passed, 1 skipped` full related; `make check` and `make typecheck` pass.

The PR description is updated as well, including the overlap/boundary with #4067. When you have a chance, could you please take another look?

python/python/lancedb/table.py
```diff
@@ -618,6 +618,79 @@ def _field_extension_name(field: pa.Field) -> Optional[str]:
...
+        )
+        return pa.field(
+            field.name,
+            json_type,
+            nullable=field.nullable,
+            metadata=field.metadata,
+        )
+
+    storage_type = (
+        pa.large_string() if pa.types.is_large_string(field.type) else pa.string()
```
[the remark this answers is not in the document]
LH-kevin: Reproduced this against actual PyArrow 16.1.0 before changing the code: `pa.string_view()` is available, `pa.json_` is not, and casting the promoted field fails with `ArrowNotImplementedError: Unsupported cast from string_view to utf8 using function cast_string`.

Fixed in `310a9dd2`. `_coerce_json_scannable` now detects only the legacy no-`pa.json_` StringView JSON fields and normalizes each affected batch column to Utf8 with a conversion that works on PyArrow 16.1 before handing the reader back to the existing schema-cast path. Null values and `arrow.json` field metadata are preserved; modern `pa.json_` and non-StringView paths are unchanged.

Added `test_coerce_json_scannable_normalizes_legacy_string_view_before_cast`, which constructs a real `pa.string_view()` array, removes `pa.json_` for the test, asserts normalization happens before the common cast, and checks the values including null. It failed before the fix and passes after it.

After integrating the latest `main`, current PR head is `fa6fae67`: targeted JSON/Blob/LSM coverage is `93 passed, 1 skipped`; the full related suite is `270 passed, 1 skipped, 32 warnings`; `make check` and repository `make typecheck` both pass.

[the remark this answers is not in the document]
LH-kevin: Fixed in `481e13d9`. I moved the legacy StringView materialization out of `_coerce_json_scannable` and into the shared `_cast_to_target_schema` pre-cast boundary, immediately before `Table.cast()`. This covers paths that sanitize first, including `add(..., on_bad_vectors="fill")` and standard `merge_insert()`.

I reproduced both public paths against actual PyArrow 16.1.0 before the change; each failed at the shared cast with `Unsupported cast from string_view to utf8`. After the change, both succeed against the same persisted `lance.json` table, and post-write `json_extract` queries return the expected rows for the sanitizing add, merge update, and merge insert.

The regressions now use real `pa.string_view()` input on the public sanitizing-add and merge-insert paths and assert that the legacy StringView-to-Utf8 materialization occurs before the common cast. The obsolete helper-only regression was removed.

Validation: targeted JSON/Blob/LSM suite `92 passed, 1 skipped`; full related suite `271 passed, 1 skipped, 32 warnings`; `make check` passes; repository `make typecheck` reports `0 errors, 0 warnings, 0 informations`.

python/python/tests/test_table.py
```diff
@@ -4211,3 +4437,140 @@ async def test_refresh_column_async_job_async_table(tmp_path):
...
+    from lancedb.table import _coerce_json_field
+
+    source = pa.field("j", source_type)
+    target = pa.field(
+        "j",
+        pa.large_binary(),
+        metadata={b"ARROW:extension:name": b"lance.json"},
+    )
+
+    monkeypatch.delattr(pa, "json_")
```
geruh, line 4479: can we use `raising=False` here and in `test_coerce_json_scannable_detects_metadata_only_change`? On older PyArrow, `pa.json_` is already absent, so these tests raise before reaching the fallback they're testing.

python/python/lancedb/table.py
```diff
@@ -432,6 +432,35 @@ def _cast_to_target_schema(
     def gen():
         for batch in reader:
             batch = _coerce_blob_write_columns(batch, reordered_schema)
+
+            # PyArrow 16 exposes StringView but cannot cast it directly to Utf8.
```
geruh, line 436: nit: can we move the legacy StringView normalization into a helper next to the JSON coercion code? That would keep this loop easy to follow like coerce blobs, normalize JSON, then cast.

python/python/tests/test_table.py
```diff
@@ -3029,6 +3029,232 @@ async def test_add_sanitization_encodes_json(mem_db_async: AsyncConnection):
...
+                ['{"k": 5}', '{"k": 7}'],
+                type=pa.string_view(),
+            ),
+        }
+    )
+
+    original_array = pa.array
+    normalized = False
+
+    def spy_array(values, *args, **kwargs):
```
geruh, line 3229: nitty: can we share the `spy_array` setup between these two tests? The normalization tracking is duplicated.

PR #4043, opened 2026-08-24
feat: bind a materialized view refresh to the view incarnation

wkalt:
A caller that queues a refresh and executes it later can only tell the
view it captured from a drop-and-recreate by comparing the definition and
version. A recreated view with the same definition and an equal or higher
version passes that check, and the check runs before the refresh reloads
the view, so it never sees the state it commits against.

This mints an `mv.incarnation` token in the schema metadata at each
physical creation of a view table.
`RefreshMaterializedViewBuilder::expect_incarnation` carries the captured
token into the refresh, which compares it against the latest stored
manifest before planning and again immediately before each commit
(publish, fragment swap, watermark stamp), refusing to land in a different
incarnation. A view with no token -- declared before tokens existed, or
its metadata replaced wholesale -- is refused under a bound refresh with
its own wording and is minted one by its next unbound refresh. The token
is exposed through `MaterializedView::incarnation`; refreshes without an
expectation are unchanged.

This is best effort: the token is not part of lance's commit condition,
so a recreation landing between the final pre-commit read and the commit
itself is not caught. Closing that window needs a base-manifest
precondition in lance's commit path.

### DIFF
--- a/rust/lancedb/src/materialized_view.rs
+++ b/rust/lancedb/src/materialized_view.rs
@@ -37,6 +37,13 @@ pub use refresh::{RefreshMaterializedViewResult, RefreshMode};
 /// Schema metadata key holding the view definition, as kind-tagged JSON.
 pub const DEFINITION_META_KEY: &str = "mv.definition";
 
+/// Schema metadata key holding the view's incarnation: a token minted at each
+/// physical creation of a view table, so a view dropped and recreated under
+/// the same name and definition is still told apart from the one a caller
+/// captured. A view whose metadata was replaced wholesale, or one declared
+/// before tokens existed, carries none until its next refresh mints one.
+pub const INCARNATION_META_KEY: &str = "mv.incarnation";
+
 /// Schema metadata key holding the source table version the view was last
 /// refreshed to. Absent until the first refresh.
 pub const SOURCE_VERSION_META_KEY: &str = "mv.source_version";
@@ -612,8 +619,17 @@ impl PreparedDeclaration {
     pub async fn create(self, name: &str) -> Result<MaterializedView> {
         let empty: Vec<std::result::Result<arrow_array::RecordBatch, arrow_schema::ArrowError>> =
             vec![];
+        // Minted here, not at preparation: a declaration can be cloned and
+        // create more than one physical table, and each needs its own token.
+        let incarnation = uuid::Uuid::new_v4().to_string();
+        let mut metadata = self.schema.metadata().clone();
+        metadata.insert(INCARNATION_META_KEY.to_string(), incarnation.clone());
+        let schema = Arc::new(ArrowSchema::new_with_metadata(
+            self.schema.fields().clone(),
+            metadata,
+        ));
         let reader: Box<dyn arrow_array::RecordBatchReader + Send> =
-            Box::new(arrow_array::RecordBatchIterator::new(empty, self.schema));
+            Box::new(arrow_array::RecordBatchIterator::new(empty, schema));
         let mut request = CreateTableRequest::new(name.to_string(), Box::new(reader));
         let write_params = request
             .write_options
@@ -648,6 +664,7 @@ impl PreparedDeclaration {
         Ok(MaterializedView {
             table,
             definition: self.definition,
+            incarnation: Some(incarnation),
         })
     }
 }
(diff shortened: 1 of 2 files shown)


LuQQiu:
One more lifecycle hole beyond the commit-fencing discussion: **the sanctioned definition-replacement flow silently erases the incarnation token.**

The crate's own way of redefining a live view — `replace_schema_metadata` with just `DEFINITION_META_KEY`, as modeled in `test_definition_replaced_between_refreshes_is_adopted` and `test_definition_view_schema_mismatch_is_refused` (refresh.rs ~2612-2653) — replaces the entire schema metadata map, so `mv.incarnation` is dropped. `execute_refresh` explicitly supports this flow (the comment above the definition reload: "The definition a handle cached at open may since have been replaced"), yet nothing preserves or re-stamps the token across it.

Consequences for a view that was never dropped:

- every subsequent refresh carrying the previously captured token fails with "it was dropped and recreated" — a false diagnosis;
- reopened handles return `incarnation() == None` forever, so `expect_incarnation` can never be used on that view again.

`stamp_watermark` deliberately uses the merging `update_schema_metadata` so the token survives refresh stamps (and the new test asserts that); the replacement path needs the same treatment — either carry the existing `mv.incarnation` forward when the definition is replaced, or treat definition replacement as a new incarnation and mint a fresh token there deliberately. Either is coherent; losing the key by accident is not.

Related nit: when the stored key is absent (`actual == None` — a pre-token view, or one that hit the path above), the error at the incarnation check still asserts "it was dropped and recreated". Under this PR that case can never be a drop+recreate (recreation always mints a token), so the message points operators at the wrong cause; the `None` case deserves its own wording ("the view carries no incarnation token").

LuQQiu:
Adding concrete evidence for the two changes the gatekeeper review asks for — we traced both mechanisms end to end.

**1. Why the preflight check alone cannot deliver the doc's guarantee (commit-time fencing is needed).**

The check at the top of `execute_refresh` runs under `refresh_lock()`, but that lock is an in-process, per-URI mutex serializing only refreshes — `drop_table` never acquires it, and cross-process it doesn't exist. Nothing on the publish or stamp path re-reads the incarnation: `publish` and `stamp_watermark` operate on the dataset cached before the check. The full escape:

1. Process A passes the check at view version N, then spends the compute/stage phase (minutes) building the increment.
2. Process B drops and recreates the view. `drop_table` is a plain `remove_dir_all` on the deterministic `{db_uri}/{name}.lance` path, so the recreated table reuses the exact URI; its `_versions/` restarts at 1.
3. A commits at N+1: lance's conflict detection only lists manifests **newer than A's cached version N** — the fresh table's 1-2 are older, so it finds nothing; the conditional put (`PutMode::Create`) succeeds because `_versions/{N+1}` doesn't exist in the new lineage; and since latest resolution takes the max version, N+1 > 2 becomes the recreated view's head.

So old-incarnation rows (plus the old schema metadata, definition, and token — the manifest is built from A's cached one) land as the new incarnation's latest version, which is exactly what the `expect_incarnation` doc says never happens. The check is still a strict improvement (it catches drops completed before the refresh starts), but closing the window needs the incarnation enforced as part of the commit condition (or a lifecycle fence / per-creation path randomization). If commit-time enforcement is deferred to a follow-up, the doc wording should be softened to best-effort in the meantime.

**2. Why "mint in the actual create path" matters even within this PR's API surface.**

The UUID is minted in `prepare_declaration` (once per `PreparedDeclaration`), and `PreparedDeclaration` is `#[derive(Clone)]` with a consuming `create(self, ..)`. Both `prepare_declaration` and `create` are `pub` and documented with exactly the prepare-then-create pattern, so this is reachable API, not test-only:

```rust
let prepared = prepare_declaration(&conn, &query).await?;
let v1 = prepared.clone().create("x").await?;   // token T
conn.drop_table("x", &[]).await?;
let v2 = prepared.create("x").await?;           // same token T
// refresh().expect_incarnation(T) now accepts the recreated view
```

Two physical creations share one token, contradicting `INCARNATION_META_KEY`'s own doc ("minted once per declaration, so a view dropped and recreated under the same name and definition is still told apart"). The builder path is unaffected (`CreateMaterializedViewBuilder::execute` prepares fresh each time). The fix is tiny either way: mint the UUID inside `create()`, or remove the `Clone` derive — nothing in the crate clones `PreparedDeclaration` today.

jackye1995 (approved):
looks good!

LuQQiu (approved):

PR #4051, opened 2026-08-25
fix(python): correct pylance test extra pin

### LINKED ISSUE #4045: tests extra pins pylance==9.0.0rc1, which was never published to PyPI — blocks pip install -e ".[tests]"
`python/pyproject.toml`'s `tests` extra pins an exact version that doesn't exist on PyPI:

```
pylance==9.0.0rc1
```

`pip install -e ".[tests]"` fails for everyone, on any platform:

```
ERROR: Could not find a version that satisfies the requirement pylance==9.0.0rc1; extra == "tests" (from lancedb)
ERROR: No matching distribution found for pylance==9.0.0rc1; extra == "tests"
```

`9.0.0rc1` was never published — pylance's rc-tagged releases stop at `2.0.1rc1`; every release since, including `9.0.0` and `9.0.1`, shipped final only. Looks like a leftover from before the real `9.0.0` shipped.

Bumping the pin to `pylance==9.0.0` (or `>=9.0.0`) should fix it — both are already on PyPI.

Risto0211:
Closes #4045

## What changed
- replace the unpublished pylance 9.0.0rc1 test-extra pin with the published 9.0.0 release
- restore dependency resolution for editable installs using the tests extra

## Validation
- downloaded pylance==9.0.0 from PyPI with pip --no-deps
- parsed python/pyproject.toml with tomllib and verified the tests extra
- git diff --check


Risto0211:
Updated `python/uv.lock` with uv 0.8.15 so the committed lockfile now resolves Pylance 9.0.0. Validation: `uv lock --check` and `git diff --check` both pass.

wjones127 (approved):

PR #4070, opened 2026-08-27
feat: add asynchronous remote SQL queries

jackye1995:
## Summary

Add SQL execution to remote LanceDB connections. On the standard synchronous connection, `execute_query` waits for the initial result stream and returns its Arrow reader. `execute_query_async` is called without Python `await` and immediately returns a query handle for status inspection, streaming, or cancellation. Local databases report that SQL is not supported.

The transport and query lifecycle live in Rust. Python exposes native-backed synchronous and asynchronous connection methods and query wrappers; it does not use PyArrow's Flight client.

## User experience

The standard synchronous connection supports both direct reads and background query execution:

```python
db = lancedb.connect(
    "db://analytics",
    api_key="ldb_...",
    sql_host_override="grpc+tls://sql.example.com:10026",
)

# Direct execution waits only until the initial result stream is available.
# Later batches continue streaming as the query progresses.
reader = db.execute_query(
    "SELECT * FROM events",
    default_namespace_path=["production"],
)
for batch in reader:
    print(batch.num_rows)

# Background execution returns a query handle immediately. Despite the
# `_async` suffix, no Python `await` is needed on a synchronous connection.
query = db.execute_query_async("SELECT * FROM events")
print(query.id)

description = db.describe_query(query.id)
print(description.status)
print(description.progress)
print(description.expires_at)

# Start reading as soon as the service advertises partial results. The reader
# continues polling and yields newly available record batches until the query
# and all result endpoints are complete.
reader = query.reader()
for batch in reader:
    print(batch.num_rows)

# Or cancel a different still-running query. Its status becomes "cancelling"
# while the server is still working, then "cancelled" once confirmed.
cancelled_query = db.execute_query_async("SELECT * FROM large_events")
cancelled_query.cancel()
```

The less commonly used asynchronous connection exposes the same operations as coroutines:

```python
async_db = await lancedb.connect_async(
    "db://analytics",
    api_key="ldb_...",
    sql_host_override="grpc+tls://sql.example.com:10026",
)
query = await async_db.execute_query_async("SELECT * FROM events")
async for batch in await query.reader():
    print(batch.num_rows)
```

The UUIDv7 query id is scoped to the connection that submitted it. The connection retains lightweight shared query state used by `query.describe()` and `db.describe_query(query.id)`; the id does not encode SQL or a Flight continuation token and is not a cross-connection resume token. Abandoned state has bounded retention, and terminal state remains available briefly.

Unqualified table names use the connected database and the `public` namespace by default. `default_namespace_path` accepts a list such as `["production", "events"]`. SQL can still use qualified names to reference other databases and namespaces available to the deployment.

## Design

- Uses Arrow Flight `PollFlightInfo` for submission and long polling, `DoGet` for results, and `CancelFlightInfo` for cancellation. Each `PollInfo.info` is treated as the cumulative set of currently available endpoints, so advertised tickets are consumed once and batches can be delivered before execution is complete.
- Serializes result completion and cancellation into one lifecycle. A server-accepted request reports `cancelling` and wakes blocked status/result work; a later retry can confirm `cancelled`. Result retrieval is rejected after cancellation is accepted, while cancellation after a result was already delivered is a no-op.
- Assigns a time-ordered UUIDv7 connection-scoped query id and retains only shared evolving lifecycle state, keeping SQL, Flight continuation tokens, and Arrow result data out of public ids and the registry.
- Leaves admission control to the server while honoring server expiration and a local fallback retention window for abandoned entries.
- Retains terminal ids for five minutes so they remain available for connection-level description.
- Keeps one lazily initialized SQL client on each remote database connection and attaches fresh authentication, routing, namespace, and request metadata to every operation.
- Applies the configured overall timeout to each execution, description, reader, and cancellation operation. A result reader carries one absolute deadline from `reader()` through the end of streaming; connect and read timeouts continue to bound their individual phases.
- Returns a bounded, backpressured, single-consumer Arrow stream rather than collecting the full result in memory. Dropping the reader stops downloading but does not implicitly cancel the server query.
- Preserves typed schemas for empty result sets through the stream schema.
- Accepts Flight result messages up to 1 GiB so a valid row containing a large blob, string, or vector is not rejected by tonic's 4 MiB default receive limit.
- Supports the Python client first while keeping the authoritative implementation in the Rust core.


rust/lancedb/src/sql.rs
```diff
@@ -0,0 +1,85 @@
...
+use async_trait::async_trait;
+use chrono::{DateTime, Utc};
+
+use crate::{Result, arrow::SendableRecordBatchStream};
+
+/// A point-in-time description of a submitted SQL query.
+#[derive(Clone, Debug, PartialEq)]
+pub struct QueryDescription {
+    /// The stable, connection-scoped identifier assigned when the query was submitted.
+    pub id: String,
```
wjones127: suggestion: this seems to always be constructed as as UUID, why not make it a UUID type?


```suggestion
    /// The stable, connection-scoped identifier assigned when the query was submitted.
    pub id: UUID,
```

rust/lancedb/src/remote/sql.rs
```diff
@@ -0,0 +1,2680 @@
...
+const DEFAULT_SQL_TLS_PORT: u16 = 10026;
+const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_secs(120);
+const DEFAULT_READ_TIMEOUT: Duration = Duration::from_secs(300);
+const STATUS_POLL_TIMEOUT: Duration = Duration::from_secs(1);
+const MIN_POLL_INTERVAL: Duration = Duration::from_millis(50);
+const MAX_SQL_MESSAGE_SIZE: usize = 1024 * 1024 * 1024;
+const QUERY_CACHE_CAPACITY: u64 = 10_000;
+const TERMINAL_QUERY_RETENTION: Duration = Duration::from_secs(300);
+const ABANDONED_QUERY_RETENTION: Duration = Duration::from_secs(24 * 60 * 60);
+const QUERY_ID_PREFIX: &str = "lq1_";
```
wjones127: question: what is the prefix for? If we remove it, we can simply use a UUID type. If there's a good justification for it, could you please add that in a comment?

rust/lancedb/src/remote/sql.rs
```diff
@@ -0,0 +1,2680 @@
...
+        with_overall_timeout(timeout, "SQL query submission", async {
+            validate_namespace_path(default_namespace_path)?;
+            let permit = self.queries.reserve()?;
+            let command = CommandStatementQuery {
+                query: query.to_string(),
+                transaction_id: None,
+            };
+            let descriptor = FlightDescriptor::new_cmd(command.as_any().encode_to_vec());
+            let poll_info = self.inner.poll(descriptor, default_namespace_path).await?;
+            let query_id = format!("{QUERY_ID_PREFIX}{}", uuid::Uuid::new_v4().simple());
```
wjones127: suggestion: have you considered using v7 UUIDs here? If we decide to log queries here, it would make them naturally clustered by their ids in the logs, which would make it faster to query. Something to think about.

rust/lancedb/src/sql.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// The lifecycle state: `running`, `finished`, `cancelling`, or `cancelled`.
+    pub status: String,
```
wjones127: question: we have that nice `enum QueryLifecycle` elsewhere, why not use that here instead?

rust/lancedb/src/sql.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Server-reported completion progress, when known.
+    pub progress: Option<f64>,
```
wjones127: suggestion: would be nice to know what this means.

```suggestion
    /// Server-reported completion progress, when known. In [0.0, 1.0], with 1.0
    /// meaning complete.
    pub progress: Option<f64>,
```

rust/lancedb/src/connection.rs
```diff
@@ -405,6 +445,51 @@ impl Connection {
...
+    ///
+    /// ```no_run
+    /// # async fn query(db: &lancedb::Connection) -> lancedb::Result<()> {
+    /// use futures::TryStreamExt;
+    ///
+    /// let query = db
+    ///     .execute_query_async("SELECT * FROM events LIMIT 10")
+    ///     .default_namespace_path(["public"])
+    ///     .execute()
+    ///     .await?;
```
wjones127, line 465: praise: great to see a tested example in the API docs.

rust/lancedb/src/remote/sql.rs
wjones127, line 1: suggestion: this module is over 2.5k lines long. Consider breaking it up into smaller modules.

rust/lancedb/src/remote/sql.rs
```diff
@@ -0,0 +1,2680 @@
...
+    api_key: String,
+    host_override: Option<String>,
+    sql_host_override: Option<String>,
+    client_config: ClientConfig,
+    client: Arc<OnceCell<SqlConnection>>,
+}
+
+struct SqlConnection {
+    channel: Channel,
+    client: FlightServiceClient<Channel>,
```
wjones127, line 63: question: is the reason for the duplicate channel so you can access it? Would be nice to have a comment explaining this.

rust/lancedb/src/remote/sql.rs
```diff
@@ -0,0 +1,2680 @@
...
+        });
+    }
+    if !value.is_ascii() || value.bytes().any(|byte| !(0x20..=0x7e).contains(&byte)) {
+        return Err(Error::InvalidInput {
+            message: format!("SQL metadata must be printable ASCII: {key:?}"),
+        });
+    }
+    headers.insert(key, value.to_string());
+    Ok(())
+}
```
wjones127: thought: we should probably switch to using `http::HeaderMap` instead of hand-rolling header validation. It has really efficient handling of well-known headers.

rust/lancedb/src/remote/sql.rs
```diff
@@ -0,0 +1,2680 @@
...
+            .map(|entry| entry._permit)
+    }
+
+    fn capacity_error(&self) -> Error {
+        Error::Runtime {
+            message: format!(
+                "This connection already retains {} active SQL queries",
+                self.max_capacity
+            ),
+        }
```
wjones127: question(blocking): Why do we need this? Is server-side rate-limiting not sufficient?

rust/lancedb/src/remote/sql.rs
```diff
@@ -0,0 +1,2680 @@
...
+        let timeout = self.inner.overall_timeout()?;
+        with_overall_timeout(timeout, "SQL query submission", async {
+            validate_namespace_path(default_namespace_path)?;
+            let permit = self.queries.reserve()?;
+            let command = CommandStatementQuery {
+                query: query.to_string(),
+                transaction_id: None,
+            };
+            let descriptor = FlightDescriptor::new_cmd(command.as_any().encode_to_vec());
+            let poll_info = self.inner.poll(descriptor, default_namespace_path).await?;
```
wjones127, line 198: suggestion: in the future, it would be nice to be able to bypass the `GetFlightInfo` step and just submit the SQL text directly to `DoGet` as the ticket. That's how [ClickHouse does this](https://clickhouse.com/docs/concepts/features/interfaces/arrowflight#doget) to achieve queries running in a single RPC. To start with, this could just be an explicit flag, but maybe we could automatically activate that path in some situations.

rust/lancedb/src/remote/oauth.rs
```diff
@@ -554,10 +554,10 @@ impl OAuthHeaderProvider {
 impl HeaderProvider for OAuthHeaderProvider {
     async fn get_headers(&self) -> Result<HashMap<String, String>> {
         let token = self.get_valid_token().await?;
-        Ok(HashMap::from([(
-            "authorization".to_string(),
-            format!("Bearer {token}"),
-        )]))
+        Ok(HashMap::from([
+            ("authorization".to_string(), format!("Bearer {token}")),
+            ("x-lancedb-credential-type".to_string(), "oidc".to_string()),
```
wjones127, line 561: question: what's the reason behind this change? Is this something the FlightSQL server requires for oidc?

wjones127 (changes requested):
This PR is a great step forward! I have several questions, including one blocking one about the client-side concurrency limiting.

rust/lancedb/src/sql.rs
jackye1995: Done. The Rust API now uses `uuid::Uuid` for both `Query::id()` and `QueryDescription::id`, and the typed value flows through `Connection::describe_query`. Python stringifies/parses only at the binding boundary.

rust/lancedb/src/remote/sql.rs
jackye1995: Agreed. I removed the `lq1_` prefix and made the connection registry and Rust public API use `Uuid` directly.

rust/lancedb/src/remote/sql.rs
jackye1995: Done. Query IDs now use UUIDv7, which keeps them opaque while providing time ordering for future logging/indexing.

rust/lancedb/src/sql.rs
jackye1995: Agreed on making this typed. I added the public `QueryStatus` enum and changed `QueryDescription.status` to use it. The private `QueryLifecycle` remains intentionally more detailed because it distinguishes `Ready` from `Completed` for result-consumption and cancellation linearization; both map to the public `Finished` status.

rust/lancedb/src/sql.rs
jackye1995: Added the range and completion meaning to the field documentation.

rust/lancedb/src/connection.rs
jackye1995: Thanks!

rust/lancedb/src/remote/sql.rs
jackye1995: Addressed the largest structural issue here by moving the roughly 1,000-line in-process SQL protocol test harness into `remote/sql_test.rs`. The production module is now about 1.5k lines; I kept the tightly coupled lifecycle state machine together for this revision.

rust/lancedb/src/remote/sql.rs
jackye1995: Yes. Arrow `FlightClient` does not expose its underlying transport, while cancellation needs a fresh low-level client with a per-call interceptor so we can distinguish requests that reached dispatch from pre-dispatch timeouts. I added a comment explaining why the channel is retained separately.

rust/lancedb/src/remote/sql.rs
jackye1995: Done. SQL metadata is now assembled and validated with `http::HeaderMap`, `HeaderName`, and `HeaderValue`; dynamic provider maps are normalized into it at the boundary.

rust/lancedb/src/remote/sql.rs
jackye1995: Agreed. I removed the client-side 10,000-query semaphore, admission rejection, and pressure eviction. Admission/rate limiting is now left to the server. The connection registry only retains lightweight state for ID lookup and applies the existing server-expiration and abandonment cleanup.

rust/lancedb/src/remote/sql.rs
jackye1995: Agreed as a useful future optimization. The current async lifecycle relies on `PollFlightInfo` so it can expose progress, cancellation, and partial endpoints. A direct SQL-as-DoGet-ticket path can be added later as an explicit or negotiated fast path for servers that implement that contract, without changing this API.

rust/lancedb/src/remote/oauth.rs
jackye1995: Yes. The LanceDB SQL service uses this marker to distinguish an OIDC bearer credential from API-key credential handling. I expanded the `OAuthHeaderProvider` documentation to make that requirement explicit.

jackye1995:
Addressed all of the latest review feedback in the current revision:

- removed the client-side 10,000-query admission limit and left rate limiting to the server
- switched Rust query IDs to typed UUIDv7 values end-to-end (Python remains string-based at the binding boundary)
- added a typed public `QueryStatus` and documented progress semantics
- replaced hand-rolled metadata validation with `http::HeaderMap`
- documented why cancellation retains the underlying channel separately
- clarified the OIDC credential marker requirement
- moved the large in-process SQL protocol test harness into a separate test module
- recorded direct SQL tickets in `DoGet` as a future negotiated/opt-in fast path

All inline threads have replies and are resolved. The branch is rebased onto current upstream, targeted Rust SQL tests pass, Python builds with 17 SQL tests passing, and the post-rebase SELF/Claude/GLM reviews are clean.

@wjones127 ready for re-review.

rust/lancedb/src/sql.rs
```diff
@@ -0,0 +1,112 @@
...
+    Finished,
+    /// The server accepted cancellation but has not confirmed it yet.
+    Cancelling,
+    /// The server confirmed cancellation.
+    Cancelled,
+}
+
+impl QueryStatus {
+    /// Return the stable lowercase representation used by language bindings.
+    pub fn as_str(self) -> &'static str {
```
wjones127: suggestion: perhaps this should just be a `Display` implementation.

python/python/lancedb/sql.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class AsyncQuery:
+    """A handle to a submitted SQL query on an asynchronous connection."""
+
+    def __init__(self, inner: "_lancedb.SqlQuery"):
+        self._inner = inner
+
+    @property
+    def id(self) -> str:
```
wjones127: question(non-blocking): what made you choose `str` for the Python representation instead of `uuid.UUID`?

wjones127 (approved):
Looks good. Only non-blocking feedback now. Nice work on this!

rust/lancedb/src/sql.rs
jackye1995: Done — `QueryStatus` now implements `Display` with the same stable lowercase values, and the Python binding uses `status.to_string()`. I also added coverage for all four rendered states.

python/python/lancedb/sql.py
jackye1995: Agreed that `uuid.UUID` is the better public representation. The earlier string conversion was only boundary convenience; PyO3 already supports native UUID conversion. This revision preserves UUID typing for `Query.id`, `QueryDescription.id`, and `describe_query(...)` end to end, updates the stubs/wrappers, and tests both UUID results and rejection of string inputs.

PR #4074, opened 2026-08-27
feat(node)!: require Node >= 22 and drop npm lockfiles

wjones127:
The bindings are built, installed and published with pnpm everywhere, but a parallel npm dependency graph was still being maintained beside it. This removes it, raises the supported Node floor to the versions we actually test, and gives Dependabot the npm coverage it was missing.

## Dropping npm

`nodejs/package-lock.json` was regenerated by `ci/update_lockfiles.sh` on every release commit and read by nothing — no workflow runs `npm ci` or `npm install` in `nodejs/`, and npm never publishes a lockfile in a package tarball. It could not even agree with the real install, since npm does not see pnpm's `overrides`. Because GitHub's dependency graph parses `package-lock.json`, it was also reporting vulnerabilities for a tree we neither install nor ship.

`docs/package.json`, `docs/package-lock.json` and `docs/tsconfig.json` go too. They depend on `file:../node` and `file:../node/node_modules/apache-arrow` — the `node/` directory was removed long ago — the tsconfig compiles `src/*.ts` where no TypeScript files exist, and nothing installs any of it. `docs.yml` only referenced the lockfile to configure an npm cache for an install it never ran.

Two `workflow_dispatch` workflows for regenerating those lockfiles are removed as well. Both were already broken: they `uses:` composite actions at `.github/workflows/update_package_lock{,_nodejs}` that do not exist, so dispatching either failed immediately.

The remaining `npx` calls become direct `node_modules/.bin/...` invocations. These were already running locally installed binaries rather than resolving anything, but naming the binary removes the npm CLI from the loop and does not depend on which Node version is active. `dev.yml`'s commitlint check was the last place doing real npm dependency resolution — an unpinned `npm install @commitlint/config-conventional` that also bypassed the `minimumReleaseAge` hold configured for `nodejs/` — and is now a pinned `pnpm dlx`.

## Node support

Node 18 and 20 both reached end-of-life, in April 2025 and April 2026. The matrix moves to 22, 24 and 26, and `engines` rises from `>= 18` to `>= 22` so the declared floor is one the matrix actually covers. Node 22 is LTS until April 2027; 24 is LTS; 26 is Current and becomes LTS in October 2026.

This also removes the reason the workflows reached for `npx` in the first place: pnpm 11 requires Node >= 22.13, which every matrix version now satisfies.

The prebuilt-binary smoke test in `npm-publish.yml` moves from Node 20 to Node 22 — the floor, where a napi ABI problem would surface first — rather than fanning out across all three, to keep the publish matrix from tripling.

## Dependabot

There were no npm-ecosystem entries at all, which is why the advisories behind #4073 went unnoticed. Both pnpm lockfiles are now watched — `nodejs/` and `nodejs/examples/`, which is a separate install — using the same `lockfile-only` strategy as the existing cargo and pip entries, so version ranges in `package.json` are left alone.

## Pre-commit biome

The hook ran `npx @biomejs/biome@1.8.3` while `nodejs/package.json` resolved 1.9.4. The two disagree about formatting, so the hook rejected code that `pnpm lint` accepts, and failed on unmodified `main` for anyone touching `nodejs/`. It now uses the pnpm-managed biome, which fixes the drift with no source changes.

## Testing

`dev.yml`'s commitlint job does not check out the repo, so it runs in an empty workspace, and I could not verify `pnpm/action-setup` there locally. It triggers on `pull_request_target`, so this PR exercises it directly — worth confirming green before merge. I did verify the `pnpm dlx` invocation itself locally: it accepts a conventional title and rejects a non-conventional one with exit 1.

Node 26 is new enough that the examples job may surface gaps in prebuilt native binaries (`onnxruntime-node`, `sharp`) before their maintainers publish for it.

## Not included

`nodejs/examples/` still pins `sharp: "0.33.5"` and has its own audit findings. Raising the Node floor unblocks that work — sharp 0.35 requires Node >= 20.9, which the matrix now satisfies — but it is a dependency bump rather than tooling cleanup, so it is left separate.

## Breaking changes

`@lancedb/lancedb` now requires Node >= 22; previously >= 18. The `@types/node` peer range moves from `>=18` to `>=22` to match. Users on Node 18 or 20 must upgrade their runtime; both have been end-of-life for some time. Existing installs are unaffected, since `engines` is only checked on install.


.pre-commit-config.yaml
```diff
@@ -20,7 +20,10 @@ repos:
     hooks:
       - id: local-biome-check
         name: biome check
-        entry: npx @biomejs/biome@1.8.3 check --config-path nodejs/biome.json nodejs/
+        # Use the biome from nodejs/package.json rather than a separately
+        # pinned one: the two drifted apart and disagreed on formatting, so
+        # this hook rejected code that `pnpm lint` accepted.
+        entry: nodejs/node_modules/.bin/biome check --config-path nodejs/biome.json nodejs/
```
wjones127, line 26: question(blocking): does this break the pre-commit hook for someone who hasn't run `pnpm install`?

wjones127:
Fixing failed builds in https://github.com/lancedb/lancedb/pull/4077

.pre-commit-config.yaml
wjones127: Confirmed, and I reproduced it: with no `nodejs/node_modules`, the hook fails with `Executable \`nodejs/node_modules/.bin/biome\` not found`. The old `npx @biomejs/biome@1.8.3` self-provisioned, so it worked from a bare checkout.

Keeping it as-is. The blast radius is people editing `nodejs/`, who need `pnpm install` anyway to build or run `pnpm lint`, and `nodejs/CONTRIBUTING.md` already lists `pnpm install` under "Initial setup" ahead of `pre-commit install`. Anything that self-provisions (`npx`, `pnpm dlx`) has to name a version literally, which is the drift that made the old hook reject code `pnpm lint` accepts — `package.json` says `^1.7.3`, so the real version comes from the lockfile. Verified `pre-commit run local-biome-check --all-files` passes against the installed biome 1.9.4.

nodejs/package-lock.json
```diff
@@ -1,11106 +0,0 @@
-{
```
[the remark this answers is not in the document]
wjones127: Fixed in e39bc9e5d. `docs/README.md` now uses `pnpm install` for both `nodejs/` and `nodejs/examples/`, and `pnpm run docs` for the typedoc step (`pnpm docs` would hit pnpm's built-in). Same treatment for the two contributor instructions in `AGENTS.md` and a stale comment in `docs-link-check.yml`.

Verifying the documented path turned up a second break: the examples' `test` script ran `node node_modules/.bin/jest`, and under pnpm that path is a shell shim rather than a symlink to the JS entry, so node died with `SyntaxError: missing ) after argument list`. CI never hit it because it invokes the shim directly. The script now points at `node_modules/jest/bin/jest.js`, which works under both package managers.

Ran the whole sequence end to end: `pnpm install` + `pnpm build` in `nodejs/`, then `pnpm install` + `pnpm test` in `nodejs/examples/` — passes.

Left alone deliberately: `npm install @lancedb/lancedb` in the READMEs and in the openai embedding error message, which are end-user instructions for a package we do publish to npm, and `npm install -g @openai/codex` in the codex workflows.

wkalt (approved):

PR #4085, opened 2026-08-28
feat(functions): support GPU resource requirements

Xuanwo:
Functions can describe their Python environment today, but cannot declare accelerator requirements. That prevents Sophon from scheduling computed-column UDF refreshes onto GPU workers from the immutable Function definition.

Add `num_gpus` to Python `@udf` through a typed `FunctionResourceRequirements` value and represent resource-aware definitions with the `python_v2` runtime discriminator. CPU Functions retain their existing `python` encoding and canonical identity.

The new discriminator is intentional for mixed-version safety: deployments that do not understand execution resources reject the runtime instead of accepting a new field and silently running the Function on CPU. Required resources are part of Function version identity; priority, concurrency, and retry policy remain Job concerns. The actual resource scheduling remains owned by Sophon.

### DIFF
--- a/python/python/lancedb/functions.py
+++ b/python/python/lancedb/functions.py
@@ -51,8 +51,31 @@ from pydantic import (
 
 _Int32 = conint(strict=True, ge=-(2**31), le=2**31 - 1)
 _UInt32 = conint(strict=True, ge=0, le=2**32 - 1)
-_PositiveUInt32 = conint(strict=True, gt=0, le=2**32 - 1)
 _UInt64 = conint(strict=True, ge=0, le=2**64 - 1)
+
+
+def _validate_gpu_wire_requirement(value: Any) -> str:
+    if not isinstance(value, str):
+        raise ValueError("runtime.gpu must be a string")
+    if not value or value != value.strip():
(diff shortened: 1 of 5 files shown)


python/python/lancedb/functions.py
```diff
@@ -1035,6 +1064,10 @@ def udf(
         Environment variables included in the Function definition.
     python_version : str, optional
         Remote Python major/minor version. Defaults to the client version.
+    num_gpus : int, optional
+        Number of whole NVIDIA GPUs required for every remote execution. Must
+        be greater than zero. The requirement is part of the immutable
+        Function version.
```
jmhsieh: Are we just porting  previous verison 1's?

We should consider changing this or deprecating it.   -- some thing more generic like `resources={}` arg so we can have flexibility to take on modal's more specific gpu settings.  https://modal.com/docs/guide/gpu

jmhsieh (commented):
We should decide if we want to just port the old name/mechanism and migrate to something that allows decisions to be more expressive.  (looking at modal as an example). 

something like 
`@udf(..., gpu= "H100:8" } ) `

jackye1995 (approved):

PR #3755, opened 2026-07-31
feat: connection-level job operations

wkalt:
Adds job operations to the connection surface, building on the Job
handle from #3742: job(id), list_jobs, get_job, cancel_job, and
job_history, plus a non-blocking Job.status(). Implemented on the
Database trait (defaulting to NotSupported), the remote backend
(/v1/jobs), and the Python and Node bindings; job_history returns Arrow
batches.

errors() and progress() are not included.

Tested with mocked endpoints in all three languages.

PR #4100, opened 2026-08-31
fix: gate job.rs's Error::Http arms behind the remote feature

### LINKED ISSUE #4096: job.rs references Error::Http unconditionally though the variant is remote-gated — fails to compile without the "remote" feature
## Summary

Building the `lancedb` crate at `v0.38.0` with `default-features = false` and without
the `remote` feature fails to compile:

```
error[E0599]: no variant named `Http` found for enum `error::Error`
  --> rust/lancedb/src/job.rs:56:40
   |
56 |             Some(request_id) => Error::Http {
   |                                        ^^^^ variant not found in `error::Error`

error[E0599]: no variant named `Http` found for enum `error::Error`
  --> rust/lancedb/src/job.rs:66:40
   |
66 |             Some(request_id) => Error::Http {
   |                                        ^^^^ variant not found in `error::Error`
```

## Root cause

- `error.rs`'s `Error::Http` (and `Error::Retry`) variants are `#[cfg(feature = "remote")]`-gated.
- `job.rs`'s `TerminalResult::decode` matches on `Error::Http { .. }` in two places (lines 56
  and 66) with no `#[cfg(feature = "remote")]` guard.
- `lib.rs:188` declares `pub mod job;` unconditionally (it is not itself gated behind
  `remote`), so `job.rs` is compiled for every feature combination, and the reference to the
  remote-only `Error::Http` variant fails to resolve whenever `remote` is off.

## Reproduction

```toml
[dependencies]
lancedb = { git = "https://github.com/lancedb/lancedb.git", tag = "v0.38.0", default-features = false }
```

```
cargo build -p lancedb
```

Fails with the two `E0599` errors above. This affects any downstream Rust consumer that
builds the crate without the `remote` feature (our case: a Go/CGo binding that pins
`default-features = false, features = ["metrics-otel"]`).

`v0.37.1` built fine under the same feature set; this is a regression introduced by the
`job.rs` additions in this release cycle (e.g. #3742, #3755, #3939, #4013).

## Suggested fix

The two `Error::Http { .. }` arms in `TerminalResult::decode` are only reachable when
`request_id` is `Some(..)`, which only happens via `TerminalResult::remote(...)`. Gating
those two arms behind `#[cfg(feature = "remote")]` (with a `#[cfg(not(feature = "remote"))]`
fallback to `Error::Runtime`, matching the existing `None => Error::Runtime { .. }` arm)
should be enough to make `job.rs` compile cleanly without `remote`.

## Environment

- lancedb tag: `v0.38.0`
- rustc `1.98.0`
- `Cargo.toml`: `default-features = false`, `remote` feature not enabled

hellower:
## What is the bug?

Fixes #4096.

Building `lancedb` with `default-features = false` and without the `remote` feature fails
to compile:

```
error[E0599]: no variant named `Http` found for enum `error::Error`
  --> rust/lancedb/src/job.rs:56:40
error[E0599]: no variant named `Http` found for enum `error::Error`
  --> rust/lancedb/src/job.rs:66:40
```

`TerminalResult::decode` (added by the recent job-handle work, e.g. #3742, #3755, #3939,
#4013) matches on `Error::Http { .. }` unconditionally in two places. `Error::Http` (and
`Error::Retry`) are `#[cfg(feature = "remote")]`-gated in `error.rs`, but `lib.rs`'s
`pub mod job;` is not itself feature-gated, so `job.rs` is compiled for every feature
combination — including any downstream consumer that doesn't enable `remote` (our own
case pins `default-features = false, features = ["metrics-otel"]`).

## How does this PR fix the problem?

`request_id: Option<String>` on `TerminalResult` is only ever `Some` via
`TerminalResult::remote(..)`, which is itself only constructed from `remote`-gated code
(`crate::remote::job`). So the two `Some(request_id) => Error::Http { .. }` arms are gated
behind `#[cfg(feature = "remote")]`, with a `#[cfg(not(feature = "remote"))]` fallback arm
producing `Error::Runtime` for the (unreachable without `remote`) case — matching the
existing `None => Error::Runtime { .. }` arm's message.

## Why did CI not catch this?

Worth flagging separately: the "Run clippy (without remote feature)" step in
`.github/workflows/rust.yml` runs `cargo clippy --workspace --tests` (no `--all-features`),
but `python/Cargo.toml` and `nodejs/Cargo.toml` both declare
`default = ["remote", ...]` on their `lancedb` dependency. Workspace-wide feature
unification means that default re-enables `remote` on the shared `lancedb` build unit even
in this step, so it never actually builds `lancedb` without `remote`. Scoping that step to
`-p lancedb` does catch this class of bug (confirmed locally), but it also currently trips
on ~10 pre-existing, unrelated `dead_code` warnings elsewhere in the crate (`job.rs`'s own
`TerminalResult::remote`/`value`, `query.rs`, `table/computed_columns.rs`,
`table/lsm_stats.rs`, `utils/mod.rs`) once it gets far enough to run clippy's lints — so
I've left that CI change out of this PR to keep it scoped to the actual bug. Happy to open
a follow-up for that if it's useful.

## Validation

All run against this branch, from `rust/`:

```
$ cargo check -p lancedb --tests
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 21.69s
# (only the pre-existing, unrelated dead_code warnings noted above; no errors)

$ cargo test -p lancedb --lib job::
running 2 tests
test job::tests::mapped_spawned_job_preserves_cancellation ... ok
test job::tests::mapped_spawned_job_reuses_outcome ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 737 filtered out

$ cargo check -p lancedb --tests --features remote
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 08s
# clean, confirming the remote-gated arms still compile and are exercised correctly

$ cargo fmt --all -- --check
# clean
```

No existing tests needed changes; `job.rs`'s own unit tests don't exercise the `remote`
path (nothing in that file constructs `TerminalResult::remote`), so this is purely a
compile-time fix.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,6 +63,13 @@ jobs:
         run: cargo clippy --profile ci --workspace --tests --all-features -- -D warnings
       - name: Run clippy (without remote feature)
         run: cargo clippy --profile ci --workspace --tests -- -D warnings
+      - name: Check lancedb without remote feature (package-scoped)
+        # Workspace-wide feature unification means the step above still
+        # builds `lancedb` with `remote` enabled, because `python/Cargo.toml`
+        # and `nodejs/Cargo.toml` both default to it. Scope this check to
+        # just the `lancedb` package so `remote`-gated code paths (e.g.
+        # `rust/lancedb/src/job.rs`) are actually exercised without it.
+        run: cargo check --profile ci -p lancedb --tests --no-default-features
 
   deny:
     # Supply-chain checks: advisories, licenses, banned crates, and source
--- a/rust/lancedb/src/job.rs
+++ b/rust/lancedb/src/job.rs
@@ -52,21 +52,36 @@ impl TerminalResult {
     }
 
     fn decode<T: DeserializeOwned>(self) -> Result<T> {
+        // `request_id` is only ever `Some` via `TerminalResult::remote`, which
+        // is only constructed behind the `remote` feature (see
+        // `crate::remote::job`). `Error::Http` is likewise `remote`-gated, so
+        // the `Some` arms below must be too — without `remote` the case is
+        // unreachable, but the match still needs a value of type `Error`.
         let value = self.value.ok_or_else(|| match &self.request_id {
+            #[cfg(feature = "remote")]
             Some(request_id) => Error::Http {
                 source: "successful typed job response did not contain a result".into(),
                 request_id: request_id.clone(),
                 status_code: None,
+            },
+            #[cfg(not(feature = "remote"))]
+            Some(_) => Error::Runtime {
+                message: "successful typed job did not contain a result".to_string(),
             },
             None => Error::Runtime {
                 message: "successful typed job did not contain a result".to_string(),
             },
         })?;
         serde_json::from_value(value).map_err(|error| match self.request_id {
+            #[cfg(feature = "remote")]
             Some(request_id) => Error::Http {
                 source: format!("failed to parse typed job result: {error}").into(),
                 request_id,
                 status_code: None,
+            },
+            #[cfg(not(feature = "remote"))]
+            Some(_) => Error::Runtime {
+                message: format!("failed to parse typed job result: {error}"),
             },
             None => Error::Runtime {
                 message: format!("failed to parse typed job result: {error}"),


hellower:
@lancedb-gatekeeper addressed the request: added a package-scoped `cargo check --profile ci -p lancedb --tests --no-default-features` step to `.github/workflows/rust.yml` (right after the existing "Run clippy (without remote feature)" step).

This closes the gap you flagged — the workspace-wide "without remote" lint still built `lancedb` with `remote` enabled because `python/Cargo.toml` and `nodejs/Cargo.toml` both default that feature on, so feature unification masked the regression. The new step scopes to just the `lancedb` package, so `remote`-gated code (`rust/lancedb/src/job.rs`) is actually compiled without the feature going forward.

Verified locally:
```
$ cargo check --profile ci -p lancedb --tests --no-default-features
    Finished `ci` profile [optimized + debuginfo] target(s) in 3m 42s
```
(only the pre-existing, unrelated `dead_code` warnings you noted; no errors)

Pushed as d91a3f81.

wjones127 (approved):

PR #4106, opened 2026-09-01
feat: add DataFrame query API

jackye1995:
Adds an always-available, lazy DataFusion-style DataFrame API implemented in Rust, with thin Python bindings and a Rust core that can be exposed in Node later.

Users create a DataFrame from an opened table. The table supplies its schema, database, namespace, and table provider; transformations retain that context and execution needs no plan bytes, version, connection, or namespace arguments.

Supported plan operations include projection, filtering, aggregation, sorting, limits, distinct, aliases, column add/drop/rename, joins, and union/intersect/except. Expressions support comparisons, boolean and arithmetic operations, null checks, casts, aliases, sort expressions, and common aggregates.

```python
from lancedb import col
from lancedb import sql_functions as F

events = db.open_table("events", namespace_path=["production"]).to_df()
frame = (
    events.filter(col("status") == "active")
    .aggregate(["region"], [F.sum(col("amount")).alias("total")])
    .sort(col("total").sort(ascending=False))
    .limit(10)
)

# Direct result form:
reader = frame.execute()

# Lifecycle form (a separate submission):
query = frame.execute_async()
print(query.describe().status)
reader = query.reader()
query.cancel()  # an in-flight reader observes a cancellation error
```

The asynchronous API uses the same transformations:

```python
table = await async_db.open_table("events", namespace_path=["production"])
frame = (await table.to_df()).filter(col("status") == "active").limit(10)

# Direct result form:
reader = await frame.execute()

# Lifecycle form (a separate submission):
query = await frame.execute_async()
print((await query.describe()).status)
reader = await query.reader()
await query.cancel()
```

Qualified columns remain unambiguous across aliased joins:

```python
left = events.alias("left").with_column_renamed("amount", "renamed")
right = events.alias("right").with_column_renamed("amount", "renamed")
joined = left.join(right, on="id").select(
    left.col("renamed").alias("left_amount"),
    right.col("renamed").alias("right_amount"),
)
```

Local tables execute the DataFusion plan in-process. Remote tables with Flight SQL configured serialize and submit the plan through Flight SQL; other connections use the in-process provider path. The durable client-side representation is DataFusion's logical plan. Substrait is only its private Flight SQL wire encoding: plan bytes and versioning are not public SDK APIs.

The Rust planner uses one shared DataFusion session. DataFrames from different connections or namespaces cannot be joined or combined. Both local and remote DataFrames resolve table data at submission time. Creating a DataFrame from a checked-out version or branch—and checking out any source handle after creating a DataFrame—is rejected until the remote transport can preserve revision identity.

DataFrame expressions currently cover functions exposed by the client. Substrait records function names and the server resolves their implementations from its session registry, but a function known only to a newer server has no typed client-side expression yet; `execute_query` / `execute_query_async` is the escape hatch for those server-only functions.

Flight SQL also supports prepared Substrait plans with typed dynamic parameters. That flow creates a prepared plan, receives its parameter schema and opaque handle, binds values separately as Arrow batches through `DoPut`, then executes the handle. This PR uses the one-shot statement-plan path and embeds literal values; prepared parameter binding can be added without changing the public DataFrame plan representation.

This PR is client-only and builds on the asynchronous Flight SQL lifecycle merged in #4070.


docs/src/python/python.md
```diff
@@ -81,6 +81,51 @@ for batch in query.reader():
...
+    .sort(col("total").sort(ascending=False))
+    .limit(10)
+)
+
+reader = frame.execute()
+query = frame.execute_async()
+plan = frame.to_substrait()
+
+# Equivalent raw-plan submission:
+reader = db.execute_substrait(plan, default_namespace_path=["production"])
```
wjones127: question(blocking): why do we need to expose this? Can we just make the fact we use substrait an implementation detail in the SDKs?

rust/lancedb/Cargo.toml
```diff
@@ -22,7 +22,7 @@ arrow-ord = { workspace = true }
 arrow-cast = { workspace = true }
 arrow-ipc.workspace = true
 arrow-flight = { workspace = true, optional = true }
-prost = { version = "0.14", optional = true }
+prost.workspace = true
```
wjones127, line 25: question(blocking): what is `prost` becoming required even when `remote` feature isn't activated? Can it still be blocked by that? Or maybe be behind another feature flag?

rust/lancedb/src/dataframe.rs
```diff
@@ -0,0 +1,456 @@
...
+}
+
+impl DataFrame {
+    fn wrap(result: datafusion_common::Result<DfDataFrame>) -> Result<Self> {
+        result.map(|inner| Self { inner }).map_err(planning_error)
+    }
+
+    /// Create a lazy named-table scan with the supplied Arrow schema.
+    pub fn from_table(name: impl Into<String>, schema: Arc<Schema>) -> Result<Self> {
+        let context = SessionContext::new();
```
wjones127: issue(blocking): we shouldn't create a fresh context for every dataframe. Could we instead attach the `SessionContext` to either a global or else on a `Database` / `Connection`?

rust/lancedb/src/dataframe.rs
```diff
@@ -0,0 +1,456 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright The LanceDB Authors
+
+//! Immutable DataFusion logical plans for client DataFrame bindings.
+//!
+//! This module is the language-neutral implementation behind LanceDB's
+//! DataFrame APIs. Bindings translate language values to [`crate::expr::DfExpr`]
+//! and delegate planning and Substrait serialization to [`DataFrame`].
```
wjones127: issue(blocking): this is a public module, but this comments discusses as if it's internal implementation. Could you rewrite to be user-facing?

rust/lancedb/src/dataframe.rs
```diff
@@ -0,0 +1,456 @@
...
+        }
+    }
+
+    /// Return the current output Arrow schema.
+    pub fn schema(&self) -> Schema {
+        self.inner.schema().as_arrow().clone()
+    }
+
+    /// Serialize this logical plan to Substrait.
+    pub fn to_substrait(&self) -> Result<EncodedSubstraitPlan> {
```
wjones127: question: same as before—should this be public?

python/pyproject.toml
```diff
@@ -140,6 +140,9 @@ include = [
     "python/lancedb/background_loop.py",
     "python/lancedb/schema.py",
     "python/lancedb/sql.py",
+    "python/lancedb/dataframe.py",
+    "python/lancedb/expr.py",
+    "python/lancedb/sql_functions.py",
```
wjones127, line 145: praise: thank you for including the new files.

rust/lancedb/src/database.rs
```diff
@@ -350,6 +350,17 @@ pub trait Database:
...
+    /// Start executing a Substrait plan on a remote database.
+    async fn execute_substrait_async(
+        &self,
+        _plan: &[u8],
+        _version: &str,
+        _default_namespace_path: &[String],
+    ) -> Result<crate::sql::Query> {
+        Err(crate::error::Error::NotSupported {
+            message: "Substrait execution is not supported by this database".to_string(),
+        })
```
wjones127, line 363: question(non-blocking): will this be implemented soon in a follow up PR?

rust/lancedb/src/connection.rs
```diff
@@ -475,6 +511,18 @@ impl Connection {
...
     }
 
+    /// Start executing a serialized Substrait plan on a remote LanceDB database.
+    ///
+    /// `version` is the Substrait specification version used to encode `plan`.
+    /// Local connections return [`Error::NotSupported`].
+    pub fn execute_substrait_async(
+        &self,
+        plan: impl Into<Vec<u8>>,
+        version: impl Into<String>,
```
wjones127: question(blocking): Why make the user pass the version? How would they know that?

rust/lancedb/src/connection.rs
```diff
@@ -475,6 +511,18 @@ impl Connection {
...
+    ) -> ExecuteSubstraitAsyncBuilder {
+        ExecuteSubstraitAsyncBuilder::new(self.internal.clone(), plan.into(), version.into())
+    }
```
wjones127: question(blocking): It looks like the expected call is:

```rust
let source = DataFrame::from_table(
    "events",
    Arc::new(Schema::new(vec![Field::new("id", DataType::Int64, false)])),
)?;
let plan = source.filter(col("id").gt(lit(10_i64)))?.to_substrait()?;

db.execute_substrait_async(plan, version)
    .default_namespace_path(["ns_1", "ns_a"])
    .execute()
    .await?;
```

This seems odd for a few reasons:
1. The user has to provide the table's schema
2. The user has to create an intermediate substrait plan
3. The user has to know what version of sbustrait is used.
4. The namespace / database connection is bound to the query until the very end

All of this feels awkard to me. Why did you design it this way?

I would have guessed a design like this:

```rust
// Get the namespace first
let db = connect("db://prod").namespace(["ns_1", "ns_a"]).await?;
// Get the table and schema off of the namespace
let source = db.open_table("events").df()
// Just run the query, no messing around with substrait
source
   .filter(col("id").gt(lit(10_i64)))
   .execute()
```

python/python/lancedb/db.py
```diff
@@ -469,6 +470,26 @@ def create_table(
     def __getitem__(self, name: str) -> LanceTable:
         return self.open_table(name)
 
+    def table(
+        self,
+        name: str,
+        *,
+        namespace_path: Optional[List[str]] = None,
+    ) -> DataFrame:
```
wjones127: suggestion: it might be confusing to users to have both `.open_table()` and `.table()` methods. Perhaps we should just add a `.to_df()` method on `Table` instance? Then you could do `db.open_table("my_table").to_df()`

wjones127 (changes requested):
Thanks for working on this. Cool to see this happening.

I have several questions about the API. The biggest one is whether we really want substrait to be part of our public API, or left as an implementation detail.

rust/lancedb/src/connection.rs
jackye1995: Changed to this shape. Rust users now call table.to_df().await?, build immutable transformations, and call DataFrame::execute().await. The opened table supplies the schema, database, and namespace; no intermediate plan or version is exposed. Joins/set operations also validate that both sources share a connection and namespace.

python/python/lancedb/db.py
jackye1995: Implemented. The connection-level table() shortcut is removed; sync Python uses db.open_table(...).to_df(), and async Python uses await table.to_df().

rust/lancedb/src/dataframe.rs
jackye1995: Rewritten as user-facing module documentation showing Table::to_df(), transformations, and direct execution. The binding/serialization implementation language is gone.

docs/src/python/python.md
jackye1995: Agreed. I removed to_substrait and the raw connection submission APIs. The public flow is now opened table -> to_df() -> execute()/execute_async(); serialization and Flight submission stay inside the Rust implementation.

rust/lancedb/src/dataframe.rs
jackye1995: Fixed. DataFrames now reuse one process-wide LazyLock<SessionContext> instead of creating a context per source. The opened table database/namespace is carried separately as execution context.

rust/lancedb/Cargo.toml
jackye1995: prost remains unconditional deliberately because the Rust DataFrame planner is always available, including in non-remote builds, and its internal encoder is compiled with that API. We do not want a second DataFrame feature gate; unsupported backends report NotSupported only at execution.

rust/lancedb/src/dataframe.rs
jackye1995: No longer public. Plan encoding is now a private Rust method used only by DataFrame::execute.

rust/lancedb/src/connection.rs
jackye1995: Removed the version argument and the whole public raw-plan API. LanceDB derives the version from the internally generated plan and submits both internally.

rust/lancedb/src/database.rs
jackye1995: The public raw execution method is gone. Remote DataFrame execution is implemented internally in this PR; local backends intentionally return NotSupported because there is no Flight SQL endpoint.

python/python/lancedb/table.py
```diff
@@ -2656,6 +2662,10 @@ def schema(self) -> pa.Schema:
             A PyArrow schema object."""
         return LOOP.run(self._table.schema())
 
+    def to_df(self) -> DataFrame:
+        """Create a lazy DataFrame that scans this table."""
+        return DataFrame(LOOP.run(self._table._inner.to_dataframe()))
```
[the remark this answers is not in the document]
jackye1995: Fixed in 07339d34. Synchronous Table.to_df() now enters LOOP before constructing the native awaitable, and DataFrame.execute()/execute_async() defer the native async call through a coroutine. The focused Python DataFrame suite passes all 8 tests.

rust/lancedb/src/table.rs
```diff
@@ -1156,6 +1156,23 @@ impl Table {
...
+    /// Create a lazy DataFrame that scans this table.
+    ///
+    /// The returned DataFrame retains this table's database and namespace, so
+    /// [`crate::dataframe::DataFrame::execute`] needs no additional connection
+    /// or transport arguments.
+    pub async fn to_df(&self) -> Result<crate::dataframe::DataFrame> {
+        let database = self.database.clone().ok_or_else(|| Error::InvalidInput {
+            message: "this table is not bound to a database".to_string(),
+        })?;
+        crate::dataframe::DataFrame::from_open_table(
```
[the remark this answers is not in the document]
jackye1995: Fixed in 6a06951d. Because the current Flight transport cannot express a revision, Table::to_df() now rejects checked-out versions and branches instead of silently scanning current main. Native and remote tables both report revision state, and Rust regressions cover pinned main, reattaching latest, and branch handles.

rust/lancedb/src/database.rs
wjones127: > local backends intentionally return NotSupported because there is no Flight SQL endpoint.

Yeah, but we have DataFusion engine in the binary. We don't need FlightSQL, we can just run the plan in-process. So why not enable that?

jackye1995: Agreed. Implemented in 6fffe967 with the stream adapter follow-up in 1607d66c: DataFrames now use the existing BaseTableAdapter as their scan source. Local databases execute the DataFusion plan in-process; remote databases still serialize and submit it through Flight SQL. Added Rust and synchronous/asynchronous Python regressions for local execution, and updated the public docs/PR examples.

rust/lancedb/src/dataframe.rs
```diff
@@ -0,0 +1,639 @@
...
+
+use crate::{
+    Error, Result,
+    arrow::{SendableRecordBatchStream, SimpleRecordBatchStream},
+    database::Database,
+    sql::{Query, QueryDescription, QueryHandle, QueryStatus},
+    table::{BaseTable, datafusion::BaseTableAdapter},
+};
+
+static DATAFRAME_SESSION: LazyLock<SessionContext> = LazyLock::new(SessionContext::new);
```
wjones127, line 65: question(blocking): What's the implication here for functions? We have several custom functions we register on the session context server side? Can we use those functions on DataFrames, even if the client version doesn't have those? Is there any escape hatch there?

This intersects with the question of testing.

wjones127 (changes requested):
This is closer to good. I think there are three remaining blockers:

* [ ] Get consensus from engineering team that Substrait is the interchange format we want to support. This will mean we want to maintain custom plans in substrait for things like FTS and vector search.
* [ ] Integration testing plan between this and the server. There are some concerns, like how we resolve functions and UDFs that need to be addressed.
* [ ] Path forward for executing dataframes against local tables

rust/lancedb/src/dataframe.rs
jackye1995: The client does not ship a function implementation in the Substrait plan. DataFusion serializes a function anchor/name, and the server-side consumer resolves that name from its SessionContext registry, so a shared built-in/custom function executes using the server implementation. The client still has to construct a typed Expr (especially its return type), though. Today this API only exposes client-known functions; a function known only to a newer server has no safe typed DataFrame expression, and execute_query / execute_query_async is the escape hatch because parsing and type resolution happen server-side. A newer-client function against an older server will fail function resolution rather than silently change semantics. I made that compatibility boundary explicit in the PR description. I do not think we should add an untyped F.call escape hatch: DataFusion planning requires a return type, and we first need a stable name/version + signature contract for server functions. The remote end-to-end gap remains called out; the focused planner coverage verifies Substrait serialization, while server function resolution needs an integration fixture with the server registry.

rust/lancedb/src/table.rs
```diff
@@ -1156,6 +1161,31 @@ impl Table {
...
+        if self.current_branch().is_some() || self.inner.is_time_travel() {
+            return Err(Error::NotSupported {
+                message: "DataFrames do not yet support checked-out versions or branches"
+                    .to_string(),
+            });
+        }
+        let database = self.database.clone().ok_or_else(|| Error::InvalidInput {
+            message: "this table is not bound to a database".to_string(),
+        })?;
+        let snapshot = self.query_snapshot().await?;
```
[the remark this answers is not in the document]
jackye1995: Fixed by choosing consistent latest-at-submission data semantics in `89d1546e` through `913022eb`. `to_df()` no longer pins a query snapshot; local execution observes compatible writes committed before submission just as the remote named-table path does. The execution context retains every source handle (including both sides of joins/set operations), and `execute()` rejects the plan if any source was subsequently checked out to a version or branch. The new Rust and Python regressions cover observing an appended row at submission and rejecting a later checkout. The PR description now states these semantics explicitly.

jackye1995:
Clarification after digging into the protocol: this PR does **not** use Substrait as the SDK or client-side plan representation. The Rust core retains a DataFusion `LogicalPlan`; only remote `execute()` lowers it to Substrait as the private Flight SQL wire command. No plan bytes or Substrait version are exposed publicly.

Flight SQL also has a prepared-Substrait flow (`CreatePreparedSubstraitPlan`): a plan can contain typed dynamic-parameter placeholders, the server returns a parameter schema and opaque handle, and values are bound separately as Arrow batches through `DoPut(CommandPreparedStatementQuery)`. DataFusion 54 already lowers typed placeholders to Substrait `DynamicParameter`. The current PR uses the simpler one-shot plan command with embedded literals, but prepared parameter binding can be added later without changing the DataFrame representation or API.

So the compatibility commitment here is specifically Substrait at the client/server transport boundary. Future Lance-specific vector/FTS operators would still require an agreed Substrait extension (or a different server API path); they do not require exposing Substrait to SDK users. I updated the PR description to make that boundary and the parameter path explicit.

Protocol references: https://github.com/apache/arrow/blob/main/format/FlightSql.proto and https://substrait.io/expressions/dynamic_parameters/

PR #4107, opened 2026-09-01
fix: fetch blobs from the query snapshot

geruh:
This PR fixes blob fetch after compaction. Query results that include `_rowid` now carry the producing dataset version, and `fetch_blobs` reads that snapshot before resolving the addresses. The table's current checkout does not move.

Blob columns no longer enable stable row ids automatically since the normal query then fetch path no longer depends on them.

A regression test covers deferred index remap separately from query version provenance.

## Testing

* `cargo test --features remote -p lancedb --test blob_integration`
* `cargo test --features remote -p lancedb --lib query_version`
* `cargo test --features remote -p lancedb --lib explicit_producing_version`
* `cargo fmt --all`
* `ruff check python/python/lancedb/table.py`


jackye1995:
Oh I was actually not expecting you to fix this, I think it is fine to just say that user needs to checkout the table to the version they use for query to do take_blobs? Given this is just a temp thing before we fully harden the SRID path. It will just always show as a limitation if user don't enable SRID.

PR #4115, opened 2026-09-01
fix(remote): forward create index replace flag

ChilePiquin:
Remote create-index requests already expose `replace` on the builder, but the remote client did not consistently forward an explicit `replace=false` over REST. That meant create-only intent could be lost before it reached a remote server, even though local builders and Python APIs can express it. This PR forwards `replace=false` on the existing `create_index` endpoint and keeps the current default behavior unchanged for compatibility.

This was accomplished with the following changes:

- Serialize `replace: false` into the existing remote create-index request body when the builder is configured with `.replace(false)`.
- Forward `replace` through the synchronous Python remote `create_index` wrapper so `RemoteTable.create_index(..., replace=False)` reaches the repaired path.
- Continue omitting `replace` for the default path so existing remote create-index requests keep their current semantics.
- Document `name` and `replace` on the existing OpenAPI create-index request schema.
- Add coverage that verifies the remote client uses the existing `/create_index/` route and forwards `replace=false`, including the synchronous Python unified API.

### Testing

- `cargo fmt --all --check`
- `cargo test -p lancedb --features remote test_create_index_forwards_replace_false_on_existing_route --locked`
- `uv tool run maturin develop --extras tests,dev,embeddings`
- `uv run --frozen pytest python/tests/test_remote_db.py::test_remote_create_index_new_api`
- `uv run ruff format --check python/lancedb/remote/table.py python/tests/test_remote_db.py`
- `cargo build -p lancedb --features remote --locked`
- `cargo clippy -p lancedb --features remote --all-targets --locked -- -D warnings`

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -446,6 +446,15 @@ paths:
               properties:
                 column:
                   type: string
+                name:
+                  type: string
+                  description: Optional name for the created index.
+                replace:
+                  type: boolean
+                  default: true
+                  description: |
+                    Whether to replace an existing index with the same resolved
+                    name. Defaults to true.
                 metric_type:
                   type: string
                   nullable: false
--- a/python/python/lancedb/remote/table.py
+++ b/python/python/lancedb/remote/table.py
@@ -548,6 +548,7 @@ class RemoteTable(Table):
         LOOP.run(
             self._table.create_index(
                 column,
+                replace=replace,
                 config=config,
                 wait_timeout=wait_timeout,
                 name=name,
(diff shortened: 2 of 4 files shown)


ChilePiquin:
Addressed the non-blocking Gatekeeper note: the synchronous Python remote `create_index` wrapper now forwards `replace` into the async/Rust path, and `test_remote_create_index_new_api` covers `RemoteTable.create_index(..., replace=False)` sending `replace: false` over the existing `/create_index/` endpoint.

jackye1995 (approved):

PR #4122, opened 2026-09-03
docs: bring the open-source documentation into this repository

jackye1995:
Moves the open-source pages of [docs.lancedb.com](https://docs.lancedb.com) into this repository, beside the code they describe.

Part of a migration that puts documentation in the repository that owns the code. The published site is assembled from several sources; this becomes the first of them.

## Why

The pages describing this code lived in `lancedb/docs` — a repository with **10 stars**, which neither contributing guide mentions. This one has 11,288. A user who hit a wrong page while reading the source had to find a separate, near-invisible repository to fix it.

## What lands here

**`docs/web`** — 102 pages, plus the static assets and generated snippets they need. It is a **complete Mintlify site**:

```bash
npm i -g mint
cd docs/web && mint dev
```

That serves on its own with no other checkout. Enterprise pages stay in a private repository and are merged in when the published site is assembled, so links to `/enterprise/...` and the SQL and Geneva sections do not resolve here. **Every open-source page does.**

**`docs/web-tests`** — the tests the code examples are extracted from. They ran against a published wheel before; here they can run against this repository's own SDK, so an example that stops compiling fails the pull request that broke it rather than being noticed later.

## The SDK reference is untouched

Both directories sit beside `docs/src` and are invisible to it — `mkdocs.yml` sets `docs_dir: src`, so `mkdocs build`, the Pages deploy, and the nightly link check all behave exactly as before. No workflow changes.

## Verified

The assembled site is **byte-identical** to what it replaces — same 268 files, same navigation, same redirects, verified by `diff -rq` and by comparing full `mint export` bundles. Readers see nothing change.

Every section heading carries an explicit `{#anchor}`; those are how the Enterprise pages will attach their additions to the right section, so they are assigned once and not regenerated.


Xuanwo (approved):
I'm happy with this change.

PR #4123, opened 2026-09-03
feat: support nullable named function outputs

jackye1995:
Supports fully nullable named Function outputs while preserving the distinction between a valid all-null struct and a null/unassigned result.

## Concrete example

This UDF contract is now valid:

```python
@udf(
    input_schema=pa.schema([
        pa.field("text", pa.string(), nullable=False),
    ]),
    output_schema=pa.schema([
        pa.field(
            "embedding",
            pa.list_(pa.float32(), list_size=1024),
            nullable=True,
        ),
        pa.field("embedding_failure_reason", pa.string(), nullable=True),
        pa.field("embedding_failure_code", pa.int32(), nullable=True),
    ]),
)
def embed(text):
    ...
```

A successful row can return:

```text
embedding = [0.12, ...]
embedding_failure_reason = NULL
embedding_failure_code = NULL
```

If remote inference still fails after retries, it can return:

```text
embedding = NULL
embedding_failure_reason = "HTTP 429: rate limited"
embedding_failure_code = 429
```

An all-null but valid result struct is also assigned; it is not mistaken for unfinished work.

## Binding shapes

- Mapping the result to one output column stores the `StructArray` directly, including its parent validity bitmap.
- Flattening the result into top-level columns stores the parent validity in a reserved internal nullable Boolean assignment column that is not part of the UDF result mapping.
- An outer null struct remains unassigned/skipped. A valid struct remains assigned regardless of which child fields are null.
- Scalar Function outputs remain non-nullable.

The contract is preserved through Python registration, Rust application planning, persisted `FunctionBinding` metadata, schema revalidation, and Enterprise execution.

### DIFF
--- a/docs/src/python/python.md
+++ b/docs/src/python/python.md
@@ -94,6 +94,8 @@ listing a storage directory.
 
 ::: lancedb.functions.OutputMapping
 
+::: lancedb.functions.AssignmentMapping
+
 ::: lancedb.functions.FunctionBinding
 
 ::: lancedb.functions.RefreshColumnResult
--- a/python/python/lancedb/__init__.py
+++ b/python/python/lancedb/__init__.py
@@ -23,6 +23,7 @@ from .expr import Expr, col, lit, func
 from .schema import blob, vector
 from .job import AsyncJob, Job
 from .functions import (
+    AssignmentMapping as AssignmentMapping,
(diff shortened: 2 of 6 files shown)


wkalt (approved):
nice

PR #4129, opened 2026-09-04
feat: add SQL-backed DataFrame API

jackye1995:
## What this implements

This PR adds a LanceDB-owned, lazy DataFrame API in Rust with synchronous and asynchronous Python bindings. A DataFrame starts from an opened table with `to_df()` and supports immutable, chainable transformations:

- projection and filtering
- grouping and aggregation
- sorting, limits, offsets, and distinct rows
- adding, dropping, and renaming columns
- relation aliases and qualified columns for self-joins
- inner, outer, semi, and anti joins
- union, intersection, and difference operations

It also exposes common aggregate expression helpers through `lancedb.sql_functions`.

## User experience

Users compose a query without writing SQL or selecting a separate local/remote API:

```python
from lancedb import col
from lancedb import sql_functions as F

events = db.open_table("events", namespace_path=["production"]).to_df()
frame = (
    events.filter(col("status") == "active")
    .aggregate(["region"], [F.sum(col("amount")).alias("total")])
    .sort(col("total").sort(ascending=False))
    .limit(10)
)

# Direct blocking Arrow reader.
reader = frame.execute()

# Or retain the existing SQL query lifecycle handle.
query = frame.execute_async()
description = query.describe()
query.cancel()
```

The async connection exposes the same transformations and returns `AsyncQuery` / `AsyncRecordBatchReader`. The output schema is available before execution, and every transformation returns a new DataFrame so plans can be safely reused.

## Supported operations and generated SQL

The table below shows representative SQL produced for each DataFrame operation. The unparser adds identifier quoting and parentheses when required by the SQL dialect or plan shape.

| Operation | DataFrame example | Representative generated SQL |
| --- | --- | --- |
| Select columns | `users.select("name", "age")` | `SELECT users.name, users.age FROM users` |
| Filter | `users.filter(col("age") > 18)` | `SELECT users.name, users.age FROM users WHERE users.age > 18` |
| Group and aggregate | `events.aggregate(["region"], F.sum(col("amount")).alias("total"))` | `SELECT events.region, SUM(events.amount) AS total FROM events GROUP BY events.region` |
| Sort | `users.sort(col("age").sort(ascending=False))` | `SELECT users.name, users.age FROM users ORDER BY users.age DESC NULLS FIRST` |
| Limit and offset | `users.limit(10, offset=20)` | `SELECT users.name, users.age FROM users LIMIT 10 OFFSET 20` |
| Distinct | `events.select("region").distinct()` | `SELECT DISTINCT events.region FROM events` |
| Add or replace a column | `users.with_column("next_age", col("age") + 1)` | `SELECT users.name, users.age, users.age + 1 AS next_age FROM users` |
| Drop columns | `users.drop("age")` | `SELECT users.name FROM users` |
| Rename a column | `users.with_column_renamed("age", "years")` | `SELECT users.name, users.age AS years FROM users` |
| Alias and qualified self-join | `left.join(right, on="id").select(left.col("value"), right.col("value"))` | `SELECT left.value, right.value FROM events AS left JOIN events AS right ON left.id IS NOT DISTINCT FROM right.id` |
| Union all | `left.union(right)` | `SELECT left.id FROM left UNION ALL SELECT right.id FROM right` |
| Union distinct | `left.union(right, distinct=True)` | `SELECT left.id FROM left UNION SELECT right.id FROM right` |
| Intersection | `left.intersect(right, distinct=True)` | `SELECT DISTINCT left.id FROM left WHERE EXISTS (SELECT 1 FROM right WHERE left.id IS NOT DISTINCT FROM right.id)` |
| Difference | `left.except_all(right, distinct=True)` | `SELECT DISTINCT left.id FROM left WHERE NOT EXISTS (SELECT 1 FROM right WHERE left.id IS NOT DISTINCT FROM right.id)` |

## How it works

1. `Table::to_df` creates a DataFusion `TableScan` logical plan from the opened table's schema and retains its LanceDB database, namespace, and source-table context.
2. Each DataFrame operation delegates logical planning and expression validation to DataFusion while keeping LanceDB's public API independent from DataFusion's `DataFrame` type.
3. Local connections execute the completed logical plan directly in-process and wrap the Arrow stream in the same query lifecycle abstraction used by the public API.
4. Remote connections pass the logical plan through a LanceDB-owned SQL lowering layer. It normalizes plan shapes where necessary to preserve filters, aggregates, ordering, limits, set operations, qualified projections, and join semantics, then uses DataFusion's SQL unparser to produce a SQL statement.
5. The generated SQL is submitted through the existing `execute_query_async` path with the DataFrame's default namespace. Query status, cancellation, and Arrow result streaming therefore reuse the unified SQL execution path.

This removes the need for the separate Substrait/protobuf transport proposed in #4106. LanceDB owns the DataFrame surface and the SQL-lowering seam, so it can extend or specialize conversion behavior without exposing DataFusion internals.

## Current remote boundaries

- DataFrames operate on the latest table revision visible to the SQL service at submission time; checked-out versions and branches are rejected.
- Remote execution does not inherit a table handle's read-consistency interval or read-your-write freshness fence.
- Joins currently support at most two source relations. Operations that need to isolate a multi-qualified join result require selecting uniquely aliased output columns first.
- Plan shapes whose ordering cannot be represented faithfully in the generated SQL are rejected with a clear error instead of returning silently reordered results.


rust/lancedb/src/dataframe/sql.rs
```diff
@@ -0,0 +1,408 @@
...
+    Expr, JoinType, LogicalPlan, LogicalPlanBuilder, Operator,
+    expr_fn::binary_expr,
+    expr_rewriter::NamePreserver,
+    logical_plan::{Projection, Sort},
+};
+use datafusion_sql::unparser::Unparser;
+
+pub(super) fn plan_to_sql(plan: &LogicalPlan) -> datafusion_common::Result<String> {
+    let plan = collapse_join_output_projections(plan)?;
+    let plan = lift_output_sort_above_projection(plan)?;
```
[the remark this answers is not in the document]
jackye1995: Fixed in `d660caee`. Sorts over a scope-establishing projection now place a direct derived-table alias around that projection before applying `ORDER BY`, so the projected name is resolved from the derived relation instead of referenced against the source table. Added `generated_sql_sorts_projected_aliases`, which executes the generated SQL and verifies the renamed column and its ordering.

rust/lancedb/src/dataframe/sql.rs
```diff
@@ -0,0 +1,408 @@
...
+    let plan = collapse_join_output_projections(&plan)?;
+    validate_output_sort(&plan)?;
+    let plan = ensure_output_projection(plan)?;
```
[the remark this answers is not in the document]
jackye1995: Fixed in `d660caee`. SQL lowering now inserts explicit captured-output projections at the final result boundary and before `DISTINCT` / set-operation scopes. Added `generated_sql_uses_the_planned_scan_schema`, which executes a two-column planned frame against an evolved three-column source and verifies both plain-scan output and `DISTINCT` semantics remain two-column and match the planned schema.

rust/lancedb/src/dataframe/sql.rs
```diff
@@ -0,0 +1,433 @@
...
+    let plan = preserve_join_semantics(&plan)?;
+    let plan = project_table_scans(plan)?;
```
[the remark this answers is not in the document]
jackye1995: Fixed in `dc3440a1` / `d660caee`. The final lowering no longer wraps every `TableScan`. It adds explicit projections only at the final output boundary and before schema-sensitive `DISTINCT` / set-operation scopes, avoiding recursive projection growth. The full `dataframe::tests` group now passes under the default test-thread stack (15 passed), including `renders_joins_and_set_operations_as_sql` and both new regressions.

docs/src/python/python.md
```diff
@@ -81,6 +81,76 @@ for batch in query.reader():
...
+    .sort(col("total").sort(ascending=False))
+    .limit(10)
+)
+
+# Choose the blocking reader for a direct result.
+reader = frame.execute()
+
+# Or submit separately and retain a lifecycle handle.
+query = frame.execute_async()
+query.cancel()
```
westonpace, line 108: Minor nit: it's slightly confusing to use async here but not connect it to python's async.  We could call it `execute_job` instead?

docs/src/python/python.md
```diff
@@ -81,6 +81,76 @@ for batch in query.reader():
...
+reader = await frame.execute()
+query = await frame.execute_async()
+await query.cancel()
+```
+
+Local tables execute the plans in-process. Remote tables convert plans to SQL
+and submit them through the remote query service. Remote execution resolves the
+latest revisions visible to that service at submission time; it does not inherit
+per-table read-consistency intervals or read-your-write freshness fences from the
+table handles used to create the plan.
```
westonpace, line 123: This feels like LLM-speak, can we make it easier to understand (do humans even read these docs anymore)?

Do we support time-travel queries in the dataframe API yet?

```suggestion
Local tables execute the plans in-process. Remote tables convert plans to SQL
and submit them through the remote query service. Remote execution always queries the latest table version, regardless of read consistency settings.
```

docs/src/python/python.md
```diff
@@ -81,6 +81,76 @@ for batch in query.reader():
...
+avoids changing join semantics while the SQL lowering layer does not yet support
+that compound shape. Operations such as `limit` or `distinct` can also require
+isolating a join result; select uniquely aliased output columns before those
+operations when the join retains columns from both input relations.
+
+For deterministic remote ordering, apply `sort` after filters and aliases. One
+final projection directly after sorting is supported, including a projection
+that omits the sort key. Plans that would need to carry ordering through an
+intervening limit, nested limits, multiple projections, a filter, or an alias
+are rejected instead of emitting SQL whose outer query could lose the ordering.
```
westonpace, line 135: Are all these limitations exclusive to remote tables?  Or do we have some of these limitations on local tables too?

python/python/lancedb/dataframe.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def _wrap(self, inner: NativeDataFrame):
+        return type(self)(inner)
+
+    @property
+    def schema(self) -> pa.Schema:
+        """The schema produced by the current logical plan."""
+        return self._inner.schema()
+
+    def select(self, *expressions: Expression):
```
westonpace, line 45: Should these functions declare a return type?

python/python/lancedb/dataframe.py
```diff
@@ -0,0 +1,252 @@
...
+        else:
+            aggregate_values = list(aggregates)
+        if not all(isinstance(value, Expr) for value in aggregate_values):
+            raise TypeError("aggregates must contain Expr values")
+        groups = [_expression(value)._inner for value in group_values]
+        aggregate_exprs = [value._inner for value in aggregate_values]
+        return self._wrap(self._inner.aggregate(groups, aggregate_exprs))
+
+    def sort(self, *expressions: Union[Expression, SortExpr]):
+        """Sort rows using DataFusion-style sort expressions.
```
westonpace, line 88: What are "DataFusion-style sort expressions"?  I'm not sure users will understand what this means.

python/python/lancedb/dataframe.py
```diff
@@ -0,0 +1,252 @@
...
+
+        For remote execution, apply sorting after filters and aliases. One final
+        projection directly after sorting is supported, but ordering through an
+        intervening limit or multiple projections cannot yet be preserved.
```
westonpace, line 92: What is an "intervening limit"  Is that a limit applied before the sort?  Or after?

python/python/lancedb/dataframe.py
```diff
@@ -0,0 +1,252 @@
...
+    def with_column(self, name: str, expression: Expr):
+        """Add or replace a column."""
+        return self._wrap(self._inner.with_column(name, expression._inner))
+
+    def with_columns(self, **expressions: Expr):
+        """Add or replace multiple named columns."""
+        result = self
+        for name, expression in expressions.items():
+            result = result.with_column(name, expression)
+        return result
```
westonpace, line 143: I'm not sure I really understand what these functions do.  Is this just appending to the select?

python/python/lancedb/dataframe.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def drop(self, *columns: str):
+        """Remove columns from the result."""
+        return self._wrap(self._inner.drop(list(columns)))
+
+    def with_column_renamed(self, old_name: str, new_name: str):
+        """Rename a column."""
+        return self._wrap(self._inner.with_column_renamed(old_name, new_name))
```
westonpace, line 151: How is this different from alias?  Do we need two ways to solve the same thing?

python/python/lancedb/dataframe.py
```diff
@@ -0,0 +1,252 @@
...
+            self._inner.join(other._inner, keys(left_on), keys(right_on), how)
+        )
+
+    def union(self, other: "_DataFrameBase", distinct: bool = False):
+        """Union two compatible plans."""
+        self._validate_set_operation(other)
+        return self._wrap(self._inner.union(other._inner, not distinct))
+
+    def intersect(self, other: "_DataFrameBase", distinct: bool = False):
+        """Intersect two compatible plans using DataFusion set semantics.
```
westonpace, line 190: What are DataFusion set semantics?

python/python/lancedb/expr.py
```diff
@@ -190,6 +191,32 @@ def isin(self, values: "Iterable[ExprLike]") -> "Expr":
...
+        """Return True where this expression is null."""
+        return Expr(self._inner.is_null())
+
+    def is_not_null(self) -> "Expr":
+        """Return True where this expression is not null."""
+        return Expr(self._inner.is_not_null())
+
+    def between(self, low: ExprLike, high: ExprLike) -> "Expr":
+        """Return True where this expression is between two inclusive bounds."""
+        return Expr(self._inner.between(_coerce(low)._inner, _coerce(high)._inner))
```
westonpace, line 208: How are literals specified?  How would I say `between(5, 10)`?

Also, I'm pretty sure this function has no coverage in tests.

westonpace (commented):
Initial review of API

PR #4133, opened 2026-09-05
feat(python): bind named Secrets to Functions with EnvVarSecret

jmhsieh:
Lets a Python UDF use a credential without the credential entering the user's code.

A UDF that calls a model API has nowhere to put its key today. It goes in the function body — where it is packaged with the source, versioned with it, and readable by anyone who can read the function — or into `runtime.env`, which is ordinary configuration and appears wherever configuration appears. Neither is a place for a credential.

## The shape

A Secret is a database-scoped named credential. A Function binds it **by reference**, and the value is resolved by the server when the Function runs:

```python
db.create_secret("openai-prod", os.environ["OPENAI_API_KEY"])

@udf(pip=["openai==3.7.0"])
def analyze_caption(caption: str) -> str:
    client = OpenAI()                       # reads OPENAI_API_KEY from its env
    return client.responses.create(...).output_text

db.create_function(
    analyze_caption,
    secrets=[EnvVarSecret(secret="openai-prod", env_variable="OPENAI_API_KEY")],
)
```

| Call | Behavior |
| --- | --- |
| `db.create_secret(name, value)` | Fails if the name is taken. |
| `db.alter_secret(name, value)` | Fails if it does not exist. |
| `db.list_secrets()` | Names only. |
| `db.describe_secret(name)` | Name and timestamps. Never the value. |
| `db.drop_secret(name)` | Removes it; this is the revocation path. |
| `EnvVarSecret(secret, env_variable)` | Binds a Secret to the variable a Function reads. |

Sync and async. Secrets are a server-side catalog, so these raise `NotImplementedError` on a local connection, like the other Function catalog verbs.

## Design notes

**Binding by reference, not by value.** Passing a credential to `create_function` would make it part of the Function's identity, so rotating a key would mint a new Function version. Columns would stay pinned to the old version with a dead key, and the only way forward would be re-pointing each one — which recomputes the column. You would pay for a full backfill because a credential rotated. Binding by name instead means rotation reaches every bound Function with nothing re-registered and nothing recomputed, while re-pointing a Function at a *different* Secret does mint a version, because the binding is part of the identity and the value is not.

**The decorator declares nothing about secrets.** Binding is entirely a registration-time decision, so a Function's source and its registration request are byte-identical whether or not a credential is later bound to it — the same source runs unchanged in a notebook with the variable set locally. A declaration would also buy little, since it cannot catch a wrong *variable* name: nothing knows which variables a Function reads, because `OpenAI()` reads its own.

**One value carries both halves of a binding.** `EnvVarSecret(secret=, env_variable=)` rather than a mapping of variable to reference. It is a typed value so a credential cannot be passed by accident — a bare string in that position is a plausible-looking mistake with the opposite meaning, and it reads identically in a diff.

**No API returns a stored value.** `describe_secret` returns a `SecretInfo` with fields for the name and timestamps and none for a value, so the property holds structurally rather than by policy.

**Client-side rules:** at most 16 bindings per Function, one binding per environment variable, and disjointness from `env` — which is ordinary configuration stored with the Function, and would otherwise leave one variable with two values of different visibility.

## Compatibility

A Function that binds no Secret is unaffected: `bind_secrets(None)` returns the same request as before, and the golden registration fixtures match byte for byte. The wire format is unchanged — `secret_bindings` remains `{variable: secret}`.

## Testing

`cargo check --features remote -p lancedb` clean. 88 Python tests across the Function slices pass, goldens included. Exercised end to end against a server: create, describe, bind, refresh — the UDF read its variable and produced the expected rows — rotate, a rejected binding to a Secret that does not exist, and a 404 after drop.

## Why this is a draft

The branch carries seven commits that are not part of this change and are not yet on `main` — the function-assignment and nullable-named-output work, plus a lance v12.0.0-beta.11 bump:

```
c259c3b8 chore: update lance dependency to v12.0.0-beta.11
7ac93a56 feat: support nullable named function outputs
e96f7b1b docs: clarify nullable function output mappings
d73b0143 feat: preserve nullable function result assignment
377cb4fe fix: validate function assignment columns
8c4542eb docs: include function assignment mapping
e76d5b56 fix: accept function assignment metadata
```

`#4123` merged a different variant of the nullable-output work, so these are not equivalent patches and `git cherry` still reports all of them as absent from `main`. The reviewable content here is `secrets.py`, `functions.py`, `db.py`, `remote/db.py`, the `describe_secret` chain through `Connection` and the remote client, and their tests. Once the assignment work lands this wants a rebase so the diff is secrets only.

`main` has also moved three commits ahead of this branch's merge base.


jmhsieh:
Superseded by https://github.com/lancedb/lancedb/pull/4150

PR #4150, opened 2026-09-08
feat(secrets): named Secrets, bindings, and namespace addressing

jmhsieh:
Adds the client half of database-scoped named Secrets: a Secret is a name and
an opaque value stored by the service, and a Function binds one to the
environment variable its library already reads. Secrets are addressed by a
namespace path plus a name.

The UDF body is unchanged and stays portable — it reads `OPENAI_API_KEY` the
way it always did, and the binding is what puts a value there:

```python
db.create_secret("openai-prod", os.environ["OPENAI_API_KEY"])
function = db.create_function(
    analyze_caption,
    secrets=[
        EnvVarSecret(secret_name="openai-prod", env_variable="OPENAI_API_KEY")
    ],
)
function.secret_bindings   # the Secret's name, never its value
```

- `create_secret` / `alter_secret` / `list_secrets` / `describe_secret` /
  `drop_secret` on sync, async and remote connections, with the pyo3 binding
  and the Rust client behind them. Each takes `namespace_path` keyword-only,
  defaulting to the root.
- **There is no read API, by construction rather than by policy** — no code
  path returns a stored credential, and `describe_secret` answers with
  metadata only.
- `EnvVarSecret` is a pure local constructor: it contacts no server, so it
  cannot fail on a Secret that does not exist. It exists so that a bare string
  in that position — which would be a credential — is a `TypeError` rather
  than a plausible-looking mistake that reads identically in a diff.
- `create_function(..., secrets=[...])` carries the bindings as
  `secret_bindings`: a list of `SecretBinding` tagged by `kind`, so a later
  delivery mode is a variant rather than a sibling field. The value never
  travels — it is resolved by the service when the Function runs, which is what
  lets a rotation reach columns already pinned to an older FunctionVersion.
- A binding names its Secret as a `SecretReference` of `{name, namespace_path}`
  rather than one joined string, so no delimiter has to be excluded from every
  name and segment forever, and `ClientConfig.id_delimiter` cannot contradict
  an identity built on a fixed separator.
- A root namespace is omitted from the request body rather than sent empty, so
  a root request is byte-identical to one from a client that predates
  namespaces. Tests pin it.

This is the client surface the design's §4 describes; the service side lives in
sophon.

**Previously split across two PRs.** Namespace addressing was #4151, stacked on
this one; it is folded in here so the Secret identity contract — name,
namespace path, and the binding that carries both — is reviewable as one piece
rather than as a shape introduced and then replaced.

## Identifier safety, merged from #4189

**#4189 is merged into this branch**, so the client half of Secrets and the
guards on the identity it puts in the URL are one PR. What it added:

- Components are checked where the identifier is built, before a request is
  constructed. `create_secret("../jobs", value)` no longer resolves to
  `/v1/jobs/create` and delivers a credential-bearing body to a route with none
  of this one's body suppression.
- Each component is percent-encoded and joined by the delimiter, so nothing
  inside a component can end the path segment or add one.
- A component may not be empty, a relative segment (`.`, `..`, and their `%2e`
  spellings), or the delimiter itself — the three ways a component erases a
  boundary the split has to recover. `["prod", ""]` joined to `prod$`, which
  reads back as `["prod"]`.
- `$` is the only accepted `id_delimiter`, refused at client construction.
  `ClientConfig.id_delimiter` remains, since the identifier grammar comes from
  the Lance REST catalog standard, but a value that would produce identifiers no
  service splits the caller's way is now an error where it was written.
- One `build_object_identifier` and one character set serve tables, namespaces,
  Secrets, Functions and materialized views.

Components are checked for *addressability*, not a character set: the name's
own grammar stays each object's own, so a catalog database keeps the `/` that
`RemoteCatalog::validate_name` allows.

## Known shortcoming

`secret_bindings` is omitted from a registration body when empty, so a client
that binds nothing sends what a client without bindings sends. When a client
does bind a Secret and the service does not know the field, the field is
ignored: registration succeeds, the returned version carries no bindings, and
the Function fails at execution with the variable unset, far from the call that
asked for it.

`ServerVersion` is how this codebase refuses a feature the service is too old
for, and it gates five features already. It does not gate this one: it is held
per table, and registering a Function is a database-level call. Noted at the
field in `remote/db.rs`; wiring the gate is follow-up work.

**Tests:** lancedb lib 1340 passed, `first_class_function_slice1` 9,
`first_class_function_slice2` 3, plus Python tests across both slices.
Rebased onto `main` after #4176 (OCI Function identity), #4191 (`.`/`..` table
names) and #4195 (remote catalogs).


rust/lancedb/src/database.rs
```diff
@@ -249,9 +249,29 @@ fn function_catalog_not_supported<T>() -> Result<T> {
...
+/// What a database records about a Secret. Never its value.
+///
+/// Returned by [`crate::connection::Connection::describe_secret`]. There is no
+/// field for the credential and no method that could produce one.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Deserialize)]
+pub struct SecretInfo {
+    /// The Secret's database-scoped name.
+    pub name: String,
+    /// When the Secret was created, as an RFC 3339 timestamp.
+    pub created_at: String,
```
jackye1995: I think we should just have a name, audit info like created_at and updated_at are technically shared component that should be tracked across all objects, and today we expose it via audit log. I think we can keep secret bare minimum for now and tackle auditing as a whole if we want to surface this info for various objects.

rust/lancedb/src/function.rs
```diff
@@ -481,13 +493,115 @@ pub struct FunctionArtifactRequest {
     pub adapter: PythonAdapterSpec,
 }
 
+/// A Function binds at most this many Secrets to environment variables.
+///
+/// Each bound Secret is one extra read on the launch path of every fragment, so
+/// the count needs a bound for the same reason a credential needs a size limit.
+pub const MAX_FUNCTION_SECRET_ENV_BINDINGS: usize = 16;
```
jackye1995: these should be server side enforcements, not on client side

rust/lancedb/src/database.rs
```diff
@@ -249,9 +249,29 @@ fn function_catalog_not_supported<T>() -> Result<T> {
...
+
 /// The `Database` trait defines the interface for database implementations.
 ///
 /// A database is responsible for managing tables and their metadata.
+/// What a database records about a Secret. Never its value.
+///
+/// Returned by [`crate::connection::Connection::describe_secret`]. There is no
+/// field for the credential and no method that could produce one.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Deserialize)]
+pub struct SecretInfo {
```
jackye1995: should probably have a dedicated module for secrets instead of having it in database module

rust/lancedb/src/function.rs
```diff
@@ -409,6 +409,8 @@ pub struct FunctionVersion {
     runtime: PythonRuntimeSpec,
     runtime_digest: String,
     environment_digest: String,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    secret_env_bindings: BTreeMap<String, String>,
```
jackye1995: not sure what is the latest state for the design, seems like my comment disappeared there. But overall I think we should just call this `secret_bindings`, because we can bind it in different ways in function execution, env variable is just one way.

rust/lancedb/src/database.rs
jackye1995: I guess this needs some cleanup overall also for functions, but in general I think we should move towards that if possible for any new object added

rust/lancedb/src/function.rs
```diff
@@ -481,13 +493,115 @@ pub struct FunctionArtifactRequest {
...
+/// enforces. Bounded because the value is destined for a process environment.
+pub const MAX_SECRET_VALUE_BYTES: usize = 64 * 1024;
+
+/// Whether `name` is a portable POSIX environment variable name.
+///
+/// Leading letter or underscore, then letters, digits, or underscores. Names
+/// reserved by the execution sandbox are deliberately not checked here: that
+/// list belongs to the runtime that owns it, and a copy in the client would
+/// drift from it silently.
+fn is_portable_env_name(name: &str) -> bool {
```
jackye1995: similar comment, I think all validation should exist at server side

rust/lancedb/src/remote/db.rs
```diff
@@ -352,6 +352,20 @@ impl RemoteDatabase {
 }
 
 impl<S: HttpSend> RemoteDatabase<S> {
+    /// `create` and `alter` differ only in which name state the server
+    /// requires, so they share one request shape. The value is a request field
+    /// and never a path segment or query parameter, which keeps it out of
+    /// access logs and proxy traces.
+    async fn write_secret(&self, route: &str, name: &str, value: &str) -> Result<()> {
```
jackye1995: I think we should still explicitly just implement them twice, because they could diverge in the future.

rust/lancedb/src/remote/client.rs
```diff
@@ -845,7 +859,12 @@ impl<S: HttpSend> RestfulLanceDbClient<S> {
                 .headers()
                 .get("content-type")
                 .map(|v| v.to_str().unwrap());
-            if content_type == Some("application/json") {
+            if route_carries_credential(request.url().path()) {
```
jackye1995: this feels a bit backward in dependency because the client now need to know API specific details. Can we just add a suppress_body input to the log_request function, and then let per API determine if body should be suppressed during logging?

rust/lancedb/src/function.rs
jmhsieh: My thinking is that when we support other binding types (files, accessors) we add `secret_file_bindings` and `secret_accessor_bindings` with the `skip_seraialized_if` they are empty.  They likely would have have different structures that we'd want in the FunctionVersion struct and should make forward compat easier.  

Specifically this env type is changing with the follow up #4151 to be accommodate namespaced secrets.

rust/lancedb/src/database.rs
jmhsieh: sure

rust/lancedb/src/function.rs
jmhsieh: There is enforcement on the server side as well, can remove here.

rust/lancedb/src/function.rs
jmhsieh: sure, can remove from here.

rust/lancedb/src/remote/db.rs
jmhsieh: sure.  authz might be an example.

cmccabe: yes, I really think we should have two different request types (of course they can call the same function internally, etc.)

rust/lancedb/src/function.rs
cmccabe: In general, it's better to just do all enforcement on the server side, because then it's easy to unit test its presence by making a request with too many bindings, etc. If the client forbids that, it's harder to test.

rust/lancedb/src/function.rs
```diff
@@ -481,13 +493,115 @@ pub struct FunctionArtifactRequest {
...
+        .next()
+        .is_some_and(|byte| byte == b'_' || byte.is_ascii_alphabetic())
+        && bytes.all(|byte| byte == b'_' || byte.is_ascii_alphanumeric())
+}
+
+/// Reject a credential the service would refuse on size alone.
+///
+/// Checked before the request body is built, so an oversized value is never
+/// serialized or uploaded.
+pub(crate) fn validate_secret_value(value: &str) -> Result<()> {
```
cmccabe: let's unit test this and other things in this file

rust/lancedb/src/remote/db.rs
```diff
@@ -570,6 +587,21 @@ struct RemoteDropFunctionResponse {
     dropped: bool,
 }
 
+#[derive(serde::Deserialize)]
+struct RemoteListSecretsResponse {
```
cmccabe, line 621: It's a bit weird to have a `struct RemoteListSecretsResponse` but not a `struct RemoteListSecretsRequest`. Can we just add that (I know it has just one field currently)

python/python/lancedb/db.py
```diff
@@ -745,6 +779,62 @@ def drop_function(self, name: str, *, version: str) -> bool:
...
+        Fails if it does not exist. Every Function bound to the Secret uses the
+        new value from its next job, and no new Function version is created --
+        which is how a rotation reaches columns pinned to a version registered
+        before it. Local connections raise ``NotImplementedError``.
+        """
+        raise NotImplementedError(
+            "Secret operations are not supported for this connection type"
+        )
+
+    def list_secrets(self) -> List[str]:
```
cmccabe, line 806: why are we mapping this to just strings, when the actual information returned has the creation and modification dates? seems like we should return a List of some structure, not a bare string.

cmccabe:
Hmm...  In general it seems like if we are returning the creation date, etc. in the List RPC, we should surface. Or, if we're not, we should remove it from the List RPC (and make people use describe for that). It's awkward to send all that data and then throw it away.

rust/lancedb/src/remote/client.rs
jmhsieh: ok, updated.

python/python/lancedb/db.py
jmhsieh: The plan is to change this to epoch timestamp.

rust/lancedb/src/function.rs
jmhsieh: Make sense.  updating.

rust/lancedb/src/function.rs
jmhsieh: this is removed.

rust/lancedb/src/database.rs
jmhsieh: Ok, after some chat, we'll update this to use epoch time but keep it to allow users to quickly verify secret alters went through using describe.

rust/lancedb/src/function.rs
jmhsieh: I went back and forth with jack and claude on this -- will go with the union type approach since it will be easier for the next person to add the next secret binding type.

Union type approach introduces forwards compat issues, but these have been mitigated elsewhere in the codebase and the patterns can be borrowed. (#[non_exhaustive] plus an Unrecognized { type } catch-all).     You do get enum match exhaustiveness on the serverside which lets the compiler catch more for you when new types are added.

The sibling field approach is cheaper on wire and gets backwards/forwards compat for free but adding new types will be more complex because you have to find all the places to update.

rust/lancedb/src/remote/db.rs
```diff
@@ -499,6 +532,16 @@ fn build_table_identifier(name: &str, namespace: &[String], delimiter: &str) ->
...
 }
 
+/// A Secret's path identifier: its namespace path and name joined the way every
+/// other object's is.
+///
+/// Secrets are not tables, but the identifier grammar belongs to the namespace
+/// spec rather than to any one object type, and having two would mean a caller
+/// had to know which kind of name it was holding.
+fn build_secret_identifier(name: &str, namespace: &[String], delimiter: &str) -> String {
+    build_table_identifier(name, namespace, delimiter)
```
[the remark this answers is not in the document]
jmhsieh: We're going to reject any delimiter values other than '$' on the server side.

rust/lancedb/src/remote/db.rs
jmhsieh: Rest API changed and is now integrating the secrete name into the route name, and eliminating body contents and using query args.

rust/lancedb/src/remote/db.rs
```diff
@@ -499,6 +540,72 @@ fn build_table_identifier(name: &str, namespace: &[String], delimiter: &str) ->
...
+/// like the table and Function identifiers beside it. One the service would
+/// refuse is merely made to arrive, so that it can be refused.
+fn build_secret_identifier(name: &str, namespace: &[String], delimiter: &str) -> String {
+    namespace
+        .iter()
+        .map(String::as_str)
+        .chain(std::iter::once(name))
+        .map(encode_identifier_component)
+        .collect::<Vec<_>>()
+        .join(delimiter)
```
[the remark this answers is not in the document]
jmhsieh: I am asserting that the server will not accept anything except for '$' as a delimiter.

rust/lancedb/src/remote/db.rs
```diff
@@ -499,6 +540,105 @@ fn build_table_identifier(name: &str, namespace: &[String], delimiter: &str) ->
...
+///
+/// The default `$` needs no special handling: it is not unreserved, so it is
+/// already escaped inside a component and only ever appears raw as the
+/// separator. Nothing on that path changes shape.
+fn encode_identifier_component(component: &str, delimiter: &str) -> String {
+    let encoded = urlencoding::encode(component).into_owned();
+    if delimiter.is_empty() || !encoded.contains(delimiter) {
+        return encoded;
+    }
+    encoded.replace(delimiter, &percent_encode_every_byte(delimiter))
```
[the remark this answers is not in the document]
jmhsieh: So the client api is [ "prod", "vision"] but the rest should convert tp "prod$vision".   If "%24" is present in the namespace, it should be unambiguous on the server side -- ["prod$vision"]  or ["prod%24vision"] would turn into "prod$vision" on the url and point to the same secret.

rust/lancedb/src/remote/db.rs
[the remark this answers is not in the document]
jmhsieh: This should be handled now.

jmhsieh:
The listing doesn't carry dates — `ListedSecret` has only `name`, on both sides. It's names-only because a name is the object key, so a listing reads no record bodies at all; timestamps live in the body, so a metadata listing is one LIST plus one GET per Secret.

rust/lancedb/src/remote/db.rs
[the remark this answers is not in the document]
jmhsieh: Ok, this will explicitly be handled in a separate pr here https://github.com/lancedb/lancedb/pull/4189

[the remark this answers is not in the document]
jmhsieh: Reject this, we'll tackle the concern in #4189.

rust/lancedb/src/remote/db.rs
```diff
@@ -352,6 +356,35 @@ impl RemoteDatabase {
...
+    /// parameter, which keeps it out of access logs and proxy traces.
+    /// The path segment addressing one Secret.
+    ///
+    /// Written into the path as-is, the way a table identifier is. The charset
+    /// a Secret name and a namespace segment admit excludes everything that
+    /// would need escaping, and a `/v1/secret/prod$openai/drop` that reads like
+    /// the table and Function routes beside it is worth more than encoding
+    /// against a charset that does not exist yet.
+    fn secret_id(&self, name: &str, namespace_path: &[String]) -> String {
+        build_secret_identifier(name, namespace_path, &self.client.id_delimiter)
```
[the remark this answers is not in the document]
jmhsieh: Reject this, we'll tackle the concern in https://github.com/lancedb/lancedb/pull/4189.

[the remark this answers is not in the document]
jmhsieh: Acknolwedged.  Reject this, we'll tackle the concern in https://github.com/lancedb/lancedb/pull/4189.

rust/lancedb/src/remote/db.rs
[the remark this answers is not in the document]
jmhsieh: Acknowledged. Reject this, we'll tackle the concern in https://github.com/lancedb/lancedb/pull/4189.

rust/lancedb/src/remote/db.rs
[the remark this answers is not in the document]
jmhsieh: Yes, we intend to land #4189 after this lands and recognize the concern and accept it.

rust/lancedb/src/remote/db.rs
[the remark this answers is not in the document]
jmhsieh: Yes, we intend to land #4189 after this lands and recognize the concern and accept it.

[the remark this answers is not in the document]
jmhsieh: Explicitly ignore this concern in this PR.

rust/lancedb/src/remote/db.rs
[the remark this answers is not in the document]
jmhsieh: Explicitly ignore this concern in this PR.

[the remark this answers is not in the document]
jmhsieh: Explicitly ignore this concern in this PR.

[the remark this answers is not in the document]
jmhsieh: If I added a TODO marker in the code would you accept?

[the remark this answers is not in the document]
jmhsieh: I am overriding you explicitly to accpet condition because PR #4189 will address it.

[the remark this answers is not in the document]
jmhsieh: PR description has been updated noting that the concerns will be address in follow on pr.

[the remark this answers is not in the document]
jmhsieh: This is non blocking.  Will be addressed in #4189

jmhsieh:
Fixed in b791d531. An empty component is refused before the join, beside the relative segment and the delimiter — the three ways a component can erase a boundary the split has to recover.

`["prod", ""]` joined to `prod$`, which reads back as `["prod"]`, so a drop reached the parent of the namespace the caller named. The catalog's own `validate_name` already required nonempty; this is the identifier layer agreeing with it.

Per-component percent-encoding is unchanged, so a catalog database named `team/search` still addresses `/v1/namespace/team%2Fsearch/...`.

python/python/lancedb/functions.py
```diff
@@ -227,6 +228,33 @@ class FunctionOutput(_OpenRemoteValue):
...
+    time either grows a new way to be configured. Naming the parts settles all
+    of that: nothing here is parsed, so nothing can parse two ways.
+    """
+
+    name: str
+    namespace_path: tuple[str, ...] = ()
+
+
+class SecretBinding(_RemoteValue):
+    """How a Secret reaches the Function that binds it.
```
jackye1995, line 245: nit: this can not just apply to function, more like "how a secret is delivered in an execution environment"?

jackye1995 (approved):

PR #4151, opened 2026-09-08
feat(secrets): address Secrets by namespace path

jmhsieh:
Stacked on #4150 — review that first; the diff here is against it.

A Secret is identified by a namespace path plus a name, and resolution is
exact: a Secret under `["prod"]` is not visible from `["prod", "vision"]` and
never falls back to a parent.

```python
db.create_secret("openai-prod", key, namespace_path=["prod", "vision"])
function = db.create_function(
    analyze_caption,
    secrets=[EnvVarSecret(
        secret="openai-prod",
        env_variable="OPENAI_API_KEY",
        namespace_path=["prod", "vision"],
    )],
)
function.secret_env_bindings
# {'OPENAI_API_KEY': SecretReference(name='openai-prod',
#                                    namespace_path=('prod', 'vision'))}
```

**A binding names a Secret by its parts, not by a joined id.** An earlier draft
of this PR joined the path and the name with `$`. That needs a delimiter, and a
delimiter has to be excluded from every name and every segment forever, agreed
on by both sides, and re-agreed each time either side grows a new way to be
configured. `ClientConfig.id_delimiter` is exactly that: it already governs
table identifiers and is advertised on every request, so an id joined on a
fixed `$` would contradict a delimiter the same request declares.

`SecretReference {name, namespace_path}` costs one object and settles it —
nothing is parsed, so nothing can parse two ways, and the character set is the
service's storage rule rather than a delimiter's shadow. Periods are legal for
that reason: LanceDB namespace names already permit one, and excluding it would
leave Secrets unaddressable inside a namespace a user already has.

- Every verb takes `namespace_path` keyword-only, defaulting to the root, and
  so does `EnvVarSecret`. Keyword-only from the start, so a later parameter
  cannot be mistaken for the path and the dual-spelling migration `open_table`
  needed is avoided.
- `EnvVarSecret` pins the path at construction, so a worker resolves the Secret
  it was handed and never re-resolves against its own default namespace — the
  same Function resolves the same Secret wherever it runs.
- `FunctionRegistrationRequest::validate` checks the name and every segment at
  the shared Rust boundary, so a hand-rolled envelope reaching the low-level
  entry point is refused the same way the typed API is.

**A root path is omitted from the request body rather than sent empty**, so a
root binding is byte identical to one written before namespaces existed. A test
pins that both ways.

**The service implements this.** Sophon holds one Secret catalog per database
with a namespace as a real prefix inside it — a directory per segment, the name
as the leaf — so listing a namespace is a non-recursive list and the exactness
above falls out of the layout rather than being enforced on top of it. Every
verb round-trips through the path it was given, and a name in one namespace is
not a name in another.

**Tests:** the wire-shape test above, Python and Rust coverage for reference
validation and local segment rules, and a REST round-trip proving a namespaced
create/list/alter/drop reaches that Secret and no other. Full suites green —
lancedb `remote` 406, function slices 8 + 6, 94 Python tests.


rust/lancedb/src/function.rs
```diff
@@ -410,7 +410,7 @@ pub struct FunctionVersion {
     runtime_digest: String,
     environment_digest: String,
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
-    secret_env_bindings: BTreeMap<String, String>,
+    secret_env_bindings: BTreeMap<String, SecretReference>,
```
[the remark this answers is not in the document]
jmhsieh: The compatibility window between 4150 and 4151 is not a valid concern.   They are both net new and will essentially land together before the sophon enterprise side lands.

jmhsieh:
> ❌ **Gate recommendation: request changes.**
> 
> The description now reports Sophon support, so the earlier 501 statement is no longer the current basis. The cited cross-service evidence is still not verifiable from this change: its namespaced Secret request coverage uses in-process handlers/mocks, and no reviewable service implementation or executed end-to-end result is linked. That support is the acceptance boundary because merged [#4047](https://github.com/lancedb/lancedb/pull/4047) removed an unreleased Function-secret surface when Sophon did not implement it.
> 
> Provide reviewable service support and a namespaced round-trip covering the Secret verbs plus structured Function binding, with a service-first or atomic rollout; then merge [#4150](https://github.com/lancedb/lancedb/pull/4150) and rebase or retarget this PR.

Reject this.  The update is WIP on the lancedb enterprise side and you should assume that service support is present there or will send an error message here.

python/python/lancedb/secrets.py
```diff
@@ -92,21 +118,34 @@ def env_variable(self) -> str:
...
-            f"env_variable={self._env_variable!r})"
+            f"env_variable={self._env_variable!r}{path})"
         )
 
     def __eq__(self, other: object) -> bool:
         return (
             isinstance(other, EnvVarSecret)
             and other._secret == self._secret
             and other._env_variable == self._env_variable
+            and other._namespace_path == self._namespace_path
```
jackye1995: I think in this case we should name it as `EnvVarSecretBinding.secret_name` , `EnvVarSecretBinding.secret_namespace_path`

python/src/connection.rs
```diff
@@ -708,45 +708,67 @@ impl Connection {
...
+                .await
+                .infer_error()
         })
     }
 
     pub fn alter_secret(
         self_: PyRef<'_, Self>,
         name: String,
         value: String,
+        namespace_path: Vec<String>,
```
jackye1995: In general we want `Optional<Vec<String>>` so that Python side can just input None for simplicity if there is no namespace.

jmhsieh: done.

python/python/lancedb/secrets.py
jmhsieh: Updated in #4150 (on `EnvVarSecret`).  resolving this and now merging as suggested.

jmhsieh:
Folded into #4150.

The namespace commit now sits on `jon/secrets-client-api` directly, so the Secret identity contract — name, namespace path, and the binding that carries both — is reviewable as one piece rather than as a shape introduced in one PR and replaced in the next. Nothing is dropped; `b9902df3` is unchanged apart from Jack's `secret_name` / `secret_namespace_path` rename and the `Option<Vec<String>>` change on the pyo3 secret methods.

Both review items from this PR are addressed there:
- `EnvVarSecret.secret_name` / `.secret_namespace_path`
- the five pyo3 secret methods now take `Option<Vec<String>>` with a `None` default, matching every other namespace-taking method on `Connection`

PR #4157, opened 2026-09-10
chore: follow DataReplacement's new source_fields and replaced_offsets

wkalt:
Lance's `DataReplacement` operation now carries the fields its values were computed from and the offsets it changed. The local refresh path declares neither yet, and the materialized-view classifier only reads the replacements, so both just acknowledge the new fields.

Depends on the lance change that adds them; the pin moves when it lands.


wkalt:
Closing: the lance fields this followed are being superseded by cell preconditions (lance-format/lance#9121).

PR #4166, opened 2026-09-13
fix(remote): align Function CRUD routes

jackye1995:
Align the experimental Function HTTP transport with the equivalent Table CRUD API shape. This is an intentional breaking change to the experimental Function routes; public Rust and Python APIs remain unchanged.

## Route comparison

| Operation | Function before | Function after | Equivalent Table API |
| --- | --- | --- | --- |
| Create | `POST /v1/functions/create` | `POST /v1/function/{id}/create` | `POST /v1/table/{id}/create` |
| Describe | `POST /v1/functions/describe` | `POST /v1/function/{id}/describe` | `POST /v1/table/{id}/describe` |
| List | `POST /v1/functions/list` | `GET /v1/namespace/{id}/function/list` | `GET /v1/namespace/{id}/table/list` |
| Drop | `POST /v1/functions/drop` | `POST /v1/function/{id}/drop` | `POST /v1/table/{id}/drop` |

## Contract details

- Create, describe, and drop use a singular resource path. Their `{id}` path parameter is the URL-encoded Function name, and the duplicate Function identifier is removed from each request body.
- Create continues to accept `202 Accepted`.
- List changes from a POST with a JSON body to a namespace-scoped GET. Its `{id}` path parameter is the namespace identifier rather than a Function name.
- Functions do not support nested namespaces yet, so the client lists against the root namespace identifier (`$` with the default delimiter). A non-root namespace is rejected.
- The optional list filter is named `name`. `limit`, `page_token`, and `include_definition` remain available as query parameters.
- The paginated list response shape is unchanged.


jackye1995:
Thanks — I’m treating this as a release-ordering constraint rather than adding dual-route compatibility. The Function API is still experimental, and this change intentionally removes the old plural/body-name shape so the canonical contract stays aligned with Table CRUD. The matching server routes should be deployed before publishing a client containing this change.

Xuanwo (approved):

PR #4167, opened 2026-09-13
fix: align function REST CRUD routes on beta.16

jackye1995:
Backport the experimental Function REST CRUD wire-shape alignment to the beta.16-compatible LanceDB line used by Sophon. Function identifiers move into singular item routes, operation bodies contain only operation fields, and listing uses GET query parameters.

Mainline counterpart: https://github.com/lancedb/lancedb/pull/4166

Sophon companion: https://github.com/lancedb/sophon/pull/7848


jackye1995:
Agreed on the server-first sequencing. I am not adding dual-route compatibility because this Function API is experimental and the approved contract intentionally removes the old plural/body-name wire shape. This backport should land only as the durable Sophon prerequisite, and no compatible client should be published before the companion server routes are deployed.

PR #4177, opened 2026-09-14
fix(deps): update rustls and cap aws-smithy-types to unbreak CI

jackye1995:
Two upstream dependency releases broke CI on `main`. Both fixes are dependency constraints, so they ride together.

## `deny` — RUSTSEC-2026-0285

rustls 0.23.40 accepts TLS 1.3 handshake messages sent at the wrong encryption level ([advisory](https://rustsec.org/advisories/RUSTSEC-2026-0285)), patched in 0.23.45.

rustls 0.23.45 requires `aws-lc-rs >= 1.18`, which the nodejs crate pinned to `=1.16.3`, so this also bumps that pin and its `aws-lc-sys` companion to `=1.18.1` / `=0.45.0`. The pin comment already calls for periodic updates on security patches.

The workspace's other rustls (0.21.12) is below the advisory's affected range (`unaffected = ["< 0.23.13"]`).

## `build-no-lock` — aws-smithy-types 1.7.0

`aws-smithy-types` 1.7.0 and `aws-smithy-json` 0.64.0 both released 2026-09-14. 1.7.0 made `Document` `non_exhaustive`, which `aws-smithy-json` 0.63 does not compile against:

```
error[E0004]: non-exhaustive patterns: `&_` not covered
  --> aws-smithy-json-0.63.0/src/serialize.rs:36:15
note: `aws_smithy_types::Document` defined here
  --> aws-smithy-types-1.7.0/src/document/mod.rs:91:1
```

Every `aws-sdk-*` crate moved to `aws-smithy-json ^0.64`, but `aws-config` 1.12.0 still requires `^0.63`, so a lockfile-free resolve pairs json 0.63.0 with types 1.7.0 and fails. This caps `aws-smithy-types` below 1.7 as a constraint-only dev-dependency, matching the existing `aws-smithy-runtime` entry. Revert once `aws-config` moves to `aws-smithy-json` 0.64.

Note this break is not specific to this PR — `build-no-lock` fails the same way on unrelated branches (e.g. `jon/secrets-client-api` run 34903587364), which passed it hours earlier.

## Verification

Resolution only, no local build:

- Locked resolve unchanged: `aws-smithy-types` stays 1.4.8; the only `Cargo.lock` delta from the cap is the new dev-dep edge.
- Fresh resolve (`rm Cargo.lock`): `aws-smithy-json` 0.63.0 with `aws-smithy-types` 1.6.3, `aws-sdk-*` one release back, `rustls` 0.23.45 retained.


rust/lancedb/Cargo.toml
```diff
@@ -111,6 +111,14 @@ aws-sdk-s3 = { version = "1.55.0" }
 aws-sdk-kms = { version = "1.48.0" }
 aws-config = { version = "1.5.10" }
 aws-smithy-runtime = { version = "1.9.1" }
+# Not used in code -- this caps aws-smithy-types so `build-no-lock` resolves.
+# 1.7.0 made `Document` non_exhaustive, which aws-smithy-json 0.63 does not
+# compile against, and aws-config 1.12.0 still requires aws-smithy-json ^0.63.
+# The lower bound only keeps the cap inside the 1.x line; without it cargo
+# satisfies this with a 0.x release and leaves everything else on 1.7. Keep it
+# low enough for the MSRV job, which pins aws-smithy-types to 1.3.6.
+# Drop once aws-config moves to aws-smithy-json 0.64.
```
hamersaw: comment feels a bit long for a version bump.

hamersaw (approved):

rust/lancedb/Cargo.toml
jackye1995: Trimmed to three lines — kept only why it's capped and when to drop it. Left one line on the bounds because both are load-bearing: no lower bound and cargo satisfies this with a 0.x release while the rest of the graph stays on 1.7, and a floor above 1.3.6 breaks the MSRV job's pin. I hit both of those before landing on this.

PR #4184, opened 2026-09-15
fix(node): wait for clock tick before cleanupOlderThan cutoff in test

BubbleCal:
## Problem

The Node test `when optimizing a dataset › cleanups old versions` is flaky on `main` since #4160. It fails in the NPM Publish workflow's Linux test jobs (`Expected: 2, Received: 1`), for example on #4183 and on `gatekeeper/fix-2866-1`, both of which do not touch this code.

The test captures `new Date()` immediately after the last commit made in `beforeEach`. Lance stores version timestamps with nanosecond precision and prunes with `manifest.timestamp() < before_timestamp`, while a JS `Date` truncates to milliseconds. When the commit and the cutoff land in the same millisecond, the cutoff compares as earlier than that version, the version survives, and only one old version is pruned.

## Change

Wait for the wall clock to advance to the next millisecond before capturing the cutoff, so it is strictly later than every prior commit. Test-only change.

## Validation

- Looped the original scenario 200 times on an M-series Mac: the `new Date()` cutoff pruned the wrong count in 26/200 runs, the fixed cutoff in 0/200.
- `node_modules/.bin/jest __test__/table.test.ts -t "when optimizing a dataset"` passes (3 consecutive runs).
- `pnpm lint`


BubbleCal:
Superseded: the same test fix was pushed directly onto #4183 (commit 872dfb29).

PR #4188, opened 2026-09-15
fix: keep computed-column freshness across branches and clones

wkalt:
A branch or shallow clone recomputed every computed column on its first refresh: the input signature carried each data file's raw base id, which the clone commit rewrites, and the signature sidecar was looked up under the branch's own tree, where nothing was ever written.

A file is now identified by where its store resolves it, so the source's own root and a clone's registered reference to that root sign the same while different bases stay distinct; compaction products are matched the same way. Sidecars live in one `_computed/` at the table root shared by main and its branches; a shallow clone reads through the base it was cloned from and keeps a copy. Pruning collects references across every branch, and the staleness walk treats the versions a branch does not hold as unknown. Stamps written before this change no longer match, so the first refresh after upgrading recomputes once; a deep clone, which copies no sidecar, still starts over.


wkalt:
Addressed in `2f4fae9a1`. A file is now identified by where its store resolves it: `Bases` maps the dataset's own files to `<uri>/data` and each registered base to its URI (with `/data` under a dataset root), normalized through `uri_to_url` with the query dropped, so the source's implicit root and a clone's registered reference to that root sign the same while different bases stay distinct. Compaction-product matching in `step_at` uses the same identity. Unit test `a_file_signs_by_its_resolved_location` covers the equivalence, the distinction, and an unknown base id.

BubbleCal (approved):

PR #4189, opened 2026-09-15
feat: check object identifier components in the Rust client

jmhsieh:
Splits the client-side identifier checks out of #4150, and extends them to
tables, namespaces and Functions — the failures were the same for all three and
reachable from the public Rust API.

Stacked on #4150.

## The problem

An object is addressed as `<object>/{id}/<action>`, where `{id}` is the
namespace path and the name joined by a delimiter. The client performs that
join, and **the join is what decides identity**. The service can only judge the
components the *split* produced, so it is asked about the wrong object, or not
asked at all:

```rust
drop_table("../jobs", &[])       -> /v1/jobs/drop/           // left the route
drop_table("a$b", &[])           -> /v1/table/a$b/drop/
drop_table("b", &["a"])          -> /v1/table/a$b/drop/      // the same URL
create_secret("../jobs", value)  -> /v1/jobs/create          // a credential, to a foreign route
```

`a$b` is not a legal table name — `$` is outside the charset. It just never
reaches a validator, because by then it has already become two valid
components.

## What this adds

**The dot-segment rule moves into `validate_table_name` and
`validate_namespace_name`**, so the local path in `database/listing.rs` and the
service — which imports both — inherit it with the client. `.` and `..` pass
every character test and still address nothing: they are resolved as relative
path segments, and *after* percent-decoding, so `%2E%2E` is resolved exactly as
`..` is. No spelling survives.

**`build_table_identifier` and `build_namespace_identifier` become fallible**
and validate each component before joining — one place rather than nine call
sites. Both also refuse a component holding the configured delimiter, the rule
that cannot live in a shared validator: the default `$` is outside every
charset here, but a configured delimiter may be any string, and the readable
choices — `.`, `-`, `_` — are all inside them.

**Secrets get their own component validator** in `lancedb::utils`, applied on
all five `Connection` verbs, plus percent-encoding of components as a second
line.

**Functions need only the dot rule.** Their identifier is percent-encoded,
which covers the rest, and they have no namespace to join — so no boundary a
delimiter could erase.

## Why the client

This is not a second opinion on whether a name is acceptable. It is what makes
the service's opinion reachable — an unencoded `../jobs` never arrives for the
service to reject.

- It is one definition, not two. The service already imports this crate's
  `validate_table_name` and `validate_namespace`; the Secret rule is placed
  beside them so it can be imported the same way.
- It covers every binding. Python and Node wrap this crate. Today the checks
  are scattered — Secrets validated in Python, Functions at registration,
  tables in exactly one method, Rust nowhere.
- `validate_table_name` is already called on the local path in
  `database/listing.rs`. The remote path is the inconsistent one.

## One test changed rather than added to

`test_table_names_with_nested_namespace` had its mock return `ns1$ns2$table1`
from a namespace listing. The service returns bare names there, as the sibling
test and the service's own tests both show. The client joins the namespace back
on to build each table's identifier, so the joined form was being joined twice
into `ns1$ns2$ns1$ns2$table1`. Nothing caught it because nothing validated the
result.

## Tests

A dot-only or delimiter-bearing table name, namespace segment, Secret component
and Function name are each refused before a request is built. An admissible
identifier is unchanged by the encoding, so the route still reads
`/v1/secret/prod$vision_2$openai-prod.v1/drop`.

lancedb lib 1230, `first_class_function_slice1` 9, `first_class_function_slice2` 3,
353 Python tests across the identifier-relevant files.


rust/lancedb/src/remote/client.rs
```diff
@@ -562,10 +600,12 @@ impl RestfulLanceDbClient<Sender> {
             host,
             retry_config,
             sender: Sender,
-            id_delimiter: client_config
-                .id_delimiter
-                .clone()
-                .unwrap_or("$".to_string()),
+            id_delimiter: validate_id_delimiter(
```
jackye1995: if this is always set from client config, why not just validate when constructing client config?

rust/lancedb/src/utils/mod.rs
```diff
@@ -81,6 +81,32 @@ impl PatchReadParam for ReadParams {
...
+/// decoding -- `%2E%2E` is a double-dot segment as surely as `..` is, so no
+/// escaping of one survives to reach a route. An object so named could be
+/// created and then never addressed again, and on the way there the request
+/// goes somewhere else: `drop_table("..")` resolves to `/v1/drop/`, delivering
+/// a body meant for one route to whatever handler is left at the other.
+///
+/// Deliberately no wider than that. `...` is an ordinary segment and addresses
+/// perfectly well; refusing it would make an object that works today stop
+/// working on upgrade, for a hazard it does not have.
+fn reject_relative_segment(what: &str, value: &str) -> Result<()> {
```
jackye1995: I think we can just update the table name regex? This is what gpt gives me:

To reject consecutive dots while preserving leading/trailing single dots:

```regex
^(?:\.?[A-Za-z0-9_-]+(?:\.[A-Za-z0-9_-]+)*\.?|\.)$
```

Allows `a.b.c`, `.foo`, `foo.`, and `.`; rejects `a..b` and `..`.

rust/lancedb/src/utils/mod.rs
```diff
@@ -126,9 +153,71 @@ pub fn validate_namespace_name(name: &str) -> Result<()> {
...
+}
+
+/// The characters a Secret name or a Secret namespace segment may contain, and
+/// the longest one either may be.
+///
+/// Deliberately the set the service admits, and no positional rule on top of
+/// it: a segment may begin with `_`, `-` or `.`, because LanceDB namespaces
+/// already do, and anything narrower would put Secrets out of reach inside
+/// namespaces that already exist.
+static SECRET_NAME_REGEX: std::sync::LazyLock<regex::Regex> =
```
jackye1995: I think we should just have a string regex for all objects to check table name, secret name, function name, MV name, and each part of the namespace path.

rust/lancedb/src/utils/mod.rs
```diff
@@ -126,9 +153,71 @@ pub fn validate_namespace_name(name: &str) -> Result<()> {
...
+    std::sync::LazyLock::new(|| regex::Regex::new(r"^[A-Za-z0-9_.\-]{1,255}$").unwrap());
```
jackye1995: name length should be checked at server side if possible.

rust/lancedb/src/utils/mod.rs
jmhsieh: We should reject '.' and '..', but I don't see why we should reject 'foo..bar'. '/v1/secret/foo..bar/drop' is valid.

rust/lancedb/src/utils/mod.rs
jmhsieh: Currently seems like the serverside uses these to enforce names as well, with no regex based length enforcement. Removing form regex has the serverside 512 bytes limit apply.

rust/lancedb/src/utils/mod.rs
jackye1995: Yeah then just use the right regex

jmhsieh: sure.  I guess you are saying for namespaces, tables, secrets, views just fold the '%2e' and '.', '..' checks into the regex?   Functions require first char to be [A-Za-z_] so it wouldn't apply to it

rust/lancedb/src/remote/client.rs
jmhsieh: sure.

rust/lancedb/src/utils/mod.rs
jmhsieh: table name, namespace name, view name and secret name should be able to have the same rule.  Functions in practice have python rules which require first char to be [A-Za-z_] but we could accept broader name here.

rust/lancedb/src/remote/client.rs
```diff
@@ -438,6 +438,69 @@ enum BodyLogging {
     Suppressed,
 }
 
+/// Check a configured identifier delimiter before anything is joined with it.
```
jackye1995, line 440: General: can we make the comments much more concise across the board?

rust/lancedb/src/remote/client.rs
```diff
@@ -438,6 +438,69 @@ enum BodyLogging {
...
+/// two path segments, and `POST /v1/secret/prod/openai/drop` matches no route
+/// at all.
+///
+/// Refusing a component that contains the delimiter is not enough to catch
+/// this, because neither component contains one -- the delimiter is the
+/// problem, not what it joins.
+///
+/// An empty delimiter is refused for the same reason from the other side: it
+/// joins `["a", "b"]` into `ab`, which no split recovers.
+fn validate_id_delimiter(delimiter: String) -> Result<String> {
```
jackye1995: What about just supporting $? We will have a lot of other issues if we allow other values for now.

rust/lancedb/src/remote/db.rs
```diff
@@ -372,8 +373,69 @@ impl<S: HttpSend> RemoteDatabase<S> {
...
+    /// The path segment addressing one Secret.
+    ///
+    /// The shape of each component is settled by then -- [`Connection`] checks
+    /// that before anything reaches a transport. What is left is the one rule
+    /// that cannot be checked there, because it depends on configuration this
+    /// layer owns: a component may not contain the delimiter it is about to be
+    /// joined with.
+    ///
+    /// [`Connection`]: crate::connection::Connection
+    fn secret_id(&self, name: &str, namespace_path: &[String]) -> Result<String> {
```
jackye1995: Can we just have a single object_id function? I think we can just turn the table ones to be general across objects.

jmhsieh (commented):
comments from previous round.

rust/lancedb/src/remote/client.rs
jmhsieh: yeah, I usually do a pass making them more concise.

rust/lancedb/src/remote/client.rs
jmhsieh: I agree it would be simpler.  

Supporting other delimiters than '$' is preexisting and introduced by you a year ago by #2603, and wired through the code base and sophon passed as a query parameter.   It talks about having the delimiter for what seem to be hypothetical cases.  

Just allowing '$' on theclient side would be a breaking change, so I was staying within the existing contract.    Should we drop the support for other delimiters?  Do you remember the rationale for the argument?

jmhsieh (commented):
Need input on if we want to drop user specified delimiters on the client side.

rust/lancedb/src/utils/mod.rs
jackye1995: > Functions in practice have python rules which require first char to be [A-Za-z_] but we could accept broader name here.

Actually would be great if we can enforce that across the board. Maybe with the exception that we should allow names starting with `_`

rust/lancedb/src/utils/mod.rs
jackye1995: I think we want the service to be able to control that as a service restriction that can be changed if needed

rust/lancedb/src/remote/db.rs
jmhsieh: done.  the different identifiers throw different errors messages so there still is a wrapper for them.

rust/lancedb/src/remote/client.rs
jackye1995: I remember we discussed about that, because the API we call into are basically the lance rest catalog open standard, which contains the delimiter config more for other catalogs to adopt to it. For enterprise specifically, the rule has been $ only.

But maybe I should not have even introduced the concept in lancedb layer if that is not gonna have another value. I am good with even just removing it as a breaking change if we want, but just having it as a single allowed value for now and propagate it across the board should also be good.

rust/lancedb/src/utils/mod.rs
jmhsieh: I'll leave it as is for now. (No explicit enforcement other than what is allowed by construction.) we can revisit when we are namespace prefixes to functions.

rust/lancedb/src/remote/client.rs
jmhsieh: Went with the single allowed value: `ClientConfig::id_delimiter` stays, and anything but `$` is refused at construction, so a mistake is reported where it was written rather than as a confusing response later.

Fixing it let a fair bit go:

- the `?delimiter=` query parameter — it only ever fired for a non-`$` value
- `RestfulLanceDbClient::id_delimiter`, now a `pub(crate) const ID_DELIMITER`
- the refusal of a component that contains the delimiter, and both its tests — `$` is outside the character set every object name admits, so no validated component can hold one

Net −138 lines. Marked as a breaking change in the commit; the only callers that notice are ones passing something other than `$`, and there are none in-tree.

rust/lancedb/src/remote/db.rs
```diff
@@ -523,33 +513,78 @@ impl From<&CreateTableMode> for &'static str {
...
     }
+    validate_table_name(name)?;
+    Ok(join_identifier(
+        namespace.iter().map(String::as_str).chain([name]),
+    ))
+}
+
+/// The path segment addressing one Function.
+///
+/// A Function has no namespace, so there is one component and no join to
```
jackye1995: we are fixing this right now, so we can remove this restriction.

rust/lancedb/src/remote/db.rs
```diff
@@ -523,33 +513,78 @@ impl From<&CreateTableMode> for &'static str {
...
-/// other object's is.
+/// The namespace a Secret listing is scoped to.
 ///
-/// Secrets are not tables, but the identifier grammar belongs to the namespace
-/// spec rather than to any one object type, and having two would mean a caller
-/// had to know which kind of name it was holding.
-fn build_secret_identifier(name: &str, namespace: &[String], delimiter: &str) -> String {
-    build_table_identifier(name, namespace, delimiter)
+/// The root is the bare delimiter, as it is for every other object's listing.
+fn build_secret_namespace_identifier(namespace: &[String]) -> String {
```
jackye1995: why do we need one dedicated for secret, if we already have build_namespace_identifier

rust/lancedb/src/remote/db.rs
```diff
@@ -523,33 +513,78 @@ impl From<&CreateTableMode> for &'static str {
...
 
-fn build_table_identifier(name: &str, namespace: &[String], delimiter: &str) -> String {
-    if !namespace.is_empty() {
-        let mut parts = namespace.to_vec();
-        parts.push(name.to_string());
-        parts.join(delimiter)
-    } else {
-        name.to_string()
+/// The path segment addressing one table: its namespace path and name.
+fn build_table_identifier(name: &str, namespace: &[String]) -> Result<String> {
```
jackye1995: given we will add related support for function, can we just have a shared `build_object_identifier`

jackye1995 (commented):
mostly looks good to me, a few nit comments and we should be good!

rust/lancedb/src/remote/db.rs
jmhsieh: Done in e9919792 — `build_object_identifier(what, name, namespace)` now serves tables, Secrets, Functions and MVs.

`build_table_identifier` survives as a two-line wrapper, only for the error type: an invalid table name reports `InvalidTableName`, which callers match on, while everything else reports `InvalidInput`.

rust/lancedb/src/remote/db.rs
jmhsieh: Gone in e9919792. `list_secrets` scopes itself with `build_namespace_identifier` like every other listing, so the dedicated Secret pair is deleted. It also picks up segment validation it did not have before, which is the right default.

rust/lancedb/src/remote/db.rs
jmhsieh: Removed in e9919792 — the builder takes a namespace path for every object, and the Function call sites pass an empty one until Functions have namespaces. When your change lands, those call sites are the only edit.

I left the Function name grammar alone, per the thread above: only the relative-segment refusal applies to it, no character set.

jackye1995 (approved):

PR #4191, opened 2026-09-15
fix: do not allow . and .. as table names

cmccabe:
Do not allow . and .. as table names. They are incompatible with the local filesystem, and confusing in cases where they are supported.


jmhsieh (approved):
+1 pending ci

ChilePiquin (approved):

PR #4199, opened 2026-09-16
feat(rust): add zonemap index builder

ChilePiquin:
Lance supports ZoneMap scalar indexes, but the LanceDB Rust API did not expose a first-class way to request one through `Table::create_index`. Users had builders for the other scalar index families, while ZoneMap was missing from the public `Index` model and remote create-index serialization. This PR adds ZoneMap as a supported scalar index option in LanceDB.

This was accomplished with the following changes:
- Added `ZoneMapIndexBuilder` in `rust/lancedb/src/index/scalar.rs`.
- Added `Index::ZoneMap` and `IndexType::ZoneMap`, including display/from-string aliases for `ZONEMAP` and `ZONE_MAP`.
- Mapped local index creation to `ScalarIndexParams::for_builtin(BuiltinIndexType::ZoneMap)` and Lance `IndexType::ZoneMap` in `rust/lancedb/src/table/create_index.rs`.
- Serialized remote create-index requests as `index_type: "ZONEMAP"` in `rust/lancedb/src/remote/table.rs`.
- Added coverage for both local ZoneMap index creation and remote request serialization.

Example:

```rust
table
    .create_index(&["my_column"], Index::ZoneMap(Default::default()))
    .execute()
    .await?;
```

### Testing

Added `test_create_zonemap_index` for local index creation and extended the remote request body test matrix for `ZONEMAP`.


rust/lancedb/src/index/scalar.rs
```diff
@@ -60,6 +60,13 @@ pub struct LabelListIndexBuilder {}
 #[derive(Debug, Clone, Default, serde::Serialize)]
 pub struct FmIndexBuilder {}
 
+/// Builder for a ZoneMap index.
+///
+/// A ZoneMap index stores min/max summaries for ranges of rows and can
+/// accelerate range predicates by pruning zones that cannot match.
+#[derive(Debug, Clone, Default, serde::Serialize)]
+pub struct ZoneMapIndexBuilder {}
```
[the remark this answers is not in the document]
ChilePiquin: Addressed in e85ae50c by adding a compile-checked rustdoc example that imports `ZoneMapIndexBuilder`, wraps the call in an async helper taking `&Table`, and passes `Index::ZoneMap(ZoneMapIndexBuilder::default())` to `Table::create_index`.

jackye1995 (approved):

rust/lancedb/src/table/create_index.rs
```diff
@@ -259,6 +259,12 @@ impl NativeTable {
                     BuiltinIndexType::Fm,
                 )))
             }
+            Index::ZoneMap(_) => {
+                Self::validate_index_type(field, "ZoneMap", supported_btree_data_type)?;
```
westonpace, line 263: The supported types for zonemap is actually a larger set than those supported for btree (e.g. binary, large string, etc.)

westonpace (commented):
Can we widen type support in a follow-up?  Supporting more types is useful because even when the zonemap doesn't provide meaningful min/max it can still provide rapid "is null" filtering.

rust/lancedb/src/table/create_index.rs
ChilePiquin: Good call, https://github.com/lancedb/lancedb/pull/4206

PR #4202, opened 2026-09-16
fix(python): add missing space in the hybrid search error message

simpleqt:
`LanceHybridQueryBuilder._validate_query`'s ValueError concatenated two adjacent string literals without a separator, rendering "...explicitly for hybrid search.**But** not both.". Added the trailing space. String-only change.


simpleqt:
Added the requested regression assertion (0148f8a2): an initial string query combined with `.vector()` now asserts the complete ValueError text, including the `search. But not both.` bytes this PR restores. Verified red/green locally — restoring the pre-fix literal fails the test, the fixed literal passes.

PR #4205, opened 2026-09-16
feat: index builders for NGram, BloomFilter, RTree

cmccabe:
Add index builders in the lancedb library for NGram, BloomFilter, and RTree indexes.


rust/lancedb/src/index/scalar.rs
```diff
@@ -82,8 +82,176 @@ pub struct FmIndexBuilder {}
 #[derive(Debug, Clone, Default, serde::Serialize)]
 pub struct ZoneMapIndexBuilder {}
 
+/// Builder for an NGram index over UTF-8 strings.
+///
+/// This index accelerates substring, `LIKE`, and regular-expression filters.
+/// It uses Lance's default trigram parameters.
```
westonpace: I don't think we accelerate all forms of like or regular expressions.  It might be nice to either say "limited" here or give specific bounds.

westonpace: It's also common-ish (I think) to have both an FTS and an NGRAM index on the same field.  We should test that (could be in a follow-up) in lancedb (lance has some tests here already)

rust/lancedb/src/index/scalar.rs
```diff
@@ -82,8 +82,176 @@ pub struct FmIndexBuilder {}
...
+/// # }
+/// ```
+#[derive(Debug, Clone, Default, serde::Serialize)]
+pub struct NGramIndexBuilder {}
+
+/// Builder for a Bloom filter index on scalar values.
+///
+/// Bloom filters accelerate equality and membership filters by skipping groups
+/// of rows that cannot match. Candidate rows are checked to remove false positives.
+/// Unset parameters use Lance's defaults.
```
westonpace: We should mention the pros / cons.  A bloom filter is much smaller than a btree or bitmap index but not as precise and is limited to equality (and !=) queries.

rust/lancedb/src/index/scalar.rs
```diff
@@ -82,8 +82,176 @@ pub struct FmIndexBuilder {}
...
+        self.probability = Some(probability);
+        Ok(self)
+    }
+}
+
+/// Builder for an R-tree index on GeoArrow geometry columns.
+///
+/// This index accelerates spatial intersection filters using geometry bounding
+/// boxes. Unset parameters use Lance's defaults. Native creation requires
+/// the `geo` feature; remote creation requires server support.
```
westonpace, line 164: What happens if the geo feature is not set?  Do we have a test for that case?

westonpace (approved):

ChilePiquin (approved):

rust/lancedb/src/index/scalar.rs
cmccabe: I'll add "certain" to make it clear that not all are accelerated.

cmccabe: Yeah, let's do that in a follow-up....

rust/lancedb/src/index/scalar.rs
cmccabe: It will get a runtime error message. I don't think we have a test. It would require multiple build variants, which we don't have an easy way to do without rebuilding everything multiple times ...

PR #4211, opened 2026-09-17
fix(python): convert objects for JSON fields

### LINKED ISSUE #4060: Feature: convert objects to JSON strings on ingestion for JSON columns
### SDK

Python

### Description

When a user configures a column for JSON, they must always pass the values into add() or merge_insert() as strings. But many times they start as objects. We should add a step in the ingestion pipeline that will convert objects into strings for fields that are configured as JSON.

Rudra-G-23:
## Summary

Closes: #4060 

  - Convert Python dict/list objects to JSON strings during ingestion.
  - Support JSON fields nested inside structs and  lists.
  - Cover `add` and `merge_insert` JSON ingestion paths.
  - Add regression coverage for nested struct and list JSON fields.

  ## Testing

  - `git diff --check`
  - Python compilation passed.
  - Focused pytest was attempted but could not complete because the native extension build stalled during `uv` bootstrap.


Rudra-G-23:
- Performance: JSON serialization now precomputes and uses only schema branches containing
    JSON. It skips vectors and non-JSON nested siblings entirely, rather than recursively
    visiting every list element.

  - Tests: all synchronous tests now use table.search().where(...).to_list() instead of
    unsupported table.query().

wjones127 (approved):
This looks good. Thanks!

PR #4226, opened 2026-09-19
feat: cap the rows per fragment a materialized view refresh writes

wkalt:
A Function column on a view is filled one fragment per worker, and a refresh writes the view's rows from one stream under lance's default cap of a million rows per file. Every view under that size is one fragment, so its Function fill runs on one worker however many the refresh was given, and a grouped view whose rows are each a whole Function call runs them one after another.

A declaration can now cap the rows per fragment; the cap is stored in the view's metadata and every refresh, rebuild or increment, writes under it. A view of Function calls declares a small cap to spread them, one row per fragment when each row is a group.

Stacked on #4223.


wkalt:
Superseded. Sizing fragments at write time made the fill's parallelism a property of the view's layout, needed a default nobody could tune, and (as the gate noted) would run into the 10,000-fragment admission cap for large grouped views. The Function-refresh planner already splits fragments across execution slots; the only thing holding a single-fragment fill to one worker was its ten-row split floor. That floor is dropped to one row in a follow-up PR on the same stack, which needs no lancedb change.

PR #4231, opened 2026-09-19
feat: retire a Function binding when its output columns are dropped

wkalt:
Dropping a column a Function binding writes was refused outright, which left a bad declaration unrecoverable: the binding is immutable, there is no rebind, and so the column could never be filled again. A drop that names every output of a binding now retires it instead.

A binding lives in three places -- the envelope in schema metadata, a `computed_column.*` marker on each output field, and the columns themselves -- and readers cross-check the first two, so removing one and leaving the others is a table that refuses every write. The retirement therefore commits in two steps, each landing a state that stands on its own: one `UpdateConfig` rewrites the envelope and clears the markers together, leaving the outputs as ordinary columns holding their last values, and the drop follows. Interrupted between them, the columns are still there to be dropped again. Folding the metadata into the drop's own `Project` was the obvious alternative and does not work: the transaction proto records `Project` as fields alone, so the edit would survive only in the writer's memory.

Naming one output of a multi-output binding is refused and names the missing siblings, since one refresh writes them in one commit. A multi-output binding's hidden `__function_assignment_*` column goes with it. Inputs stay protected while a surviving binding reads them, and drop in the request that retires the last one.


wkalt:
Closed the input-error half of the risk in `e19d5a3c`: the two checks `drop_columns` makes -- unknown column, and a drop that would remove every column -- now run before the retirement commit, using the same lookup and the same rule, so a drop accepted here is one lance accepts. An input error no longer costs the binding.

An I/O failure in the projection can still land between the two commits, and that part stays as designed: the columns and their values survive it, and re-issuing the drop finishes the job.

Worth recording that the first version of the regression passed against the unfixed code. It asserted on `table.version()`, and the error path never calls `dataset.update`, so the handle reported the old version whether or not the unbind had committed. It reads durable state now, and fails without the fix.

wkalt:
Right, and thank you -- that was a real hole. `7d87ab9b` decides the surviving-column question by resolved field id instead of raw text, so one identity per column however it is spelled. A nested path resolves to the child and leaves its parent standing, which is the same answer the projection gives.

The quoted all-column drop is now a case in the regression, and under the old text comparison it is the only one of the three that fails -- the unknown-column and bare all-column cases stay green, so the case is pinning the identity rule rather than just asserting that everything errors.

This is the third instance of the same class in this change: the route check, then `ensure_not_an_input_of`, now this. I swept the rest of the patch for raw-spelling comparisons and the remaining paths all canonicalize (`field_root`, `whole_column`, or binding-supplied names rather than caller text).

Xuanwo (approved):

